PR #22, opened 2018-10-02
WIP mdBook

### LINKED ISSUE #18: possibly use mdBook to render md files online
one possibility for the specification now that it is in md is to use [mdBook](https://github.com/rust-lang-nursery/mdBook). it's a project from the Rust team and it renders [book-like](https://rust-lang-nursery.github.io/mdBook/) html site.

teonbrooks:
closes #18. here's a proof of concept of mdBook rendered [here](https://teonbrooks.github.io/bids-specification/book/index.html).


teonbrooks:
cc: @chrisfilo

chrisfilo:
That's an awesome proof of concept. Thanks for sharing.

@choldgraf do you have some experience with such solutions? WDYT?

choldgraf:
very cool! the general layout is very similar to what we've been doing w/ Jekyll textbooks (here's an example texbook: https://www.textbook.ds100.org/)

I'm +0 on choosing a specific build system.

On the one hand, mdbook has first class support for "book" structures. On the other, it uses Rust which is an awesome tool/community, but is very unfamiliar to the scientific analytics community. 

On the one hand, Jekyll is fairly familiar and has first-class github support, on the other, it's not "designed" to host books etc.

One potential difference is the templating engine used under the hood...does mdbooks use liquid?

choldgraf:
also @teonbrooks - what's the closest "standard" license to the mozilla public license? Is it basically MIT?

chrisfilo:
It uses handlebars for templating - see https://rust-lang-nursery.github.io/mdBook/format/theme/index.html for details.

I think we should move fast with splitting the document and adopting a framework (we already have people sending content based PRs). mdbook seems to be easy to implement - are there any red flags?

teonbrooks:
> also @teonbrooks - what's the closest "standard" license to the mozilla public license? Is it basically MIT?

I think the closest would be in fact MIT but don't quite. MPL is a very permissive license from my understanding.

teonbrooks:
> I think we should move fast with splitting the document and adopting a framework (we already have people sending content based PRs). mdbook seems to be easy to implement - are there any red flags?

I agree. I think it would make build a static site easy and it would make diffing even easier and clearer.

teonbrooks:
found [this](http://veldstra.org/2016/12/09/you-should-choose-mpl2-for-your-opensource-project.html) blogpost about the [MPL](https://www.mozilla.org/en-US/MPL/2.0/) license.

tl;dr: all the things outside of the mdBook project can be whatever license we select (MIT likely). if we modify the source code of mdBook, then the changes go back to the project under the MPL. it's seen as win-win for everyone 🎉

choldgraf:
@teonbrooks one thing I think we should do if we use mdBook is write up some dev docs so that others in the community know how the general build system works. Does that sound reasonable?

teonbrooks:
@choldgraf yes, agreed. I think we should make it a separate PR bc I think the restructuring from the monolithic specification to sections should happen soon especially since people are turning their eyes to the repo

choldgraf:
+1 on a separate PR.

teonbrooks:
closing this pr for now to hash out discussion on #18. we've separated out the md splitting (#28) from the rendering engine

teonbrooks:
we can re-open this if we decide to go the way of mdBook

PR #26, opened 2018-10-02
Clarifying trial_type and HED columns in event.tsv

VisLab:
Added a sentence at end of Appendix III explaining the propose of trial_type and HED columns and why you want to allow both.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1631,6 +1631,7 @@ Example:
   }
 }
 ```
+When both the trial_type column and the HED column are included, the HED tags from both columns are used to annotate the event instance. This usage allows convenient specification of common event properties while also allowing specification of event instance details in a convenient manner.
 
 Appendix IV: Entity table
 =========================


specification.md
```diff
@@ -1631,6 +1631,7 @@ Example:
   }
 }
 ```
+When both the trial_type column and the HED column are included, the HED tags from both columns are used to annotate the event instance. This usage allows convenient specification of common event properties while also allowing specification of event instance details in a convenient manner.
```
sappelhoff: What do you mean by "the HED tags grom both columns are used ..." --> the trial_type column might contain content that does not qualify as a valid HED tag?

VisLab: Yes, this needs additional clarification.  The idea is that trial_type values should be HED tagged and contain common event instances.  In the HED system, the hedtags field of EEG.event contains instance specific HED tags, while the usertags field of EEG.event contains tags common to study-specific event codes. The tools downstream merge these fields for analysis. My understanding from discussions with Chris was that the trial_type column would be provided in conjunction with an additional table that maps trial_type to HED tags.

Would it be better to say:
When both the trial_type column and the HED column are included, the HED tags associated with the trial_type value and the HED tags in the HED tag column are merged to annotate the event. ...

sappelhoff: I am not sure that we are on the same page here. For example, search for "trial_type" in the [specification](http://bids.neuroimaging.io/bids_spec.pdf). It will yield:

> OPTIONAL. Primary categorisation of each trial to identify them as instances of the experimental conditions. For example: for a response inhibition task, it could take on values "go" and "no-go" to refer to response initiation and response inhibition experimental conditions.

A trial can consist of several events, for example:
1. onset of fixation cross
2. presentation of stimulus
3. reaction of participant
4. presentation of feedback

let's assume that these four events are part of a trial "no-go" ... in our `event.tsv` file, we could have 4 rows, each corresponding to one of the events within the trial. The `trial_type` column would have the same value for each of these rows, whereas the `HED` column would differ for each of these rows.

To me, `trial_type` and `HED` are two different variables, describing events at different scales.

I would write:
> When both the trial_type column and the HED column are included for each event (i.e., row in events.tsv), HED tags may be used to specify the event instance in a detailed and exhaustive manner, while the trial_type column provides a convenient overarching description of the trial that the event in question is a part of.

teonbrooks:
heads up @VisLab, the `specification.md` is going away (#29). I would suggest updating your PR against https://github.com/bids-standard/bids-specification/blob/master/src/99_appendices/03_hed.md

VisLab:
It is true that trial-type and HED describe at different scales. However, as explained in the section above, trial-types have HED tags stored in a map.  The point is that HED tags can be specified in both places and can assume to be merged by downstream tools.  I am not sure that I understand the previous comment as I have just started participating. I am going to update the branch suggested with new wording.

PR #33, opened 2018-10-04
update hed description to clarify trial_type

VisLab:
Still trying to clarify tags associated with general classes of events and those associated with event instances.

### DIFF
--- a/src/99_appendices/03_hed.md
+++ b/src/99_appendices/03_hed.md
@@ -34,3 +34,4 @@ Example:
   }
 }
 ```
+Both the trial_type column and the HED column may be included for each event (i.e., as columns in "_events.tsv"). In this case, an individual event is described by the union of the tags in the HED column of events.tsv and the tags associated with the dictionary entry for the trial_type specified in data dictionary "_events.json". The tags in the HED column are often specific to the event instances, while the common properties associated with the trial_type are encapsulated in the trial_type dictionary. Downstream tools should not distinguish between tags specified using the different mechanisms.
--- a/src/99_appendices/03_hed.md
+++ b/src/99_appendices/03_hed.md
@@ -14,18 +14,18 @@ sub-control01/
         sub-control01_task-emoface_events.tsv
 ```
 ```
-onset duration  trial_type  HED
+onset duration  event_type  HED
 1.2 0.6 fixationCross Event/Category/Experimental stimulus, Event/Label/CrossFix, Event/Description/A cross appears at screen center to serve as a fixation point, Sensory presentation/Visual, Item/Object/2D Shape/Cross, Attribute/Visual/Fixation point, Attribute/Visual/Rendering type/Screen, Attribute/Location/Screen/Center
 5.6 0.008 target  Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)
 500 0.008 nontarget Event/Label/Non-target image, Event/Description/A non-target image is displayed for about 8 milliseconds, Event/Category/Experimental stimulus, (Item/Natural scene/Arial/Satellite, Participant/Effect/Cognitive/Expected/Non-target, Sensory presentation/Visual/Rendering type/Screen/2D), Attribute/Onset
 ```
 
-Alternatively if the same HED tags apply to a group of events with the same `trial_type` they can be specified in the corresponding data dictionary (`_events.json` file) using the following syntax:
+Alternatively if the same HED tags apply to a group of events with the same `event_type` they can be specified in the corresponding data dictionary (`_events.json` file) using the following syntax:
 
 Example:
 ```JSON
 {
-  "trial_type": {
+  "evebt_type": {
     "HED": {
       "fixationCross": "Event/Category/Experimental stimulus, Event/Label/CrossFix, Event/Description/A cross appears at screen center to serve as a fixation point, Sensory presentation/Visual, Item/Object/2D Shape/Cross, Attribute/Visual/Fixation point, Attribute/Visual/Rendering type/Screen, Attribute/Location/Screen/Center",
       "target": "Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)",
@@ -34,4 +34,6 @@ Example:
   }
 }
 ```
-Both the trial_type column and the HED column may be included for each event (i.e., as columns in "_events.tsv"). In this case, an individual event is described by the union of the tags in the HED column of events.tsv and the tags associated with the dictionary entry for the trial_type specified in data dictionary "_events.json". The tags in the HED column are often specific to the event instances, while the common properties associated with the trial_type are encapsulated in the trial_type dictionary. Downstream tools should not distinguish between tags specified using the different mechanisms.
+Both the event_type column and the HED column may be included for each event (i.e., as columns in "_events.tsv"). In this case, an individual event is described by the union of the tags in the HED column of events.tsv and the tags associated with the dictionary entry for the event_type specified in data dictionary "_events.json". The tags in the HED column are often specific to the event instances, while the common properties associated with the event_type are encapsulated in the event_type dictionary. Downstream tools should not distinguish between tags specified using the different mechanisms.  
+
+Note: An "_events.tsv" file can have several named columns that have HED data dictionaries in "_events.json" files. All dictionaries should be applied.  Further, the normal BIDS inheritance rules apply so these data dictionaries can appear higher in the BIDS hierachy.  
--- a/src/99_appendices/03_hed.md
+++ b/src/99_appendices/03_hed.md
@@ -4,7 +4,7 @@ Events in a given study are often annotated using study-specific codes, making i
 
 HED is a controlled vocabulary of terms describing events in a behavioural paradigm. HED was originally developed with EEG in mind, but is applicable to all behavioural experiments. Each level of a hierarchical tag is delimited with a forward slash ("/"). A HED string contains one or more HED tags separated by commas. Parentheses (brackets) group tags and enable specification of multiple items and their attributes in a single HED string (see section 2.4 in [HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)). For more information about HED and tools available to validate and match HED strings, please visit [www.hedtags.org](http://www.hedtags.org). Since dedicated fields already exist for the overall task classification (`CogAtlasID` and `CogPOID`) in the sidecar JSON files HED tags from the Paradigm subcategory should not be used to annotate events.  
 
-There are several ways to associate HED annotations with events within the BIDS framework.  The most direct way is to use the `HED` column of the `_events.tsv` file to annotate events:  
+There are several ways to associate HED annotations with events within the BIDS framework.  The most direct way is to use the `HED` column of the _events.tsv file to annotate events:  
 
 Example:  
 
@@ -15,7 +15,7 @@ onset  duration  HED
 ...
 
 ```
-The direct approach requires that each line in the event file be annotated. Since there are typically thousands of events in each experiment, this method of annotation is usually not convenient unless the annotations are automatically generated. In many experiments, the event instances fall into a much smaller number of categories and often these categories are labeled with numerical codes or short names. It is usually more convenient to associate the HED annotations with the categories and allow the analysis tools to make the association with individual event instances during analysis. To use this approach, your `_events.tsv` file should associate a category (often called an event code) with each event instance. Since BIDS allows an arbitrary number of columns to be included in an `event.tsv` file, you can make this association by including a column with the event category in your `_events.tsv` file.
+The direct approach requires that each line in the event file be annotated. Since there are typically thousands of events in each experiment, this method of annotation is usually not convenient unless the annotations are automatically generated. In many experiments, the event instances fall into a much smaller number of categories, and often these categories are labeled with numerical codes or short names. It is usually more convenient to associate the HED annotations with these categories and allow the analysis tools to make the association with individual event instances during analysis. To use this approach, your _events.tsv file should associate a category (often called an event code) with each event instance. Since BIDS allows an arbitrary number of columns to be included in an _event.tsv file, you can make this association by including columns representing various types of event categories in your _events.tsv file.
 
 Example:  
 
@@ -27,7 +27,7 @@ onset  duration  mycodes     HED
 ...
 
 ```
-If you provide an `_events.json` file somewhere in your data hierarchy that has a HED map for `mycodes`. The HED tags associated with a given `mycodes` value can then be associated with the events in that category.  You may also provide a `HED` column and multiple categories.  The union of the relevant HED tags will then be associated with the event instance.
+If you provide an _events.json file somewhere in your data hierarchy that has a HED map for `mycodes`. The HED tags associated with a given `mycodes` value can then be associated with the event instances in that category.  You may provide a `HED` column and multiple category columns.  The union of the relevant HED tags will then be associated with the event instance.
 
 Example:  
 
@@ -37,12 +37,10 @@ Example:
     "HED": {
       "Fixation": "Event/Category/Experimental stimulus, Event/Label/CrossFix, Event/Description/A cross appears at screen center to serve as a fixation point, Sensory presentation/Visual, Item/Object/2D Shape/Cross, Attribute/Visual/Fixation point, Attribute/Visual/Rendering type/Screen, Attribute/Location/Screen/Center",
       "Target": "Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)"
-      "Button": ...
+      "Button": "..."
     }
   }
 }
 ```
 
-Both the event_type column and the HED column may be included for each event (i.e., as columns in "_events.tsv"). In this case, an individual event is described by the union of the tags in the HED column of events.tsv and the tags associated with the dictionary entry for the event_type specified in data dictionary "_events.json". The tags in the HED column are often specific to the event instances, while the common properties associated with the event_type are encapsulated in the event_type dictionary. Downstream tools should not distinguish between tags specified using the different mechanisms.  
-
-Note: An "_events.tsv" file can have several named columns that have HED data dictionaries in "_events.json" files. All dictionaries should be applied.  Further, the normal BIDS inheritance rules apply so these data dictionaries can appear higher in the BIDS hierachy.  
+The tags in the `HED` column are often specific to the event instances, while the common properties associated with categories such as `mycodes` are encapsulated in the _events.json dictionary. Downstream tools should not distinguish between tags specified using the different mechanisms. Further, the normal BIDS inheritance rules apply so these data dictionaries can appear higher in the BIDS hierarchy.  
--- a/src/99_appendices/03_hed.md
+++ b/src/99_appendices/03_hed.md
@@ -25,7 +25,7 @@ Alternatively if the same HED tags apply to a group of events with the same `eve
 Example:
 ```JSON
 {
-  "evebt_type": {
+  "event_type": {
     "HED": {
       "fixationCross": "Event/Category/Experimental stimulus, Event/Label/CrossFix, Event/Description/A cross appears at screen center to serve as a fixation point, Sensory presentation/Visual, Item/Object/2D Shape/Cross, Attribute/Visual/Fixation point, Attribute/Visual/Rendering type/Screen, Attribute/Location/Screen/Center",
       "target": "Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)",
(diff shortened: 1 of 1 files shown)


chrisfilo:
This looks good to me. To clarify - does this PR replace #26?

VisLab:
Yes, this is a replacement for #26

sappelhoff:
> However, as explained in the section above, trial-types have HED tags stored in a map

Could somebody point me to this "map" or the `trial_type dictionary` as you name it in this PR? I can't find anything on this in the specification. @chrisfilo can you help out?

and even *if* there is such a map ... who updates this map? As long as there are no clear rules for the `trial_type` column, people can come up with whatever description they deem sensible (which is good!). The map would thus become rather hard (read: impossible) to maintain ... and on the other hand - if we *had* strict rules for the trial_type column, we might as well just use the HED rules.

VisLab:
,
Lines 29 through 34 of 03-hed.md   define this.  This was there when I
joined the group. I think it is a good idea. It is the only really
practical way to do Hed tagging.  The natural way (outside of BIDS) is to
provide a spreadsheet of unique event codes and columns that give
corresponding HED tags.  This is the way HED validators work.  The user
then only has to tag a small number of items.  Tools map these into the
event structure and assure consistency across events. The user can also
provide additional spreadsheets that have one line for each event to give
instance specific tags (which are often used to just give the
instance-specific annotation of the event such as numerical).  Tools then
map everything. The code-specific HED goes into the .usertags field of the
EEG.event structure and the instance specific goes into the .hedtags
field.  This allows consistent editing of tagging later.

sappelhoff:
[link for convenience: HED appendix](https://github.com/bids-standard/bids-specification/blob/6b22fbe93a04a8c38ced701b48dc83f6c09bf2d8/src/99-appendices/03-hed.md)

Okay, thanks for the explanation! I think I see what you mean but I am not sure how it will work.

See this `events.tsv` file ... I have added a column *actual_event* just for the sake of the example ... imagine it was not actually present in the file.


```
onset    duration    trial_type    *actual_event*
1.1    n/a    go    fixation cross onset
1.3    n/a    go    stimulus onset left of fix cross
1.5    n/a    go    participant reaction to stimulus
2.1    n/a    no-go    fixation cross onset
2.3    n/a    no-go    stimulus onset left of fix cross
...
```

Now with your approach, we would provide an `events.json` file with a key `trial_type` that has as its value a json object `{}`.

One key of this trial_type json object is `HED`, which is containing a mapping between each `trial_type` and an HED tag:
```
{
  "trial_type": {
    "HED": {
      "go": "some/HED/tag; that/goes/like/this",
      "no-go": "some/HED/tag; that/goes/differently",
    }
  }
}
```

Did I understand you correctly up to this point? If yes, there are two questions that I have:

1. Concretely: How would you define the HED tags in the present example (i.e., please replace  `some/HED/tag; that/goes/like/this` by the actual HED tag one should use in such a case)
2. What would you do if this mapping was not supplied by the dataset curator ... i.e., no "HED" key in the `events.json`, or even no `events.json` at all?

VisLab:
Hmmm --- that isn't what I thought trial_type meant. I see where the
confusion is.  I think there is a problem here.  You need to have the
facility to have users HED tag event codes as well as individual event
instances. The trial_type can be HED tagged but that conveys different
information than is used during a practical implementation of HED tags.  I
think you need to have maps for both trial_type and actual_event.  In
answer to item (2) above, the user can give any or all of the information
and the tags are merged.  When we write tools to map bids into the
EEG.event structure, we would map the trial_type and HED column tags into
the .hedTags field and the actual-event tags into the userTags field.  I
would call the actual event column   'event_type'.   At  analysis time,
there is no distinction between the two types of tags --- they are merged
in a consistent way. However, from a practical tagging viewpoint, almost
everyone will just tag event_type for most experiments.  The HED tools take
whatever is there.

go:    Attribute/Condition/Go

No go:  Attribute/Condition/No-go

Event_type:

Fixation cross:  Event/Category/Miscellaneous, Event/Label/Fixation,
Event/Description/Display fixation cross at center of screen to indicate
start of trial, Item/2D shape/Cross, Attribute/Visual/Fixation point

Stim left: Event/Category/Experimental stimulus, EventLabel/Square,
Event/Description/Display a red square on the left of the screen, /Item/2D
shape/Rectangle/Square, Attribute/Visual/Color/Red,
Attribute/Location/Screen/Left


I could do a better job tagging if I had a description of the experiment.

sappelhoff:
Thanks for providing the example tags! Of course these are minimal because the example is constraint, but they give a good idea of what it would be like to use HED tags here.

> You need to have the facility to have users HED tag event codes as well as individual event instances. 

I don't understand what you mean by that. Are you saying that a single column "HED" is not sufficient to exhaustively describe an event?

>  I think you need to have maps for both trial_type and actual_event. 

Sorry, but I don't think I can follow this reasoning. I think we have now clarified that the `trial_type` column is truly separate from the `HED` column. Just think of it as two ways to describe events.

- The `trial_type` is arguably very limited and idiosyncratic ... but quick and simple.  (imagine an experiment where a participant lifts a hand during a trial)

```tsv
onset    duration    trial_type
1.1    n/a    left
2.3    n/a    left
4.5    n/a    right
...
```

In the accompanying `events.json`, we can have the following map:

```json
{
    "trial_type": {
        "LongName": "trial type",
        "Description": "Which hand was lifted by the participant.",
        "Levels": {
            "left": "Left hand was lifted",
            "right": "Right hand was lifted"
        }
}
```

now completely separate from this, you can describe each event using the HED tag column ... it's also possible not to use `trial_type` at all. The HED column allows for a standardized and exhaustive description of the event, albeit at the cost that it is harder to do and there is currently no inbuilt validator (see this issue: https://github.com/bids-standard/bids-validator/issues/537). Note that for the HED column, a "map" in the `events.json` is not really possible, because all HED tags are already standardized and defined - there is no room for interpretation (which is the big plus over the simple `trial_type` column), and with no room for interpretation there is no need to document in the `events.json`. Instead, users can simply refer to the HED tag documentation at http://www.hedtags.org/ or similar.

With these premises, could you try to explain what the problem is?

VisLab:
I think we need to have additional discussion about this.  I don't think
having not having a validator at this point is really an issue.  We are
starting to work on a javascript validator --- and I don't see any
obstacles, but it will take a while to do and there are a number of
questions about what should be validated. However, I think we need to get a
usable incorporation of HED into the BIDS standard.

I am not sure we are talking about the same things and I am VERY concerned
that the single HED column does not cover 99% of the use cases of HED in
practice.  In practice, a typical experiment will have on the order of
10,000 events, but these events are characterized by a small number of
event classes (types of codes). Usually not more than 100.  The current tag
each event individually doesn't work from a manual perspective.  People are
not going to write 10,000 lines of HED tags.  They will tag there unique
codes and then they will have to write a script to write the tags into the
spreadsheet one line at a time. What happens if you have to change or
refine the tags?  Do you write something to make sure that you have the
same tags for the same kinds of events --- its a mess.

The trial_type as given by the example specifies an experimental condition
and not an event class and so it does not solve the problem.

My proposed solution is this:
1)  HED column has HED tags.
2) Any other named column (except things like latency and duration) can
have a map in event.json and if one of the key's is HED, those HED tags are
also used.
3) All of the HED tags from all of the relevant columns will be validated.
(This is not at all difficult.  Just look for HED in the events.tsv and
look for maps corresponding to the named columns in events.json.   Then you
can have trial_type   or event_code  or whatever you want.)

The downstream tools are supposed to union the tags from the various
sources.  Those are validated against the schema. The statement "Note that
for the HED column, a "map" in the events.json is not really possible,
because all HED tags are already standardized and defined - there is no
room for interpretation (which is the big plus over the simple
trial_type column), "
 doesn't make any sense to me.

Thanks!

chrisfilo:
@sappelhoff let me try to explain what the problem is

Imagine the following scenario

`_events.tsv`
```
onset    duration    trial_type HED
1.1    1    left  HedTagA
2.3    1    left  HedTagB
4.5    1    right HedTagC
```

Accompanied by the following `_events.json`
```
{
    "trial_type": {
        "HED": {
            "left": "HedTagB",
            "right": "HedTagD"
        }
}
```

Currently the spec is unclear how the information about HED tags from `_events.tsv` and `_events.json` should be combined. The proposal in this PR states that a sum of two sets should be taken therefore:

- Event 1 HED tags: HedTagA and HedTagB
- Event 2 HED tags: HedTagB
- Event 3 HED tags: HedTagC and HedTagD

Does this make more sense now?

chrisfilo:
I think spec needs clarification and this PR provides it. I just added an example to the discussion to help @sappelhoff understand the context.

VisLab:
I'm not seeing any new examples on the spec..... could you point me in the
right direction?

To continue on (sorry if I am belaboring a point) .....I am still concerned
about usability...... I have participated in the tagging of more than 30 of
other people's EEG studies and then used the results in analysis so I have
some experience in what works and what doesn't work and what you can
reasonably persuade people to do to annotate their studies. I would like
this to be as convenient and transparent as possible.

As I mentioned in a previous posting, most experimenters work by having
spreadsheet with a table of study-specific "event codes" or "event labels"
that identify event classes common to all of the recordings in the study.
They also keep spreadsheets of events for each recording with the
instance-specific information such as the event latency and duration and
the event class (or multiple classes)--- there might also be other columns
with specific information about events.  (The latter recording-specific
spreadsheets are currently what BIDS  specifies for events, but BIDS misses
the underlying global event information.)

Experimenters then write scripts to map the study-specific spreadsheet and
the recording-specific spread sheets into the structures needed for
whatever analysis software they are using. People look at the
study-specific event table when they want to understand or enhance the
event descriptions.

I would like there to be event.json files allowed at higher levels in the
BIDS file structure that have these maps that apply to lower levels in the
hierarchy.  The typical use case is that the experimenter specifies one
events.json file at the top level that specifies the event classes (again
--- it doesn't have to be a single event-type as you could specify for
example two or three of these (say   event_type and trial_type)  each with
HED tag maps.  The recording specific xxx_event.tsv  could have columns
with some or all of event_type and trial_type and HED and all that are
specified would be used.

chrisfilo:
> I'm not seeing any new examples on the spec..... could you point me in the
> right direction?

When I mentioned "examples" I was referring to the one I posted in this thread as a comment (https://github.com/bids-standard/bids-specification/pull/33#issuecomment-427706151). If you think it would be useful to add this to the spec please update the pull request (https://egghead.io/lessons/javascript-how-to-update-a-pull-request-on-github).

> I would like there to be event.json files allowed at higher levels in the
> BIDS file structure that have these maps that apply to lower levels in the
> hierarchy.  The typical use case is that the experimenter specifies one
> events.json file at the top level that specifies the event classes (again
> --- it doesn't have to be a single event-type as you could specify for
> example two or three of these (say   event_type and trial_type)  each with
> HED tag maps.  The recording specific xxx_event.tsv  could have columns
> with some or all of event_type and trial_type and HED and all that are
> specified would be used.

This is already allowed since the [inheritance principle](https://github.com/bids-standard/bids-specification/blob/master/src/02-common-principles.md#the-inheritance-principle) applies to events.tsv and events.json. If you think it's useful it might be worth fleshing this out in the context of HED and "global tags" in the HED appendix.

sappelhoff:
Thanks for providing the example @chrisfilo. I understood the point now - adding the mentioned example to the spec would be a valuable contribution.

Regarding @VisLab's previous comment:

> The statement "Note that for the HED column, a "map" in the events.json is not really possible, because all HED tags are already standardized and defined - there is no room for interpretation (which is the big plus over the simple trial_type column), " doesn't make any sense to me.

replace "not really possible" with "not really necessary"

VisLab:
Sorry, I was reading these threads from my email rather than github and the comments got out of sync.  I will make a first pass at a clarifying explanation of global tags. I agree the HED column shouldn't have a map.  It is designed for instance-specific tags.

VisLab:
I added something on current branch.  However, in doing so, I realized that I don't completely understand the BIDS naming convention.  Could someone clarify the naming convention for  _events.json files at higher levels in the hierarchy.  Thanks....

src/99_appendices/03_hed.md
```diff
@@ -14,18 +14,18 @@ sub-control01/
...
 ```
 
-Alternatively if the same HED tags apply to a group of events with the same `trial_type` they can be specified in the corresponding data dictionary (`_events.json` file) using the following syntax:
+Alternatively if the same HED tags apply to a group of events with the same `event_type` they can be specified in the corresponding data dictionary (`_events.json` file) using the following syntax:
 
 Example:
 ```JSON
 {
-  "trial_type": {
+  "evebt_type": {
```
chrisfilo: typo?

src/99_appendices/03_hed.md
```diff
@@ -14,18 +14,18 @@ sub-control01/
...
 5.6 0.008 target  Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)
 500 0.008 nontarget Event/Label/Non-target image, Event/Description/A non-target image is displayed for about 8 milliseconds, Event/Category/Experimental stimulus, (Item/Natural scene/Arial/Satellite, Participant/Effect/Cognitive/Expected/Non-target, Sensory presentation/Visual/Rendering type/Screen/2D), Attribute/Onset
 ```
 
-Alternatively if the same HED tags apply to a group of events with the same `trial_type` they can be specified in the corresponding data dictionary (`_events.json` file) using the following syntax:
```
chrisfilo: I don't see the need to introduce a new prescribed column name. Why not use `trial_type`?

chrisfilo:
This segment should explain the naming convention for higher levels of hierarchy: https://github.com/bids-standard/bids-specification/blob/master/src/02-common-principles.md#the-inheritance-principle Please let me know if something is not clear.

src/99_appendices/03_hed.md
VisLab: Sorry evebt_type was a typo.  I think that trial_type is confusing --- there are usually several events in a trial (e.g., stimulus and response).  The example given of trial_type was of an experimental condition rather than something that would naturally correspond to an "event code" that would be coming off of an acquisition system.  I think it would be natural to have both.

VisLab:
Yes that is clear.  However, can you have something above subj01.... that is the global HED files usually apply to all subjects.  Can you do it above subject?  Otherwise if you have 20 subjects in your experiment, you have to repeat 20 times.

Also, I corrected the typo, but pressed the commit before writing the explanatory comment.  Sorry, I'll be more careful next time....

src/99_appendices/03_hed.md
chrisfilo: I would suggest opening a separate PR proposing the addition of `event_type` - it's going to be a much more controversial (it is potentially backward incompatible, it requires a clear definition in https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/03-task-events.md).  The issue of combining tags between JSON and TSV is much more straightforward and consensus will be easier to reach.

chrisfilo:
You can have `task-nback_events.json` in the root of your dataset and it will apply to all events files for task `nback` across participants and sessions. This is where global HED tags could be specified.

src/99_appendices/03_hed.md
VisLab: Does trial_type have some special meaning in BIDS?  What I was proposing that the user could add any columns that they wanted to ( --- may I should call the column giraffes in the example?) and the idea would be that HED maps that were defined at various levels would be applied if they matched column names. I didn't mean to conflict with a BIDS standardized column.  It is not meant to be a prescribed column....

chrisfilo: `trial_type` is defined here: https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/03-task-events.md. I see your point and I guess this uncovers another lack of clarity in the current description - can one annotate any column from events.tsv with HED tags in events.json or only `trial_type`? If the answer is any column this needs to be clarified and could be part of this PR.

VisLab:
You should be able to annotate any column.  I'll try to clarify in the
spec.

I would like to comment that IMO the current BIDS specification of events
will make analysts very unhappy.   The trial_type isn't something you can
time-lock to. Every EEG analyst that I know time-locks to events of
specified type when doing EEG analysis.  trial_type does not correspond to
event_type.  The only saving grace is that with the extension we are
proposing with HED, they can get around this by defining their own column
and then annotating with HED.

VisLab:
The issue may be my understanding of trial_type as indicated by the example
in the spec. To me, this says that it is a condition.  Say we have an
experiment in which in one condition the subject is supposed to press a
button when they see a square and in another condition, they are just
supposed to think about pressing the button when they see a square.  The
trial itself consists of 3 standard event types (cue, square, maybe button
press) and possibly some other not connected events such as blink events.
All of the events in a given trial will have the same trial_type.  The
analyst might want to time-lock to any of these events (including blinks)
or use all of them (for temporal regression).   There is nothing in the
standard except the optional HED that will give an analyst any idea of the
what these events are. With the standard as it stands --- I would have no
idea what to do with the data.  It's almost as if there is an underlying
assumption that there will be only one event in each trial?

nicholst:
I don’t see the problem with your example. The trial_type column will have 3 different values (“cue”, “square”, “maybe”) (or a 4th, “blink”), allowing the analyst to select the events of a desired type.  

What values were you thinking trial_type would have?

sappelhoff:
> With the standard as it stands --- I would have no idea what to do with the data. 

With BIDS EEG we are introducing a new column called `event_value`. This marks a TTL marker value to an event. Then, there will be a json object in the corresponding `_events.json` which will map the TTL marker values to their meaning. Please read the BEP 006, which is very relevant to the HED discussion as well --> https://docs.google.com/document/d/1ArMZ9Y_quTKXC-jNXZksnedK2VHHoKP3HCeO5HPcgLE

Example for the `events.tsv`

```tsv
onset    duration    trial_type    event_value
1.1    n/a    left    1
1.3    n/a    left    2
1.6    n/a    left    3
2.0    n/a    left    4
...
```

Example for the corresponding `events.json`

```json
{
    "trial_type": {
        "LongName": "trial type",
        "Description": "Which hand was lifted by the participant.",
        "Levels": {
            "left": "Left hand was lifted",
            "right": "Right hand was lifted"
        }
    },
    "event_value": {
        "LongName": "event value",
        "Description": "Value of a TTL trigger that was sent to mark the event",
        "Levels": {
            "1": "onset of fixation cross",
            "2": "onset of stimulus",
            "3": "response of participant",
            "4": "feedback displayed",
        }
    }
}

```

I think that all the "infrastructure" we need for HED tagging the events is already present but might be a bit scattered across BEPs and the spec / or not adequately documented. This is also why this PR is important.

VisLab:
Stefan's example clears things up --- however, I think there is confusion about what trial_type can be. I read the EEG proposed extension and am basically happy with it, though there are several points that are not perfectly clear.  

Does the BIDS standard allow the researcher to include as many additional named columns as they want and can those named columns can have optional maps to HED tags in the _events.json?   This would give the most flexibility.  Where should this be clarified?  Should there be examples?   When the EEG bids extension is adapted, will the event_value column be included in BIDS as a whole, or only when it pertains to EEG studies.

chrisfilo:
> Does the BIDS standard allow the researcher to include as many additional named columns as they want?

Yes. As per https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/03-task-events.md "An arbitrary number of additional columns can be added."

> Can those named columns can have optional maps to HED tags in the _events.json? This would give the most flexibility. Where should this be clarified? Should there be examples? 

This is currently unclear, but I don't see why not. This could be clarified in this PR in the HED appendix. Examples will be useful.

> When the EEG bids extension is adapted, will the event_value column be included in BIDS as a whole, or only when it pertains to EEG studies.

This is a bit of a separate discussion - I don't know what is the current scope of the EEG proposal.

chrisfilo:
@VisLab this seems to be close to ready to merge. Only a couple of clarifications and rebase to resolve conflicts left. Are there any outstanding questions we can help with?

VisLab:
I will make another pass at it tomorrow and let you know when it is ready.

chrisfilo:
Great! Thanks!

VisLab:
Okay, I think this is ready to merge unless you spot some issues.

chrisfilo:
I still see conflicts - could you rebase or merge master?

src/99_appendices/03_hed.md
```diff
@@ -1,36 +1,46 @@
...
 
-Since dedicated fields already exist for the overall task classification (`CogAtlasID` and `CogPOID`) in the sidecar JSON files HED
-tags from the Paradigm subcategory should not be used to annotate
-events.
+HED is a controlled vocabulary of terms describing events in a behavioural paradigm. HED was originally developed with EEG in mind, but is applicable to all behavioural experiments. Each level of a hierarchical tag is delimited with a forward slash ("/"). A HED string contains one or more HED tags separated by commas. Parentheses (brackets) group tags and enable specification of multiple items and their attributes in a single HED string (see section 2.4 in [HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)). For more information about HED and tools available to validate and match HED strings, please visit [www.hedtags.org](http://www.hedtags.org). Since dedicated fields already exist for the overall task classification (`CogAtlasID` and `CogPOID`) in the sidecar JSON files HED tags from the Paradigm subcategory should not be used to annotate events.
```
sappelhoff:
```suggestion
HED is a controlled vocabulary of terms describing events in a behavioural paradigm. HED was originally developed with EEG in mind, but is applicable to all behavioural experiments. Each level of a hierarchical tag is delimited with a forward slash ("/"). A HED string contains one or more HED tags separated by commas. Parentheses (brackets) group tags and enable specification of multiple items and their attributes in a single HED string (see section 2.4 in [HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)). For more information about HED and tools available to validate and match HED strings, please visit [www.hedtags.org](http://www.hedtags.org). Since dedicated fields already exist for the overall task classification (`CogAtlasID` and `CogPOID`) in the sidecar JSON files, HED tags from the Paradigm subcategory should not be used to annotate events.  
```

sappelhoff: note, I just added a comma between 

> ... in the sidecar JSON files

and

>  ... HED tags from the

src/99_appendices/03_hed.md
```diff
@@ -1,36 +1,46 @@
...
+
+Example:  
+
 ```
-onset duration  trial_type  HED
-1.2 0.6 fixationCross Event/Category/Experimental stimulus, Event/Label/CrossFix, Event/Description/A cross appears at screen center to serve as a fixation point, Sensory presentation/Visual, Item/Object/2D Shape/Cross, Attribute/Visual/Fixation point, Attribute/Visual/Rendering type/Screen, Attribute/Location/Screen/Center
-5.6 0.008 target  Event/Label/Target image, Event/Description/A white airplane as the RSVP target superimposed on a satellite image is displayed., Event/Category/Experimental stimulus, (Item/Object/Vehicle/Aircraft/Airplane, Participant/Effect/Cognitive/Target, Sensory presentation/Visual/Rendering type/Screen/2D), (Item/Natural scene/Arial/Satellite, Sensory presentation/Visual/Rendering type/Screen/2D)
-500 0.008 nontarget Event/Label/Non-target image, Event/Description/A non-target image is displayed for about 8 milliseconds, Event/Category/Experimental stimulus, (Item/Natural scene/Arial/Satellite, Participant/Effect/Cognitive/Expected/Non-target, Sensory presentation/Visual/Rendering type/Screen/2D), Attribute/Onset
+onset  duration  mycodes     HED
```
sappelhoff:
```suggestion
onset  duration  mycodes
```

sappelhoff: you are not using the HED column here, so I'd remove it

VisLab:
These suggestions look good to me.  It is good to go.  I would appreciate someone else doing the actual merge.  Thanks.

chrisfilo:
Thank you @VisLab.

I am pretty busy right now - @sappelhoff would you mind helping out and opening a new PR based on @VisLab branch but synced with master and with conflicts resolved. It would be a great help.

PR #42, opened 2018-10-08
Mkdocs configuration and RTD setup

choldgraf:
Here's a demo of the bids specification:

https://bids-specificaion.readthedocs.io/en/latest/

It's currently linked on RTD against my account, since I had to build the docs off of my fork. If we like how it looks, then we'll have to figure out how to transfer it to `bids-standard`.

I haven't figured out how to get the PDF/EPUB downloads working, though this should theoretically be possible in a relatively straightforward way.


choldgraf:
relevant to #42 and #18

chrisfilo:
Thanks.

Can you see `pdflatex` running in RTD logs? Maybe switching the theme to RTD will enable the version/downloads menu making the PDF accessible.

choldgraf:
Actually I opened an issue in the RTD repo and it turns out that `mkdocs` does **not** support PDF generation:

https://github.com/rtfd/readthedocs.org/issues/4739

we could try to build with sphinx, in which case we'd get PDF generation. It shouldn't be too hard, but we won't have access to the theme you were using.

http://www.writethedocs.org/guide/tools/sphinx-themes/

chrisfilo:
I also have a suspicion that tables will be problematic in the context of PDF generation.

choldgraf:
@chrisfilo shall we merge this and I can make sure the readthedocs build works? I don't think we'll be able to demo it on the bids-standard org until it's merged into master

Issue #2, opened 2018-09-28
Replace table of contents with autogenrated solution


chrisfilo:
https://github.com/jonschlinkert/markdown-toc via CircleCI

PR #43, opened 2018-10-09
update sections 3 8.3 8.4 with line break

CPernet:
just added 'space - space'


chrisfilo:
Thank you for the PR, but I think the plan is to replace the TOC with an automated solution (see #2).

Issue #69, opened 2018-10-26
make ce-<label> optional for bold/sbref modalities


dasturge:
Some studies make use of mion/ferumoxytol as a contrast enhancing agent for fmri data, e.g. https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4843437/

should also be considered for the epi modality, as it could be present while this field map is acquired.

chrisfilo:
Sounds reasonable to me. Could you open a Pull Request with those changes? Thanks in advance.

PR #70, opened 2018-10-26
add ce-<label> for fmri data

dasturge:
added contrast enhancement as optional custom label for task fmri
added contrast enhancement as optional custom label for pepolar field maps 
closes issue #69


chrisfilo:
Thanks. Could you also update the table in https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/04-entity-table.md?

chrisfilo:
Thx. Pinging @bids-standard/bep_leads for comments. If no one objects I will merge this on 2018-11-02.

yarikoptic:
I have no concrete use case but if ce admitted, and func runs complemented with dwi or other scans, shouldn't we allow ce (although I am not sure if it has effect or not) optional for all imaging modalities?

chrisfilo:
I personally would prefer the spec development to remain use case driven and add only features that are needed. This could be a topic for a new issue though.

PR #78, opened 2018-10-30
ENH: allow _dir for other EPI (func, dwi) sequences

### LINKED ISSUE #47: BEP: make `dir` optional for other than fmap modalities
ATM, "Phase encoding direction" (`dir`) is a [required field](https://github.com/bids-standard/bids-specification/blob/6b22fbe93a04a8c38ced701b48dc83f6c09bf2d8/src/99-appendices/04-entity-table.md) for `fmap(epi)`.  But it is relevant for other sequences (`fmap`, `dwi`) but is not available in BIDS specification for them. To overcome this specification people resort to list the value in a flexible and omni-present `acq` field:
```shell
$> datalad -c datalad.search.index-egrep-documenttype=all search path:.*_acq-ap.*
search(ok): /home/yoh/datalad/openfmri/ds000221/sub-010001/ses-02/func/sub-010001_ses-02_task-rest_acq-AP_run-01_bold.json (file)
...
search(ok): /home/yoh/datalad/openfmri/ds000244/sub-01/ses-00/func/sub-01_ses-00_task-ArchiSocial_acq-ap_bold.nii.gz (file)
...
```
So it seems that demand is there and I do not see a reason why `dir` should not be allowed to be used in those cases.

yarikoptic:
Closes #47

This commit also demonstrates ability to reference other sections within the
document (when referencing fmap Case 4 where _dir described in greater detail
ATM


chrisfilo:
Pinging @bids-standard/bep_leads for comments.

satra:
`func` already has a few ways of disambiguating (run, task, etc.,.). as such my preference would be to start nudging people to get metadata from the sidecar instead of encoding in filename. perhaps we could go the other way and remove `dir` from dwi and leave it in the json sidecar.

it almost seems like we just want a disambiguation mnemonic. but i feel that naming things in filenames should be as restrictive as possible.

alternatively, we can adopt more generally what we discussed in derivatives meeting that any key and value can be added to the filename in pairs, but in general only a disambiguation mnemonic suffices.

yarikoptic:
While doing it I thought that may be detailed descriptions for all those key/pairs shared by different modalities could be brought into a single location, where it would be easy to reference them. Eg as subsections of src/99-appendices/04-entity-table.md ?

yarikoptic:
hm, ... wise words @satra ;-)  
Indeed I do not generally envision people to perform identical task twice (thus kinda demanding the same `task` and `run`) just for the sake of having both field encoding directions.  I am (re)thinking...

chrisfilo:
@yarikoptic what's the status after @satra comments? Would you be ok closing this PR?

yarikoptic:
I thought / looked at it a bit more and was provided some counter examples, e.g. HCP dataset where the same task would have LR and RL "runs".  Sure thing it could be encoded within `_run` but again that would loose the specificity/semantic which `_dir` carries.
Other "examples" include some studies where they could have changed direction within the study/across sessions (ppl do crazy stuff out there in the wild ;-)).  Without reflecting it in the name, it would become not obvious (not caught by the bids-validator as well?) that such variability is present.
So I still feel that `_dir`, if desired, should be allowed for since it provides a clear semantic on how otherwise possibly identical (within or across subjects/sessions) files differ.  Stuffing that information into `run` (if applicable at all) or `acq` would retain the same filename length while loosing clear semantic association.

Please let me know what you think @chrisfilo and @satra

chrisfilo:
I still don't think it's necessary and `acq` is the right place to put it (not `run` since it's restricted to numbers), but I don't feel strongly enough about the topic to block this change. It also been three weeks and with no further feedback from the community - this does not seem like a breaking change.

yarikoptic:
Thank you @ericearl . Taking it with @kaczmarj 's comment at https://github.com/bids-standard/bids-specification/issues/47#issuecomment-432284191  I think there is enough community support/interest to see it implemented.  I will refurbish this PR for merge

chrisfilo:
Thanks!

PR #81, opened 2018-11-05
Add CC-BY 4.0 license

KirstieJane:
I couldn't find any reference to the licence for the specification itself in the github repo but the google doc had a CC-BY 4.0. If that's still the case then here's a little license file to make that clearer for new folks 😸


chrisfilo:
Thanks!

PR #93, opened 2018-11-20
RF harmonize and thus shorten templates etc

yarikoptic:
Otherwise all of those are used inconsistently, make templates longer for no apparent benefit, etc,


yarikoptic:
Dear @bids-standard/everyone (seems a designated team with "merge" privileges would be useful to have) -- please provide your feedback and/or merge.

Cheers!

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -142,7 +142,7 @@ run labels). When there is only one scan of a given type the run key MAY be
 omitted. Please note that diffusion imaging data is stored elsewhere (see
 below).
 
-The OPTIONAL `acq-<label>` key/value pair corresponds to a custom label the user
+The optional `acq-<label>` key/value pair corresponds to a custom label the user
```
chrisfilo: Any reason why lowercase `OPTIONAL`? It's supposed to be an indication that this is a RFC2119 label. If you are proposing to change that it should be done in another PR.

chrisfilo:
I gave you merge rights - if anyone else wants that please email me.

yarikoptic:
> Personally I like the verbosity of the original names and don't see the benefit for shortening the templates, but it's just a subjective opinion.

The main problem is/was that it was not consistent to any stretch.  Some were long, some short, etc, and I believe some both ways (I could be wrong now).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
yarikoptic: ha, also found https://tools.ietf.org/html/rfc8174  ;-) so indeed then should be all UPPERCASE to be keywords.
Again -- I was just aiming for consistency and there were uses both ways, I will just push (later today) a fix to get them all UPPERCASED

yarikoptic:
ok, pushed more for consistent use of OPTIONAL and REQUIRED... didn't bother splitting into two PRs to avoid conflicts and messing with already resolved ones (life is too short ;-))

yarikoptic:
BTW, while doing linting locally I get:

```shell
$> node_modules/remark-cli/cli.js src/*.md src/*/*.md --frail
...
src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
  277:1-277:24  warning  Do not use headings with similar content (65:1)  no-duplicate-headings  remark-lint
...

$> node_modules/remark-cli/cli.js --version                                        
remark: 9.0.0, remark-cli: 5.0.0
```
which is due to that `Timing Parameters` subsection which is imho kosher since sections are different:
```
### Scanner Hardware
#### Sequence Specifics
#### In-Plane Spatial Encoding
#### Timing Parameters
...
### Task (including resting state) imaging data
#### Required fields
#### Other RECOMMENDED metadata
##### Timing Parameters
```
how to suppress it? (thought to put that call into precommit hook to avoid keep messing things up)

emdupre:
Personally I liked the longer, more explicit label names, but I'm happy to roll what whatever makes sense to the broader community. Thanks for thinking about this, @yarikoptic !! :sparkles:

yarikoptic:
heh, "broader community" seems to be just me here... I took a look back at current specification, and decided to stay firm that such unification would be helpful in the long run, there is now just too much variance (`run-<index>` and `run-<run_index>` etc) which cannot be useful.  I am going to resolve the conflict and then possibly take advantage of given to me recently merge powers. Ready to bear the blame or pride ;)

Issue #89, opened 2018-11-19
BEP022 numbering conflict for Spectroscopy BEP


yarikoptic:
[BIDS Extension Proposal 22 (BEP022): Magnetic Resonance Spectroscopy (MRS)](https://docs.google.com/document/d/1pWCb02YNv5W-UZZja24fZrdXLm4X7knXMiZI7E2z7mY/edit#) was [announced](https://groups.google.com/forum/#!topic/bids-discussion/sNw1yZgMdmE) awhile back by Dickson Wong (couldn't find github handle). I thought to add it to the registry of existing BEPs within this repo but found that BEP022 [was already taken for "PET Preprocessing derivatives"](https://github.com/bids-standard/bids-specification/blob/master/src/01-introduction.md#extensions).  Conflict was already identified by Roeland Hancock and confirmed by @filo but not reported here.

Hboni:
Oh I didn't saw this BEP. When I was asked to add [BIDS Extension Proposal (BEP): Computed tomography scan (CT)](https://docs.google.com/document/d/1fqnJZ18x5LJC8jiJ8yvPHUGFzNBZ6gW2kywYrUKWtuo/edit#), I found also that a [BIDS Extension Proposal 22(BEP022): ](https://docs.google.com/document/d/1yzsd1J9GT-aA0DWhdlgNr5LCu6_gvbjLyfvYq2FuxlY/edit#) was created so I added it, without checking the existence of others BEP. 
It's possible to change BEP number for BEP022 and BEP023 to add MRS as BEP022.

chrisfilo:
Sounds good to me - I sent an update to the PET Preprocessing Derivatives google doc changing it to 23. Could you send another PR adding MRS under 23 and changing PET derivatives to 22? Thanks!

Hboni:
I think you mean PR adding MRS under 22 and changing PET derivatives to 23 and Computed Tomography to 24 ?

chrisfilo:
Yes

Hboni:
Can be closed ?

PR #94, opened 2018-11-22
Change BEP numbers to include MRS

Hboni:
See #89 

Change BEP numbers to include MRS as BEP022 and so update BEP numbers of PET derivatives and Computed Tomography.


chrisfilo:
Thanks!

Issue #96, opened 2018-11-26
Instructions for how to update the specification with a BEP


choldgraf:
Is there a place with instructions for how to update the BIDS specification with a new BEP? I think we're ready to go soon with BIDS-iEEG, and wonder if we can use this as an opportunity to improve the docs around contributing to the markdown etc.

emdupre:
I know there's the [BIDS Contributor Guide](https://docs.google.com/document/d/1pWmEEY-1-WuwBPNy5tDAxVJYQ9Een4hZJM06tQZg8X4/edit) -- I think we decided against putting this in the starter kit, though I can't quite remember why -- maybe @KirstieJane would know ?

Regardless, there's still a lot here that references google docs, and I'm not sure if the whole thing should be converted to Markdown !

chrisfilo:
Indeed the Contributor Guide explains the procedure and has been updated a while ago to mention the need for sending a Pull Request with the BEP in Markdown form.

I think it is a great idea to improve the docs - starting from porting Contributor Guide to Markdown and incorporating it into this repo.

choldgraf:
As a first start, how about we add a section to the bids-specification website for this, and just link to the google doc?

chrisfilo:
Sounds good to me!

emdupre:
Right now it's linked in the text under the table for current BEPs -- would
you think that part should all be moved to its own section ? Or just this
guide ?

choldgraf:
Ahh I see it - IMO those could live in their own separate section so that they show up in the navbar on the left and are more discoverable that way. What if we:

1. Made a section for "Extending the BIDS specification" at the end and move the current text there.
2. Replace that text with a short description of the extension process and link out to the new page for more information.

emdupre:
I'd +1 having all of the BEPs (and the guide) as their own section from the navbar ! So that sounds good to me :+1:

emdupre:
It looks like the links added here aren't rendering correctly :(

https://bids-specification.readthedocs.io/en/latest/06-extensions.html

choldgraf:
ahhh I bet it's because of the line break that I added to try and not violate line length limits...

emdupre:
Should we re-open this issue, @chrisfilo ?

PR #97, opened 2018-11-27
adding extensions page

choldgraf:
closes https://github.com/bids-standard/bids-specification/issues/96

This is an attempt at splitting off the "contributing to BIDS / extensions" page into its own separate page so that the content is a bit more discoverable (originally mentioned in https://github.com/bids-standard/bids-specification/issues/96). @emdupre and @chrisfilo what do you think?


chrisfilo:
Thanks for doing this. LGTM.

src/06-extensions.md
```diff
@@ -0,0 +1,47 @@
+# Contributing to and extending the BIDS specification
+
+This page lists some ways that you can get involved with the BIDS community.
+
+## The BIDS Contributor guide
```
emdupre: I think here it would also make sense to point to and distinguish from the BIDS starter kit, since people can contribute to BIDS without proposing an extension. WDYT ?

choldgraf:
good call @emdupre - how's the latest push look?

emdupre:
LGTM, thanks ! :+1:

chrisfilo:
All that's left is to fix some linter errors: https://travis-ci.com/bids-standard/bids-specification/builds/92754960#L490

choldgraf:
fixed!

PR #98, opened 2018-11-28
[MRG] number of small corrections to the specification

robertoostenveld:
These are mostly formatting changes and fixes to external links.


src/02-common-principles.md
```diff
@@ -33,10 +33,10 @@ clarify them here.
     scanning sequence/protocol.
 
 1.  Data type - a functional group of different types of data. In BIDS we define
-    five data types: func (task based and resting state functional MRI), dwi
+    six data types: func (task based and resting state functional MRI), dwi
     (diffusion weighted imaging), fmap (field inhomogeneity mapping data such as
     field maps), anat (structural imaging such as T1, T2, etc.), meg
-    (magnetoencephalography).
+    (magnetoencephalography), beh (behavioural).
```
yarikoptic: while at it may be fix those "incorrect" spellings:
```
$> git grep behavioral
CHANGES.md:-   Added section on behavioral experiments with no accompanying MRI
src/04-modality-specific-files/05-behavioural-experiments.md:In addition to logs from behavioral experiments performed along imaging data
src/99-appendices/03-hed.md:HED is a controlled vocabulary of terms describing events in a behavioral
src/99-appendices/03-hed.md:all behavioral experiments. Each level of a hierarchical tag is delimited with
```
for consistency

src/02-common-principles.md
```diff
@@ -128,7 +128,7 @@ participant can exist only at participant level directory, i.e
 specific to a participant is to be declared only at top level of dataset for eg:
 `task-sist_bold.json` must be placed under `/dataset/task-sist_bold.json`
 
-Example 1: Two JSON files at same level that are applicable for NIfTI file.
+Example 1: Two JSON files that are erroneously at the same level.
```
yarikoptic: looking at example -- it is even more screwed up I guess (on purpose or not) -- it is `sub-01` but example .json files on top are for `sub-test`...  
text then doesn't make it clearer and also talks about `task-overtverbgeneration_run-2_bold.json` which is not there exactly (I feel that `sub-test_` was appended incorrectly at some point without due care)

chrisfilo:
Thanks for sending this along, FYI - here are the linter errors Travis spotted: https://travis-ci.com/bids-standard/bids-specification/builds/92812807

src/02-common-principles.md
robertoostenveld: Thanks, good point. There are now 7x "behavioral" and 9x "behavioural" in the whole spec  (including my edits). The slightly shorter behaviour and behavior also both happen.

I suppose you prefer the American English, I actually don't care (as long as it is consistent). Is there a formal preference for spelling?

src/02-common-principles.md
robertoostenveld: I agree that the naming error should not be part of it, the demonstration of the error should be on the inheritance only. I have corrected it.

yarikoptic:
FWIW +1 on formalizing to use American English, and mention that in common principles of it isn't there yet

yarikoptic:
@robertoostenveld would you have time to resolve outstanding issues causing CIs to fail?  I think we are ready otherwise and should merge!

yarikoptic:
@robertoostenveld - I have pushed a commit resolving introduced problems.  If would be all kosher - I would merge or we would be in jeopardy to introduce conflicts

robertoostenveld:
The Christmas period delayed work on this. I have updated to the HEAD of the master branch and resolved the conflicts. I see that circleci passes, but that travis fails: I will investigate.

robertoostenveld:
Whereas travis initially was showing "Incorrect list-item indent: add 2 spaces", after me changing it it to 2 spaces it was showing "Incorrect list-item indent: add 1 space". Also the table fence around the emojis seems impossible to get right. I am lost! 

Content wise it all seems fine to me, so I suggest to merge.

chrisfilo:
Consider using https://github.com/prettier/prettier to automatically fix syntax

robertoostenveld:
... so I managed to get all checks to pass!

@chrisfilo: the opinion of `prettier` regarding the table fence turns out not consistent with that implemented using remark-lint in Travis. I guess it is because the different emoji having different byte counts, and the tools not agreeing on how this is to be dealt with. 

Furthermore, remark-lint gives me `Error: Could not find module lint-list-item-indent`, both on the command line and in Atom.  So linting - although a nice idea - does not work for me (as a non computer scientist). I wonder how other neuroscientists are dealing with this, especially the ones for whom WordPad is the default tool for the editing of text files.

chrisfilo:
I agree, but this is a good topic for a new issue (perhaps proposing less strict linting).

chrisfilo:
@yarikoptic @robertoostenveld is this read to merge?

src/99-appendices/01-contributors.md
```diff
@@ -46,7 +46,6 @@ your name is not listed, please add it.
 -   Elizabeth DuPre 📖💡🔍🤔
 -   Eric A. Earl 🤔
 -   Anders Eklund 📖📢💻
--   Oscar Esteban 📖🔧🤔💬💻
```
sappelhoff: Why is this line deleted?

robertoostenveld: No idea why this line went missing, I have added Oscar back

robertoostenveld:
this is ready to merge.

chrisfilo:
@yarikoptic thanks.

This is minor, but please wait 5 days since the last commit before merging next time (as per https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules). At least until someone automates this (see https://github.com/bids-standard/bids-specification/issues/147).

yarikoptic:
Ok, noted

PR #99, opened 2018-11-28
ENH: First(?) good practice recommendation. No excessive overrides in Inheritance principle

yarikoptic:
It is often useful to look at the top level `task*.json` files to see what is
"common" among all subjects/sessions.  Listing only the fields with common
values is what we also implemented (albeit incorrectly, fix is pending
https://github.com/nipy/heudiconv/pull/279) in ReproIn.

Upon quick search I also found no "good practice" recommendations in the
spec.  It might be arguable either spec should or should not include them.
I think it should since we already RECOMMEND things.  We of cause could
place recommendations into a separate subsection somewhere but then they
would not be linked to the specific topics and thus would not be attended
to. Hence I placed this one right at the section where it is pertinent

This PR is largely an RFC ;-)

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -186,6 +186,14 @@ apply to all bold runs. However, if there is a key with different value in
 `sub-01/func/sub-01_task-xyz_acq-test1_run-1_bold.json`, the new value will be
 applicable for that particular run/task NIfTI file/s.
 
+### Good practice recommendations
+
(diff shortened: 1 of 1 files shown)


sappelhoff:
I agree that good practice recommendations are valuable for users! This particular PR looks good to me as well, it seems like an obvious recommendation we should make.

PR #100, opened 2018-11-28
Added contributor information

jgrethe:
Based on utilization of BIDS within ReproNim and testing of BIDS in other communities (SPARC). Also developed ReproNim BIDS tutorial presented at OHBM and SfN during ReproNim training course.


sappelhoff:
thanks for contributing!

Issue #3, opened 2018-09-28
Auto generate change log


chrisfilo:
Based on pull requests. Listed on github via releases.

franklin-feingold:
Starting to work on this - @teonbrooks @choldgraf may you have any recommendations to create this automated solution? Any suggestions will be greatly appreciated!

emdupre:
Have you already looked at [release-drafter bot](https://github.com/apps/release-drafter) ?

franklin-feingold:
Thank you for the recommendation! This looks like a good option. It appears that it will output onto the releases tag? The hope is that the merged PR's will be automatically added to the CHANGES.md, with the CHANGES.md linked in the readthedocs.

emdupre:
It _will_ output to the release notes ! You can see them as a draft before the release, looking something like this:

![screenshot from 2018-11-07 20-01-30](https://user-images.githubusercontent.com/15017191/48171009-ffa25c80-e2c7-11e8-9561-fe5e07dc913a.png)


One thought: It looks like you're building your RTD site continuously, rather than only at releases. If that's true, then my concern would be that a continuously updating changelog (by directly adding merged PRs to a CHANGES.md) could yield confusion as to what is in the "beta" specification as opposed to the official release.

I _think_ the solution used in other software such as `FMRIPREP` (and I know the solution we use in `tedana`) is to manually update the changelog at releases. But I'm not sure what your considerations are for minting a release, here !

emdupre:
Actually, that's a broader issue, then, if any merged PRs are going live on the specification as soon as they're merged. Maybe the docs *should* only be built at releases ? Or I might just have totally misunderstood the current situation, sorry !

emdupre:
Yes, I'd +1 having RTD build `stable` ! AFAICT that's not where we are currently, but it would make it much easier to keep track of changes between releases as PRs are merged in.

That being said, it won't solve your question @franklin-feingold, of having the CHANGES.md file automatically update as well. Manually pulling from the drafted release notes is one option, but there might be other, better ones !

franklin-feingold:
Thank you for your explanation @emdupre ! Very useful to know about that tool and this possible solution moving forward. I will continue looking for other possible solutions that may update directly into the CHANGES.md file.

PR #103, opened 2018-12-06
Autogenerate CHANGES.md

franklin-feingold:
PR will close #3 

This solution will collect the merged PR and put it under the unreleased section. It will also update the unreleased section to the new release number after the release has occurred when a new PR has been merged (the next one after the release PR, with that one now being under the unreleased section). The src/CHANGES.md is how it looks ran on this repo in the current state. 

The changelog will only generate and push when there is a merged PR (it will not push on a release PR or regular commits). A release PR will contain one change (changing the Unreleased title to x.x.x [the version number])

The changelog has also been added to mkdocs. Some of the specification documentation was updated to pass the updated .remarkrc file.

To complete the addition - a github token needs to be added to the CircleCI environmental variables named CHANGE_TOKEN. I can do that if this solution looks good before the merge.


.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+           working_directory: ~/build
+           command: |
+             if git log -1 --pretty=%s | grep Merge* ; then
+             if git log -1 --pretty=%b | grep release ; then
+             echo "New release, do nothing"
+             else
+             mv ~/build/tmp/CHANGES.md ~/build/src/CHANGES.md
+             git config credential.helper 'cache --timeout=120'
+             git config user.email "franklin.feingold@gmail.com"
+             git config user.name "franklin-feingold"
```
chrisfilo: Probably better to change it to something like "changelog bot".

.remarkrc
```diff
@@ -1,6 +1,9 @@
 {
   "plugins": [
     "preset-lint-markdown-style-guide",
-    ["lint-no-duplicate-headings", false]
+    ["lint-no-duplicate-headings", false],
+    ["lint-list-item-indent", "tab-size"],
+    ["lint-emphasis-marker", "consistent"],
+    ["lint-maximum-line-length", false]
```
chrisfilo: Why are those changed to the linter necessary for the changelog generation?

.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+      - setup_remote_docker:
+           version: 17.11.0-ce
+      - checkout
+      - run:
+           name: Build changelog
+           working_directory: ~/build
+           command: |
+             if git log -1 --pretty=%s | grep Merge* ; then
+             if git log -1 --pretty=%b | grep release ; then
+             echo "New release, do nothing"
```
chrisfilo: Is this check necessary? I wonder if there could be a situation where PR title would include the word "release" (with a different meaning).

.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+           npm install remark-cli@5.0.0 remark-lint@6.0.2 remark-preset-lint-recommended@3.0.2 remark-preset-lint-markdown-style-guide@2.1.2
+      - run: # remark the auto generated changes.md
+          name: remark on autogenerated CHANGES.md
+          command: |
+             cd ~/project
+             if git log -1 --pretty=%s | grep Merge* ; then
+             if git log -1 --pretty=%b | grep release ; then
+             echo "New release, do nothing"
+             mkdir ~/project/src/tmp
+             touch ~/project/src/tmp/empty.txt
```
chrisfilo: This seems identical to the "not a pr merge commit" path - logic could be simplified.

chrisfilo:
Great work! Thanks for putting this together!

.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+  search_build:
+    jobs:
+      - build
+      - github-changelog-generator
+      - remark:
+          requires:
+               - github-changelog-generator
+      - push:
+          requires:
+               - remark
```
effigies: Do you want to run this on every commit? I think limiting to `branch: master` will make more sense:

```suggestion
               - remark
           filters:
               branches:
                   only: master
```

This should probably apply to `github-changelog-generator` and `remark` as well.

.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+             if git log -1 --pretty=%b | grep release ; then
+             echo "New release, do nothing"
+             else
+             mv ~/build/tmp/CHANGES.md ~/build/src/CHANGES.md
+             git config credential.helper 'cache --timeout=120'
+             git config user.email "franklin.feingold@gmail.com"
+             git config user.name "franklin-feingold"
+             git add ~/build/src/CHANGES.md
+             git commit -m "autobuild CHANGES.md"
+             git push https://${CHANGE_TOKEN}@github.com/bids-standard/bids-specification.git master
```
effigies: I might suggest pushing to another branch than `master`, in case you want to modify before pushing to `master`. But if the plan is just to clean up later, then no big deal.

.circleci/config.yml
```diff
@@ -25,3 +25,124 @@ jobs:
...
+             if git log -1 --pretty=%s | grep Merge* ; then
+             if git log -1 --pretty=%b | grep release ; then
+             echo "New release, do nothing"
+             else
+             mv ~/build/tmp/CHANGES.md ~/build/src/CHANGES.md
+             git config credential.helper 'cache --timeout=120'
+             git config user.email "franklin.feingold@gmail.com"
+             git config user.name "franklin-feingold"
+             git add ~/build/src/CHANGES.md
+             git commit -m "autobuild CHANGES.md"
```
effigies: I would prefer a cleaner/more descriptive message. Something like `[DOC] Auto-generate changelog entry for PR #XYZ`.

.circleci/config.yml
franklin-feingold: This check is necessary to make sure that the changelog will only generate on merged PR's. It will skip for commits and release PRs. For a release PR title we could have a controlled vocabulary or set of words to denote only a new specification release ("New main specification release"? with the new version followed i.e. New main specification release 1.2.3). I agree there can be a situation where a PR title could include the word "release" in it and mean something else. 

The release tag could work. The git describe --tags outputs the `"latest release tag"-# of commits since release-g"(commit string)"` . This can be parsed down to get the latest release tag, but I noticed that the git tag --list will output whatever tag you provide it, so the test will be true for all cases. I may not see how the test will skip the changelog generation for a new release?

.circleci/config.yml
franklin-feingold: Across the two jobs it is a similar path. The jobs have to be separated because of the docker environment differences between the two jobs.

.circleci/config.yml
franklin-feingold: The cleaning up happens automatically so it should be all ready for the master branch (with no linter errors). The hope is that this is an automatic process not needing manual intervention. If the automation cannot be realized then pushing to another branch and then be PR'd over to master would work.

.remarkrc
franklin-feingold: Those changes are necessary for the auto generated changelog (after it was corrected through remark) to pass the linter style guide

.circleci/config.yml
franklin-feingold: I have applied this change in my recent commit

.circleci/config.yml
franklin-feingold: I have applied this update to my recent commit

.circleci/config.yml
effigies: I think he meant change the name of `user.name`.

.circleci/config.yml
effigies: Try in some project with tags (say 0.1.0):

```Bash
for REF in 0.1.0 master; do
    git checkout -q $REF
    DESC=$(git describe --tags)
    if [ "$(git tag --list $DESC)" = "$DESC" ]; then
        echo "We're on a tag: $DESC"
    else
        echo "Not a tag, after all: $DESC"
    fi
done
```

In Pybids, for example:

```
We're on tag: 0.1.0
Not a tag, after all: 0.6.5-229-gbd23eaa
```

This is because `git tag --list 0.6.5-229-gbd23eaa` doesn't match any tag, and produces an empty string.

effigies: Oh, I see. You plan to do release prep PRs. In which case, I would recommend using a template like "REL: <version>" for the PR name (or similar), and then looking for the "REL: " string in the merge commit. Otherwise you'll have to be vigilant about any PRs that have "release" somewhere in the title, which may show up innocently. "REL: " is much less likely to happen by accident.

chrisfilo:
RTM @franklin-feingold ?

chrisfilo:
Many thanks!

PR #114, opened 2018-12-20
Added AcquisitionArea as metadata field

ulrikehorn:
The field AcquisitionArea can be used to indicate what can be seen in the corresponding files, e.g. brain or parts of the brain, spinal cord, peripheral nerves, etc.
Guidelines for reporting fMRI studies (Poldrack et al 2008) also recommend using descriptions of the acquisition area when the images do not encompass the whole brain, the field AcquisitionAreaDescription can be used for that.


Remi-Gau:
I think that this is not the only type of suggested metadata from Poldrack 2008 and the COBIDAS report that could be added to BIDS. I have been wondering for a while if this worth it and if so, whether we should add each of them as separate PRs or try to pull them together in the same PR to be worked on.

chrisfilo:
@josator2 any thoughts?

sappelhoff:
Closing due to inactivity, feel free to re-open as an issue to discuss the overlaps with BEP025 and the [comment](https://github.com/bids-standard/bids-specification/pull/114#issuecomment-448951585) above

PR #109, opened 2018-12-16
[ENH] Derived (processed) MR data

chrisgorgo:
This PR culminates over two years of discussions and work on BIDS Derivatives. I derived it from our [RC1 Google Document](https://docs.google.com/document/d/17ebopupQxuRwp7U7TFvS6BH03ALJOgGHufxK8ToAvyI/edit#) with the following changes:

- I removed the `Space` JSON metadata field - it was never properly described and seemed redundant with `ReferenceMap`
- I opted not to include the transformation files specification in this PR. Without specifying what file format to use to store transformations it has limited use - hopefully, it could be added later in another PR.
- I replaced `desc-manual` restricted key/value pair with a JSON field (for segmentation)
- I limited the file formats for surface data to GIFTI or CIFTI (previously they were only recommended which lead to too much flexibility).
- I stipulated for `label` to use `Abbreviation` in the context of segmentation metadata (instead of index, context or name which was too flexible)
- I opted not to include `qform`/`sform` in tractography metadata in this PR since it seems not well described and quite controversial.
- Similarly, I could not find a proper description for `Param1`, `Param2` etc. so I also left them out.
- Limited allowed suffixes for the scalar derivatives.

Calling @effigies @satra @sgiavasis @oesteban @edickie @francopestilli @nicholst @tyarkoni @ahoopes for validation and comments. Please send suggestions via PRs to https://github.com/chrisfilo/bids-specification/tree/enh/derivatives.

I'll try to work on examples and the validator over the xmas break. Many thanks to everyone who contributed to this work.

PR #118, opened 2018-12-22
[WIP] adding first BEP010 spec

choldgraf:
This is an attempt at adapting the [BEP010 spec](https://docs.google.com/document/d/1qMUkoaXzRMlJuOcfTYNr3fTsrl4SewWjffjMD5Ew6GY/edit#) for this repository. There are probably a bunch of things to update, but it'd be helpful if I could get some high-level feedback on this one.

For example, I basically just tried to adapt word-for-word the google docs version. However, there's probably some redundant information with the base BIDS spec etc. Feel free to recommend sections to remove, move around, etc. Other than that, let me know what I should do. @chrisfilo do you have thoughts on this? Or can you ping somebody who went through this with the MEG spec?


chrisfilo:
I pointed out some obvious sections with redundancy, but there are plenty of others. Especially referring to BIDS-fMRI, BIDS-MRI, BIDS-iEEG is confusing - those should be replaced with links to specific sections of the spec. You should check out https://github.com/bids-standard/bids-specification/pull/108 and https://github.com/bids-standard/bids-specification/pull/109 for inspiration.

PS I don't quite remember who did the reformatting and merging of MEG.

robertoostenveld:
I think I was involved in merging of the MEG spec (although I was not the only one, I think Guio did most work). But that was taking snippets from a google doc and pasting them into another google doc (the main spec), which at that time was not yet so heavily visited and commented upon, so the main spec was a relatively clean document and the MEG changes quite overseeable.

I think that in this case we might want to have an intermediate version (which would become 1.1.2) to merge the current draft changes into the main spec. This is under the assumption that these are just formatting changes. And then we make a branch starting from that, which would become version 1.2 (including EEG and iEEG).

sappelhoff:
@robertoostenveld @choldgraf 

re: https://github.com/choldgraf/bids-specification/pull/2#issuecomment-453249039 and https://github.com/choldgraf/bids-specification/pull/2#issuecomment-453330478

Yes, you can close this one (#118) and add the contents as a PR to #108 ... I'll merge it and then we can integrate EEG and iEEG at once. I think this makes sense, although it will be a huge PR of course ...

so if you make the PR to #108, please make sure to rebase your changes from #118 off the #108 branch so that we do not run into trouble later on.

I will start a PR on the validator to remove the "optional flags" for EEG and iEEG simultaneously as well.

choldgraf:
ok I think I've got the PR to your branch ready to go:

https://github.com/sappelhoff/bids-specification/pull/5

Let me know how that looks and if I need to change anything up.

choldgraf:
(ping @sappelhoff in case you're not receiving notifications about this)

Issue #63, opened 2018-10-23
Adding CBV support 


TheChymera:
Hi @chrisfilo , thank you so much for moving development to Git. This has motivated me to finally get to adding CBV contrast support, as we had discussed on the Google docs page.

Would [this](https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md) be the adequate page to edit for my addition? Is there any model for documenting the CBV field?

chrisfilo:
No worries - thanks for your contributions.

Yes that would be the right place to add CBV.

TheChymera:
@chrisfilo I have trouble understanding the type/modality distinction in BIDS.

Are “CBV” / “BOLD” / “T2w” *modalities*, and “func”, “anat”, “dwi” types? Is there any point in the document where the names of these fields are clearly defined (wasn't able to find it)?

In MR parlance the former are actually known as *contrasts*, and I think using this term would definitively dispel any ambiguity with the “type” field. Is it ok if I refer to this field as the *contrast* field in my update?

chrisfilo:
Those two terms are blurry and I doubt we can untangle them without breaking backward compatible. The definition of type is here: https://github.com/bids-standard/bids-specification/blob/master/src/02-common-principles.md, but suffix takes different meaning from one type to another.

TheChymera:
Yep, that clears one thing up :) 

But **modality**, **suffix**, and **contrast** remain undefined defined there. Ideally, going forward, I would strongly prefer to call the contrasts **contrasts**, if you oppose this I would go for **modalities** second and **suffix** last (suffix isn't really descriptive here, might be confused by casual users with the extension, and will certainly cause even more confusion if the statistic type is added as a suffix to statistic derivatives).

TheChymera:
If you accept the commit, I'd follow it up with a section better documenting the “image contrast” field. Contextualizing it as such would not just help disambiguate the nomenclature, but also clarify for users the distinction between type (high-level categorization), contrast (measure of interest which the voxel intensities provide information on), sequence (the technical instructions which allow the measurement of the contrast), and contrast enhancer (chemical substances which aid the sequence in increasing the contrast).

chrisfilo:
Closing this since https://github.com/bids-standard/bids-specification/pull/82 was merged.

PR #119, opened 2018-12-26
Documenting the contrast suffix and disambiguating it from the catch-all modality categorization

TheChymera:
Renaming and clarifying the scope, utility, and usage of the suffix field hitherto ambiguously named “modality” — as [discussed previously](https://github.com/bids-standard/bids-specification/issues/63#issuecomment-435168888).

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -303,6 +303,30 @@ sub- prefix is not part of the subject label, but must be included in file names
 MUST be integers. Those labels MAY include zero padding, but this is NOT
 RECOMMENDED to maintain their uniqueness.
 
+## The contrast suffix
+
+In addition to the `*<label_identifier>-<custom_label>*` pairs (e.g. `sub-001`),
+it is RECOMMENDED that file names contain a suffix of the form `*_<suffix>.`,
+which directly precedes the file format delimiter. The string values of this
+suffix indicate  the contrast of the file in question.
+
+This specification is often needed, as the NIfTI header does not allow us to
+specify what unit or contrast the voxel values correspond to. The decision of
+BIDS to track this information in the file name rather than in the JSON sidecar
+file, is based on (a) the high relevance of this field for rapid file selection
+(b) the low number but large scope of brain imaging contrasts. For non-NIfTI
+files, this suffix provides more user-friendly recognition of the feature
+documented by the file. A file ending in `_events.tsv` is more unambiguously
+understood as documenting the events than if this information had to be inferred
+purely from the `.tsv` format.
+
+It is particularly noteworthy that there can only be one contrast suffix per file,
+as the voxel values can only correspond to one contrast. This field is not
+designed to track historical information, and the suffix `_cbv_tstat.*` is
+nonsensical. If multiple contrasts (e.g. `_cbv.*` and `_bold.*`) are acquired in
+one session, the corresponding files MUST be differentiated by other labels
+(e.g. `*run-<label>`).
+
 ## Units
 
 All units SHOULD be specified as per International System of Units (abbreviated
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -112,14 +112,14 @@ Template:
 ```Text
 sub-<label>/[ses-<label>/]
     anat/
-        sub-<label>[_ses-<label>][_acq-<label>][_ce-<label>][_rec-<label>][_run-<index>]_<modality_label>.nii[.gz]
+        sub-<label>[_ses-<label>][_acq-<label>][_ce-<label>][_rec-<label>][_run-<index>]_<contrast_label>.nii[.gz]
         sub-<label>[_ses-<label>][_acq-<label>][_ce-<label>][_rec-<label>][_run-<index>][_mod-<label>]_defacemask.nii[.gz]
 ```
 
 Anatomical (structural) data acquired for that participant. Currently supported
-modalities include:
+contrast include:
 
-| Name               | modality_label | Description                                                                                                                                                                                                                             |
+| Name               | contrast_label | Description                                                                                                                                                                                                                             |
 | :----------------- | :------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | T1 weighted        | T1w            |                                                                                                                                                                                                                                         |
 | T2 weighted        | T2w            |                                                                                                                                                                                                                                         |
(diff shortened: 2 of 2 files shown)


src/02-common-principles.md
```diff
@@ -303,6 +303,30 @@ sub- prefix is not part of the subject label, but must be included in file names
...
+documented by the file. A file ending in `_events.tsv` is more unambiguously
+understood as documenting the events than if this information had to be inferred
+purely from the `.tsv` format.
+
+It is particularly noteworthy that there can only be one contrast suffix per file,
+as the voxel values can only correspond to one contrast. This field is not
+designed to track historical information, and the suffix `_cbv_tstat.*` is
+nonsensical. If multiple contrasts (e.g. `_cbv.*` and `_bold.*`) are acquired in
+one session, the corresponding files MUST be differentiated by other labels
+(e.g. `*run-<label>`).
```
chrisfilo: This new rule breaks a lot of existing datasets and I am not convinced it is necessary. I would recommend removing it.

src/02-common-principles.md
```diff
@@ -303,6 +303,30 @@ sub- prefix is not part of the subject label, but must be included in file names
 MUST be integers. Those labels MAY include zero padding, but this is NOT
 RECOMMENDED to maintain their uniqueness.
 
+## The contrast suffix
```
chrisfilo: I don't think the term `contrast` applies to all data types covered by BIDS (think about MEG, fieldmaps, physiological recordings etc.). I think this clarification fits better into anatomical MR section.

src/02-common-principles.md
TheChymera: How then would you adapt the suffix in the described use case?

src/02-common-principles.md
TheChymera: The term contrast should apply equally well to any multidimensional array of data. Perhaps I should clarify that each image intrinsically has a contrast and that this does not imply the administration of a “contrast agent” (which is simply an experimental manipulation which increases the dynamic range of the contrast which the image has either way)? Is this what you were thinking of?

For fieldmaps the contrast is the b0 strength, for physiological recordings it's motion, or heart rate, etc. The only place where I see an inconsistency is with tabular data representation such as for the event files, which I attempted to clarify.

src/02-common-principles.md
chrisfilo: Let me clarify - right now it's allowed to have `sub-01/anat/sub-01_T1w.nii.gz` and `sub-01/anat/sub-01_T2w.nii.gz` in the same dataset. The rule you added in this PR would require adding `run` or other keywords to those filenames. Was this your intention?

TheChymera: I see, my concern was the following. Let's say you have:

```
sub-01/anat/sub-01_bold.nii.gz
sub-01/anat/sub-01_cbv.nii.gz
```

And you perform some statistic, meaning that the values in your NIfTI will no longer code for bold or cbv but for a t-statistic. Concatenating the fileds would be confusing since (unlike subject, session, etc.) they don't have a prefix label to identify what they are. Furthermore, only the last one would actually specify what the values are representing. A t-statistic is a t-statistic regardless of the nature of the data it was applied to.

So you would need the uniquely identified run:

```
sub-01/anat/sub-01_run-0_bold.nii.gz
sub-01/anat/sub-01_run-1_cbv.nii.gz
```

to arrive at an unambiguous statistic representation:

```
sub-01/anat/sub-01_run-0_tstat.nii.gz
sub-01/anat/sub-01_run-1_tstat.nii.gz
```

chrisfilo: I see your concern now. This filename conflict on the level of derivatives, can be resolved by a combination of `desc` filename keyword and `RawSource` JSON keyword. See https://github.com/bids-standard/bids-specification/pull/109 for details.

src/02-common-principles.md
chrisfilo: I only listed some and there are many others. They are all described through the spec - you can look them up. I agree that it would've been nice to have a coherent definition of the meaning of suffix, but I don't think this is currently possible without breaking backward compatibility. At the end of the day, suffix is an “ad-hoc identifier which helps us select the files” - not necessarily by design, but we have to live with this now.

src/02-common-principles.md
```diff
@@ -303,6 +303,30 @@ sub- prefix is not part of the subject label, but must be included in file names
...
+
+In addition to the `*<label_identifier>-<custom_label>*` pairs (e.g. `sub-001`),
+it is RECOMMENDED that file names contain a suffix of the form `*_<suffix>.`,
+which directly precedes the file format delimiter. The string values of this
```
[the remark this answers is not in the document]
chrisfilo: Each data type has separate specific file naming rules. As far as I know there are no data types that allow something along the lines of `<key>-<value>.<ext>`.

src/02-common-principles.md
TheChymera: hm, I get that you probably put a lot of thought in to this, but `desc` sounds like yet another ad-hoc field which can contain whatever and be overall confusing and limit reusability of workflows because there's no way you can use this key and predict what it might hold. Sounds like mitigating the suffix problem by duplicating it :( Are you totally sold on this?

src/02-common-principles.md
TheChymera: I understand your point. Ok, so if there is no unifying feature about this field other than being a suffix, we can perhaps just rename it from “modality” to just that, “suffix”?

src/02-common-principles.md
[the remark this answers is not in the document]
TheChymera: ok, so there is always a suffix, corresponding to no paritcular type of terms. But in the case of NIfTI files specifically, we can comment that the suffix specifies the contrast?

src/02-common-principles.md
chrisfilo: This discussion would be better placed on https://github.com/bids-standard/bids-specification/pull/109

src/02-common-principles.md
chrisfilo: Sure!

src/02-common-principles.md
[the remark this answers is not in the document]
chrisfilo: Perhaps, but `phase1`, `phase1`, `magnitude1`, `defacemask` and other weird suffixes are being used for nifti files. Would "contrast" still cover them?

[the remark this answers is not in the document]
TheChymera: yes, they would, though `defacemask` --- while it can certainly be conceptualized as a contrast --- is named in a way which does not inspire the concept of contrast. Just to give you an example, something along the lines of `anonymousanatomyregion` would do that. Still, unfortunate naming at most would require a sentence of clarificantion, it is in the end a feature which is being contrasted.

src/02-common-principles.md
TheChymera: so you would be ok with me synthetizing the discussion we had here about the intrinsic variability of the `_suffix` and its role primarily as a disambiguator for cursory users, but adding a paragraph on how for NIfTI files this is equivalent to the contrast, and how the contrast in this sense is separate from both the “contrast agent” and the “acquisition” (which both have their own fields). y?

chrisfilo: Go ahead - however, I am not convinced this is the most pressing issue to work on and I am worried that the description of `suffix` with its exceptions will be of limited use.

Working on a PR adding support for CBV data might be a better investment of your time and skills.

TheChymera: I thought that was already merged?

chrisfilo: indeed it was - my bad!

sappelhoff:
I agree with Chris' opinion that `contrast` is not a good replacement for `modality` in parts of the proposed changes. Feel free to move the discussion to #109, I am closing this PR because of disagreement and lack of participation from other members to find a consensus.

PR #121, opened 2019-01-09
REL: v.1.1.2

chrisfilo:
There are no major changes in this release. It serves two purposes:
- testing the release mechanism (especially RTD component contributed by @choldgraf)
- creating a minor release prior to EEG and iEEG merges

### DIFF
--- a/src/CHANGES.md
+++ b/src/CHANGES.md
@@ -1,6 +1,6 @@
 Changelog
 
-## [Unreleased](https://github.com/bids-standard/bids-specification/tree/HEAD)
+## [1.1.2](https://github.com/bids-standard/bids-specification/tree/1.1.2)
 
 -   Update 01-contributors.md [#120](https://github.com/bids-standard/bids-specification/pull/120) ([oesteban](https://github.com/oesteban))
 -   Global fields in data dictionaries [#117](https://github.com/bids-standard/bids-specification/pull/117) ([chrisfilo](https://github.com/chrisfilo))


src/CHANGES.md
```diff
@@ -1,6 +1,6 @@
 Changelog
 
-## [Unreleased](https://github.com/bids-standard/bids-specification/tree/HEAD)
+## [1.1.2](https://github.com/bids-standard/bids-specification/tree/1.1.2)
```
choldgraf:
```suggestion
## [Unreleased](https://github.com/bids-standard/bids-specification/tree/HEAD)

## [1.1.2](https://github.com/bids-standard/bids-specification/tree/1.1.2)
```

choldgraf:
Only one nitpick, which is to keep an "unreleased" section :-)

src/CHANGES.md
chrisfilo: @franklin-feingold would this change be preserved by the automatic changelog scripts?

chrisfilo: Actually, I don't think this will work. @choldgraf the CI scripts will add the "unreleased" section after this is merged.

chrisfilo: See the protocol: https://github.com/bids-standard/bids-specification/blob/master/Release_Protocol.md

choldgraf: sounds good, +1 to merge from me then

choldgraf:
seems that folks are happy w/ the changes so let's see how this goes, wheeeee :shipit:

src/CHANGES.md
franklin-feingold: This change would not be preserved by the automatic changelog scripts. There will be an unreleased section as soon as another PR is merged after the release (this PR will not trigger an unreleased section generation, the next one will). 

The release appears to have worked! My changelog generator scripts didn't run.

franklin-feingold:
Though, there are two last minor things is to finish the release: create a github release and add today's date to the CHANGES.md (the date will be automatically added after the next merged PR). 

I created the release (https://github.com/bids-standard/bids-specification/releases/tag/v.1.1.2) and committed today's date to the CHANGES.md (https://github.com/bids-standard/bids-specification/commit/5e2b11304ecfaa98cc72ed992b87e3261e45b08f)

chrisfilo:
I know that we do not have the decision rules finalized, but we should leave much more time (at least 5 working days) between proposing to make a new release and actually doing it. This allows the community to chime in and potentially prevent releasing something incorrect that we will have to support in the future.

choldgraf:
My bad - I thought we wanted to get this merged quickly to test whether the release mechanism worked

PR #126, opened 2019-01-11
Update to Release_Protocol.md

franklin-feingold:
Updated figures and added 'v' in front of the version number


franklin-feingold:
I think for releases that can be good and more useful pointing to the rendering because more than likely the reader will not need to see the github repo in that state, just the specification (if so one may navigate to it on the browser, it does not need to be the default). I have reflected that in my recent commit. (Only the first one has the '.' between v and version, for the future we'll follow github convention, i.e. v1.1.2)

Issue #134, opened 2019-01-24
The readthedocs rendering of the spec should advertise the version


sappelhoff:
[https://bids-specification.readthedocs.io/en/stable/](https://bids-specification.readthedocs.io/en/stable/) does not advertise its version on the main landing page, we should probably change this.

Perhaps display it somewhere in this corner?

![image](https://user-images.githubusercontent.com/9084751/51684546-91c72080-1fec-11e9-98a7-bc7b6dc0d671.png)

sappelhoff:
> Perhaps it could go into the text as a third paragraph indicating the latest stable release? (under the get started paragraph)

yes, that would be an easy to do alternative. But in the corner would be preferable I think, because it can be seen regardless of the sub-page a reader might currently be on.

franklin-feingold:
We are able to edit the corner header by changing the site_name in the [mkdoc.yml](https://github.com/bids-standard/bids-specification/blob/master/mkdocs.yml). However, it is not immediately clear how this permeates across stable and latest releases (I think to both, but further simulations are needed to confirm my suspicion). Regardless though I think latest and stable will be part of the same version because stable takes the latest release and latest is current to master. Master should not be version(s) ahead (i.e. stable at v1.1.2 and master at v1.1.3 without a release).

If we do find there are different versions across necessitating more dynamic version control site_name, It appears this has been discussed over at [mkdocs since 2014](https://github.com/mkdocs/mkdocs/issues/193). There appears to be a solution called [mike](https://pypi.org/project/mike/) that could handle this exact versioning issue.

sappelhoff:
> Regardless though I think latest and stable will be part of the same version because stable takes the latest release and latest is current to master. Master should not be version(s) ahead (i.e. stable at v1.1.2 and master at v1.1.3 without a release).

yes, master should never be a version ahead of the latest release in my opinion. I think of it as:
- stable 1.1.2
- master 1.1.2-dev

i.e., only adding `-dev` to the stable version to indicate that this is the latest version.

would be cool if we could add the `-dev` suffix to the version for the "latest" rendering of readthedocs.

franklin-feingold:
That sounds good to me. I think this all can be integrated and be a part of the Release Protocol. I will start putting together how that would look and submit a PR.

sappelhoff:
@franklin-feingold since #137 has been merged, this issue can be closed - correct? If I understand it correctly, we will have the version advertised when making a new release?

[...]

PR #137, opened 2019-01-26
Update site_name and release protocol

franklin-feingold:
Addressing https://github.com/bids-standard/bids-specification/issues/134

This will have the readthedocs specification indicate what version it is at. This is adding to our release protocol. For latest we will be in vX.X.X-dev and when we are ready for a release (stable) the mkdocs.yml site_name will change to reflect the new version. Then after the github release, submit a PR to bring the site_name back to dev, because stable will have captured the new version, but we will be back in dev at that moment.


sappelhoff:
@franklin-feingold is this ready for review? Perhaps some other people aside from the two of us should have a look at it?

franklin-feingold:
Yup! This is ready for review. Yeah, I think once Chris reviews and see his thoughts, that could be sufficient

PR #151, opened 2019-02-15
RFC: represent units in ascii

satra:
This is a draft request for comments.

In a former discussion on the google doc, we had considered representing all units in ascii notation.

The ASCII representation is taken from 
[here](http://people.csail.mit.edu/jaffer/MIXF/CMIXF-12), where additional units 
may be found. The primary reason for choosing an ASCII representation, is that 
certain characters can have multiple unicode representations that look very 
similar. (e.g., https://graphemica.com/search?q=mu)


sappelhoff:
I have no objection against including the ascii representation of units, but I would not have them as a replacement to the unicode units (this would also be a backwards compatibility breaking change I suppose).

I suggest having both representations side by side, so that each user can choose which to use.

We should introduce a test in the bids-validator that checks for the correctness of units, i.e., compares the unit to the set of prefixes, names, and suffixes we accept

effigies:
I for one am content with any of the following:

1) ASCII following the CMIXF12
2) Unicode with explicit code points specified, copy-pasteable from any generated spec documents, and the validator checking these
3) Both

Probably would prefer (1), just because I don't want to write unicode that doesn't appear on my keyboard. On the other hand, we have been specifying Unicode characters, so we may end up invalidating compliant datasets. Which would lead us to (3).

chrisfilo:
I agree with the backward compatibility issue raised by @sappelhoff.

satra:
@chrisfilo - in general, how are we dealing with backwards compatibility? aren't many changes in upcoming 2.0 not backwards compatible? since bids adopts a validation driven approach, perhaps the validator can indicate that a dataset is compatible with bids version x -- y, z and a linter could suggest what to change to expand/change compatibility.

i'm not a big fan of 3 (supporting both), since that piece of logic now has to go into every software. it's always possible for software to detect version of bids dataset and do something or not do something. but having lived through six and future, and many `if sys.version < (3, )` and all the fun unicode things, i'm strongly in favor of dropping unicode characters in units. 

in this particular case i would say the friction this causes (adopting 1) is probably worth the headaches it reduces later, and the simplicity we gain.

chrisfilo:
> @chrisfilo - in general, how are we dealing with backwards compatibility?

Current master is 1.x and strives to maintain backward compatibility. Ideas/suggestions/proposals that would break backward compatibility are collected in the [BIDS 2.0 document](https://docs.google.com/document/d/1LEgsMiisGDe1Gv-hBp1EcLmoz7AlKj6VYULUgDD3Zdw/edit).

> aren't many changes in upcoming 2.0 not backwards compatible?

Yes - by definition.

> since bids adopts a validation driven approach, perhaps the validator can indicate that a dataset is compatible with bids version x -- y, z and a linter could suggest what to change to expand/change compatibility.

Indeed that would be technically possible, but I expect that developing and maintaining such solution would incure a substantial effort. This is probably not the best place to discuss such approach though (since this PR will eventually be merged and forgotten). If you are interested in this topic perhaps you can open a new issue?

> in this particular case i would say the friction this causes (adopting 1) is probably worth the headaches it reduces later, and the simplicity we gain.

Such arguments have been successful in the past (for example with hotfixes to MEG spec). It is really the question of which datasets and tools this change would impact.

sappelhoff:
Some progress on this matter, I opened an issue on validating units, see: https://github.com/bids-standard/bids-validator/issues/777

Specifically, we would want the bids-validator to enforce the units we are deciding on. 

I also cross-post @ftadel's opinion here from a discussion in https://github.com/bids-standard/bids-examples/pull/155#discussion_r281095667 

> It's probably better to accept less options for units formatting. If we accept too many variations, the programs processing automatically BIDS datasets might miss the interpretation of some of the extra options. This is "useless" freedom (who cares if the file has to use microV instead of uV?) might make development/maintenance/debugging more complicated.
But the validator should enforce this properly then.

satra:
@sappelhoff - this was the PR/discussion on units. perhaps we can continue here instead of https://github.com/bids-standard/bids-examples/pull/175:

fyi: https://www.unicode.org/Public/security/revision-05/confusables.txt

across the bids domains it's more likely that units will come up in combos. so i think deciding on what the validator should do would be good.

sappelhoff:
@satra after reading @effigies comment: https://github.com/bids-standard/bids-examples/pull/175#issuecomment-580323160 and further discussing with him in person I am now of the impression that we should go ahead with your PR (that you proposed almost a year ago :wink:  )

- obviously we stay with SI units for BIDS
- however, we now clarify how we want them to be formatted
- namely:
    - either exactly like in out Appendix currently (i.e., permit unicode characters U+00B5 (µ) and U+00B0 (°) and U+2126 (Ω)... and ONLY those)
    - or according to this document: http://people.csail.mit.edu/jaffer/MIXF/CMIXF-12

For this PR that would mean concretely that the intro pragraph would have to be rephrased to account for the three unicode characters we permit ... and to change their table entries back to unicode.

This change in the spec would involve changes with the validator, which actually refreshed this whole discussion through bringing up https://github.com/bids-standard/bids-examples/pull/175 cc @rwblair

sappelhoff:
okay, seems like merging https://github.com/bids-standard/bids-specification/pull/408 blew up this PR ... :-/ 

At this point it seems like picking the changes from https://github.com/bids-standard/bids-specification/pull/151/commits/bdb3cd5eed7ba19a724a782195462f0841cb2ebc and https://github.com/bids-standard/bids-specification/pull/151/commits/9f22dc9f65c76f7f4d93467264775580d6828b78  and opening a new PR would be cleanest.

I'll just go ahead and do that.

sappelhoff:
please see https://github.com/bids-standard/bids-specification/pull/411 and continue discussion there.

PR #155, opened 2019-02-19
Fix example misalignment

### LINKED ISSUE #154: File hierarchy is unclear : location of “raw data” / “BIDS-managed data”?
I may be wrong  but the standard is unclear or not clear enough about the location of NIfTI “raw data” or “BIDS-managed data”.

I understand:
* We store source data such as DICOM files in subfolder `sourcedata`.
* We store processed data such as normalized NIfTI files in `derivatives`.
This is explained in [Common principles | Source vs. raw vs. derived data](https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#source-vs-raw-vs-derived-data):
> BIDS in its current form is designed to harmonize and describe raw (unprocessed or minimally processed due to file format conversion) data. During analysis such data will be transformed and partial as well as final results will be saved. Derivatives of the raw data (other than products of DICOM to NIfTI conversion) MUST be kept separate from the raw data. This way one can protect the raw data from accidental changes by file permissions. In addition it is easy to distinguish partial results from the raw data and share the latter. Similar rules apply to source data which is defined as data before harmonization and/or file format conversion (for example E-Prime event logs or DICOM files).

However it is unclear where we should store the `sub-<label>` subfolders with the raw NIfTI files. Directly under the root folder of the dataset?

If so, [Directory structure | Single session example](https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#single-session-example) is misaligned and thus misleading:
```
sub-control01/
    anat/
    .
    .
    .
    code/
        deface.py
    derivatives/
    README
    participants.tsv
    dataset_description.json
    CHANGES
```

DimitriPapadopoulos:
Partially fixes #154. This fixes the example. However I believe a more “theoretical” discussion is required, explaining that raw data are stored directly under the dataset root folder, the same way source data are stored in subfolder `sourcedata` and derived data in subfolder `derivatives`.

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -370,13 +370,13 @@ sub-control01/
         sub-control01_magnitude1.nii.gz
         sub-control01_scans.tsv
 
-    code/
-        deface.py
-    derivatives/
-    README
-    participants.tsv
-    dataset_description.json
-    CHANGES
+code/
+    deface.py
+derivatives/
+README
+participants.tsv
+dataset_description.json
+CHANGES
 ```
 
 Additional files and folders containing raw data may be added as needed for


Remi-Gau:
I had made a figure (when this whole thing was on google doc) to help explain how `raw` / `derivatives` / `source` relate to one another. It is now sitting on a [PR / issue](https://github.com/bids-standard/bids-starter-kit/issues/105) on the bids starter kit repo.  

I think that I had in mind the idea of putting in the BIDS specs but never got around to to it.

I think @chrisfilo had made some comments on the figure being too "specific" and might lead people to assume that the content of `source` and `derivative` had to the one described in the figure.

I am happy to change it if there is interest in adding this to the specs.

DimitriPapadopoulos:
Indeed I cannot find in the standard a clear statement about the location of **raw data**. As a result colleagues of mine believe you _may_ put raw data in a subfolder of the root folder such as `rawdata`.

 - I do see that **source data** belong to `sourcedata` and **derived data** to `derivatives`. Fixing the example should clarify a little bit.

I would also like to add a `sourcedata` subfolder to the example to complete it. Would you agree to adding `sourcedata`?

Perhaps we don't need to show the contents of the subfolder though. Should we show the contents of `sourcedata` and other subfolders as in the above figure? I am all for it. This is an **example** after all!

Finally what about adding bow-drawing characters to this example - like the `tree` command? Something like this:
```
├── sub-control01
│   ├──  anat
.    .
.    .
.    .
├──  code
│   └──  deface.py
├── derivatives
├── README
├── CHANGES
├── participants.tsv
└── dataset_description.json
```
GitHub doesn't render it nicely but perhaps _Read the Docs_ does. Perhaps a real figure like in #105 is preferable.

chrisfilo:
Let's merge this as is since it clearly solves an existing problem (alignment).

I totally agree that we need more clarification on where source vs raw data should go. @Remi-Gau I think your plots were very close and maybe we should bring them back. I also like the idea of listing `sourcedata` and `derivatives` folders without contents.

@DimitriPapadopoulos consider also sending a PR adding yourself to https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md and https://github.com/bids-standard/bids-specification/blob/master/CODEOWNERS

PR #158, opened 2019-02-20
Get rid of broken remark lint rule

DimitriPapadopoulos:


DimitriPapadopoulos:
From the _remark-lint-table-pipe-alignment_ [readme](https://github.com/remarkjs/remark-lint/blob/master/packages/remark-lint-table-pipe-alignment/readme.md):
> Aligning cells perfectly is impossible as some characters (such as emoji or Chinese characters) are rendered differently in different browsers, terminals, and editors. You can pass your own [`stringLength`](https://github.com/remarkjs/remark/tree/master/packages/remark-stringify#optionsstringlength) function to customize how cells are aligned. In which case this rule must be turned off.
To me this rule is by essence poorly written/defined. It is better to manually align tables (supposing all characcters are 1-character-wide) than having this rule push misalignments that look bad in most text editors.

sappelhoff:
> It is better to manually align tables (supposing all characters are 1-character-wide) than having this rule push misalignments that look bad in most text editors.

I have also experienced cases in which this rule did not work as expected (some unicode / emoji / other byte-value characters) - but as @chrisfilo points out, I have experienced many more cases where the rule has alerted me to fixing my tables.

I don't think that your suggestion of asking people to "manually align tables" will work without automatic enforcement. What was your idea to make sure everybody is aligning their tables?

DimitriPapadopoulos:
The rule enforces a misaligned table in the one file I had to modify. I probably don't see the big picture.

sappelhoff:
I suggest to close this based on what was said before. Would you be okay with that @DimitriPapadopoulos? Or do you want to make a new argument?

DimitriPapadopoulos:
I'm in favour of removing broken automatic tools that enforce wrong rules and pushing contributors to do the right thing  manually - until automatic tools are fixed. But I see your point. I'll close this issue.

Instead I'll focus on fixing _remark-lint-table-pipe-alignment_ and upstream dependencies, or directly in _bids-standard_. I think we need to override the default [_lengthNoop()_](https://github.com/wooorm/markdown-table/blob/b3929dc/index.js#L237-L239) function.

sappelhoff:
> Instead I'll focus on fixing remark-lint-table-pipe-alignment and upstream dependencies, or directly in bids-standard

That would be valuable indeed!

Issue #150, opened 2019-02-14
PR titles


franklin-feingold:
@bids-standard/everyone 

Hi BIDS Community!

Regarding the PR titles - when opening a PR, please have the PR title provide context to what is being merged. These titles will appear in our [changelog](https://bids-specification.readthedocs.io/en/latest/CHANGES.html) automatically (when the PR has been merged). 

Thank you!

effigies:
As examples, it would probably be good to look at [pregh-changes.md](https://github.com/bids-standard/bids-specification/blob/master/src/pregh-changes.md) to get a sense of the level of description that's useful.

It would also be good to have some kind of tag, like `[minor]` or `[nospec]` that we can put in titles for things that don't affect the semantics of the spec, like modifying the code that performs the rendering or continuous integration. (That said, we might still want those PRs to be credited in the [release](https://github.com/bids-standard/bids-specification/releases) blurbs.)

sappelhoff:
How about tags like [MRG] in this regard? Perhaps it would be nicer to have this removed in the future --> especially with the new github feature of internally tagging a PR as WIP or MRG without having to convey it in the title: https://github.blog/2019-02-14-introducing-draft-pull-requests/

effigies:
Yes, I personally would want the titles at merge time to be exactly what we want in the changelog, so tags like `[MRG]` and `[WIP]` are definitely no good.

Though we may also be able to automatically filter those out. If we can filter, perhaps a better approach would be to have `[SPEC]` in entries we actively *want* in the spec changelog. If that's the case, dropping `[*]` would be an easy addition, so any tags could continue to be used.

edickie:
Yeah, I think I (and other novices) are going to need some more guidance
about this. Can we add a "Naming your pull request" section to the
CONTRIBUTING.md file?  In this section can we have:


   1. a table of tags that are recommended and what they mean (I still
   haven't figured out what [MRG] stands for..)
   2. Some examples of commit messages and how they will render to the
   change log?

effigies:
Yes, we should definitely have docs for describing these and giving advice, once we settle on the process.

I'm not sure what all bracket tags will make sense with a document repository instead of a software repository, but `[WIP]` and `[MRG]` are "Work in Progress" and "Merge" as in "Ready-to-merge". These conventions seem to float freely between initialisms and severe contractions.

[...]

PR #161, opened 2019-02-25
REL: v1.2.0

chrisfilo:
After months of reviews, EEG and iEEG support was merged. This seems like a good time to release.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,4 +1,4 @@
-site_name: Brain Imaging Data Structure v1.1.2-dev
+site_name: Brain Imaging Data Structure v1.2.0
 theme: material
 plugins:
     - search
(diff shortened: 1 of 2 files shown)


KirstieJane:
I find the [CHANGES.md](https://github.com/chrisfilo/bids-specification/blob/rel/1.2.0/src/CHANGES.md) file difficult to follow - I think a human readable version more similar to those of v1.1.1 and lower (v1.1.0, v1.0.2 etc) would be really valuable - as discussed (but not completed) in #150.

@nicholst - did you make any progress on drafting a process for a release?

As a reminder, the decision making process is the same for a release as for any other decision, so it doesn't require any additional hoops even though - in my opinion - it carries much more weight than most other PR merges.

KirstieJane:
Commented and then decided against it. Congratulations on the next release.

sappelhoff:
@chrisfilo once this release is done, could you also make a release of the `bids-validator` so that EEG/iEEG work out of the box?

chrisfilo:
Yes. That's the plan!

sappelhoff:
According to the rules this could be merged now. Are we waiting for something @chrisfilo ?

PR #162, opened 2019-03-04
[FIX] Removing a leftover file

chrisfilo:
This file was moved tp `04-modality-specific-files/07-behavioral-experiments.md` and is no longer referenced.


sappelhoff:
Thanks @chrisfilo!

PR #178, opened 2019-03-15
[ENH] Alternative folder organization for raw, derived, and source data

### LINKED ISSUE #159: Location of raw data: rawdata subfolder?
_BIDS_ is centred around **raw data**. This probably used to make sense but nowadays I see more and more users working with **derived data** - perhaps thanks to _BIDS_ and other efforts to standardize raw data and code. Many users in my lab disagree with hiding derived data in `derivatives` and storing **raw data** directly under the root folder of the project.

In addition to the imbalance between the different categories of data (_source_, _raw_ and _derived_) important files such as `README` or `participants.tsv` are lost alongside `sub-<label>` subfolders. In projects with thousands of subjects stored on an NFS share a mere `ls` in the root folder may require many seconds.

Would you consider an **option** to store raw data in a `rawdata` subfolder instead of the root folder?

chrisfilo:
Closes #159.


chrisfilo:
This PR is much more likely to be merged faster and thus I would prefer to tackle your question in https://github.com/bids-standard/bids-specification/pull/109 after this PR (#178) is merged.

PR #180, opened 2019-03-18
[FIX] make section headings and ToC consistent between meg and eeg specs

robertoostenveld:
I removed the `space-<xxx>` from the filename in the section headings, it is still part of the actual specification. 

This should make the following headings and table of contents more similar
- https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html
- https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html
- https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html


chrisfilo:
thanks!

PR #185, opened 2019-03-21
[FIX] change file for definition of electrical stimulation labels from _electrodes.json to _events.json

### LINKED ISSUE #184: Description of intracranial stimulation parameters in *events.json
On the [specification](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#electrical-stimulation) it says that the description of the electrical_stimlation_\<label> columns should be placed in the *_electrodes.json file.

_Electrical stimulation parameters can be described in columns called electrical\_stimulation\_\<label>, with labels chosen by the researcher and optionally defined in more detail in an accompanying \_electrodes.json file (as per the main BIDS spec)._

 Shouldn't it be placed in *_events.json instead?

ezemikulan:
closes #184 

The text mentioned that the description of the electrical\_stimulation\_\<label> columns had to be placed on \_electrodes.json, but it should be placed on \_events.json. 

This PR fixes the issue.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -11,4 +11,5 @@
 # Individual sections
 /src/01-common-principles.md @chrisfilo @DimitriPapadopoulos
 /src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md @chrisfilo
-/src/04-modality-specific-files/03-electroencephalography.md @sappelhoff
+/src/04-modality-specific-files/03-electroencephalography.md @sappelhoff @ezemikulan
+/src/04-modality-specific-files/04-intracranial-electroencephalography.md @ezemikulan
--- a/src/04-modality-specific-files/04-intracranial-electroencephalography.md
+++ b/src/04-modality-specific-files/04-intracranial-electroencephalography.md
@@ -496,7 +496,7 @@ onset, duration), the `_events.tsv` file can contain the electrical stimulation
 parameters in addition to other events. Note that these can be intermixed with
 other task events. Electrical stimulation parameters can be described in columns
 called `electrical_stimulation_<label>`, with labels chosen by the researcher and
-optionally defined in more detail in an accompanying `_electrodes.json` file (as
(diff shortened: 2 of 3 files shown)


dorahermes:
thanks! lgtm

chrisfilo:
@ezemikulan if you want you can add your name to https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md AND your github id to https://github.com/bids-standard/bids-specification/blob/master/CODEOWNERS

ezemikulan:
@chrisfilo yes, I'll add it, thanks

choldgraf:
nice! thanks for noticing @ezemikulan :-)

@chrisfilo bugfixes don't fall under the governance rules of the org, right?

chrisfilo:
All PRs are treated the same way, with the exception of new releases (which require additional approval from a Maintainer). A triage system could be put in place in the future, but is not currently part of decision-making rules.

However, if you want this PR to be merged now without having to wait 5 days to gain community exposure a Maintainer (@sappelhoff) has the power to merge it as is.

choldgraf:
I think we should clarify this in the governance docs, either:

1. The waiting period only applies to changes that actually change the content of the spec, not that clarify or de-bug the language

or 

2. One or a subset of individuals has the authority to override the governance at their discretion.

In its current state if governance says "every PR needs 5 days of comment" (or whatever amount), then we should follow it.

chrisfilo:
> 1. One or a subset of individuals has the authority to override the governance at their discretion.

I believe this is pretty clear in the recently added point 5 of the rules: https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules

If something is not clear or you would like to propose a triage system, please open a new issue or send a PR. This PR is not the best place to discuss this (most people will miss it).

choldgraf:
cool, that looks fine to me, I'm happy with @sappelhoff merging

sappelhoff:
thanks @ezemikulan!

PR #190, opened 2019-03-26
[FIX] Unclear caption of example 3

pvdemael:
The caption of example 3 is unclear compared to the code.
There is only one JSON file in the example.
A better title could be "Single json file at top level for same task and acquisition parameters

If the change is not committed, then change the json to capitals.

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -191,8 +191,7 @@ apply to different runs and rec files. Also if the JSON file
 (`task-xyz_acq-test1_bold.json`) is defined at dataset top level directory, it
 will be applicable to all task runs with `test1` acquisition parameter.
 
-Example 3: Multiple json files at different levels for same task and acquisition
-parameters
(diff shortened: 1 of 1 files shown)


sappelhoff:
Thanks for pointing this out, however I think that the true issue was a mistake when carrying the spec over from Google Docs to Markdown.

Please go to [the spec from before changing infrastructures](https://bids.neuroimaging.io/bids_spec1.1.0.pdf), and in the pdf that'll pop up, search for `However, if there is a key` --> this will bring you to that section you are currently trying to fix (alternatively, go manually to "Inheritance Principle" section ... Example 3 is called "case 2" in the outdated spec).

This also alerts me to the fact that https://bids.neuroimaging.io/bids_spec.pdf is linking to version 1.1.0 of the spec in PDF format, instead of our most recent spec. Perhaps we should change that.

sappelhoff:
@pvdemael did you see my comment? If you don't have time to make the changes, we can turn this into an issue and have somebody else take over fixing it :-)

PR #144, opened 2019-02-09
Update Pipfile.lock

franklin-feingold:
update pyyaml to 4.2b4

PR #208, opened 2019-04-17
[MISC] update pipfile.lock

franklin-feingold:
updating jinja2 in the pipfile.lock file


effigies:
This looks fine, but is there a justification?

franklin-feingold:
this is to fix a github alert of a potential security concern to upgrade this version of jinja2

sappelhoff:
I need some background to understand this PR:

Was `Jinja2` previously installed via the `mkdocs-material` package in our [Pipfile](https://github.com/bids-standard/bids-specification/blob/master/Pipfile)?

If so, is this PR the result of you (Franklin) doing a manual upgrade of `Jinja2` in your local pipenv and then pushing the updated `Pipfile.lock`?

Would another solution be to wait for `mkdocs-material` to update their dependencies (`Jinja2`) and wait until it passes down to us?

franklin-feingold:
this is to settle the github alert regarding the security concern. I grabbed the hashes from a different upgraded repo. not sure what exactly it traces back to, but its purpose is to settle the github alert

sappelhoff:
Could you link from where you got the hashes? 

The hashes are simply there to specify the exact version of `Jinja2` on [pypi](https://pypi.org/).

It's probably not the intended way of `pipenv` to copy the hashes from somewhere else. But it seems to work? :man_shrugging: not sure if this practice is a bit insecure in itself though...

effigies:
Looks like mkdocs's only jinja2 [constraint](https://github.com/mkdocs/mkdocs/blob/7c0d0b762451660622164b9e85ecff956a948075/requirements/project.txt) is `>=2.7.1`.

That said, lockfiles probably shouldn't be edited by hand. Can we upgrade our constraint to `>=2.10.1` and let pipenv recreate the lockfile?

sappelhoff:
We are not specifying a version of `Jinja2` anywhere in our project. See https://github.com/bids-standard/bids-specification/blob/191e9045ceedb97ff0bb213b9f384a8dd649cd10/Pipfile#L1-L13

effigies:
Okay. Then we can probably just re-lock, or whatever the pipenv operation is, and let the most recent version of jinja2 get pinned.

sappelhoff:
Yes, I think so too. That's what I wanted to clarify in my [comment](https://github.com/bids-standard/bids-specification/pull/208#issuecomment-484592949) above:

> If so, is this PR the result of you (Franklin) doing a manual upgrade of Jinja2 in your local pipenv and then pushing the updated Pipfile.lock?

franklin-feingold:
this is to fix our potential security vulnerability. I tried in the past to let pipenv do it and there was an assortment of issues I ran into. (previously seen vulnerability fix https://github.com/bids-standard/bids-specification/pull/144). These concerns pop up once in a while and the solution has been to upgrade the little parts. I have tried on my few other repo copies of the specification (i.e. https://github.com/franklin-feingold/bids-specification-test) to confirm this will fix the vulnerability.

PR #207, opened 2019-04-16
[ENH] Derived (processed) MR data

effigies:
This replaces #109, tracking the state of the derivatives BEP as we finalize.

The current state of this branch will be rendered on https://bids-specification.readthedocs.io/en/derivatives/

Depends on #205, #167.

PR #213, opened 2019-04-26
[FIX] Revise template-generated coordinate systems

oesteban:
I've made several changes to the table of template identifiers:

  - [x] Reordered template identifiers alphabetically
  - [x] Removed the distinction between surface templates and volumetric
        templates. After all, to properly apply a surface template one
        should first spatially normalize to a particular volumetric
        template. Although it is possible to create mixed templates
        (e.g. using ``MNI152Lin`` to generate a mapping for volumetric
        coordinates and then, say, ``fsLR`` for the surface after some
        surface mapping process), this option should be discouraged.
        Anyways, this discussion does not affect how the names of BIDS
        entities should be specified, since one can always describe how
        the template coordinates are generated with the information
        described in this PR.
  - [x] Removed ``MNI152NLin6AsymConte69`` because it does not define
        any new coordinate system (see
        https://github.com/bids-standard/bids-specification/pull/207#discussion_r278911304),
  - [x] Marked DISCOURAGED ``fsaverage[3|4|5|6|sym]``
  - [x] Added ``fsaverage`` and ``fsaverageSym`` corresponding to such
        deprecations.
  - [x] Removed ``FS305`` as it is now covered by ``fsaverage``.
  - [x] DISCOURAGED modifiers of ``UNCInfant``
  - [x] Reorganized subsections in "standard" vs "nonstandard" spaces.


src/99-appendices/08-coordinate-systems.md
```diff
@@ -116,47 +116,35 @@ below, the origin is at the AC and the orientation of the axes is RAS. Unless
...
+| Coordinate System | Description                                                                                                                                                                                                                                                           |
+| ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| individual        | Participant specific anatomical space (for example derived from T1w and/or T2w images). This coordinate system requires specifying an additional, participant-specific file to be fully defined. In context of surfaces this space has been refered to as `fsnative`. |
```
satra: just clarifying that the individual anatomical space can come from any MR/PET modality. 

also, i know we are doing fsnative, but how would we handle things like CIVET or other algorithms that generate surface spaces. is `fsnative` the default key for native space, independent of the algorithm used to produce it?

oesteban:
@edickie 
> OK looks good. Two things:
> 
> 1. I'm OM with deprecating all the density-specific tags BUT I do think we need to provide _in this doc_ the correspondence between the "new" `den-` values and many types of fsaverage (i.e. space-fsaverage5 --> fsaverage_den??) So that users do not become confused.

Sure, I agree. Thanks for the tip, will modify accordingly. BTW, what does "I'm OM" mean...?

> 2. One section (the introduction of the `volspace-<space>` key) of the anatomical derivatives section still refers to separate "Surface" and "Volume" template tables. I remember expanding this section with more examples in another PR, so I would make sure you don't get conflicts down the road.  Note: this key was introduced to capture the use case where the surface files (gifti surface coordinates) files where the surface points have been transformed correspond with the T1w image `space-fsLR_den-32k_volspace-individual_pial.surf.gii` from the one where the points are in MNI-space `space-fsLR_den-32k_volspace-MNI152Lin6Sym_pial.surf.gii`

That's my next step here, revising that other PR and make sure everything is consistent.

@satra 
> just clarifying that the individual anatomical space can come from any MR/PET modality.

Agreed, I just reorganized what it was written already. But I'll have a second pass to state this.

> also, i know we are doing fsnative, but how would we handle things like CIVET or other algorithms that generate surface spaces. is `fsnative` the default key for native space, independent of the algorithm used to produce it?

This is a good question. I'm not very fond of `fsnative` as a different space for surface, but we need to keep backward compatibility. I'd be happy with any new solution. What would you suggest?

edickie:
..oops I meant to write I'm OK..

src/99-appendices/08-coordinate-systems.md
```diff
@@ -116,47 +116,35 @@ below, the origin is at the AC and the orientation of the axes is RAS. Unless
...
-| MNI152NLin6AsymConte69                              | Template based on MNI152NLin6Asym using Conte69 data. Built by Human Connectome Project. [https://github.com/Washington-University/HCPpipelines/tree/master/global/templates](https://github.com/Washington-University/HCPpipelines/tree/master/global/templates)                                                                                                |
+| MNI152NLin6Sym                                      | Also known as symmetric ICBM 6th generation (non-linear coregistration). Used by SPM99 - SPM8. [http://www.bic.mni.mcgill.ca/ServicesAtlases/ICBM152NLin6](http://www.bic.mni.mcgill.ca/ServicesAtlases/ICBM152NLin6)                                                                                                                             |
```
effigies: Not alphabetical...

src/99-appendices/08-coordinate-systems.md
effigies: It looks like fsnative is not in the current standard, so it is reasonable to me to change it, if people want. What about `recon` (short for `individualsurface`)?

Speaking of `individual`, is this a proposal, or was there previous discussion? Were there other options? It just feels long...

src/99-appendices/08-coordinate-systems.md
```diff
@@ -116,47 +116,35 @@ below, the origin is at the AC and the orientation of the axes is RAS. Unless
...
+| Coordinate System | Description                                                                                                                                                                                                                                                           |
+| ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| individual        | Participant specific anatomical space (for example derived from T1w and/or T2w images). This coordinate system requires specifying an additional, participant-specific file to be fully defined. In context of surfaces this space has been refered to as `fsnative`. |
+| custom            | Custom space defined using a group/study-specific template. This coordinate system requires specifying an additional file to be fully defined.                                                                                                                        |
```
effigies: So I've conceded on the controlled vocabulary issue, but I would say if that's the case, we need to have more than just `custom`, as you'll want to distinguish between some custom cases. Here we identify a specific use case: study-specific templates. Why not add `study`?

Are there any others that are common that are worth carving out?

oesteban: This is a great point. I'll add `study` and try to think of other possibilities.

src/99-appendices/08-coordinate-systems.md
edickie: yeah I also feel like "individual" is too long..but do we have a good alternative?

effigies:
This should probably be rebased onto `common-derivatives` (#265).

oesteban:
@effigies, I've rebased to `common-derivatives` but shouldn't this PR be done against the main spec?

src/99-appendices/08-coordinate-systems.md
oesteban: > @satra:
> how would we handle things like CIVET or other algorithms that generate surface spaces

what you mean by "surface space"? how those operate differently w.r.t. FreeSurfer and fsnative/fsaverage?

effigies:
You were previously doing it against derivatives. I assumed that was intentional.

oesteban:
Right. Okay, not intentional at all. Should I send a new PR?

effigies:
If it's strictly against things that aren't in derivatives, then I'd just rebase onto master and set master as the upstream. If it also touches on derivatives, I'd say set the upstream to common-derivatives.

src/99-appendices/08-coordinate-systems.md
satra: yes, they are different surfaces similar to fs32LR being different from fsnative/fsaverage

oesteban: I am probably being obtuse here, my apologies. But fsLR is a template space, for CIVET we would then add the corresponding ID for their template (for instance).

How that maps on to the individual space?

satra: i'm not a user, so it would be difficult for me to say (perhaps some of the folks north of the border can chime in). in terms of common templates, i would definitely say the freesurfer templates are much more widely used.

oesteban:
I'm having real trouble rebasing. I separated out this in two PR:

- #306 Updates current BIDS "official" spec.
- #307 Updates the common-derivatives BEP

PR #214, opened 2019-04-27
[INFRA] Enable permalink urls to appear at (sub)section headings

yarikoptic:
There is no convenient way to obtain a permalink to reference a specific (sub)section
in the page ATM.  Only top level section urls available in the TOC. With this permalink
those IDs, which are already in HTML anyways, become exposed so it is easy to get
permalink for any section.  I chose # symbol since that is what is used in the URL to
separate out the fragment to point to that ID in the page, e.g.
image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,7 +2,7 @@ site_name: Brain Imaging Data Structure v1.2.0-dev
 theme: material
 markdown_extensions:
     - toc:
-        permalink: "#"
+        anchorlink: true
 plugins:
     - search
 docs_dir: 'src'
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,8 @@
 site_name: Brain Imaging Data Structure v1.2.0-dev
 theme: material
+markdown_extensions:
+    - toc:
+        permalink: "#"
 plugins:
     - search
 docs_dir: 'src'


effigies:
I've often seen the paragraph symbol (¶) for this purpose, but I don't know that it's any more meaningful.

Is there any way to just make the header text itself the link? That would be most intuitive to me.

effigies:
But to be clear, I am very supportive of this move and have felt the lack of these links in any text.

sappelhoff:
Thanks @yarikoptic +1 for this PR. Like @effigies I would prefer if the header text could be the link.

mkdocs.yml
```diff
@@ -1,5 +1,8 @@
 site_name: Brain Imaging Data Structure v1.2.0-dev
 theme: material
+markdown_extensions:
+    - toc:
+        permalink: "#"
```
effigies:
```suggestion
        anchorlink: true
```

yarikoptic: Great, if that does the trick I all all for it! Thanks!

yarikoptic:
Looks great! Thanks!

yarikoptic:
where could I find the current master build of the spec?  I guess the one on https://bids-specification.readthedocs.io is the released one so doesn't have anchors yet

effigies:
https://bids-specification.readthedocs.io/en/latest/

PR #220, opened 2019-05-06
[FIX] Trivial column header fix 

nicholst:
Replace underscore-separated column headers with capitalised, space-separated headers.

Just fixing some snake-case formatted column labels (modality_label & contrast_label).


nicholst:
If matching the original label string was intentional, then I'd recommend making it 'code' text, as in `modality_label` instead of modality_label and `contrast_label` instead of contrast_label.

chrisgorgo:
> If matching the original label string was intentional, then I'd recommend making it 'code' text, as in `modality_label` instead of modality_label and `contrast_label` instead of contrast_label.

Sounds good to me.

nicholst:
Done!  Now just changes the font.

chrisgorgo:
Thanks! I think you need to fix the table width now. See https://travis-ci.com/bids-standard/bids-specification/builds/110783874#L481 and https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md#writing-in-markdown

nicholst:
Lint should be happy now.

PR #234, opened 2019-05-16
REL: 1.3.0-dev

### LINKED ISSUE #233: [INFRA] Latest version shows 1.2.0-dev
In most cases the `-dev` suffix is considered pre-release state, so `1.2.0-dev` is for versions leading up to the 1.2.0 release. We are post-1.2.0, but the listed version is `1.2.0-dev`.

Do we consider this a problem?

effigies:
Closes #233.

@franklin-feingold If you have a process for generating the before/after figures, could you change them to `1.2.0-dev -> 1.2.0` and `1.2.0 -> 1.3.0-dev`, or similar?


franklin-feingold:
added in the before/after images for the release protocol - https://github.com/effigies/bids-specification/pull/1

effigies:
Sorry, I rebased to get rid of a bunch of ugly merge commits, so I'll need new reviews.

I'm doing a rewrite of the release protocol that's a bit heavier on exact git commands. That can go in this or wait for a new PR, whichever y'all prefer.

franklin-feingold:
I don't have a preference for either option. I think it can go into this one unless it will take some time to flesh out then it can go into the next one

effigies:
Nah, should finish in the morning. Just a question of whether you want to potentially do back and forth on this PR.

franklin-feingold:
oh! it may be good in that case to separate the two for historical tracking, we can have one PR changing the dev name and another to do the release protocol rewrite

effigies:
Yeah, I think I'm going to propose some potentially substantive changes to the release procedure, so we should split it from this.

Issue #229, opened 2019-05-13
[INFRA] Use annotated tags for releases


yarikoptic:
Streamlines use of `git describe` etc

```
$> git tag                    
v.1.1.2
v1.1.2
v1.2.0

# 1.1.2 is proper annotated:
$> git describe e21e619       
v1.1.2-227-ge21e619

# but 1.2.0 (github does mis-service in producing those while doing releases on github web ui)
# so --tags is needed to get proper one for 1.2.0
$> git describe --tags e21e619
v1.2.0-67-ge21e619
```

So ATM older tag is an annotated one while the latest one is not.
I guess we fell into the trap of github where it doesn't create annotated tags when releases made from the web UI.

franklin-feingold:
Thank you for raising this! If I am interpreting correctly - the tags created due to a release (via github web) are not annotated to the release because it is done via web ui? To fix this and have annotated tags for releases, how may you recommend the [release protocol](https://github.com/bids-standard/bids-specification/blob/master/Release_Protocol.md) be altered to enable this?

yarikoptic:
I have searched again and I think github still doesn't allow anyhow to create annotated tags via web ui.  So the only way I know, and possible way is to 
- create them locally (just annotated or even signed) via `git tag -a <RELEASE> [<COMMIT>]` or `git tag -s  <RELEASE> [<COMMIT>]` (for signed tag). 
   - I also typically add changelog for the release within tag description, so it is right there for me when I want to see the release details.  The problem though that markdown `##` section annotations are treated by git as comments
- push to github `git push origin <RELEASE>` (`git push --tags origin` might be other tags present locally which might be undesirable)
- and only then go to releases web ui and make the release from that tag ("Edit" existing tag, "Publish the Release" or alike -- whatever green button it is in that interface).

NB there is a "rogue" tag `v.1.1.2`, (the same as v1.1.2), I would recommend `git push origin :v.1.1.2` ;-)

effigies:
Bad tag removed and release updated to the correct v1.1.2 tag.

effigies:
@yarikoptic Should I create an annotated tag for v1.1.2?

yarikoptic:
I am confused... 

> Bad tag removed and release updated to the correct v1.1.2 tag.
> @yarikoptic Should I create an annotated tag for v1.1.2?

are you talking about   v.1.1.2 (when saying "removed") ?  but there was one annotated one already

```
$> git show v1.1.2
tag v1.1.2
[...]

Issue #164, opened 2019-03-05
Allow modification of the changelog post merge


chrisgorgo:
The changelog is currently generated automatically without a possibility to manually modify an incorrect PR title post-merge. This would be useful to correct typos and improve the clarity of the changelog.

franklin-feingold:
The changelog generator implemented will process PR updates (i.e. modifying the PR title) and be updated in the next merged PR. The generation happens at the next merged PR because that is the action that triggers a changelog build and deploy.

chrisfilo:
Ok so if I understand you correctly you are saying that we can go to github web interface and changed names of historical PRs that have already been merged. Upon next merge of a new PR those changes will be picked up by the changelog generator and the changelog will be updated.

This sort of works, but it does require a new PR (non release one) to be merged to triger regeneration of the changelog. I was envisaging tweaking the changelog as part of the release process. @effigies what do you think?

effigies:
On the whole, I'd prefer that there was an opportunity to manually set a ground truth, instead of manipulating the algorithm's inputs. But this does sound like a relatively easy first pass adjustment.

franklin-feingold:
Correct and the reason it is on merged PR's is that we do not want the changelog to be generated on commits because then it will get into a circle due to the changelog being pushed automatically into the repo. I think in line with your thinking - as part of the release PR there will also be an updated changelog as well?

If I'm interpreting what you are describing - in the PR review process we would have a set of controlled vocabulary that the label can be and before the PR is merged confirm that the label is there and part of the vocabulary?

effigies:
Just to link in discussion in https://github.com/bids-standard/bids-specification/pull/232#issuecomment-493122650:

> I think there should be two changelogs: the repository changelog and the specification changelog. When somebody's reading the spec, what they care about is what new features are there in the specification, and absolutely not what changes were made to the build system.
> 
> IMO, the specification changelog should ONLY include `[FIX]` and `[ENH]` changes, and should strip out the tags. I also think we should be able to go in and retroactively fix the changelog to be more human friendly without our changes getting destroyed every time a PR gets merged.

franklin-feingold:
My concern regarding two changelogs is the maintenance and piping for each to stay up to date. I can envision using the pieces we have of the current changelog generator to create both of these logs, then having a pruning call to clean it up and keep what each needs.

> I also think we should be able to go in and retroactively fix the changelog to be more human friendly without our changes getting destroyed every time a PR gets merged.

[...]

Issue #66, opened 2018-10-24
[DISCUSSION] acknowledge contributors in DOIed specs on Zenodo


tyarkoni:
To facilitate citation, we should auto-generate Zenodo DOIs for each spec, with author list extracted from the contributors list in the docs.

---

edit by @sappelhoff  2020-05-06: PDF versions of the BIDS spec are now stored on zenodo: https://zenodo.org/record/3720628 ... we still have to adjust the metadata on Zenodo to properly acknowledge contributors.

chrisfilo:
What authorship order would you suggest?

tyarkoni:
Good question... we should probably weight each of the different types of contribution some specified amount, sum up each person's contributions, and then order in descending order of total score. (With the option to manually fix some people--e.g., you should always go first in the core BIDS spec, I should go first in the BIDS-Model spec, etc.).

tyarkoni:
I'm not sure assigning a single DOI makes sense. The degree of modularity is already pretty high. I guess I think of BIDS-Derivatives, BIDS-Models, etc., as modular components in a broader ecosystem—much like dependencies of a package. I think maintaining a single DOI is going to make proper credit assignment very difficult in two respects: (a) when people cite the spec, we'll have essentially no idea what part of it they're making use of, and (b) it'll dilute many people's contributions (because you might have played a central role in BIDS-Derivatives, but never touched the other documents).

chrisfilo:
I agree with the credit argument. I guess we could use Zenodo API to issue multiple DOIs upon spec release and associate subsets of the md files with each one. Versioning would have to be the same though if we are to live in the same repo. It's more complex, but not impossible.

Please also mind that Zenodo is not indexed by Google Scholar, so I would still rely on individual papers dedicated to each BEP for credit purposes.

tyarkoni:
Joint versioning sounds reasonable.

I think it's reasonable to assume that in the not-too-distant future, Google Scholar will index Zenodo (see https://github.com/openjournals/brief-ideas/issues/132). If I read the linked thread correctly, they may already be indexing text-heavy repos, which this certainly would be.

chrisfilo:
Ok - could you propose a formula for authorship order? It would be a good start of the discussion.

tyarkoni:
Where are the contributions listed now?

tyarkoni:
As a first pass, maybe just weight everything equally? We can iterate later...

tyarkoni:
[...]

PR #235, opened 2019-05-17
[INFRA] Update release protocol

effigies:
Started amending the release protocol in #234, found that it could be made clearer with explicit `git` commands. Then felt that it could stand some more comprehensive revision.

The three substantial changes I'm proposing are:

1) Merging other pull requests is permitted during a pull request if the discussants agree that the pull requests should be merged before a release. In particular, this is intended to allow for edits that restore consistency without forcing the pull request to be deleted and re-started at some later date.
2) Releases should be given annotated tags (see #229).
3) Cleanup of the changelog should be made in the release branch, as it won't get demolished by the changelog generator. (See some discussion in #164.)

As a subsidiary to change 2, if we want to use GitHub releases at all is kind of an open question at this point. We can continue doing so if people like them, but there's no technical advantage and we haven't been putting much effort thus far into making them informative documents.

To see this document rendered: https://github.com/effigies/bids-specification/blob/process/release/Release_Protocol.md

Leaving this PR as a draft for now because I haven't finished (e.g., I haven't incorporated change 3), but I would like to get comments on the issues raised above before investing more time.


sappelhoff:
+1 for the proposed changes, thank you for the clarification work and integrating annotated tags.

> As a subsidiary to change 2, if we want to use GitHub releases at all is kind of an open question at this point. We can continue doing so if people like them, but there's no technical advantage and we haven't been putting much effort thus far into making them informative documents.

I agree that currently the GitHub releases do not add much. However, once we start to integrate with [Zenodo](https://guides.github.com/activities/citable-code/) to generate DOIs for the spec, the releases will become meaningful, because they will trigger a new archiving on Zenodo.

related to: https://github.com/bids-standard/bids-specification/issues/66

franklin-feingold:
Re: the 3 changes -
1) I think that sounds good to me. We may want to constrain and identify these type of cases (or at least an example) to be more clear what is meant. The PR can be opened before and merged after a PR so not to lose the comments and history of work. 

2) LGTM. I have no strong opinion with GitHub releases. Seemed like an easy quick way to see a historical record, but not concerned if we decide against. Zenodo could change that work process decision. 

3) With the release cleanup - so the cleanup happens in the branch and then it is committed over? The commit would allow for the generation to be skipped and then could correspond with the same day of Release to be most up to date in `stable`. 

Thank you for putting this together!

effigies:
Alright, finished this. LMK what you think.

franklin-feingold:
Thank you for putting together! Looks good to me! Will be good to see in practice for the next release and iterate if need be

effigies:
Cool. Thanks for the review.

PR #236, opened 2019-05-23
[FIX] remove ManufacturersAmplifierModelName (again)

robertoostenveld:
- this was discussed in the past and supposed to be already removed in version https://github.com/bids-standard/bids-specification/blob/master/src/CHANGES.md#111 but apparently slipped through 
- it is documented in the text properly, where it states "Note that if EEG is recorded with a separate amplifier, it should be stored separately under a new /eeg data type"


choldgraf:
to confirmations and this LGTM so merging away!

PR #273, opened 2019-07-13
[ENH] Adding active discussions/enhancements and help wanted section to readme

franklin-feingold:
Added an active discussions/enhancements and help wanted section to readme. This can help orient/focus the community around what is being actively discussed and where help is needed. 

I captured a few for each to give an idea of how it may look. Can be expanded upon.  

Note: the `&nbsp;` is for spacing


franklin-feingold:
> make more use of the "Help wanted" tag on the issues page

How would you propose we do this? We have help wanted tags, but currently little to no activity on those threads. 

> "pin" issues --> we can pin up to three dominant issues

This could perhaps work, but would these pined issues be active discussions or help wanted or both? How would these be easily differentiated, probably a tag?

> having a brief section in the README referring to the issues page: What it is and how we use it (not only for bugs, but also discussions).

Ok. we can try this and evaluate.

Overall, the goal of this is to reduce the barrier to contributing and quickly be able to see where the project is at and where the efforts are currently being directed toward. This can be difficult to see for newer members of the community. I think it needs to be very clear and easy to locate. The fluctuation is fine it can be updated, but we can try having a brief explanation of `issues` and reevaluate.

Issue #189, opened 2019-03-25
Use a Discourse instead of an email listserv


choldgraf:
The recent spat of spam job listing emails on the BIDS listserv made me remember an idea I had a while back:

What do folks think about using a Discourse forum instead of an email listserv? A few points of information from me:

* Discourse is free for open projects (https://blog.discourse.org/2018/11/free-hosting-for-open-source-v2/)
* it provides a platform for communication that you have much more control over compared to an email listserv. For example, you can define multiple threads/topics/tags, people can mark posts as "questions" or "answers", you can run polls, etc
* It lets you respond to / get notifications of issues via email in case people want to keep communicating in that way.
* It lets you moderate more effectively and control who can post what based on their previous activity in the community
* It gives people more of a personal identity, and can help people build their own personal image within the community

Given that BIDS is a project that is largely community-driven and requires lots of community conversations, I think it could particularly be useful here. We've been using it over at Jupyter (discourse.jupyter.org) and have generally really liked the results.

Would be curious what people think about this!

satra:
@choldgraf - we are already using discourse for neurostars.org (which handles a lot of the software around bids). perhaps we could just continue to use that instead of another discourse?

sappelhoff:
> recent spat of spam job listing emails on the BIDS listserv

@choldgraf just to be clear: are you talking about the ["google forum/maillist"](https://groups.google.com/forum/#!forum/bids-discussion)?

I personally like discourse as well, e.g., the one for [psychopy](https://discourse.psychopy.org/) or as @satra mentions, our [neurostars](https://neurostars.org/). However, I think the Google forum/mailinglist helps to reach a few people who want to only interact via email instead of jumping on a new platform. What do others think?

nicholst:
I was under the impression that Discourse can be configured so that you can
interact with it in an email-only fashion, if that's your preference.

choldgraf:
@nicholst I believe that's the case, yes.

@sappelhoff yep, that's the listserv I was talking about.

@satra if there's already a place that could be used for this discussion, I'm +1 on using something like this instead of creating a new thing

emdupre:
Just to say -- I'm :+1::+1: on moving the mailing list over to a discourse board ! 

Using NeuroStars makes sense to me. Right now there are only a small number of folks who consistently actively engage on NeuroStars (just looking at [the Thank You badge](https://neurostars.org/badges/38/thank-you)), so this could be a good way to encourage more community buy-in.

[...]

PR #278, opened 2019-07-18
[ENH] Starter Kit dropdown menu

franklin-feingold:
Added direct links to - starter kit GitHub repo, tutorials, forum, and wiki. Here is the [rendering](https://807-150465237-gh.circle-artifacts.com/0/home/circleci/project/site/index.html) from the index. 

This can bring more attention to the starter kit by directly linking to these resources.


mkdocs.yml
```diff
@@ -38,4 +38,8 @@ nav:
         - MEG systems: 99-appendices/07-meg-systems.md
         - Coordinate-systems: 99-appendices/08-coordinate-systems.md
       - Changelog: CHANGES.md
-    - The BIDS Starter Kit: https://github.com/bids-standard/bids-starter-kit
+    - The BIDS Starter Kit:
+      - GitHub repository: https://github.com/bids-standard/bids-starter-kit
+      - Tutorials: https://github.com/bids-standard/bids-starter-kit/wiki/Tutorials
+      - Forum: https://neurostars.org/tags/bids
```
sappelhoff:
```suggestion
```

I think the neurostars forum is now sufficiently advertise (and framed in a better way) in #279

franklin-feingold: I think that this is still fine to publicize here because it is part of the starter kit and probably better to keep them together as a group

sappelhoff: In how far is Neurostars a part of the BIDS-starter-kit? I see them as separate entities and think that combining them here will lead to confusion between the two.

franklin-feingold: Regarding part of the starter kit - I think it is a good feature to publicize that we are using this. I agree that they are separate entities, but I think helpful to bring attention to this great resource. Several of us are active on NeuroStars and good to further promote as another resource for those to learn and pose questions. I think from the perspective of learning and starting out, putting the NeuroStars forum is useful to be there. Also depending on the conclusion of https://github.com/bids-standard/bids-specification/issues/189 may make it more of a need to add it in. 

To clarify the concern - the confusion can be that NeuroStars is for all neuroimaging questions and not just BIDS related questions?

effigies: I think the concern is simply that the BIDS Starter Kit is a related set of documents, and NeuroStars is a useful resource, and the way this is currently written makes it appear that NeuroStars is a sub-structure of the starter kit.

I am +1 for removal, myself.

franklin-feingold: I agree, in the current setup that can be misunderstood as solely part of the BIDS. 

Perhaps moving it to be it's own button? To clarify - the starter kit would have the drop down menu then under this (at the same level as the starter kit) have the NeuroStars forum to directly click and direct to our BIDS tag on NeuroStars. I think drawing attention to this forum is helpful.

sappelhoff: I think it's okay to advertise the BIDS starter kit on the specification - and I saw it as a compromise between not putting too many links to other sites onto the specification and yet still doing a *bit* of advertising.

I agree that Neurostars is a great resource, perhaps we can advertise it on the starterkit Wiki page, which is a spot we are linking to.

franklin-feingold: We can experiment with this setup (removing NeuroStars) and iterate from there. The starter kit does link out to NeuroStars. My concern is missing the opportunity to provide the most help and resources to get them into the environment but can experiment and learn

PR #286, opened 2019-07-29
[ENH] Added discontinuous datatype for EEG and iEEG

wouterpotters:
**Problem**: EDF+ files (and other file types like BrainVision) can be `continuous` (EDF+C, BrainVision without discontinuous boundary markers) or `discontinuous` (EDF+D, BrainVision with discontinuous boundary markers). This can be essential information in processing of data (i.e. not all EDF readers support EDF+D or discontinuous BrainVision data).

See also [https://www.edfplus.info/specs/edfplus.html](https://www.edfplus.info/specs/edfplus.html) and [https://pressrelease.brainproducts.com/markers/#4](https://pressrelease.brainproducts.com/markers/#4)

**Suggestion**: mark data as `continuous` or `discontinuous` (or `epoched`)

**Note**: `discontinuous` data is not the same as `epoched` data!

_Please let me know what you think!_


robertoostenveld:
BrainVision files can also include markers that indicate the time of discontinuous boundaries between otherwise continuous segments, e.g. when pausing the recording. I don't think it applies to the other formats.

sappelhoff:
@wouterpotters our linter complains that the table pipe characters in the markdown (`|`) are not aligned properly, see: https://travis-ci.com/bids-standard/bids-specification/builds/121065179

Can you fix this, please? See here for a Markdown cheat sheet on tables: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

Apart from this I agree with @robertoostenveld that a short part added in the text will be helpful.

wouterpotters:
Dear @robertoostenveld and @sappelhoff, thank you for the useful suggestions and your time. Tried to incorporate all of the suggestions. The checks now pass successfully (didn't check them when I submitted this; will do so next time). 
Do not know about the BrainVision format, but as far as I can tell this pull request could apply to both edf and brainvision without explicitly naming them in the text.

wouterpotters:
> Thanks for the updates @wouterpotters I am fine with the changes made 👍
> 
> One more request: Could you please add yourself to the contributors file? --> https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md
> 
> We use this list to acknowledge people who have helped the BIDS standard to progress :-)

Done.

src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -57,6 +57,11 @@ also encourage users to provide additional meta information extracted from the
 manufacturer specific data files in the sidecar JSON file. Other relevant files
 MAY be included alongside the original EEG data in `/sourcedata`.
 
+Note that data can have different RecordingTypes; noncontinuous data can be
```
robertoostenveld: I propose to change this into this, with an explicit definition for all three values:

"Note the RecordingType, which depends on whether the data stream on disk is interrupted or not.  Continuous data is by definition 1 segment without interruption. Epoched data consists of multiple segments that all have the same length (e.g., corresponding to trials) and that have gaps in between. Discontinuous data consists of multiple segments of different length, for example due to a pause in the acquisition."

wouterpotters: will do so. Thank you for the improvements.

wouterpotters:
Included all suggestions by @robertoostenveld and @sappelhoff. Thank you for your time and effort into the BIDS standard. Great initiative!

As far as I can see it's now ready to merge into the main branch.

sappelhoff:
Thanks for your contribution @wouterpotters!

PR #291, opened 2019-08-02
REL: v1.2.1

franklin-feingold:
This marks the beginning of pushing a new release: 1.2.1 . We have several new modifications and enhancements to the specification. 

Checklist of PRs to merge before release

- [x] https://github.com/bids-standard/bids-specification/pull/272
- [x] https://github.com/bids-standard/bids-specification/pull/277
- [x] https://github.com/bids-standard/bids-specification/pull/284

All milestones reached / migrated to next release?

- [x] https://github.com/bids-standard/bids-specification/milestone/2

PR #294, opened 2019-08-05
[ENH] Clarify requirements in Release Protocol

franklin-feingold:
Added a git clone command. This was found when opening https://github.com/bids-standard/bids-specification/pull/291 . This command was needed to complete the rest of the procedure when starting from a clean directory

### DIFF
--- a/Release_Protocol.md
+++ b/Release_Protocol.md
@@ -14,9 +14,11 @@ $ git remote get-url upstream
 git@github.com:bids-standard/bids-specification.git
 ```
 
-If you do not, add it with:
+If you do not, clone the Master branch of the BIDS-specification and add it to `upstream` with:
 
 ```Shell
+$ git clone https://github.com/bids-standard/bids-specification.git
+$ cd bids-specification
 $ git remote add upstream git@github.com:bids-standard/bids-specification.git
 ```
 
--- a/Release_Guideline.md
+++ b/Release_Guideline.md
@@ -4,7 +4,7 @@ This document captures guidelines to follow when considering a new release of th
 
 ## Background
 BIDS has generally followed a semantic versioning (semver) approach.
-The canonical semver text can be found at [https://semver.org/](https://semver.org/), and an adaptation to documents can be found at [https://semverdoc.org/](https://semver.org/).
+The canonical semver text can be found at https://semver.org/, and an adaptation to documents can be found at https://semverdoc.org/.
 A specification falls somewhere between these two, as documentation does not require
 backwards compatibility, while software has notions of bugs and features which do not
 map to the reasons for updating a specification.


Release_Protocol.md
```diff
@@ -14,9 +14,11 @@ $ git remote get-url upstream
 git@github.com:bids-standard/bids-specification.git
 ```
 
-If you do not, add it with:
+If you do not, clone the Master branch of the BIDS-specification and add it to `upstream` with:
 
 ```Shell
+$ git clone https://github.com/bids-standard/bids-specification.git
+$ cd bids-specification
```
sappelhoff: I usually add remotes using: `git remote add upstream https://github.com/bids-standard/bids-specification` --> for that I do not need to clone the upstream that I want to add.

I think @effigies wrote this part of the protocol, perhaps he can weigh in.

effigies: Yeah, I don't think this is necessary. But we could add a clone to before the "You should have a remote", if it's not clear to people that they need to be in the repository. I assume most maintainers will already, though...

franklin-feingold: I think adding that can be useful. I agree those that are charged with releasing will likely know this information already. Thought it may be helpful to have the protocol self-contained

sappelhoff: Okay, fine with me to include it. But wouldn't a developer/maintainer always clone their fork, rather than the upstream?

for me:
```shell
git clone https://github.com/sappelhoff/bids-specification.git
cd bids-specification
git remote add upstream https://github.com/bids-standard/bids-specification
```

franklin-feingold: your path is what I took too. I think it is easier to git clone

sappelhoff: @franklin-feingold do you want to incorporate these changes in this PR? Or do you want to take this PR in another direction and we should do the revision somewhere else?

franklin-feingold: what change may you be referencing? the change discussed above is incorporated - the git add upstream is the line below this comment. What you commented previously that was missing is what is added (the clone and cd). I did not change the `git@github.com:bids-standard/bids-specification.git` to `git remote add upstream https://github.com/bids-standard/bids-specification` because that is what Chris had. Can commit that change, I used Chris's process when processing the release (https://github.com/bids-standard/bids-specification/pull/291)

sappelhoff: Ah I see, yes that works - although in this first line `git clone https://github.com/bids-standard/bids-specification.git` the GitHub user `bids-standard` would usually be the username of the maintainer. I don't know if that **needs** adjustment, ... might be okay to just leave it as is for now.

franklin-feingold: using `bids-standard` will ensure it is the most up to date. could be safe to assume the maintainers doing this process will have their fork synced but can be amended later if we find this is not needed

sappelhoff: @franklin-feingold I have made a PR to your PR-branch, see https://github.com/franklin-feingold/bids-specification/pull/27

I think it is better to mention a fork and a clone as the requirements for the release protocol, because the practical implementation of this may differ between maintainers/developers.

franklin-feingold: merged in - thank you

sappelhoff:
Thanks @franklin-feingold

PR #265, opened 2019-07-09
[ENH] BEP 003: Common Derivatives

effigies:
This PR is the first to replace #207, as discussed in #254.

In this PR, we will focus on the (mostly) modality-agnostic components of the derivatives specification, and the necessary changes to the broader spec. This should hopefully keep the scope to a level that several people can read it all the way through, with a particular lookout for:

1) Backwards-incompatible changes
2) Excessive restriction that will make it difficult for apps to produce valid datasets with as-yet-unspecified data types (related to discussion in #264)
3) Inconsistencies with how the modality-specific sub-BEPs have been treating an issue

There are a lot of conversations on #207. If any of them are relevant to the common derivatives, we should link to the old conversation, ideally with a summary of the state of discussion.

The branch is being rendered on https://bids-specification.readthedocs.io/en/common-derivatives/.

---

## 9/18 State of the BEP (by: @effigies)

I consider the following PRs essential to the completion of this BEP:

* [x] #301 (submitter: @oesteban)
* [x] #307 (submitter: @oesteban)
* [x] #334 (submitter: @effigies)

The following PR into `master` is a pre-condition for merging this BEP:

* [x] #306 (submitter: @oesteban)

The following issues are worth considering in this context:

* [ ] #309 (submitter: @satra)

## 8/22 State of the BEP (by: @franklin-feingold)
Remaining PRs to meet at a consensus. This will enable the finalization of merging the common derivatives extensions.
https://github.com/bids-standard/bids-specification/pull/300
https://github.com/bids-standard/bids-specification/pull/301 (offshoot discussion in https://github.com/bids-standard/bids-specification/issues/309)
https://github.com/bids-standard/bids-specification/pull/306
https://github.com/bids-standard/bids-specification/pull/307
https://github.com/bids-standard/bids-specification/pull/310

PR #300, opened 2019-08-13
[ENH] Storing basic provenance on JSON sidecars of derivatives

oesteban:
## Summary

In the context of BEP 016 we found ourselves prescribing what metadata fields should/could be included within derivatives. We then realized that some of them would easily extend to other _Imaging derivatives_ (see #310), and also tapped on the quite controversial `SkullStripped` metadata field.

## Proposal

Since BIDS has always been reluctant to include much provenance information, this PR proposes to make all these metadata fields OPTIONAL, and provides some suggestions for naming.

## Original Content of this PR
Mapped from @Lestropie's bids-standard/bids-bep016#5
Closes bids-standard/bids-bep016#2

I have removed some keys that I believe should remain in BEP16 for being
applicable to dwi only.

I have also changed _FieldInhomogeneity*_ with
_SusceptibilityDistortion*_ since I think the latter will be more
familiar to a wider audience.

EDIT: fields that remained in BEP16 are - `EddyCurrentCorrection`, `GradientNonLinearityQSpaceCorrection`, `SliceDropoutDetection`, `SliceDropoutReplacement`

### DIFF
--- a/src/05-derivatives/02-common-data-types.md
+++ b/src/05-derivatives/02-common-data-types.md
@@ -62,9 +62,19 @@ makes them invalid (e.g., if a source 4D image is averaged to create a single
 static volume, a SamplingFrequency property would no longer be relevant). In
 addition, all processed files include the following metadata JSON fields:
 
-| **Key name**  | **Description**                                                                                 |
-| ------------- | ----------------------------------------------------------------------------------------------- |
-| SkullStripped | REQUIRED. Boolean. Whether the volume was skull stripped (non-brain voxels set to zero) or not. |
+| **Key name**                           | **Description**                                                                                                                                   |
+| ---------------------------------------| ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| SkullStripped                          | REQUIRED. Boolean. Whether the volume was skull stripped (non-brain voxels set to zero) or not.                                                   |
+| Denoising                              | OPTIONAL. String. Denoising method                                                                                                                |
+| GibbsRingingCorrection                 | OPTIONAL. Boolean. Removal of Gibbs ringing artifacts                                                                                             |
+| MotionCorrection                       | OPTIONAL. String. Motion correction; allowed values: `none`, `volume`, `slice`                                                                    |
+| EddyCurrentCorrection                  | OPTIONAL. String. Eddy current distortion correction; reserved values: `none`, `linear`, `quadratic`, `cubic`                                     |
+| IntensityNormalizationMethod           | OPTIONAL. String. Method (if any) used for intensity normalization                                                                                |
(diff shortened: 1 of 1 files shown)


oesteban:
cc/ @bids-standard/derivatives-mri-dwi

effigies:
Since this is common derivatives, we should be pinging @bids-standard/derivatives.

These questions in particular are relevant to @bids-standard/derivatives-mri, but there has been discussion (apologies for not linking at the moment; I'll try to find it) with @bids-standard/derivatives-electrophys that `SkullStripped` is problematic, as it only applies to MRI derivatives. That seems to be true here, as well, although at least they are optional.

I guess I would say that either we should merge this quickly and address the MRI-centricity in #265, or address it there first, and hold off on this until that's resolved.

robertoostenveld:
`SkullStripped` but also `FieldInhomogeneity` etc. would never apply to MEG, EEG, iEEG, but also not to [BEH](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/07-behavioral-experiments.html), which is also a full-featured (but probably often overlooked) BIDS data type. 

I don't think that any datatype specific metadata fields should be required as per common-derivatives.

oesteban:
I drafted this PR to move out the discussion from BEP16, as we felt we were talking about metadata that could easily be present in other MRI derivatives.

That said, my opinion is that we should remove all these metadata bits for two reasons:
- They do not apply to non-MRI data
- They keep track of provenance, which is something beyond the scope of BIDS. Although it is difficult to keep provenance away in all the cases, I think in this particular time ``Skullstripped`` and the ones I added in this PR should be removed. IMHO it is a task for the workflow developers to allocate a dedicated provenance tracking (NIDM?) and/or insert the sufficient metadata information in more openly-scoped metadata fields/ READMEs / documentation / citation boilerplates.

> I guess I would say that either we should merge this quickly and address the MRI-centricity in #265, or address it there first, and hold off on this until that's resolved.

I'm fine either way, but if ``Skullstripped`` is going away or a better solution is found in #265, then this PR does not make any sense. So I think I'm inclined to figure this out in #265 first and then downstream the decision here and into BEP16 (because we are defining there more of these metadata tracking keys - the non MRI side of it does not apply anymore, but the provenance argument is still valid).

sappelhoff:
@oesteban this PR is still marked for the common derivatives milestone.

How do you want to proceed with this PR? Can it be closed? Should it become an issue for further discussion? Some other idea?

satra:
@oesteban - i do think there is a need to curate a vocabulary around processes. perhaps we can work on that as a separate PR after the common derivatives is merged. as this vocabulary could be used in the provenance BEP #487 .

oesteban:
Sure, what's the course of action regarding BEP016? Should we just remove mentions to these provenance entries?

PR #301, opened 2019-08-13
[ENH] Add ``res`` and ``den`` keywords to indicate resolution of resampled data

oesteban:
Covers the use case when one would like to have several resamplings of a derivative (i.e., keep high-res and low-res versions of data).

cc/ @bids-standard/derivatives


src/05-derivatives/02-common-data-types.md
```diff
@@ -8,7 +8,7 @@ Template:
 <pipeline_name>/
     sub-<participant_label>/
         anat|func|dwi/
-        <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.<ext>
+        <source_keywords>[_space-<space>][_desc-<label>][_res-<index>][_den-<label>]_<suffix>.<ext>
```
effigies: I'm inclined to keep these as close to space as possible.

```suggestion
        <source_keywords>[_space-<space>][_res-<index>][_den-<label>][_desc-<label>]_<suffix>.<ext>
```

src/05-derivatives/02-common-data-types.md
```diff
@@ -62,9 +71,49 @@ makes them invalid (e.g., if a source 4D image is averaged to create a single
...
+| Resolution    | REQUIRED if `res` keyword present. Dictionary of Indexes to Strings. Specifies the interpretation of the resolution keyword. |
+| Density       | REQUIRED if `den` keyword present. Dictionary of Strings to Strings. Specifies the interpretation of the resolution keyword. |
+
+Example JSON file corresponding to `pipeline1/sub-001/func/sub-001_task-rest_run-1_space-MNI305_bold.json` above:
+
+```JSON
+{
+  "SkullStripped": true,
+  "Resolution": {
+    "1": "MNI305 1mm isotropic resolution",
```
effigies: MNI305 seems to be shoehorning the `Space` or `CoordinateSystem` metadata flag in here. If I understood correctly, the point was to dissociate space from sampling.

edickie: Agreed - would limit the description to "1mm isotropic" 

Aside: we have any example/suggestions for what to use for 0.9mm or 0.7mm (this is a pretty common one for anat data)?

robertoostenveld:
Would this also apply to temporal resampling? That can be done both for 4D nifty BOLD, and electrophys?

src/05-derivatives/02-common-data-types.md
oesteban: @effigies - I agree it is a bad example, not so sure it really shadows `Space` or `CoordinateSystem`. Probably something like "Matches MNI305 1mm isotropic resolution" would be better. For the sake of clarity, I'll get MNI305 out of the way. Will fix soon.

@edickie I'll add an example for that. BTW it is important to note that res/den are only to be present if necessary (because several resolutions or densities are generated).

oesteban:
> @robertoostenveld 
> Would this also apply to temporal resampling? That can be done both for 4D nifty BOLD, and electrophys?

The intent here is to describe only spatial samplings. Feel free to suggest modifications to represent temporal samplings too or a new keyword (`time-`?).

Another question is whether there is a use case where you have more than one time-samplings. I can see pipelines having both (spatial) high- and low-resolution pathways, but I haven't seen such a thing with the time dimension of fMRI.

oesteban:
> Why no "space" keyword in the CIFTI example?

Following the spec, that would mean data remain in the original BOLD "space". And that is completely correct regardless of from where and how you brought in the prior knowledge (i.e., the location of volumetric and surface sampling positions). For simplicity, I preferred to keep the concept of space away.

> isn't space-fsaverage still required in the filename or the .json?

Following the previous point, space-fsaverage here would be a misrepresentation of the data

> Also - the CIFTI example should also remove the reference template (space) from the densitiy description?

I can add the language _Matched with_ as I did in previous examples, if that feels clearer. The intent of this description is to quickly tell how densely you sample the manifold ("what we are looking at? - a surface of 41k points per hemisphere with one corresponding fMRI time series per vertex"), it is not to keep track of the provenance ("okay, this is 41k vertices per hemisphere because it comes from fsaverage6").

robertoostenveld:
> The intent here is to describe only spatial samplings. Feel free to suggest modifications to represent temporal samplings too or a new keyword (`time-`?).
> 
> Another question is whether there is a use case where you have more than one time-samplings. I can see pipelines having both (spatial) high- and low-resolution pathways, but I haven't seen such a thing with the time dimension of fMRI.

It is the principle rather than the specific topic of resampling along a specific dimension: I believe that keywords limited to one modality are better not be part of common derivatives, but be under the modality specific derivatives. Otherwise all keywords of all modalities (including future BEPs for other data types) should be documented in common derivatives, which would become a mess. 

Common derivatives should describe metadata that is common and that - in principle - could apply to all modalities. E.g. the operating system on which the pipeline ram, the software details, component versions, the [type of CPU or GPU](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0038234),  etc. Hence I would expect `derivatives/01-introduction.md` to be relatively small and `derivatives/02-mri.md` more elaborate.

oesteban:
This addition applies to PET/SPECT, CT, derived contrast maps, or any 3D
object with samples in a regular grid or a surface..

robertoostenveld:
> This addition applies to PET/SPECT, CT, derived contrast maps, or any 3D object with samples in a regular grid or a surface..
> […](#)

yes, which means that right now it only covers 1 of the 5 modalities in the specification  (since CT and PET are not yet part of the specification). 

I can see the point that some derivative extensions apply to more than one (i.e. MRI, PET and CT; we would also have that with MEG, EEG and iEEG, and I would also expect BEH and EYE to have that overlap. So `common-imaging`, and `common-electrophys` and `common-behavioral` make sense. But `common-to-all` should be distinguished from imaging.

robertoostenveld:
> This addition applies to PET/SPECT, CT, derived contrast maps, or any 3D object with samples in a regular grid or a surface..
> […](#)

Yes, which means that it only applies to 1 out of the 5 modalities that is presently included in the specification (PET and CT are not yet). 

It makes sense to have  `common-imaging-derivatives` (for MRI, PET and CT), `common-electrophys-derivatives` (for MEG, EEG and iEEG) and `common-behavioral-derivatives` (for BEH and eyetracking), but `common-for-all-derivatives` should not prescribe details for either one, unless it really applies to that derivative.

effigies:
Okay, so it sounds like @robertoostenveld's primary concern here is with the `res` keyword being specifically spatial, as temporal resolution is the resolution of interest in electrophysiological analysis. Would it be reasonable to recast this `res` as `sres` for spatial resolution (or `vres` for volumetric resolution) and save `tres` for temporal resolution?

They're not quite as readable (especially `tres`, which is an actual word...), so I'm happy to hear alternative suggestions.

Another option is for `res-` to have different meanings in different modalities, but this would require it to be the case that you won't generally want to combine data across modalities and indicate that you have, for example, estimated a 50Hz time series in a 1mm isotropic voxel grid. I don't know how likely that would be, so I can't really issue a preference.

> Common derivatives should describe metadata that is common and that - in principle - could apply to all modalities. E.g. the operating system on which the pipeline ram, the software details, component versions, the type of CPU or GPU, etc. Hence I would expect derivatives/01-introduction.md to be relatively small and derivatives/02-mri.md more elaborate.

@robertoostenveld I'm going to be reworking the MR-specific sections of common derivatives to make just this sort of distinction. I'll ping all derivatives folks to have a look when I have done.

robertoostenveld:
I don't have a preference for a specific name or so, and don't object to `res` itself. It is just that it should be clear that the keyword applies only to imaging data, not to behavioral or electrophys. I have had many electrophys people expressing confusion about BIDS documentation that is MRI biased, asking whether fields like `acq`, `rec`, `dir`, `mod` etcetera also apply for their data.

Oh, now that I think of it, we put quite some effort into making [this table](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html), not only to document their possible presence and order, but also to ensure consistency between similar types (like EEG and iEEG). The table is relevant to keep in mind for mapping keys (as in `<key>-<value`) to datatypes/formats. 

Regarding the `res` equivalent for electrophys: that is something to discuss with the electrophys folks, but I expect them to phrase this as "resampling" rather than "changing the resolution", and  that the value would rather be expressed in Hz (i.e. rate) rather than seconds (i.e. duration between samples). I.e., I suspect `fsample` to be preferred by the electrophys field. 

If temporal resampling were an issue for processing 4D NIFTI (which it is not at the moment), then it would not surprise me if fMRI folks would prefer that to be expressed in seconds (duration between samples, i.e. TR) rather than Hz (rate). In that case I would also be fine with another keyword than `fsample` for derived 4D bold data. 

This example also makes me realize that the proposed `res` could not apply to the time axis any way, since the `<value>` must be specified in mm. ... or would the value be a string including the units? How about the value for `den`? If the units are not specified in the value, where would they be specified? Here I am thinking e.g. about high-res [MRI at 100um](https://www.ncbi.nlm.nih.gov/pubmed/12385770) or [CT at 30um](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6662504/). [Units](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#units), see also [the appendix](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html) are tricky to get documented properly.

oesteban:
The `res` and `den` keywords apply to any 2D and 3D data, regardless of the origin. `den` could be used, for example, in EEG data if having two processing paths where one uses the full array of receptors and the other some subsample. And `res` could have a use in non-parametric solutions to the inverse problem in EEG. Resolution/surface density alludes to signal processing concepts rather than modalities. We are talking of the 80 part of the 80/20. No way this is a 5% use case.

> This example also makes me realize that the proposed res could not apply to the time axis any way, since the <value> must be specified in mm. ... or would the value be a string including the units? How about the value for den? If the units are not specified in the value, where would they be specified? Here I am thinking e.g. about high-res MRI at 100um or CT at 30um. Units, see also the appendix are tricky to get documented properly.

This is not what has been proposed here. The proposed spec is `res-<index>` and index maps to a dictionary where index are the keys and the values are just descriptions. Similarly for `den-<label>`. The units must be encoded in the file format metadata.

Regarding `tres`/`fsample`: this is why I asked for examples at first. Although I don't see the use (this key is for disentangling resolutions when derivatives with several of them must be generated). If only one resolution is present, then that was previously well captured by the spec.

My fear here is that dividing the common derivatives again by broader data types will really extend the already long time it is taking for us to get this finished.

edickie:
I believe that HCP publish source reconstructed MEG derivatives as CIFTI files. For these images the `res` and `den`  would definitely apply..

I second the idea that the temporal resolution is also something that can be modified in many workflows (esp EEG) - but also physio confounds from MR and stimulus presentation.. So it does make sense to add a general keyword for it... CIFTI file format uses "timestep" to refer to this in their metadata..

oesteban:
> @edickie 
> I second the idea that the temporal resolution is also something that can be modified in many workflows (esp EEG) - but also physio confounds from MR and stimulus presentation.. So it does make sense to add a general keyword for it... CIFTI file format uses "timestep" to refer to this in their metadata

I'm also totally in favor of adding a keyword for temporal resolution. The broader question is whether that keyword is going to be ever used. If EEG (or whatever modality) do not play with several resolutions at a time, then they are not necessary.

My thinking behind `res` (and probably with even more reasons `den`) is to allow fMRIPrep users to generate preprocessed data:
 - in MNI space; and
 - with two versions: one keeping the original resolution, one with the resolution matching MNI space.
In such a situation, the only option is to use `desc` to differentiate those two (and drop the `desc-preproc` we currently use). Although using `desc` for this seems pretty inappropriate.

jasmainak:
The discussion is already too involved. But I have a couple of broad questions:

- what is the criteria for something to be added as a keyword rather than a key in the sidecar with a new `desc` label?

> And res could have a use in non-parametric solutions to the inverse problem in EEG. Resolution/surface density alludes to signal processing concepts rather than modalities. We are talking of the 80 part of the 80/20. No way this is a 5% use case.

- what is the criteria for deciding if something is 80 or 20? :-) In MEG source modeling, the canonical way we talk about resolution is in source spaces and how many times it has been subdivided rather than in terms of mm. See here for instance: https://martinos.org/mne/stable/manual/cookbook.html?highlight=ico5#anatomical-information. But this is surface source spaces and probably doesn't apply to CIFTI/NIFTI but I can't say from my superficial read of the PR.

nicholst:
> what is the criteria for something to be added as a keyword rather than a key in the sidecar with a new desc label?

Simplest answer I can think of is: By default, a metadata element should go to the side car; only becomes a filename key if (1) the element is needed to avoid two distinct files having the same name, or (2) the element is simply so fundamental that use of the file would be impaired if omitted.

jasmainak:
Right, however you could use `desc` to distinguish between files having the same name. From the specification:

> The desc keyword is a general purpose field with freeform values. To distinguish between multiple different versions of processing for the same input data the desc keyword should be used.

oesteban:
> @jasmainak In MEG source modeling, the canonical way we talk about resolution is in source spaces and how many times it has been subdivided rather than in terms of mm.

Then the proposed `den` (_density_) keyword applies to MEG. As I mentioned above, the concept of units is irrelevant as these keywords can only be used when there are two or more resolutions/densities as a result of your pipeline. E.g., your pipeline generates a derivative with two densities: `'oct6'` and `'ico5'`. This would allow these two files representing two "_twin_" objects of the same data type, and same all other properties except for the sampling, to be called `den-4k` and `den-10k`. Please look at the examples for surfaces in this PR.

> @jasmainak what is the criteria for deciding if something is 80 or 20? :-)

I don't think there is. I just meant that, if I were asked what the coverage (or application) of this PR is then I'd say it falls on the 80 side because I believe this is necessary for imaging data and potentially for many other modalities. I'd see this a 20-side if it applied only to MRI, for instance.

This PR applies to any nonparametric 2D and 3D sampled data regardless of the modality. Happy to extend it to cover 1D or ND+t if we can foresee the utility, and to find alternative names to the keywords if they misrepresent some modality.

> @jasmainak  Right, however you could use `desc` to distinguish between files having the same name.

Can you give an example of how would you use `desc` here? - We discussed this at large in the BIDS meeting last year. The impression I got is that `desc` should be avoided if a keyword in a controlled vocabulary can be defined to cover some specific and relevant concept (e.g., `space`). Otherwise, we should drop `space` and, since several `desc` are not allowed (I believe), then end up with monstruosities like `desc-MNI152NLin6Asym2mmBrainSmoothed`.

I propose this after a large discussion in the area of imaging templates, where several versions of the template (which differ in resolution) can be used. If space is an entity (physical world), I thought resolution/density should have their entity (digital world).

oesteban:
BTW, with the changes of the latest commit I think `res` could be used for timeseries too.

jasmainak:
the changes + examples seem fine to me.

regarding the `desc` label, it probably ties into the discussion here: https://github.com/bids-standard/bids-specification/issues/309 where @satra quotes discussion from the derivative meeting:

> filename globbing is not a recommended pattern outside of the base filename 

I don't quite agree with the above statement. But if this were true, then `desc-MNI152NLin6Asym2mmBrainSmoothed` shouldn't matter since you use it only to disambiguate and the real information is in the sidecar.

That said, I don't want to block this PR any further if @robertoostenveld and others are okay with the changes.

src/05-derivatives/02-common-data-types.md
```diff
@@ -8,7 +8,7 @@ Template:
 <pipeline_name>/
     sub-<participant_label>/
         anat|func|dwi/
-        <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.<ext>
+        <source_keywords>[_space-<space>][_res-<index>][_den-<label>][_desc-<label>]_<suffix>.<ext>
```
effigies: The indices thing feels weird, as opposed to a (potentially) descriptive label, given that the indices don't mean anything.

src/05-derivatives/02-common-data-types.md
```diff
@@ -62,9 +71,50 @@ makes them invalid (e.g., if a source 4D image is averaged to create a single
...
 addition, all processed files include the following metadata JSON fields:
 
-| **Key name**  | **Description**                                                                                 |
-| ------------- | ----------------------------------------------------------------------------------------------- |
-| SkullStripped | REQUIRED. Boolean. Whether the volume was skull stripped (non-brain voxels set to zero) or not. |
+| **Key name**  | **Description**                                                                                                              |
+| ------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| SkullStripped | REQUIRED. Boolean. Whether the volume was skull stripped (non-brain voxels set to zero) or not.                              |
+| Resolution    | REQUIRED if `res` keyword present. Dictionary of Indexes to Strings. Specifies the interpretation of the resolution keyword. |
+| Density       | REQUIRED if `den` keyword present. Dictionary of Strings to Strings. Specifies the interpretation of the resolution keyword. |
```
effigies: Our usual way to disambiguate entities is something like:

`res-1_bold.json`:
```
{
  "Resolution": "abc"
}
```

`res-2_bold.json`:
```
{
  "Resolution": "xyz"
}
```

The switch to a dictionary indexed by strings doesn't appear to be justified anywhere in the proposed text. And for resolution, if it's an index, should it not be an integer, rather than a string of digits?

oesteban: I'm very happy to elaborate more on this:

- Why an index: to allow developers to impose an implicit hierarchy. However, I would agree with alternatives such as `res-high`, `res-low`, `res-mid`... but that vocabulary would be very limited. I wouldn't allow `res-2x2x2` for various reasons. First is that very often we will need decimals and `res-07x07x07` seems like a recipe for disaster. Second, file formats generally address this issue, so it gives plenty of room for the discrepancy between this value and what you find in the file's header. To me, the index solution seemed easy.

However, to allow human-readable descriptions in the JSON file, they need to be converted to string indexes for the dictionary (JSON does not allow integer keys, does it?).

I can add your metadata proposal, but I would not replace the dictionary because that allows you to place the metadata at a higher level of the hierarchy (ie., describe res-1 and res-2 in the same, unique file).

- I like the idea of free-form descriptions of each index. The rationale is that - software should read the actual resolution (and units that seem to be of concern) from the file format. Humans, however, might check quickly this description field without the need of opening the file (e.g., `nib-ls` for niftis with nibabel). I would be very happy to change my mind if a better idea is proposed, or improving this PR with a better description of the current proposal - which seems necessary at this point.

src/05-derivatives/02-common-data-types.md
oesteban: As mentioned below, indices allow for the developer to impose some explicit order they want.

robertoostenveld:
> In MEG source modeling, the canonical way we talk about resolution is in source spaces and how many times it has been subdivided rather than in terms of mm. See here for instance: https://martinos.org/mne/stable/manual/cookbook.html?highlight=ico5#anatomical-information.  But this is surface source spaces and probably doesn't apply to CIFTI/NIFTI but I can't say from my superficial read of the PR.

Yes, this does also apply to MEG source results in [CIFTI](https://www.nitrc.org/forum/attachment.php?attachid=341&group_id=454&forum_id=1955) (which I happen to have co-authored) and to HCP. In HCP we share processed results both as "dense" (i.e. regular volumetric for beamforming, surface-based for minimum norm estimates) and as "parcellated" (esp. for the connectivity matrices). The `res` and `den` apply to the spatial aspect of the dense representation in CIFTI.  

In HCP we don't consider source reconstructions as (merely) _pre_-processed, but as (fully) _processed_, since a lot decisions have to be made in the source-level processing that fundamentally affect the output. See [this figure](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3784249/figure/F2/) for HCP-MEG: the functional preprocessing is considered to be the box at the top (shared between task and rest) and the anatomical preprocessing the box at the left. `Res` and `den` play a role in the box at the left, and thereby also affect the functional outcomes. 

Similar as other key-value pairs that we tried to make [consistent between modalities](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html), we should do the same for 'res' and 'den'. They would apply to MEG/EEG source level results- the output of a multi-modal analysis with both MRI and electrophys  - but that does not mean we have to stretch `res` and `den` to EEG-only datasets like [this one](https://github.com/bids-standard/bids-examples/tree/master/eeg_matchingpennies) from @sappelhoff.

src/05-derivatives/02-common-data-types.md
robertoostenveld: Where is the index made explicit, i.e. where is the lookup-table that allows me to see what index value "1" signifies? Is the lookup table a TSV or a JSON?

Is the label in `den-label` free to choose by the researcher, or is it limited to a controlled vocabulary?

robertoostenveld: since the format only specifies `anat|func|dwi`, does this mean that `res` and `den` don't apply to `meg|eeg|ieeg|beh`?

emdupre: I believe the index look-up is in the sidecar, [specifically here](https://github.com/bids-standard/bids-specification/pull/301/files#diff-ee997314dc6a77ebab0d882ed7945467R77).
On my reading, it seems that `den` is free text (rather than a controlled vocabulary). I'm not sure what the controlled vocabulary would be, here.

src/05-derivatives/02-common-data-types.md
emdupre:
```suggestion
| Density       | REQUIRED if `den` keyword present. Dictionary of Strings to Strings. Specifies the interpretation of the density keyword. |
```

src/05-derivatives/02-common-data-types.md
oesteban: Updated to include the additional modalities.

I'm also thinking this should be updated when the BEP about connectivity is merged in, as resolution is something connectivity analyses surely care about.

robertoostenveld: thanks @emdupre, that clarifies.

I think it would be better if both `res` and `den` would allow strings rather than digits. This allows them to carry explicit meaning (like "1mm", "2mm"). Digits to indicate indices suffer from some ambiguities, i.e. should they start Python/C style 0-offset or MATLAB/Fortran style 1-offset? And is it allowed to use them in non-successive order like 1, 3, 5, 7?  The developer can (and probably should) always start by reading the JSON, and subsequently loop over the keys of the dictionary.

oesteban: The proposal of an index for `res` here follows that of `run` (https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#the-run-entity), i.e., start at 1, non-successive order disallowed.

To find the actual meaning of each resolution, the user MUST look into the file format, to avoid that you say `res-1x1x1mm` but then the headers say that the voxel sizes are 0.7x0.7x0.7.

Having labels like "1mm" seemed pretty limited to me, and quite misleading - for instance if your sampling is oblique (i.e., a NIfTI file with a shearing affine).

The JSON description field is free-from to preempt using the dictionary to shadow this information in the file format.

satra: @oesteban - unlike `run` which has some order to it, `res` would not have any specific order.   if the intent is to define a mapping to something to something more interpretable, then the use of the `index` is for disambiguation. in this scenario, wouldn't `desc` suffice for disambiguation?

if `index` is not a **controlled** vocabulary, people would have to look up, to do search. i.e. a glob pattern would not necessarily work.

if the intent is simply to say this is a regularly sampled n-dgrid, perhaps then `res-700mm` would be ok, with number of dimensions not being important for disambiguation. this makes the value an interpretable whole number without decimals and with units (we can follow the common units terms). this does pose the additional constraint that the number should match whatever is in the file, but that is something a future validator could check.

oesteban: > wouldn't `desc` suffice for disambiguation?

Only when it is not already used for disambiguation of some other semantic. The concept of resolution/density on digital data [is fundamental](https://github.com/bids-standard/bids-specification/pull/301#issuecomment-522264904) and justifies extracting it from `desc`. In my opinion, `desc` is more about the actual content of the file and `res` tells you about the structure of the data. However, in my opinion, the actual resolution (regardless number of dimensions, or units) is a detail that is better kept with the file format (and unnecessary to fully understand that we have two or more resolutions of the same reality).

> if `index` is not a controlled vocabulary, people would have to look up, to do search. i.e. a glob pattern would not necessarily work.

Okay, you mean that you could have two `res-1` pointing to different actual resolutions? Well, I don't think that is a problem, as long as you don't give `res` more powers than: _there is another file with the same content but sampled differently_. If you write a JSON metadata dictionary at the top level to describe the indexes then you are enforcing that all files sharing the index of `res` MUST have the same resolution and that all the subfolders have all the resolution indexes when `res` is present.

My conflict with `res-700mm` is that, like `desc` talks more about the content, not the structure of the data (and I see these labels very impractical too).

These explanations are probably not sufficiently described in the pull request, happy to iterate over it and make them more clear (or change my mind if I'm missing the point).

satra: > you could have two res-1 pointing to different actual resolutions?

indeed. yes, we would need a validator that enforces this, and figures out inheritance principles and such mundane things.

on the more general note, i think the reason you want to include `res` is for disambiguation (i.e. the use case of a file in two different resolutions of MNI template/fsaverage/etc.,.), but the disambiguating info is the content (the resolution of the data), which is likely already in the file itself. so whether `res` talks resolution via an `index`, or not, it is really talking about the data !

i think you already agreed with me :)
> concept of resolution/density on digital data is fundamental and justifies extracting it from desc

oesteban: After thinking a bit more about this (and similar ideas from @robertoostenveld above) - would you agree in relaxing the value of `res` to be a `<label>` (and hence, consistent with `den-`, which is nice), but including the recommendation to use one-based, consecutive indexes?

satra: i'm still missing what the 1-based consecutive index recommendation really helps with, given that labels can be arbitrary. and how would you differentiate across datasets that follow the recommendation and those that don't?

src/05-derivatives/02-common-data-types.md
```diff
@@ -62,9 +78,54 @@ makes them invalid (e.g., if a source 4D image is averaged to create a single
...
+
+Example JSON file corresponding to
+`pipeline1/sub-001/func/sub-001_task-rest_run-1_space-MNI305_bold.json` above:
+
+```JSON
+{
+  "SkullStripped": true,
+  "Resolution": {
+    "hi": "Matched with high-resolution T1w (0.7mm, isotropic)",
+    "lo": "Matched with original BOLD resolution (2x2x3 mm^3)"
```
satra: i like these above examples as readable descriptors of resolution.

src/05-derivatives/02-common-data-types.md
```diff
@@ -62,9 +78,54 @@ makes them invalid (e.g., if a source 4D image is averaged to create a single
...
+```JSON
+{
+    "SkullStripped": true,
+    "Resolution": {
+        "1": "Matched with MNI152NLin6Asym 1.6mm isotropic",
+        "2": "Matched with MNI152NLin6Asym 2.0mm isotropic"
+    },
+    "Density": {
+        "10k": "10242 vertices per hemisphere (5th order icosahedron)",
+        "41k": "40962 vertices per hemisphere (6th order icosahedron)"
```
satra: just like these say `10k` and `41k` someone could have used `1p6` and `2p0` for `Resolution` as well, right? or conversely i could have used `1` and `2` for `Density`.

btw, this may be the first use of a dictionary as a value in bids. i know this was discussed a long time back in the context of pybids, and in our derivatives meeting to stay away from, so may be worth a little ping to bep leads and others.

oesteban: > just like these say 10k and 41k someone could have used 1p6 and 2p0 for Resolution as well, right? or conversely i could have used 1 and 2 for Density.
Sure

> btw, this may be the first use of a dictionary as a value in bids. i know this was discussed a long time back in the context of pybids, and in our derivatives meeting to stay away from, so may be worth a little ping to bep leads and others.

I just added a suggestion from @effigies that would provide an alternative to dicts. However, it seems useful to have all resolutions in just one place.

src/05-derivatives/02-common-data-types.md
oesteban: > i'm still missing what the 1-based consecutive index recommendation really helps with, given that labels can be arbitrary.

1. Gives you an idea of how many different resolutions there are with a quick look. Other options don't.
2. Allows the developers to impose an order they might be interested in (e.g., higher resolution first).
3. Preempts inconsistency between the label and the data (e.g., having `res-333mm` for a NIfTI with 2x2x2 or that has an affine with shearing)

> how would you differentiate across datasets that follow the recommendation and those that don't?

The recommendation does not allow you to remove the metadata in the JSON. Other than that, why would you need to know whether the dataset sticks with the recommendation or not?

edickie: Umm - I'm a little confused - I also believe that it should be a `res-<label>`  and not `res-<index>`.

I simple alphabetical sort of labels is probably all the ordering we would need?

I really hate the idea of res "2x2x2mm" being associated with index 3...that seems way too confusing.

oesteban: I modified the PR and now it proposes `res-<label>`. However, I left a recommendation indicating the use of indexes. The question is whether that should be removed or not. I could also expand the [recommendation box in the docs](https://1193-150465237-gh.circle-artifacts.com/0/home/circleci/project/site/05-derivatives/02-common-data-types.html#processed-coregistered-andor-resampled-volumes) to include the three reasons I gave above.

src/05-derivatives/02-common-data-types.md
```diff
@@ -30,13 +30,29 @@ Note that even though `space` and `desc` are
...
+**Sampling**. When two or more instances of a given derivative are provided
+with resolution (or surface sampling density) being the only difference
+between them, then the `res` (for *resolution* of regularly sampled N-D data)
+and/or `den` (for *density* of non-parametric surfaces) SHOULD be used to avoid
+name conflicts.
+Note that only files combining both regularly sampled (e.g., gridded) and surface
+sampled data (and their downstream derivatives) are allowed to present both `res`
+and `den` keywords simultaneously.
+
+!!! tip "Recommendation"
```
robertoostenveld: I think this is better phrased using RFC 2119 terms for consistency with documentation elsewhere and [overall common principles](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html).

robertoostenveld: Note that `RECOMMENDED` is an option there.

robertoostenveld:
Should the common derivatives description explain that derived data can be of the same type, but also might be of an other type than the 8 types listed for BIDS-raw in the [common principles](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html)?

_5. Data type - a functional group of different types of data. In BIDS we define eight data types: func (task based and resting state functional MRI), dwi (diffusion weighted imaging), fmap (field inhomogeneity mapping data such as field maps), anat (structural imaging such as T1, T2, etc.), meg (magnetoencephalography), eeg (electroencephalography), ieeg (intracranial electroencephalography), beh (behavioral)._

src/05-derivatives/02-common-data-types.md
oesteban: I'll revise and define this as OPTIONAL

effigies:
I have merged `common-derivatives` into this branch. @oesteban please verify that this still reflects your intended change.

Could all participants in this discussion re-review this? It was refactored significantly due to the separation of imaging derivatives.

src/05-derivatives/02-common-data-types.md
```diff
@@ -8,7 +8,7 @@ Template:
 <pipeline_name>/
     sub-<participant_label>/
         <datatype>/
-            <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.<ext>
+            <source_keywords>[_space-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.<ext>
```
effigies: From previous discussion, it looks like perhaps this ought to be moved inside 03-imaging...

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,119 @@ extent and resolution.
 
 ## Preprocessed, coregistered and/or resampled volumes
 
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<participant_label>/
+        <datatype>/
+            <source_keywords>[_space-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.<ext>
```
effigies: Moved res/den to imaging.

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,119 @@ extent and resolution.
...
+```
+
+
 Volumetric preprocessing does not modify the number of dimensions, and so
 the specifications in [Preprocessed or cleaned data][common_preproc]
 apply.
-In addition, all processed files include the following metadata JSON fields:
+The use of surface meshes and volumetric measures sampled to those meshes is
+sufficiently similar in practice to treat them equivalently.
```
effigies: This could probably be worded better.

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,119 @@ extent and resolution.
...
+*density* of non-parametric surfaces) SHOULD be used to avoid name conflicts.
+Note that only files combining both regularly sampled (e.g., gridded) and surface
+sampled data (and their downstream derivatives) are allowed to present both `res`
+and `den` keywords simultaneously.
+
+!!! tip "Recommendation"
+    Although the `res` entity accepts any BIDS-valid `<label>`, it is
+    recommended to use one-based, consecutive indices as label to avoid
+    inconsistencies between the value of `res` and the metadata embedded
+    within the actual file.
```
effigies: While on 02-common-data-types, there was this discussion: https://github.com/bids-standard/bids-specification/pull/301#discussion_r319862076

I'm not sure that this still applies, as the example is now `hi`/`low`, not one-based... I think we can probably just drop this tip.

src/05-derivatives/02-common-data-types.md
```diff
@@ -33,14 +33,6 @@ be defined to avoid name conflict with the raw file.
...
 Examples:
 
-```Text
-pipeline1/
-    sub-001/
-        func/
-            sub-001_task-rest_run-1_space-MNI305_bold.nii.gz
-            sub-001_task-rest_run-1_space-MNI305_bold.json
-```
-
```
effigies: Should probably restore this?

src/05-derivatives/03-imaging.md
oesteban: I feel this recommendation is still relevant, but not as relevant as to think it through to get a strong argument in favor - so I'm fine either way.

src/05-derivatives/02-common-data-types.md
oesteban: sure

src/05-derivatives/03-imaging.md
oesteban: Where do you see the problem is? (just to propose amends in the right direction)

src/05-derivatives/03-imaging.md
effigies: Removed it for now. If it seems advisable, we can re-add. It just doesn't seem to match the surrounding text.

src/05-derivatives/03-imaging.md
effigies: Just feels a little awkwardly worded. Maybe it could be written more persuasively. But I'm inclined to go for it and let people address wording later.

effigies:
@oesteban @edickie What's the status of this conversation?

> > Why no "space" keyword in the CIFTI example?
> 
> Following the spec, that would mean data remain in the original BOLD "space". And that is completely correct regardless of from where and how you brought in the prior knowledge (i.e., the location of volumetric and surface sampling positions). For simplicity, I preferred to keep the concept of space away.
> 
> > isn't space-fsaverage still required in the filename or the .json?
> 
> Following the previous point, space-fsaverage here would be a misrepresentation of the data
> 
> > Also - the CIFTI example should also remove the reference template (space) from the densitiy description?
> 
> I can add the language _Matched with_ as I did in previous examples, if that feels clearer. The intent of this description is to quickly tell how densely you sample the manifold ("what we are looking at? - a surface of 41k points per hemisphere with one corresponding fMRI time series per vertex"), it is not to keep track of the provenance ("okay, this is 41k vertices per hemisphere because it comes from fsaverage6").

Now that `fsLR` is in common derivatives, should we add a space to the CIFTI files?

effigies:
@robertoostenveld @satra @emdupre Any remaining issues on your end before merging to `common-derivatives`?

satra:
@effigies - without a reference to  `space` in the cifti json file, how would i know the underlying geometry of the cifti file, both the surface part and the volume part?

effigies:
@satra If I understand [Common file level metadata fields](https://bids-specification.readthedocs.io/en/common-derivatives/05-derivatives/01-introduction.html#common-file-level-metadata-fields) correctly, if `space-` is a [standard template](https://bids-specification.readthedocs.io/en/common-derivatives/99-appendices/08-coordinate-systems.html#standard-template-identifiers), then that defines it. If it is not, then it requires a `SpatialReference` metadata field.

@oesteban?

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,111 @@ extent and resolution.
...
+{
+    "SkullStripped": true,
+    "Resolution": {
+        "1": "Matched with MNI152NLin6Asym 1.6mm isotropic",
+        "2": "Matched with MNI152NLin6Asym 2.0mm isotropic"
+    },
+    "Density": {
+        "10k": "10242 vertices per hemisphere (5th order icosahedron)",
+        "41k": "40962 vertices per hemisphere (6th order icosahedron)"
+    }
```
satra: in this example, how would spatial reference be included?

each density could have its own spatial reference.
each resolution could also have its own spatial reference.

oesteban: What do you mean?

This?
```Text
pipeline1/
    sub-001/
        func/
            sub-001_task-rest_run-1_den-10k_bold.dtseries.nii
            sub-001_task-rest_run-1_den-41k_bold.dtseries.nii
            sub-001_task-rest_run-1_bold.dtseries.json
```

satra: my question is how do i get the geometry (faces/triangles) of the `den-10k` vertices?

effigies: That's a fair point, and seems to argue against this accumulated metadata mode.

I'm also unclear on what `SpatialReference` would point to for a grayordinate CIFTI. It could point to something like https://github.com/templateflow/tpl-fsLR/tree/f82f453, but that has more than three files in it.

We could allow `SpatialReference` to be an object, e.g.,

```
"SpatialReference": {
    "left": "/path/to/hemi-L.surf.gii",
    "right": "/path/to/hemi-R.surf.gii",
    "subcortical": "/path/to/subcortical_atlas.nii"
}
```

robertoostenveld:
> @robertoostenveld @satra @emdupre Any remaining issues on your ends before merging to `common-derivatives`?

I had a look at https://bids-specification.readthedocs.io/en/common-derivatives/05-derivatives/01-introduction.html and the other two pages. 

In the "Introduction" I am not so happy with the phrasing *SpatialReference = REQUIRED when a custom template or file is used*. 

I would prefer *REQUIRED in case a custom reference image was used*.  

Other than that I am fine with it. From the @bids-standard/derivatives-electrophys  perspective I believe that the common derivatives and the anatomical MRI will allow a further/future specification for derivatives of that data as well.

src/05-derivatives/03-imaging.md
oesteban: I apologize it took me almost one week to mull over this.

- :+1: to @effigies' suggestion for the `SpatialReference`.
- Regarding @satra's question, the more I think it through, the more convinced I am that "greyordinates" are a new "manifold?" supporting a sampling scheme. Therefore, to be consistent they would allow only the density key. I believe HCP Pipelines made a good call for the valid values (e.g., den-170k, den-91k), based on the total number of samples. WDYT? This, in conjunction with the `atlas-fsLR` would allow for univocal designation of reference spaces without specifying a `SpatialReference`. Then, for Frankenstein grayordinates like _fMRIPrep_'s, one could always use @effigies's spec for `SpatialReference`.

oesteban:
I'm fine with @robertoostenveld suggestion. Will update including that, when https://github.com/bids-standard/bids-specification/pull/301#discussion_r356308259 is more clear - if everyone is :+1: with the suggestion.

src/05-derivatives/03-imaging.md
edickie: I think I agree with where things are going with spatial reference - although we may have the problem that the surface for the spatial reference can be multiple things..(the individuals pial surface, the individual white surface, the individual midthickness surface - as well as the versions of those within template flow). Although they _should_ all have the same density (# faces and # vertices) so maybe that is no big deal?

oesteban: @satra how would you go around the problem you pointed out?

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,111 @@ extent and resolution.
...
+
 Volumetric preprocessing does not modify the number of dimensions, and so
 the specifications in [Preprocessed or cleaned data][common_preproc]
 apply.
-In addition, all processed files include the following metadata JSON fields:
+The use of surface meshes and volumetric measures sampled to those meshes is
+sufficiently similar in practice to treat them equivalently.
+
+When two or more instances of a given derivative are provided with resolution
+(or surface sampling density) being the only difference between them, then the
```
effigies:
```suggestion
or surface sampling density being the only difference between them, then the
```

src/05-derivatives/03-imaging.md
```diff
@@ -5,14 +5,111 @@ extent and resolution.
...
+            sub-001_task-rest_run-1_space-MNI305_bold.json
+```
+
+The following metadata JSON fields are defined for preprocessed images:
+
+| **Key name**  | **Description**                                                                                                              |
+| ------------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| SkullStripped | REQUIRED. Boolean. Whether the volume was skull stripped (non-brain voxels set to zero) or not.                              |
+| Resolution    | REQUIRED if `res` keyword present. Dictionary of Strings to Strings. Specifies the interpretation of the resolution keyword. |
+| Density       | REQUIRED if `den` keyword present. Dictionary of Strings to Strings. Specifies the interpretation of the density keyword.    |
```
effigies: Inconsistent with a per-res/den JSON sidecar, in which case it's just a string.

Assuming we want to permit two approaches, given that it's complicated, I think we could describe this as:

```
REQUIRED if <entity> is present. String or dictionary of label to string. See <section name> for details.
```

effigies:
Unfortunately, my computer crashed with my notes from our meeting last week and I did not use a program with auto-save, so if anybody has those, feel free to share.

The upshot of the meeting, however, was that I would be working to put together a concrete use case and a tool that would allow us to go from a simple JSON representation of the relevant spaces to a sparse, but usable, Connectome-Workbench-style `.wb.spec` file.

To this end, I've processed some data with fMRIPrep 20.0.0 with `--cifti-output` enabled, and will be manually augmenting with a JSON sidecar.

```
fmriprep/sub-01
├── anat
...
├── func
...
│   ├── sub-01_task-repeatchange_run-1_space-fsLR_den-91k_bold.dtseries.json
│   ├── sub-01_task-repeatchange_run-1_space-fsLR_den-91k_bold.dtseries.nii
...
```

Currently the JSON file contains:

```JSON
{"space": "HCP grayordinates", "surface": "fsLR", "volume": "MNI152NLin6Asym", "surface_density": "32k", "grayordinates": "91k"}
```

As this is a case where the image is not part of the outputs, it makes sense to use URLs to indicate where a file can be retrieved. I would suggest making `fmriprep/space-fsLR_den-91k_bold.dtseries.json`:

```JSON
{
    "SpatialReference": {
        "VolumeAll": "https://github.com/templateflow/tpl-MNI152NLin6Asym/blob/796ab78/tpl-MNI152NLin6Asym_res-02_T1w.nii.gz",
        "CortexLeft": "https://github.com/templateflow/tpl-fsLR/blob/f82f453/tpl-fsLR_hemi-L_den-32k_midthickness.surf.gii",
        "CortexRight": "https://github.com/templateflow/tpl-fsLR/blob/f82f453/tpl-fsLR_hemi-R_den-32k_midthickness.surf.gii"
    }
}
```

For now I'll assume that URLs are acceptable, and say this should translate into:

```XML
<?xml version="1.0" encoding="UTF-8"?>
<CaretSpecFile Version="1.0">
   <MetaData>
   </MetaData>
   <DataFile Structure="CortexLeft"
             DataFileType="SURFACE"
             Selected="true">
      https://github.com/templateflow/tpl-fsLR/blob/f82f453/tpl-fsLR_hemi-L_den-32k_midthickness.surf.gii
   </DataFile>
   <DataFile Structure="CortexRight"
             DataFileType="SURFACE"
             Selected="true">
      https://github.com/templateflow/tpl-fsLR/blob/f82f453/tpl-fsLR_hemi-R_den-32k_midthickness.surf.gii
   </DataFile>
   <DataFile Structure="Invalid"
             DataFileType="VOLUME"
             Selected="true">
      https://github.com/templateflow/tpl-MNI152NLin6Asym/blob/796ab78/tpl-MNI152NLin6Asym_res-02_T1w.nii.gz
   </DataFile>
</CaretSpecFile>
```

A second, per-subject use case might look like `fmriprep/sub-01/func/sub-01_bold.dtseries.json`:

```JSON
{
    "SpatialReference": {
        "VolumeAll": "../anat/sub-01_desc-preproc_T1w.nii.gz",
        "CortexLeft": "../anat/sub-01_space-fsnative_hemi-L_den-32k_midthickness.surf.gii",
        "CortexRight": "../anat/sub-01_space-fsnative_hemi-R_den-32k_midthickness.surf.gii"
        }
}
```
and

```XML
<?xml version="1.0" encoding="UTF-8"?>
<CaretSpecFile Version="1.0">
   <MetaData>
   </MetaData>
   <DataFile Structure="CortexLeft"
             DataFileType="SURFACE"
             Selected="true">
      ../anat/sub-01_space-fsnative_hemi-L_den-32k_midthickness.surf.gii
   </DataFile>
   <DataFile Structure="CortexRight"
             DataFileType="SURFACE"
             Selected="true">
      ../anat/sub-01_space-fsnative_hemi-R_den-32k_midthickness.surf.gii
   </DataFile>
   <DataFile Structure="Invalid"
             DataFileType="VOLUME"
             Selected="true">
      ../anat/sub-01_desc-preproc_T1w.nii.gz
   </DataFile>
</CaretSpecFile>
```

Before I get working on a tool, does this seem reasonable to everybody? I'm using the `T1w` above, but possibly `dseg`s make more sense?

effigies:
Following a meeting with @satra, @oesteban and @mgxd, we established the following:

1) wb_view will take URLs, as long as those URLs resolve to valid GIFTI/NIfTI files. I have separately confirmed that they will take relative paths, as well.
2) Any conventions promoted here should be made with reference to the CIFTI-2 specification, not conventions followed by ConnectomeWB. It is the responsibility of tools to convert between these, when necessary.
3) Additional measures should be taken on the TemplateFlow fMRIPrep side as reference implementations to ensure that good URLs that point directly to the necessary GIFTI (Conte69) and NIfTI (MNI152NLin6Asym) can be retrieved programmatically. This is not necessary for the spec work, but valid and persistent URLs will make examples more valuable.

I will try to polish up this pull request ASAP, which might need to wait until late next week based on my schedule.

Please add any notes you think are important.

cc @sappelhoff @franklin-feingold Status update.

effigies:
@satra Reading the [CIFTI-2 spec](https://www.nitrc.org/forum/attachment.php?attachid=333&group_id=454&forum_id=1955), it looks like we have neither `CortexLeft` or `CORTEX_LEFT`, but rather `CIFTI_STRUCTURE_CORTEX_LEFT`. Do you think we could say "BrainStructure values with the prefix `CIFTI_STRUCTURE_` omitted` or simply that we suggest using "BrainStructure values from page 10 of the CIFTI-2 specification"?

I plan to finish drafting this in the next 1-2 days.

satra:
@effigies - don't know a good answer. if every future structure starts with that prefix, then sure. but in the absence of any guarantees seems like aligning to the CIFTI -2 spec in this case makes sense.

effigies:
@edickie @satra @oesteban @robertoostenveld Can you please review this PR when you get a few minutes? If there are any conceptual issues rather than just drafting issues, can we schedule a call ASAP?

edickie:
Love it so far! I have one more clarification question re: inheritance principle. For the case of spaces - we could have a top level json called:

```
space-fsLR_den-91k_bold.json
space-fsLR_den-91k_thickness.json
sub-01/
    anat/
        sub-01/anat/sub-01_hemi-L_space-fsLR_den-91k_thickness.shape.gii
    func/
         sub-01_task-rest_space-fsLR_den-91k_bold.dtseries.nii
````

or 

```
space-fsLR_den-91k.json
sub-01/
    anat/
        sub-01/anat/sub-01_hemi-L_space-fsLR_den-91k_thickness.shape.gii
    func/
         sub-01_task-rest_space-fsLR_den-91k_bold.dtseries.nii
```

Do we need to clarify this in an extra example?

File: src/05-derivatives/01-introduction.md [hunk 16]
```diff
@@ -148,14 +148,18 @@ share the following (non-required) ones:
...
 
 ### SpatialReference key allowed values
 
-| **Value name** | **Description**                                                                                                               |
+| **Value**      | **Description**                                                                                                               |
 | -------------- | ----------------------------------------------------------------------------------------------------------------------------- |
-| orig           | A (potentially unique) per-image space. Useful for describing the source of transforms from an input image to a target space. |
-| uri or path    | This can be used to point to a specific file.                                                                                 |
+| `"orig"`       | A (potentially unique) per-image space. Useful for describing the source of transforms from an input image to a target space. |
```
oesteban, line 157:
```suggestion
| `orig`         | A (potentially unique) per-image space. Useful for describing the source of transforms from an input image to a target space. |
```

src/05-derivatives/01-introduction.md
```diff
@@ -173,6 +177,27 @@ sub-01/func/sub-01_task-rest_space-individual_bold.json
...
+
+```Text
+sub-01/func/sub-01_task-rest_space-fsLR_den-91k_bold.dtseries.nii
+sub-01/func/sub-01_task-rest_space-fsLR_den-91k_bold.json
+```
+
+```JSON
+{
+    "SpatialReference": {
+        "VolumeReference": "https://templateflow.s3.amazonaws.com/tpl-MNI152NLin6Asym_res-02_T1w.nii.gz",
```
oesteban, line 194: should we stick with constant names from CIFTI-2 for volume as well?
```suggestion
        "VOLUME": "https://templateflow.s3.amazonaws.com/tpl-MNI152NLin6Asym_res-02_T1w.nii.gz",
```

satra: i'm ok with changing to `VOLUME` but @oesteban where in cifti-2 did you find that constant?

satra: just in case it carries other semantics

effigies: Perhaps I can explain my reasoning:

1) A CIFTI-2 file, as currently generated by fMRIPrep, has 22 BrainModels. `CORTEX_LEFT`, `CORTEX_RIGHT` and a bunch of volumetric ones.
2) There is a `CIFTI_STRUCTURE_ALL_GREY_MATTER` and `CIFTI_STRUCTURE_OTHER_GREY_MATTER`.
   1. These are not in an ontology, where it can be deduced that, when looking for the reference for some grey matter structure, one should also check this entry.
   2. These are hypothetically structures that could be directly encoded, which would lead to an ambiguity of whether the reference was meant specifically for those structures or, again, all grey matter.
3) In the absence of a relevant CIFTI-2 constant, we are imposing something in BIDS. The BIDS convention for JSON keys is CamelCase.
4) In the event that this becomes relevant to another file format, it is not guaranteed that we will get constants that in CAPITAL_SNAKE.

I'm not deeply opposed to VOLUME, but unlike the others, using doesn't reduce the impedance mismatch for CIFTI-2. So I'd lean towards BIDS-like conventions.

effigies:
@edickie...

> ```
> space-fsLR_den-91k.json
> sub-01/
>     anat/
>         sub-01/anat/sub-01_hemi-L_space-fsLR_den-91k_thickness.shape.gii
>     func/
>          sub-01_task-rest_space-fsLR_den-91k_bold.dtseries.nii
> ```
> 
> Do we need to clarify this in an extra example?

Are sidecars without suffixes permitted? I can't think of an example within the rest of the spec or derivatives off the top of my head. My inclination is that we don't need this, but happy to be wrong, or to include something you suggest.

effigies:
I think the final two comments can be summarized:

1) We may want to add discussion of top-level sidecars for `SpatialReference`.
   * My overall feeling is that this is a general principle, and unless we're introducing something new WRT inheritance, we don't have a particular need to duplicate this in the example. I don't *think* we do that here.
2) We could use VOLUME instead of VolumeReference.
   * Agreed, but for reasons outlined in https://github.com/bids-standard/bids-specification/pull/301#discussion_r428960886, I don't see a strong argument for it.

I think that making these changes, if the community deems them necessary, would be sufficiently small as to be doable on the main PR (#265). To get things moving on the final review, I'm going to merge shortly.

PR #299, opened 2019-08-12
[INFRA] Adding contributor appendix sentence to PR template 

franklin-feingold:
closes https://github.com/bids-standard/bids-specification/issues/298, related to https://github.com/bids-standard/bids-specification/pull/284

PR #304, opened 2019-08-14
[ENH] Release protocol notes

franklin-feingold:
I noticed while completing the release a few guidance notes would be helpful on a few steps. @effigies I know there was potentially some RTD intervention for rebuilding. From the RTD build notes it appeared that were was some synchronized building that did occur during the release so left off those steps. I can add if that should be our process to be sure that stable is updated and the versioned rendering are generated.

Edit: added a minor step for ensuring the Contributors appendix is also up to date with the wiki (related to https://github.com/bids-standard/bids-specification/pull/299)


Release_Protocol.md
```diff
@@ -35,7 +35,7 @@ $ git fetch upstream
 $ git checkout -b rel/1.2.0 upstream/master
 ```
 
-### 2. Update the version in the changelog and mkdocs.yml
+### 2. Update the version in the changelog and mkdocs.yml
```
sappelhoff:
```suggestion
### 2. Update the version in the changelog and mkdocs.yml
```

Release_Protocol.md
```diff
@@ -68,6 +72,14 @@ $ git commit -m 'REL: v1.2.0'
 $ git push -u upstream rel/1.2.0
 ```
 
+If the Contributors needed to be updated:
+
+```Shell
+$ git add src/CHANGES.md mkdocs.yml src/99-appendices/01-contributors.md
```
effigies: Just add this to (what's currently) L70. Adding a file with no changes has no effect.

Release_Protocol.md
```diff
@@ -157,6 +169,8 @@ description:
 
 Click "Publish release".
 
+Note: It will take ReadTheDocs a few minutes to sync with the new release and build the new stable and versioned rendering.
```
effigies: I feel like this doesn't really serve much purpose here. Either they'll build without intervention or they won't. You can just say "Verify that they do", at which point you might as well point them to the right page, e.g.,

```suggestion
Verify ReadTheDocs builds complete and publish. If needed, manually
trigger [builds](https://readthedocs.org/projects/bids-specification/builds/)
for `stable` and the most recent tag.
```

Release_Protocol.md
effigies: Since we're putting the contributors here, let's just make this heading more descriptive of the section contents.

```suggestion
### 2. Update the version and the contributors list
```

Release_Protocol.md
```diff
@@ -57,6 +57,10 @@ If the version preceding the `-dev` is not the target version, update the versio
 In the figure below, we update `v1.2.0-dev` to `v1.2.0`.
 ![dev-to-stable](release_images/site_name_release_1.2dev-1.2.png "dev-to-stable")
 
+Note: this will make our continuous integration ([CircleCI](https://circleci.com/)) fail. This fails because the URL of the new ReadTheDocs rendering has not been generated at this time. It will be generated once the GitHub release has been completed. 
+
+#### 2a. Ensure the [Contributors appendix](https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md) is synced with the [Contributors wiki](https://github.com/bids-standard/bids-specification/wiki/Contributors)
```
effigies: Seems a little weird to make this a sub-heading. I suggested an update to L38. Now how about:

```suggestion
Synchronize the [Contributors appendix](https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md)
with the [Contributors wiki page](https://github.com/bids-standard/bids-specification/wiki/Contributors)
to ensure all contributors are duly credited.
Be sure not to remove credits if both have been edited.
```

franklin-feingold:
pinging @sappelhoff @effigies for review - should be set to go

PR #307, opened 2019-08-15
[FIX] Revise template-generated coordinate systems

oesteban:
This PR supersedes #213, and should be rebased with master after #306 is
merged.

W.r.t. `master` it only adds the table of _Nonstandard template
identifiers_

W.r.t. `common-derivatives`, the changes are:

  - [x] Removed the distinction between surface templates and volumetric
        templates. After all, to properly apply a surface template one
        should first spatially normalize to a particular volumetric
        template. Although it is possible to create mixed templates
        (e.g. using ``MNI152Lin`` to generate a mapping for volumetric
        coordinates and then, say, ``fsLR`` for the surface after some
        surface mapping process), this option should be discouraged.
        Anyways, this discussion does not affect how the names of BIDS
        entities should be specified, since one can always describe how
        the template coordinates are generated with the information
        described in this PR.
  - [x] Removed ``MNI152NLin6AsymConte69`` because it does not define
        any new coordinate system (see
        https://github.com/bids-standard/bids-specification/pull/207#discussion_r278911304),
  - [x] Removed ``FS305`` as it is now covered by ``fsaverage``.
  - [x] Reorganized subsections in "standard" vs "nonstandard" spaces.


effigies:
@oesteban Please review to make sure this is what you intended.

@bids-standard/derivatives If anybody would like to discuss these changes, I would ask that you do so in the next two days. After that, any further discussion should happen in #265.

effigies:
Squashed to avoid a tangle of merge commits.

PR #323, opened 2019-09-05
[FIX]: Number explicitly all cases in MRI field map section headers

jhlegarreta:
Number explicitly all cases in MRI field map section headers for the
sake of consistency.


jhlegarreta:
@effigies That's right.

effigies:
Style failure:

```
462:1-462:69  warning  Use headings shorter than `60`  maximum-heading-length  remark-lint
```

This is a bad requirement, IMO. I think we should probably add:

```
["lint-maximum-heading-length", false]
```

to:

https://github.com/bids-standard/bids-specification/blob/4c39c10fbc2e321aaf57379373a3566a8be21fe3/.remarkrc#L1-L9

jhlegarreta:
> This is a bad requirement, IMO. I think we should probably add:

Although excessively long headings should be avoided IMHO, the markup already requires a few characters, so I agree with what you propose.

effigies:
Sure, go ahead and open a PR to update the linter rules. It'll take two reviewers to merge. I don't think it's likely to be controversial.

jhlegarreta:
> Sure, go ahead and open a PR to update the linter rules. It'll take two reviewers to merge. I don't think it's likely to be controversial.

Done in #325.

effigies:
Can you go ahead and rebase this commit onto #325 in order to verify that it is the correct fix for this linting error?

jhlegarreta:
You're very welcome @franklin-feingold !

I had already added my name to the contributors wiki !

sappelhoff:
OKay, this PR is now ready for a rebase @jhlegarreta :-)

jhlegarreta:
ba93922 rebased on `master`.

sappelhoff:
Thanks! BTW: I like the way you named your branch to make this PR :wink:

PR #325, opened 2019-09-05
[INFRA] Set the maximum heading length lint check to `false`

jhlegarreta:
Set the maximum heading length `lint` check to `false`.


effigies:
Note for other reviewers: this is to accommodate #323.

sappelhoff:
thanks @jhlegarreta

PR #328, opened 2019-09-06
[FIX] Fix `Common principles` `Key/value files` section level

jhlegarreta:
Fix `Common principles` `Key/value files` section level.


jhlegarreta:
I assume it makes sense to group all considered file formats under the same section/it was an oversight to put `Key/value files` in an upper level.

sappelhoff:
Thanks @jhlegarreta

PR #334, opened 2019-09-13
ENH: Clarify the position toward non-compliant derivative datasets and files

effigies:
This is a proposal in follow-up to https://github.com/bids-standard/bids-specification/pull/265#issuecomment-530972613 and the ongoing discussion. It is not intended for immediate inclusion, but as a concrete example that can be discussed and critiqued.

This is also related to #264, where discussion of non-standard files in the main specification is being updated.

@bids-standard/derivatives I would appreciate wide input on this, as it has a significant impact on the overall interpretation of the standard, with regard to tooling.


effigies:
Given the lack of critique, I'm going to propose this to be included in common derivatives. I would appreciate reviews, @bids-standard/derivatives.

src/02-common-principles.md
```diff
@@ -146,10 +147,13 @@ my_dataset/
...
-need to be BIDS compliant** datasets. This specification does not prescribe 
-anything about the contents of `sourcedata` folders in the above example - nor 
-does it prescribe the `sourcedata`, `derivatives`, or `rawdata` folder names. 
+In this example **only the `rawdata` subfolder** needs to be a BIDS-compliant
+dataset.
+The subfolders of `derivatives` MAY be BIDS-compliant derivatives datasets
+(see [Non-compliant datasets][non-compliant-datasets] for further discussion).
+This specification does not prescribe anything about the contents of `sourcedata`
+folders in the above example - nor does it prescribe the `sourcedata`,
+`derivatives`, or `rawdata` folder names.
```
sappelhoff: The last part is a bit confusing to me:

> nor does it prescribe the `sourcedata`, `derivatives`, or `rawdata` folder names.

with that you mean that I can call my BIDS dataset root, e.g. `eeg_matchingpennies` ... or whatever - and it is valid BIDS? However, WITHIN `eeg_matchingpennies`, I am *not* free to give `sourcedata` any other name than `sourcedata`.

I think this could be clarified.

robertoostenveld: No, I understand it as

```
-- whatevenyoucallthis
   |-- sourcedata_or_so
   |-- rawdata_in_bids
   `--pipeline_results_or_so
      |-- pipeline1_in_bids
      `-- pipeline2_in_bids
```
where the top level is up to you (e.g. it could be `D:`), and neither `sourcedata_or_so` nor `pipeline_results_or_so` are themselves according to BIDS. But both `rawdata_in_bids ` and each `pipelineX_in_bids` directories is a BIDS dataset. So BIDS only starts playing a role further down in the directory tree, and there are three (related) BIDS datasets here.

robertoostenveld: I think that the example might already be more clear if it would be renamed like this
```
my_project/
  originaldata/
    ...
  rawdata/
    dataset_description.json
    participants.tsv
    sub-01/
    sub-02/
    ...
  my_results/
    pipeline_1/
    pipeline_2/
    ...
```
where only `rawdata`, `pipeline_1` and `pipeline_2` are organized according to BIDS.

effigies: Yes, @robertoostenveld's interpretation is correct. I think part of the issue is that that context is missing from the diff.

Still, I'm happy to make any modifications (such as that suggested by Robert) which make that clearer.

src/02-common-principles.md
```diff
@@ -146,10 +147,13 @@ my_dataset/
     ...
 ```
 
-In this example **only the `rawdata` subfolder and all subfolders of `derivatives`
-need to be BIDS compliant** datasets. This specification does not prescribe 
-anything about the contents of `sourcedata` folders in the above example - nor 
-does it prescribe the `sourcedata`, `derivatives`, or `rawdata` folder names. 
+In this example **only the `rawdata` subfolder** needs to be a BIDS-compliant
```
effigies: @sappelhoff @robertoostenveld Would this make the situation clearer?

```suggestion
In this example, where `sourcedata` and `derivatives` are not nested inside
`rawdata`, **only the `rawdata` subfolder** needs to be a BIDS-compliant
```

effigies:
This one has had a reasonable review period, IMO. For further concerns, please comment directly on #265.

PR #338, opened 2019-09-21
[MISC] Move the PR template to a separate folder and improve contents

jhlegarreta:
Move the PR template to a separate folder and use HTML syntax.

The first action helps in reducing clutter in the root folder, and helps
identifying more clearly the GitHub-related files.

The HMTL syntax transition using HTML comments allows the PR template
text not to be part of PR message, and saves the user the task of
deleting it while being still clearly visible when opening a PR.

Minor syntax enhancements.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,9 @@
 --- PLEASE READ AND DELETE THE TEXT BELOW BEFORE OPENING THE PULL REQUEST ---
 
+See the [CONTRIBUTING](https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md) guide. Specifically:
+
 - Please keep the title of your pull request (PR) short but informative - it will
-  appear in the changelog
+  appear in the changelog.
 - Please ensure your name is credited on our [Contributors appendix](https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md).
   To add your name, please edit our [Contributors wiki](https://github.com/bids-standard/bids-specification/wiki/Contributors) and add your name with the type of contribution.
   For assistance, please contact @franklin-feingold or @sappelhoff.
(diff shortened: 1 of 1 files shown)


jhlegarreta:
The effect would be this:
https://github.com/InsightSoftwareConsortium/ITK/blob/master/.github/pull_request_template.md
whereas the raw contents are:
https://raw.githubusercontent.com/InsightSoftwareConsortium/ITK/master/.github/pull_request_template.md

Hope folks find this useful.

jhlegarreta:
The `circleci: linkchecker` CI reports two
```
Result     Error: TooManyRedirects: Exceeded 10 redirects.
```
errors, but none of the two seems to display the URLs contained in the PR template.

Any guess?

sappelhoff:
Looks interesting to me, thanks for the submission. +1 definitely for moving the template into a GitHub folder for clarity in the repository root.

Regarding using an HTML comment, I am undecided: It seems a bit cluttery and it'd distract me from writing my issue (I like to delete the template after reading). But I think that's too much a personal opinion to factor in here ... I'd be fine adding it if popular opinion is in favor.

@franklin-feingold @effigies what are your opinions?

RE: Circle CI:

> Any guess?

The HED website seems to be down. This is unrelated to your PR, I think we can ignore it for now and it'll soon be fixed by the responsible people (else, we'll deal with it soon)

effigies:
I wouldn't use HTML, because the main way people will see this is through the text editor, not the preview mode. It should be easily readable without further action.

On the whole, I think the current strategy is fine. If you want to create a template with sections that people use, then HTML comments are useful for describing what's expected and letting people choose whether or not to delete them. A big block of text that asks submitters to delete itself seems to work pretty well at getting people to do what we want.

effigies:
That'd be fine, though feel free to do it in this PR, to keep the conversation in one place.

jhlegarreta:
Done. Please have a look at the new revision dfeb98e . Thanks.

franklin-feingold:
+1 for not having the HTML, mostly on what @effigies described. 

I like moving the template away from the Root!

overall LGTM

As @sappelhoff said we can ignore the linkchecker because we know it is from HED not rendering right now, not our system. Side point: If they are down for extended time, can consider perhaps finding a way to skip them in validation so it doesn't appear we are failing our CI

Issue #135, opened 2019-01-24
pdf version of spec


sappelhoff:
[https://bids.neuroimaging.io/bids_spec.pdf](https://bids.neuroimaging.io/bids_spec.pdf) does currently *not* link to the most recent version of the spec, which is as @KirstieJane pointed out 1.1.2

This should be updated and I think we should automate this in our release cycles.

sappelhoff:
> Previously the pdf spec was on the website at that location, but I am not aware of where the pdf link currently is promoted.

right! I forgot about this. So currently we do not have a pdf spec. Correct?

franklin-feingold:
That is right as far as I know. There is a nice [plug-in](https://github.com/zhaoterryy/mkdocs-pdf-export-plugin) that exports all the markdown files into pdfs and there is an option to combine them and set the path (default is the same directory as the md files). I can investigate and see how this works so we can get a built pdf into the repo

sappelhoff:
thanks @franklin-feingold, this sounds perhaps worth an experiment. A portable (i.e., not online and a single file) of the specification would perhaps be good to have.

I'll close this issue however, because its main purpose is "solved"

KirstieJane:
Thanks @sappelhoff & @franklin-feingold.

Just commenting that I had the link in the documentation for BEP001, it isn’t linked from the website at the moment (as Franklin said). No need to track down an incorrect link 🕵🏼‍♀️😁

franklin-feingold:
I think for the portable (pdf) specification, it should be of the latest stable version (versus latest). I think we would want the portable version to be something that we support and being the latest stable version we support. This can be part of the release protocol. WDYT?

In the interest of keeping PRs to small manageable reviews and additions, I will wait until we have worked through https://github.com/bids-standard/bids-specification/pull/137 before adding in how the pdf generation will work. Though, in the meantime I will PR the latest stable version pdf (v1.1.2) before fleshing out that step in the release protocol (if you agree) so at least we have it in the repo.

franklin-feingold:
this is how it will roughly look - https://github.com/franklin-feingold/bids-specification/blob/pdf_spac/BIDS-Specification_v1.1.2.pdf - still a WIP, I want to add a TOC and brief title to more resemble the previous pdfs.

franklin-feingold:
To answer your point - 
1. We are able to copy paste from the pdf (i..e. into a web browser or or other word doc)
[...]

PR #341, opened 2019-09-28
RM: rm TOC.md - seems no longer pertinent/used

yarikoptic:
Seems to point to google doc ids etc

I could be wrong though ;-)  Let's see if build passes


effigies:
This might be useful/necessary (if we fix the links) for a PDF build.

yarikoptic:
Then probably it should be generated during pdf build

sappelhoff:
Oups, thanks for spotting @yarikoptic 

Perhaps @franklin-feingold remembers what this file was for, because he did the initial migration from Gdocs to Markdown.

franklin-feingold:
This file was originally brought over for building pdf's (https://github.com/bids-standard/bids-specification/issues/135). In the meantime has been a few challenges to doing this - I agree with Yairk that this can be generated during the pdf build so safe to delete imo

sappelhoff:
@effigies @franklin-feingold feel free to merge if you approve

Issue #349, opened 2019-10-18
[ENH] Optional alternative to bval/bvec


mattcieslak:
Trying to develop diffusion BIDS apps is made significantly more difficult by using the FSL bval/bvec format. A format that is dependent on the image axis (and even then only *sometimes*) means that a huge amount of effort is required to consistently handle orientation information. These files are ubiquitous and therefore should not be removed from BIDS, but they're a lot of trouble. 

I propose adding an optional mrtrix-style ".b" file for each dwi file. These can be robustly generated from dicoms, don't depend on how an image is stored on disk (are in world coordinates, consistent with NIfTI convention), are text-based, and are easy to rotate with other software packages (eg ANTs). 

For reference on how difficult these files can be:
- https://community.mrtrix.org/t/fsl-bvecs-handling/1434
- #243 
- There are numerous programs that exist just to test empirically how to re-orient bvecs. The fact that these are necessary points to a real problem. 
   - https://mrtrix.readthedocs.io/en/latest/reference/scripts/dwigradcheck.html
   - https://www.sciencedirect.com/science/article/abs/pii/S0730725X1830540X
   - https://afni.nimh.nih.gov/pub/dist/doc/program_help/@GradFlipTest.html

DSI Studio, camino, and others have invented their own gradient file formats that are similar to the mrtrix .b format. It would be great if the community can agree on an image-axis-independent file format to be optionally included in BIDS. Even a tsv file with the b values and x, y, z in RAS+ world coordinates would be amazing.

effigies:
Looking at [the format](http://userdocs.mrtrix.org/en/latest/concepts/dw_scheme.html#mrtrix-format), I think this in TSV would be a reasonable type of data to support. Our TSV files always have column headers, which will help make it more explicit that these are RASB columns. So it would be roughly:

```
R       A       S       B
0       0       0       0
0       0       0       0
-0.0509541      0.0617551       -0.99679        3000
...
```

We definitely need to continue to allow bval/bvec, since that's part of the standard, but I would personally be fine with permitting a more generic format that is more consistent with BIDS.

Given that it's `dwi.bval` and `dwi.bvec`, would `dwi.tsv` might make sense, or would some other suffix be clearer?

Tagging @bids-standard/raw-mri-dwi.

cookpa:
I think FSL supports the 3xN bvec format, but it has not been widely promoted.

mattcieslak:
[...]

Issue #243, opened 2019-06-11
Incomplete description of DWI bvecs format


Lestropie:
In thinking about appropriate definitions of orientation coordinates for DWI derivatives, I realised that in its current form the main BIDS spec makes the same mistake that we made in *MRtrix3* regarding the definition of the FSL `bvecs` / `bvals` format.

The best explanation of the issue was posted on our [community forum](http://www.mrtrix.org/2016/04/15/bug-in-fsl-bvecs-handling/); the [FSL FDT documentation](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FDT/FAQ#What_conventions_do_the_bvecs_use.3F) has also since been updated accordingly.

#### Long story short

The current description is:

> Inherent to the FSL format for bvec specification is the fact that the coordinate system of the bvecs is with respect to the participant (i.e., defined by the axes of the corresponding dwi.nii file) and not the magnet’s coordinate system

This is not precisely correct. The vector orientations defined in the `bvecs` file are with respect to the i,j,k axes of the corresponding NIfTI image *as though the NIfTI were to use a neurological convention*. That is: regardless of whether the associated NIfTI image is LAS (Analyze heritage; negative transform determinant) or RAS (NIfTI spec), the `bvecs` directions are with reference to the LAS equivalent of that NIfTI image.

PR #352, opened 2019-10-20
[ENH] Add world-coordinate tsv file for dwi gradient directions

mattcieslak:
Adds a dwi tsv option with gradient directions in physical coordinates instead of image axes. Addresses #349.


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -423,7 +435,9 @@ JSON example:
 ```JSON
 {
-  "TotalReadoutTime": 0.095
+  "TotalReadoutTime": 0.095,
+  "BigDelta": 0.005,
+  "SmallDelta": 0.0005
```
chrisgorgo: I could not find the definitions of those fields. Was it introduces in other PR?

mattcieslak:
I added a table describing BigDelta and SmallDelta. These address @arokem's suggestion https://github.com/bids-standard/bids-specification/issues/349#issuecomment-544201991. Although I'm not sure if the thread is suggesting a json sidecar for the gradient tsv or adding some additional fields to the dwi's json sidecar. This adds them to the existing sidecar

oesteban:
There was also some discussion about existing software (and particularly BIDS-Apps) that could potentially require changes if this PR were accepted.

Some people who might want to have a say here: @bids-standard/derivatives-mri-dwi, @ecaruyer, @jelleveraart.

mattcieslak:
Since the discussion has moved here, I asked friends who are using gradient tensor imaging how they store b-tensors and they pointed me to this repository: https://github.com/markus-nilsson/fwf_header_tools. I haven't gone through it in detail yet.

.remarkrc
```diff
@@ -1,6 +1,6 @@
 {
   "plugins": [
-    "preset-lint-markdown-style-guide",
+    "remark-preset-lint-markdown-style-guide",
```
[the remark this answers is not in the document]
mattcieslak: Added as #356

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -411,10 +413,28 @@ bval example:
...
+
+##### Pulse gradient timing information
+
+| Field name     | Definition                                                                                                                                                                                                                                                                |
+| :------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| SmallDelta     | RECOMMENDED. Acquisition pulse separation time (in seconds). For Siemens, this is in the shadow header (0029,1120) under `sSpecPara.dSpecLipidSupprBandwidth` in milliseconds. This value is necessary for estimating ensemble average propagators in physical units.     |
+| BiglDelta      | RECOMMENDED. Acquisition pulse separation time (in seconds). For Siemens, this is in the shadow header (0029,1120) under `sSpecPara.dSpecLipidSupprDeltaPos` in milliseconds. This value is also necessary for estimating ensemble average propagators in physical units. |
```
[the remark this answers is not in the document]
mattcieslak: yes, that's a typo

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -411,10 +413,28 @@ bval example:
...
 ```
 
-`.bval` and `.bvec` files can be saved on any level of the directory structure
+optional `dwi.tsv` example:
+
+```Text
+R	A	S	B
+0	0	0	0
+1.0	0	0	1000
+-0.0509541      0.0617551       -0.99679        1000
```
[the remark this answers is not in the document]
mattcieslak: Having the vectors normalized should be mandatory. I have seen bvecs that were non-normalized attempting to encode q-space coordinates, but it is much better to have the R A S be simply a unit vector specify magnitude from either the B column (in s/mm^2) or more precisely in the BigDelta and SmallDelta fields in the sidecar.

jhlegarreta:
Thanks for summarizing what has been discussed in the pas few days @oesteban . 

> Implementation of tensor b-encoding; proposed by @jhlegarreta #349 (comment)

Following the link posted by @mattcieslak, I found @filip-szczepankiewicz 's  repository's [resources](https://github.com/filip-szczepankiewicz/fwf_seq_resources) section, which seems indeed extremely useful to have freely available data examples, examples of the sequences and tools to deal with them. Have not investigated the implementation (I am not an expert on FWF/not a user yet), what the gradient data looks like, or how these fit into the `TSV`+`JSON` philosophy.

> The general discussion about backward compatibility introduced by @chrisgorgo - #349 (comment)

I also see 3 followed by 4 as the way to go.

.remarkrc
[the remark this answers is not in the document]
mattcieslak: reverted

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: Then, I guess w.r.t. the b-vals, the description that B must give the magnitude in s/mm^2 or encode it with BigDelta and SmallDelta is missing. An example of both would be greatly appreciated.

chrisgorgo:
+1 for option 3 followed by 4. 

Practically speaking it would be a form of deprecation, with the validator issuing deprecation warnings (that ideally will say when the old format will become unsupported), right?

If this works for this case it might be a general way to approach backward-incompatible changes on a case by case basis instead of waiting for 2.0.

jchoude:
Hi all,

I'm a bit late on this. I wanted to comment on the initial issue , but it is now locked, so I'll comment here, from a few points that were touched:

-  w.r.t keeping track of the transformations that might have been applied to the gradients, as @arokem mentionned, that would be more suited to the Derivatives spec. However, even there, it isn't quite clear if all steps should be encoded in the .json file. This is more provenance, and less about a user being able to correctly interpret and use the current files.
- Everyone seems to agree that gradients should be defined in the scanner RAS+ space. I also agree. I think it should be said even more explicitly in the current proposed PR, to ensure that no ambiguity is left.
- About RAS+ in scanner space: a lot of tools implicitly work with `FSL-style` encodings. To be able to still work, they would need to transform the RAS+ `.tsv` values back to `FSL-style` orientation. This means they would need to know the transform. However, I'm not clear if what is proposed here is to encode the "requested" directions (i.e. specified in the MRI interface), or the effective directions. In some cases and machines, requested directions are rotated to be aligned with the main axes of the subject's head. In this case, which directions would be saved in the `.tsv` file?
- About b-tensor style encoding, I think there is still lots of discussion to be had for this point. Considering it is still quite "early" in this field, we might want to differ decision on b-tensor encoding for a future discussion. For example, I'm not sure what would be preferable: saving the analytical trajectories that were acquired, or the effective "gradient steps" that were used by the machine to traverse said trajectory? Maybe we would need to save both, or the analytical version + some information that could be used to generate the steps if needed, such as the number of steps, etc.

Finally, in the initial discussion, there was one big point that was about deprecation, coexisting `.tsv` and `.bval/.bvec`. I think we should try to see what timeframe will exist for the cohabitation. If the "next" version that deprecates `.bval/.bvec` will not happen for quite a while, we run the risk of developpers having to support both types for quite a while, and that might not be optimal.

Thanks for this, I think it will be of grat help!

oesteban:
I have rebased this on top of #624, following conversations there. For a clean diff see https://github.com/oesteban/bids-specification/compare/enh/dwi-combinations...mattcieslak:dwi_tsv

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: @mattcieslak added a table as for other definitions - marking resolved.

oesteban:
@bids-standard/raw-mri-dwi @bids-standard/raw-mri I have recently received a notification for this (I believe) alternative BEP in development - https://arxiv.org/pdf/2103.14485.pdf which I think addresses at the very least some of the problems we are targeting here.

WDYT about the preprint overall, @mattcieslak? should we just close this and ask the authors of the preprint to submit their proposal as a PR?

francopestilli:
@arokem, @francopestilli, @Lestropie, @mattcieslak, @satra, @edickie @dlevitas at a first glance the proposal seems good. It also appears to propose additions to the current spec and not necessarily steps towards what we were trying to do (from my point of view). I will read it in more detail, but what do the others think?

arokem:
Looks like it very thoroughly addresses the [comment](https://github.com/bids-standard/bids-specification/issues/349#issuecomment-544163066) on the issue that precipitated this PR. It would be nice to start thinking about these things. 

> should we just close this and ask the authors of the preprint to submit their proposal as a PR?

I agree with @francopestilli's observation that this might be somewhat orthogonal to what is done here. Many of the things that are proposed are also not crucial for many methods. Maybe this PR could be completed in parallel (or first?)? I think there might be ways in which incorporating their ideas could build upon what is proposed here.

mattcieslak:
The preprint looks very ambitious and covers diffusion encoding schemes that are still new and fairly rare. The main goal of this PR is to have a definite mapping between gradient directions and LR/AP/IS that is independent of the image axis. This should be easy in most cases to get out of the dicoms and fits nicely into a simple tsv, without adding lots of other features

oesteban:
Thanks! Let's get this done then and we'll see how the preprint fits
independently.

francopestilli:
@oesteban @arokem @mattcieslak it looks like we agree. Interesting but orthogonal. A BIDS for Pros.

JAgho:
Hi all,

I'd planned to share this preprint explicitly rather than indirectly. Overcoming difficulties such as the subject of this PR is intended as a central aim of this BEP. Recording parametric variation throughout a diffusion experiment is a necessary evil, and while stopgap solutions resolve temporary issues, ever more intricate sequences promise ever more detailed parametric variation. We've tried to show in the manuscript how the many types of diffusion experiment cannot all be easily expressed in the same format of tabular file, and hence justify the modifiable structure proposed. 

We've structured things to suit option 2) here, extending and not superceding the bval / bvec system, which is used in a lot of pipelines we'd like to keep working. In general, the discussion here echoes a lot of conversations we've had internally, concerning paired tsv / json structures. Ultimately, this BEP would address this concern, and from the user's perspective, will mirror the suggestion proposed (of a more detailed tsv with more detailed parametric variation). 

This proposal isn't finished until it's seriously reviewed, so any analysis you offer is really appreciated. Please feel free to write to me with any suggestions.

oesteban:
Thanks @JAgho for confirming that your BEP would extend/continue from this one. That clarifies the fate of this particular PR even further.

Regarding the preprint, I think the BIDS steering committee and/or maintainers will reach out to you - particularly large BEPs typically require some discussion. Therefore, ensuring a flow of bi-directional feedback between BEP leads and the BIDS community is really important for them to succeed.

francopestilli:
@JAgho @oesteban if possible we would love to stay informed because of the work we have done before and plan to do later @arokem @mattcieslak @Lestropie @jchoude

oesteban:
Rebased after fixing conflicts.

oesteban:
@mattcieslak I have sent a PR against this branch (mattcieslak/bids-specification#1) with a reorganization following this rationale:

- I think we can make the TSV format more flexible, and recommend users to write the gradient table as it is provided by the scanners. If the scanner produces scanner coordinates, then the header is RAS+b. If the scanner produces voxel coordinates, then the header is IJK+b.
- The `SmallDelta`, `BigDelta` specification was split into two portions. I've grouped them together under the new subsection you created.
- I've corrected something about the FSL format that I think was blatantly wrong - everyone should carefully review this edit just in case it's me who is wrong.

Matt, if you are fine with my changes, please merge them ASAP so that we can push this PR to the finish line within the next couple of weeks. I think this flexibilization of the TSV format will make things a lot easier for the maintainers and software (e.g. PyBIDS) to digest, as in most of the cases this will only entail transposing the bvec and bval files and adding the IJKb column names.

oesteban (approved):
This PR addresses a known problem (see #349) and makes the specification for DWI gradients more consistent with the general BIDS specs, will accommodating some scenarios that were not contemplated before (scanners that work with the gradients on "world-coordinates"). The proposed spec is unambiguous with regards to this problem.

It also "deprecates" the utilization of custom text files, that are replaced by a TSV file. This will have a positive impact on the BIDS-Derivatives side, thanks to this better consistency with overall BIDS.

Finally, the PR adds two metadata entries that are related to the accurate reconstruction of diffusion encoding gradients.

We should push this to the finish line as soon as we can 👍 

Thanks @mattcieslak for the effort and for your patience.

oesteban:
I'm very interested in @jchoude's opinion after the latest developments:

> w.r.t keeping track of the transformations that might have been applied to the gradients, as @arokem mentionned, that would be more suited to the Derivatives spec. However, even there, it isn't quite clear if all steps should be encoded in the .json file. This is more provenance, and less about a user being able to correctly interpret and use the current files.

I think the proposed TSV file (which can be given in voxel or world coordinates) could easily be projected into the Derivatives specs and complete it there with the mentioned metadata.

> Everyone seems to agree that gradients should be defined in the scanner RAS+ space. I also agree. I think it should be said even more explicitly in the current proposed PR, to ensure that no ambiguity is left.

Hopefully, this is very clearly stated now. Both scanner RAS+ and voxel conventions are supported, and they must be encoded in the TSV file's column headers.

> About RAS+ in scanner space: a lot of tools implicitly work with FSL-style encodings. To be able to still work, they would need to transform the RAS+ .tsv values back to FSL-style orientation. This means they would need to know the transform. However, I'm not clear if what is proposed here is to encode the "requested" directions (i.e. specified in the MRI interface), or the effective directions. In some cases and machines, requested directions are rotated to be aligned with the main axes of the subject's head. In this case, which directions would be saved in the .tsv file?

This is the point where I'm most concerned. Is this PR clear about that we are talking of "effective" directions? If you feel it is not, then please add concrete suggestions that would clarify this point.

> About b-tensor style encoding, I think there is still lots of discussion to be had for this point. Considering it is still quite "early" in this field, we might want to differ decision on b-tensor encoding for a future discussion. For example, I'm not sure what would be preferable: saving the analytical trajectories that were acquired, or the effective "gradient steps" that were used by the machine to traverse said trajectory? Maybe we would need to save both, or the analytical version + some information that could be used to generate the steps if needed, such as the number of steps, etc.

My impression is that the preprint above addresses this problem, and that the problem is beyond the scope of this PR (which is very operational - remove the limitation of encoding gradients with FSL-style files only).

> Finally, in the initial discussion, there was one big point that was about deprecation, coexisting .tsv and .bval/.bvec. I think we should try to see what timeframe will exist for the cohabitation. If the "next" version that deprecates .bval/.bvec will not happen for quite a while, we run the risk of developpers having to support both types for quite a while, and that might not be optimal.

With the change that the TSV file can accommodate voxel coordinates, transitioning from .bval/.bvec to .tsv is immediate. Tools like PyBIDS could move from .bval/.bvec-centric implementations to table-centric without any overhead (instead of `get_bvec` and `get_bval` you may have something like `get_btable` which internally reads and composes .bvec/.bval if tsv is not available).

oesteban:
Okay, I have added one line about the "effective" vs "requested" gradients in (see 487032c).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -522,21 +521,78 @@ In such a case, two files could have the following names:
...
+rather than the "*requested*" values (as in, the original gradient table design that is
+configured into the scanner's MR protocol settings).
+
+The RECOMMENDED way to store gradient information is a `[*_]dwi.tsv` file corresponding
+to one or more DWI files.
+The format of this TSV file, and associated metadata, are described below.
+The gradient information MAY be stored using the `[*_]dwi.bval` and `[*_]dwi.bvec` pair
+of files.
+However, the `[*_]dwi.[bval|bvec]` files are DEPRECATED and will be
+phased out in a future release.
```
satra, line 539: Since this is a breaking change shouldn't this be under the 2.x consideration? if not perhaps the wording at present could be changed to support both the current required bval/bvec and add the tsv file.

oesteban: I see this PR more as a bugfix (and hence that takes precedence over the backward compatibility principle). Happy to see proposals on better wording though.

That said, with the current text both would be supported. TSV would take precedence over bval/bvec, and implicitly (but not explicitly), the tsv should be equivalent to bval/bvec. The only "breaking" change is lifting the mandate to have bvec/bval files (if you have the tsv)

oesteban: @satra did my reply address your concern? How would you address this issue?

satra: sorry too many things going on. my suggestion would be to continue to require bval and bvec till bids 2.0. regarding it being a bug. is it really a bug, or just that it forces a certain consideration that does not work in all situations. i.e. software have to be aware of the details of the spec. if the latter, that to  me is not a bug but an alienation problem, which can be addressed by making the tsv required/recommended moving forward. it's possible i'm not completely understanding the bug part of the spec.

francopestilli: @oesteban I am inclined on giving precedence to bvecs/bvals reason being that if we give precedence to TSV only a very narrow set of mainstream toolboxes will be able to handle BIDS DWI. @arokem @jchoude

JAgho: I think mandating bvec/bval is the only reasonable option here, plenty of tools rely on them (for better or worse) to inform the arrangement of the acquisition. Giving a secondary tsv is much friendlier imo, and gives room to grow in the future.

oesteban: I've worked a deep rewriting I hope addressing these issues: https://github.com/bids-standard/bids-specification/pull/352/commits/6c73743815b596dbe8d5c129829b83ac1463d4f6

> regarding it being a bug. is it really a bug, or just that it forces a certain consideration that does not work in all situations. (@satra)

It is not rigorously a bug, especially for scanners that write voxel coordinates because that's the information that goes into the bval/bvec. However, some scanners write world coordinates and at this moment, that requires a conversion into bval/bvec. That means, we are not re-formatting (e.g., what you would hope for most -if not all- of the DICOM -> NIfTI step), you really need to recalculate these numbers.

> which can be addressed by making the tsv required/recommended moving forward. (@satra)

Yes, this is what this PR tries to establish. Let me know if it is unclear, and please suggest edits.

> i.e. software have to be aware of the details of the spec.

Currently, the bval/bvec information could be written in world coordinates. Software would definitely digest it (as it is a de-facto standard format) but give you very wrong results if the dataset is e.g., oblique (which is common with EPI). So current specs are pretty dangerous.
This PR uses the TSV header to specify the coordinate system of the gradients, making this consideration unambiguous (this would be "the bug" we are fixing).

Another way to put it. ATM, BIDS cannot *directly* write metadata about gradients if the scanner provides scanner coordinates (as opposed to voxel coordinates). In such a scenario, you either "preprocess" the gradients (with the risk of doing it wrongly) and convert them to voxel, or even worse, you encode the metadata as voxel coordinates, leave no trace that this problem has happened.

> I am inclined on giving precedence to bvecs/bvals reason being that if we give precedence to TSV only a very narrow set of mainstream toolboxes will be able to handle BIDS DWI. (@francopestilli)

and 

> I think mandating bvec/bval is the only reasonable option here, plenty of tools rely on them (for better or worse) to inform the arrangement of the acquisition. Giving a secondary tsv is much friendlier imo, and gives room to grow in the future. (@JAgho)

IMHO the risk of having "valid" (from the spec POV) bvecs/bvals in the wrong coordinate system (without the possibility of detecting the problem) completely outweighs playing nice with existing software. Also, the conversion is so straightforward (and now including an example in the spec), that I honestly can't see any reason not to discourage the bvecs/bvals format.

oesteban: Addendum:

re: bvec/bval remain mandatory - please check the conversations during the early days of this longstanding PR. Although now there's more information to support those initial decisions, I think the new information has only confirmed the adequacy of them (IMHO).

oesteban: BTW - how does this PR breaks backward compatibility? The deprecation is exactly to tolerate the utilization of the old feature while issuing a strong recommendation of the replacement (https://en.wikipedia.org/wiki/Deprecation) - precisely to avoid breaking changes.

I'd agree with removing the `.bvec`/`.bval` only the 2.0 release (which would indeed be the breaking change).

satra: a breaking change in the specification is one where existing tooling stops working as a result of the change. typically deletions of required elements in any schema are breaking changes. additions are also breaking changes if the new keys introduced are required.

oesteban: > a breaking change in the specification is one where existing tooling stops working as a result of the change.

Yup, and since .bval/.bvec will still be allowed, a dataset that currently works will continue working on the same tool. The requirement is to have at least one of the two options (.tsv or .bval/.bvec) - so if the .tsv is not present, the specification remains exactly the same.
The new format is RECOMMENDED, and will only be mandatory when the deprecation finishes. Possibly BIDS 2.0.

The only effect of this PR is that now a BIDS dataset can be written with a much better, unambiguous format for the gradients. And that's not breaking compatibility because scanner coordinates could not be encoded (without conversion to voxels) before.

Anyhow, besides I personally think this is not breaking compatibility, there's also the fact that we indeed discussed how this deprecation would happen when Matt first submitted the PR. It would be unfortunate that we revisited those decisions (for the reasons above, and because that will set some precedent as to maybe preempt future contributions).

My impression is that although this proposal has important substance changes (I particularly like @jchoude's summary above), we seem to be focusing more on the formal aspect of the deprecation (which has largely been discussed already).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -542,52 +541,128 @@ In such a case, two files could have the following names:
...
-The `[*_]dwi.bvec` file contains 3 rows with *N* space-delimited floating-point numbers
+The `.bvec` file contains 3 rows with *N* space-delimited floating-point numbers
 (corresponding to the *N* volumes in the corresponding NIfTI file.)
 The first row contains the *x* elements, the second row contains the *y* elements and
 the third row contains the *z* elements of a unit vector in the direction of the applied
 diffusion gradient, where the *i*-th elements in each row correspond together to
 the *i*-th volume, with `[0,0,0]` for *non-diffusion-weighted* (also called *b*=0 or *low-b*)
 volumes.
-Following the FSL format for the `[*_]dwi.bvec` specification, the coordinate system of
+Following the FSL format for the `.bvec` specification, the coordinate system of
```
mattcieslak, line 626: The directions in the bvec aren't always using the image axis coordinate system. @frankyeh recently updated DSI Studio to handle the case where the image axes aren't in the FSL "RPI" orientation but the bvecs are. FSL interprets bvecs differently depending on the orientation of the image axis.

oesteban: Okay, I didn't know that. I think this just reinforces the argument that this PR is necessary. How the (recently added) conversion example would change with this? Should we mention the image axis coordinate system so that what is said is fully accurate?

mattcieslak: The conversion example will only work if the image is in LAS+ orientation. This will be true by default if dcm2niix was used, but is not guaranteed.

satra: here is the actualFSL spec: https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FDT/FAQ#What_conventions_do_the_bvecs_use.3F (it should be included in that section)

mattcieslak: Additional description here: https://community.mrtrix.org/t/bug-in-fsl-bvecs-handling/165?source_topic_id=1434

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
satra: i completely agree that this sub discussion has to do with principles of BIDS rather than the substance of this PR. however, if this PR is merged software that worked prior to this could stop working. for example, wouldn't this mean that after this change none of the scripts using the FSL tools will work on new datasets unless someone did an appropriate conversion?

most software doesn't check on bids version for a dataset given that bids has not released 2.0. i think that's the fundamental question in this thread. this PR just happens to be an instance of something that brings this up. i don't want to hold up this PR, but it seems that there are other instances where such an issue could come into play. as bids expands this could happen at a modality specific level such as this, or the common principles level (there were a few inheritance issues that were pushed off to bids 2.0). 

thus this is a meta discussion for bids as a whole, not just this PR. but merging this PR, means we have accepted this route at least for one instance.

oesteban: > for example, wouldn't this mean that after this change none of the scripts using the FSL tools will work on new datasets unless someone did an appropriate conversion?

For existing datasets, not at all - this PR does not mandate the update from bvecs/bvals to tsv.
For future datasets, you can generate the bvecs/bvals from the tsv file. Making this step more explicit will make it harder for the user, but will also ensure data is preserved as it was generated by the scanner.

If someone is motivated and has the knowledge to implement a test within the BIDS validator to ensure that the tsv and bvec/bval are consistent, that would be really helpful for the DWI community.

francopestilli:
I follow up on @satra  

The PR seems to stem from the learning and implementation of new code (I assume). The new code might have better features than the older code (coordinate system) than previous code. Yet, as it is new code, it is not embraced by the community yet, it is not known yet and it has not involved the community yet.

So this is a good general discussion. How do we propose changed to the specification that would accommodate new code and general improvements when the code is not out yet, the code has need been tested either by the BIDS contributor not by the community at large.

Good food for thought I hope.

effigies:
Is it always possible to take a new-style gradient TSV and input images and generate a correct set of FSL-style bval/bvecs? If so, then making a small tool that does that might be the cleanest option. bval/bvec can remain REQUIRED which means that old tools that take the standard at its word and assume the files can be found, while users can start with an authoritative TSV file.

We have [DEPRECATED](https://github.com/bids-standard/bids-specification/search?q=deprecated) things in recent releases, but we have not previously deprecated REQUIRED data. Probably the closest is `_phase.nii.gz` is now `_part-phase_bold.nii.gz` and `_PD.nii.gz` should now specify `_PDw.nii.gz` or `_PDmap.nii.gz`. Beyond not being REQUIRED, these are also relatively rare.

We may not be at voting stage yet, but I think I would probably be inclined to say that bval/bvec remain REQUIRED.

effigies:
If it's impracticable to validate consistency, we could say that, for tools that accept either TSV or bval/bvec, the TSV is to be considered authoritative.

mattcieslak:
> Is it always possible to take a new-style gradient TSV and input images and generate a correct set of FSL-style bval/bvecs? If so, then making a small tool that does that might be the cleanest option. bval/bvec can remain REQUIRED which means that old tools that take the standard at its word and assume the files can be found, while users can start with an authoritative TSV file.

It wouldn't be hard to make a small tool to translate between bval/bvec files and TSVs. MRtrix3 already has a really good one that even handles oblique acquisitions.

It also might not be totally correct to say that there is already a lot of software support for the bvec format. MRtrix (uses the strict FSL spec) and DIPY (uses image axes) handle them differently. DSI Studio has used handled them both ways at different times. DSI Studio, MRtrix and AFNI all have tools that permute the signs of the bvecs to figure out empirically how to interpret them

francopestilli:
@mattcieslak I understand the complexity in handling coordinate systems, please do not het ,e started :-) ... yet all of these tools you mention read in bvecs, correct?

@effigies I would say the mapping is lossless this way bvecs -> TSV (but it requires more than just BVECS to know the coordinate system) and is lossy this way TSV -> BVECS (as it loses the coordinate system). Does this help?

SyamGadde:
If I may chime in on this very important, long-due improvement:

1. Recommending that the gradient directions are stored as provided by the scanner (either scanner/realworld-space or image-space) is nice, but implicit in that text is an assumption that any scanner which stores the directions in scanner space will store them in RAS+.  But with DICOM, for example, that is clearly not the case (LPS+), and so this recommendation would be impossible for most data.  Maybe de-emphasize the "original coordinate system" recommendation and push for "closest coordinate system", since they should look pretty similar aside from some negative signs.
2. The conversion between .bvec/.bval and .tsv is complicated by the fact that it is dependent on the orientation of the input data, which was one of the major inspirations for this whole issue (i.e #243)  In some cases the conversion between bval/bvec and .tsv is not quite as simple as it is currently described because it would need to take the image orientation (and whether it was a rotation of LAS+ or a rotation of RAS+).  Pointing to the DTIFIT page doesn't do justice to the danger because most of the FSL pages assume the data is in some rotation of LAS+, and only in a few places do they mention the orientation assumptions.  It's probably not worth including support for that in the text, but a link to other documentation (like the mrtrix forum page mentioned in #243) or to a separate discussion page or even just a DANGER sign?  The bvec/bval and .tsv alternatives are not equivalent and there is no successful conversion between them without knowing the orientation of the associated image.

Thanks for your work on this!
(USENET group alt.bvec_bvals.die.die.die)

oesteban:
> If I may chime in on this very important, long-due improvement:

Of course! Thanks for doing so.



>     1\. But with DICOM, for example, that is clearly not the case (LPS+)

I may have this wrong, but LPS+ determines the orientation of the voxel cartesian space, while the physical (world) coordinates system is right-handed and defined w.r.t. the scanner's bore, right?.

I would agree though that the proposed header (R, A, S) is unfortunate because actually confuses the image's cartesian grid system with the world system. It would be better to just say X, Y, Z (as opposed to I, J, K that is proposed for voxel coordinates).





>     2\. The conversion between .bvec/.bval and .tsv is complicated by the fact that it is dependent on the orientation of the input data, which was one of the major inspirations for this whole issue [...]

Yes, this is probably one of the weakest points of this PR. Please send suggestions to clarify this.

SyamGadde:
> I may have this wrong, but LPS+ determines the orientation of the voxel cartesian space, while the physical (world) coordinates system is right-handed and defined w.r.t. the scanner's bore, right?.
>
> I would agree though that the proposed header (R, A, S) is unfortunate because actually confuses the image's cartesian grid system with the world system. It would be better to just say X, Y, Z (as opposed to I, J, K that is proposed for voxel coordinates).

Sorry I was probably mixing terminology/conventions and not as familiar with the BIDS conventions.  So to clarify my point -- independent of the matrix that defines the locations of voxels in space (which could be described in shorthand as RAS+, ALS+, RPI+, etc. where the letters indicate the rough direction in which higher-indexed voxels go), there is the coordinate system defining what the coordinate (x, y, z) means and the convention in NIFTI is where positive numbers mean R, A, and S respectively, which I would also find useful to describe as RAS+ (as opposed to DICOM's LPS+).  I actually like that the columns would use R, A, and S as the convention is then explicit in the file.  The alternative would be to embed the transformation from gradient directions into world-space  (what NRRD would call the "measurement frame") but that would diverge with the "image space" option (with I, J, K columns) and so I think the current approach is simple enough and easy to understand.

> > 1. The conversion between .bvec/.bval and .tsv is complicated by the fact that it is dependent on the orientation of the input data, which was one of the major inspirations for this whole issue [...]
> 
> Yes, this is probably one of the weakest points of this PR. Please send suggestions to clarify this.

I'll try to put some text together.  Thanks!

oesteban:
@SyamGadde any chance you find a slot to put together the edits?

SyamGadde:
Can you tell me where to target the changes?  I was having trouble figuring out how to put together a pull request for this issue and I sent one to the original branch:

https://github.com/mattcieslak/bids-specification/pull/2

Happy to resubmit, just need to know where...

oesteban:
Sorry I missed that. I'll go through it tomorrow

Lestropie:
Coming here for the first time due to bids-standard/bids-bep016#26; apologies for missing the discussion to this point after having initially lodged #243, GitHub notifications weren't making it to my inbox.

1. Unless otherwise controlled, when *writing* to a NIfTI / JSON, *MRtrix3* will actually already write the diffusion gradient table in RASB format to the JSON under the field "`dw_scheme`". This is an MRtrix key that well predates BIDS. This behaviour results from the fact that in MRtrix we have been able to store sidecar data within image headers for a long time ([documentation](https://mrtrix.readthedocs.io/en/latest/getting_started/image_data.html#mrtrix-image-formats-mih-mif)) (albeit in a string-string key-value list only), including the diffusion gradient table ([documentation](https://mrtrix.readthedocs.io/en/latest/concepts/dw_scheme.html#image-header)), and in order to provide the baseline functionality for BIDS support we essentially interpret JSON as being an alternative location for the image sidecar information to be stored rather than the image header. So if that information is present internally within *MRtrix3*, it will be written to the JSON in our internal representation.

2. The current proposed format for such data:
    ```Text
    R           A          S          b
    0           0          0          0
    1.0         0          0          1000
    -0.0509541  0.0617551  -0.99679   1000
    ....
    ```
    will not work out-of-the-box in *MRtrix3*. We expect anything not commented out to be numerical data. We don't treat files with `.b` or `.tsv` file extensions any differently to e.g. `.csv`, it all goes to a generic numerical matrix parser. It would be possible technically to support this, by permitting only the first non-comment row to be non-numeric; that's something that would need to happen at the *MRtrix3* side, just flagging that right now this won't work in *MRtrix3*-land.

3. From the PR:

    > The gradient strength and orientation information MUST correspond to "*effective*" values (as in, the actual orientations and strengths applied by the scanner), rather than the "*requested*" values (as in, the original gradient table design that is configured into the scanner's MR protocol settings).

    There's a couple of different ways this could be interpreted:
    1. Many will know that if you request a *b*=3000 acquisition, you can get a DICOM where the individual volumes have their *b*-values encoded as e.g. 2985-3015 or so, as the *b*-vector is re-calculated based on the full set of gradient waveforms. These are typically however very minor differences, and instead interpreting the data based on the requested gradient table would most likely not be deleterious to any analysis; so imposing a "MUST" is unnecessarily stringent.
    2. If the concern here is instead of the scanner hardware executing the gradient table using its XYZ gradients, but rotating the gradients according to the image FoV, then I agree this is a concern, however the statement is not accurate enough due to the ambiguity with point 1 above.
    
    > The RECOMMENDED `.tsv` file and the OPTIONAL `.bval`/`.bvec` MAY coexist in the structure.

    If these are not to be made mutually exclusive, then I would encourage App developers to think about how their software should behave in the case where both sources of data are provided, but are *different* to one another (as interpreted by their App). We have struggled with this for many years in *MRtrix3* with NIfTI's qform / sform (since we support many image formats other than NIfTI, so expect just a single image transform). Our solution is that *if both are specified*, we read and interpret *both*, and if the difference between them exceeds expected precision, we issue a warning, and notify the user as to which is being used (the behaviour of which can be [controlled by the user via a config file](https://mrtrix.readthedocs.io/en/latest/reference/config_file_options.html#cmdoption-arg-niftiusesform)). Note that such mismatches are wholly likely to occur given the `bvec` i-flipping issue.

    > Each row of the TSV file corresponds to one sampled DWI

    Would suggest "one DWI volume"; don't think I've ever seen the wards "sampled" and "DWI" alongside one another.

    > Rows corresponding to *b = 0* s/mm<sup>2</sup> strengths MUST contain four zeros.

    This to me is unnecessarily stringent. It's entirely common for DWI data in DICOM format to have a non-zero vector direction (and indeed even a non-zero *b*-value; see 3.1 above). So this places an elevated demand on conversion software to get from DICOM to BIDS. E.g. Using *MRtrix3*'s `mrconvert` manually to go from DICOM to NIfTI / JSON will currently violate this demand. To me it is up to software to take the data as they are received from the scanner (with *minimal* conversion) and interpret it as it wishes; if this includes taking volumes with "trivially small" *b*-values and interpreting them as *b*=0 and ignoring vector directions, that is the choice of the software developers.

    > It is RECOMMENDED to maintain the column ordering as described above (thus, either `I J K b`
or `R A S b`).

    Permitting column permutation here is to me unnecessary. It would also currently break *MRtrix3* support. Again, we could implement code there that would read the header line and permute accordingly, but because the file is initially read by a *generic* numerical matrix read function, it's non-trivial to support. It's a lot of demand on developers to provide flexibility that serves little to not benefit to the end-user.

4. I do not like the IJK / RAS distinction for multiple reasons:

    1. FSL `bvecs` is *not* IJK: it is IJK *if the image is interpreted as neurological*, regardless of whether it is in fact neurological or radiological. While one can argue that the contents of a `.tsv` is independent of the FSL `bvecs` / `bvals` format and can therefore do as it pleases, inevitably someone somewhere will mistakenly believe that, since both are stored with respect to image axes, they can simply translate information between the two. 

    2. This suggests that, if the first image axis most closely corresponds to subject right, and similarly for second / anterior and third / superior, then IJK and RAS are equivalent; but if the image FoV is anything other than pure axial with no angulation, this will not be the case.

    3. Fundamentally (please correct me @jdtournier if I'm wrong), the gradient directions provided in DICOM, as well as the way in which they are interpreted in *MRtrix3* is *not RAS*. It is *XYZ*: the axes of the scanner hardware. *If* the subject is in a supine position, then positive-X will be *closest to* subject right, similarly for positive-Y / subject anterior and positive-Z / subject superior. In BEP016 (Diffusion models) I currently have field "`ReferenceAxes`" as either "`ijk`" or "`xyz`" (though I'm yet to investigate whether or not there are instances in which the FSL `bvecs` `i`-flip issue can potentially apply to the outputs of diffusion models also...), names as such for this reason.
    (Edit: @oesteban [beat me to it](https://github.com/bids-standard/bids-specification/pull/352#issuecomment-879694401))

5. Regarding Chris' comments RE rollout strategy:

    1. If `.tsv` were to truly supersede `.bvec` / `.bval`, this would IMO need to happen as part of a major update (i.e. BIDS 2.0) rather than a minor one (e.g. 1.7.0). I'd personally be happy for things to go in this direction. But even then, any BIDS App developer wrapping DWI tools would need to be aware of the issues around representation of this information.

    2. If `.tsv` is implemented as a non-mutually-exclusive option alongside `.bvec` / `.bval` in a minor update, then there's a more general comment about compatibility between BIDS spec version and BIDS Apps that needs to be made (I'm unable to keep up with all BIDS-related discussions, so apologies if there's already some established principle here). If an App is written to (as best as possible given finite development resources) support BIDS data of a particular spec version, but is provided with BIDS data of a newer version, it may or may not work, depending on what changed in the spec. So maybe there needs to be a consensus about how Apps should be producing some form of warning about potential incompatibilities.
    (And this isn't even getting into the massive discrepancy between the amount of effort necessary for an App to support an exemplar BIDS dataset of a particular version, and supporting *all possible* BIDS datasets that conform to a particular spec version...)

6. From @effigies:

    > Is it always possible to take a new-style gradient TSV and input images and generate a correct set of FSL-style bval/bvecs?

    Yes. *MRtrix3* can import & export both `.bvec` / `.bval` and what's being called "RAS+b" here (which is also its internal representation); it's all based on the image transformation. It's just easy to get wrong (and we suspect many DWI software packages still do). See e.g. our [load](https://github.com/MRtrix3/mrtrix3/blob/3.0.3/core/dwi/gradient.cpp#L116-L171) and [save](https://github.com/MRtrix3/mrtrix3/blob/3.0.3/core/dwi/gradient.cpp#L177-L205) code.

jdtournier:
Just to quickly clarify a statement from @Lestropie - since he saw fit to tag me in... 

> the gradient directions provided in DICOM, as well as the way in which they are interpreted in MRtrix3 is not RAS. It is XYZ: the axes of the scanner hardware. If the subject is in a supine position, then positive-X will be closest to subject right, similarly for positive-Y / subject anterior and positive-Z / subject superior

Yes and no. First off, apologies if I'm being a bit pedantic or telling everyone what they already know... In the [_official_ DICOM specs](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.13.5.9.html), and for Philips and Siemens, the gradient vectors are indeed in XYZ, which corresponds to the axes of the scanner hardware _in the patient-centered coordinate system (PCS)_ ([official docs here](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.6.2.html))¹. This only corresponds to the _actual_ hardware coordinate system if patients are positioned head-first supine, otherwise it will be some permutation of these axes (this depends on what the radiographer specifies in the positioning field when registering the patient). But in general, this actually simplifies things, since we can _always_ rely on the assumption that (in DICOM) X is R→L, Y is A→P, and Z is I→S (i.e. LPS+, as others have said). In practice, I expect this is what most people mean when talking about the XYZ coordinate system, since we overwhelmingly deal with HFS positioning (certainly true in my case) - but in my opinion it's more accurately described as RAS+ (though I acknowledge the ambiguity that remains concerning whether we mean scanner or image axes - I mean scanner axes here). 
 
Otherwise, I would echo what others have said elsewhere:
- Important to have a single spec, and that this be independent of the images. Sticking with the same coordinate system as NIfTI is the most sensible, so I think RAS+ (in patient-centred coordinate system) is the way to go. Not a fan of the IJK option. 
- If you need to have a header line in the file, can we make it a comment? I.e. prefix with a `#`? If we have a single unambiguous spec, this is probably redundant - though nice to include for completeness. This would make our lives (in MRtrix) a bit easier... 
- The conversion from DICOM shouldn't try to impose too many assumptions. Given that manufacturers tend to store the actual _b_-value as computed using all imaging gradients, forcing _b_=0 DW volumes to have a strict zero is likely to backfire for e.g. IVIM acquisitions, which will be acquiring very low b-value images. Best to leave the decision as to what is to be interpreted as _b_=0 to the end application. 

One final point which I don't think has been raised so far is how to handle acquisitions where the magnitude of the gradient vector is modulated to achieve lower _b_-values than nominally requested - this is to allow multi-shell acquisitions on systems that would otherwise not easily support this (what we call [_b_-value scaling in MRtrix](https://mrtrix.readthedocs.io/en/latest/concepts/dw_scheme.html#b-value-scaling)). I may have missed it, but has there been any discussion as to how this should be handled in the spec - if at all?

---

¹As a side note: before the official DICOM standard, GE would store their gradients in the PRS frame.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -665,19 +740,36 @@ sub-<label>/[ses-<label>/]             # MultipartID
...
 JSON example:
 
 ```JSON
 {
   "TotalReadoutTime": 0.095,
-  "B0FieldSource": ["phasediff_fmap0", "pepolar_fmap0"]
+  "B0FieldSource": ["phasediff_fmap0", "pepolar_fmap0"],
+  "BigDelta": 0.005,
+  "SmallDelta": 0.0005
```
Lestropie, line 772: -   Given that these appear in a general sidecar metadata space rather than a DWI-specific context, the names "`BigDelta`" and "`SmallDelta`" may not be specific enough.

-   If revisiting, it would be worth considering whether it is worthwhile adding these, or instead deferring more advanced diffusion encoding information entirely to the aDWI extension.

PR #356, opened 2019-10-22
[INFRA] change plugin to remark-preset-lint-markdown-style-guide

mattcieslak:
This plugin was causing the travis build to fail.

### DIFF
--- a/.remarkrc
+++ b/.remarkrc
@@ -1,6 +1,6 @@
 {
   "plugins": [
-    "preset-lint-markdown-style-guide",
+    "remark-preset-lint-markdown-style-guide",
     ["lint-no-duplicate-headings", false],
     ["lint-list-item-indent", "tab-size"],
     ["lint-emphasis-marker", "consistent"],


franklin-feingold:
thank you @mattcieslak for raising this - perhaps what may have motivated changing the linter?

mattcieslak:
I don't know, actually. I'm not familiar with how node works, but this fixes an error I was seeing when trying to build this locally and on travis. It came up during https://github.com/bids-standard/bids-specification/pull/352#discussion_r337606741

franklin-feingold:
our linter can sometimes be challenging to get the enhancement to conform - we added a [section in our Contributing guide](https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md#fixing-travis-remark-errors) for addressing these type of errors
by chance, were there more fails than table fences? that is the common one I have seen and the Contributing guide goes through the process I use

mattcieslak:
Turns out this was only needed locally. Sorry for the confusion.

PR #375, opened 2019-11-22
[INFRA][draft PR] changes to generate pdf-version of the specs

Arshitha:
## pdf-version
[bids-specs.pdf](https://github.com/bids-standard/bids-specification/files/3881468/bids-specs.pdf)

## To generate the pdf: 
run `sh pandoc-script.sh` after cloning the changes. 

## Prerequisites: 
* `pandoc` package needs to installed. If you don't already have it installed, then running any of the following two commands should work. 
1. `brew install pandoc`
2. `conda install pandoc`

* `LaTex` needs to be installed as well. `pandoc` converts markdowns into pdf via LaTex files. 

## Changes made: 
* ToC with proper indentation
* Tables formatted to be within page margins
  * Had to increase the width of the columns of the pipe tables manually within the markdown files. It's was a one time edit, however, not sure if this might lead to problems while rendering them with RTD. 
  * Couldn't fix the entity table since it has more than 10 columns. I believe this can be fixed by converting just that markdown document to landscape and then integrate it with the rest of the markdowns with pandoc. Working on it. 
* Long lines of code fit within the margin and don't get cut-off. 

## Problems with this solution: 
* The contributors page has emojis which doesn't show up in the pdf. Working on this too. 
* Entity table still not fixed 
* Under table of contents, the title 'Modality Specific Files' doesn't show up even though all of the subsections are listed. (content not compromised) 

Most of the above problems with the solution can be potentially. However, would appreciate thoughts on how much of a priority these are and which takes priority over the other.


nicholst:
See some comments on #135, e.g. https://github.com/bids-standard/bids-specification/issues/135#issuecomment-557719394

yarikoptic:
Awesome!!! Thank you!
Indeed it is not yet "ideal" but it would be great to have it instead of none.  It might be worth (if easily possible?) to add at a disclaimer at the beginning of the pdf with pointer to known issues with PDF version (absent emojis etc) so ppl to not start filing issues. We could address/improve them later.

effigies:
Yeah, given that this has been a slow process to get fully set up, I think a good idea is to pick a location to store them, create the PDFs with the banner Yarik suggests, and then update them in place as we improve the process. Once it's set, we can see about automating it on release.

sappelhoff:
Thanks a lot @Arshitha for this PR and @nicholst for a thorough review.

+1 from my side to include this with a disclaimer and continue working from there.

Do you want to work on some of @nicholst feedback first @Arshitha ? What would be your preferred way of continung from here?

Arshitha:
Thanks @nicholst for the detailed review. I'll look into it. 

Also, thank you @yarikoptic and @effigies for the suggestions!

@sappelhoff yes, I'd like to work on some of the feedback and add a disclaimer too. I'll update this PR once I've made some progress.

yarikoptic:
oh, I somehow thought that this one is done already but it got stale and started to breed conflicts :-/  I retract my suggestion for a draft mark - let just get it in and improve upon later on!!! ;)

Arshitha:
@yarikoptic I didn't get a chance to get back to this before now. I'm working on the suggestions now so can we revisit this in a couple of days?

sappelhoff:
replaced by #400 

please reopen otherwise @Arshitha

Issue #373, opened 2019-11-22
Add "DRAFT" watermark (or similar) to non-stable rendered spec


sappelhoff:
According to @franklin-feingold, this issue was brought up in a recent steering group meeting:

### Intro

We are rendering the specification in several flavors, such as:

- [stable](https://bids-specification.readthedocs.io/en/stable/) ... the currently accepted version of the spec
- [latest](https://bids-specification.readthedocs.io/en/latest/) ... distinguished by a `-dev` appendix to the version of the spec in the header of the rendered spec. This version is not currently released/accepted, but it will become `stable` eventually.
- and previous versions
    - [v1.2.0](https://bids-specification.readthedocs.io/en/v1.2.0/)
    - ...

### Problem 

However, we are also rendering some development branches, such as the one for [derivatives](https://bids-specification.readthedocs.io/en/derivatives/). Currently, this development branch "looks" similar to the "stable" rendered spec. This is a problem, because users may accidentally use this version of the specification for formatting datasets, although this is merely a development branch and subject to change.

### Solution

A potential solution would be to add a watermark or similar sign to all renderings that are not either

- stable
- latest
- or  previously released (outdated) version

### What I have looked into so far

- potentially adding a "ribbon" to the page using html similar to the ["form me on github"](https://github.blog/2008-12-19-github-ribbons/) ribbons
    - --> turns out to be difficult with MkDocs, because we would *want* the ribbon only on specific versions of the specification
- Adding a watermark of some kind on top of our source during the rending process
    - --> I didn't find any READTHEDOCS options that support this

This leaves me with the only feasible option at this point: For every rendered development version (currently only [derivatives](https://bids-specification.readthedocs.io/en/derivatives/)), we can edit this line in the sourcecode:

https://github.com/bids-standard/bids-specification/blob/b6d7099e16861fa21a9a7f1f43cdef1a2ab13c5c/mkdocs.yml#L1

```diff
---site_name: Brain Imaging Data Structure v1.3.0-dev
+++site_name: Brain Imaging Data Structure v1.3.0-DERIVATIVES_DRAFT
```

Opinions and alternatives are welcome, I am not super happy with this solution.

KirstieJane:
Thanks for writing this up @sappelhoff & for thinking about it @franklin-feingold!

The idea we had in the meeting was potentially changing the colour of the theme....that seems more customisable, right? Maybe purple rather than blue? (something like that)

[...]

PR #383, opened 2019-12-13
ENH: Add watermark to drafts

effigies:
Addresses #373.

Created a mkdocs plugin that enables overriding config options on a per-branch basis: https://github.com/effigies/mkdocs-branchcustomization-plugin

Set the plugin to add a stylesheet to every branch but `master`. I pushed to the main repository and enabled RTD for this branch, so we can see it: https://bids-specification.readthedocs.io/en/enh-watermark/

Things that need doing:

* [x] ReadTheDocs adds CSS for their version panel that this obliterates. Fix that.
* [x] Test with tags to make sure that we aren't going to mark all releases as draft.
* [ ] Think about styling.
* [ ] Consider: Do we want two or three cases? We currently have master/other. We could have release/latest/other.


effigies:
> In the future, this would boil down to simply adding a CSS file and adjusting a line for your mkdocs plugin, right?

Yes. Something like:

```
- branch: /^master$/
  extra_css:
    - style/unreleased.css
- branch: /(?!^master$)/
  extra_css:
    - style/draft.css
```

effigies:
Cool. Thanks for the reviews. I'll go ahead and merge so I can include it in `common-derivatives` and `derivatives`,

effigies:
> In the future, this would boil down to simply adding a CSS file and adjusting a line for your mkdocs plugin, right?

Actually, I think that what I described above would work on any tag except ones that are on the latest `master` commit. We'd need to update the plugin to handle tags.

PR #354, opened 2019-10-22
[ENH] Governance structure

franklin-feingold:
adding in the governance structure document

PR #389, opened 2020-01-07
[MISC] moved list of extension proposals to the main BIDS website

robertoostenveld:
The steering group identified that information about the bids community and style of working are scattered and there is not a clear starting point. This PR aims to consolidate information on the process of BEPs on the homepage and move it away from the technical specification itself.

The page `06-extensions.md` remains (for now), but without the table, and simply points to the main website.

This also relates to https://github.com/bids-standard/bids-specification/pull/354, which proposed to add the Governance document to the specification. Rather than adding it to the specification on RTD, it should better be added to the homepage.


effigies:
:+1: I agree with keeping the spec text as focused as possible, with some linking to the rest of the community.

src/06-extensions.md
```diff
@@ -1,71 +1,8 @@
...
-This page lists some ways that you can get involved with the BIDS community.
-
-## Contributing to BIDS
-
-There are many ways to get involved with the BIDS community!
-
-### The BIDS Starter Kit
-
-If you're new to the BIDS community and you'd like to learn a bit more, we
-recommend checking out the [BIDS Starter Kit](https://github.com/bids-standard/bids-starter-kit/blob/master/README.md).
```
sappelhoff: are you going to remove the BIDS Starter Kit dropdown menu as well? 
image

On the website and in your updating PR (https://github.com/bids-standard/bids-website/pull/77), I don't see a reference to the starter kit yet - perhaps it'd be good to add it somewhere.

robertoostenveld: The starter kit is already mentioned on the website under https://bids.neuroimaging.io/getting_started.html. 

Although I did not make the initial section headings - just copied them - I suppose that "getting started" on the main website is more aimed towards _users_ and "get involved" more at _developers/contributors_. The distinction between the two is ideally not very strict, since everyone would contribute one way or another to the BIDS ecosystem (with data, knowledge, insights, skills, technical effort, etc). 

On the bids-specification site on RTD, the starter kit was not only mentioned on https://bids-specification.readthedocs.io/en/stable/06-extensions.html (even though the starter kit does not really deal with how to work on BEPs), but also still features in the top-level menu that appears on the left of the screen.

robertoostenveld: Oh, I now see your 1st question:

> are you going to remove the BIDS Starter Kit dropdown menu as well?

If it is up to me, yes. But I think that should be a separate PR on both sides, so this PR keeps that as is. We first might want to improve the way how the Starter Kit is featured on the main website.  And perhaps even reconsider the current Starter Kit implementation altogether.

sappelhoff: Okay, that's a good way to move forward.

sappelhoff:
Thanks @robertoostenveld

yarikoptic:
Fwiw, and sorry that I came too late. My 1c: I liked being able to search in the spec for the term of interest (eg ASL) and be able to see that it is not just not yet supported but that there is a bep I could go to. In my view, moving beps references to the main website actually scatters information more instead of consolidating.

PR #397, opened 2020-01-24
Adding genetic descriptor to RTD drop down menu

franklin-feingold:


effigies:
This is subsumed in #398.

PR #408, opened 2020-02-07
Units2ascii

satra:
this should not happen in the future, but the current units PR is based off a branch on the main repo. so just merging to that.


sappelhoff:
Wouldn't it be possible to rebase `satra:units2ascii` on current `master`? That way we wouldn't need another PR.

But if this is easier for you this way, why not.

satra:
@sappelhoff - just rebased.

Issue #421, opened 2020-02-19
Section numbering & consistency with mkdocs navigation labels


nicholst:
This issue raises two distinct but connected issues.

In #400 work on the pdf rendering, each section has to be treated as a "Chapter" and is explicitly given a "Chapter ?" number.  As we *already* number the Appendices, *and* as it common for  standards to have explicitly numbered sections to make it easier to reference different sections, I would like to propose that use numbering.  This can be activated during the PDF rendering, but apparently *also* with the HTML rendering?  (see [here](https://stackoverflow.com/questions/19999696/are-numbered-headings-in-markdown-rdiscount-possible/21126615)).

Secondly, as the PDF rendering is based *entirely* on the section header labels, it does not match the mkdocs HTML, left-panel navigation menu. Below is the comparison between the labels used in the left-nav HTML (left col) and the section headers (right col).

I propose these should be revised to agree exactly, and going forward require they are the same, to consistency between navigation and the headings, and between perceived navigation (HMTL) and PDF navigation.

Thoughts?

<pre>
 - Introduction                                        |Introduction
 - Common principles                                   |Common principles
 - Modality agnostic files                             |Modality<b>-</b>agnostic files
 - Modality specific files                             |
   - Magnetic Resonance Imaging                        |Magnetic Resonance Imaging <b>data</b>
   - Magnetoencephalography                            |Magnetoencephalography<b> (MEG)</b>
   - Electroencephalography                            |Electroencephalography<b>  (EEG)</b>
   - intracranial Electroencephalography               |intracranial Electroencephalography<b>  (iEEG)</b>
   - Task events                                       |Task events
   - Physiological and other continuous recordings     |Physiological and other continuous recordings
   - Behavioral experiments (with no MRI)              |Behavioral experiments (with no MRI)
 - Longitudinal and multi-site studies                 |Longitudinal <b>studies with multiple sessions (visits)</b>
 - Extending the BIDS specification                    |<b>BIDS Extension Proposals</b>
 - Appendix                                            |
   - Contributors                                      |Appendix I: Contributors
   - Licenses                                          |Appendix II: Licenses
   - HED                                               |Appendix III: <b>Hierarchical Event Descriptor (HED) Tags</b>
   - Entity-table                                      |Appendix IV: Entity table
   - Units                                             |Appendix V: Units
   - MEG file formats                                  |Appendix VI: MEG file formats
   - MEG systems                                       |Appendix VII: <b>preferred names of</b> MEG systems
[...]

PR #400, opened 2020-01-30
Updated pdf version specs

Arshitha:
Bugs Fixed: 
- Stable Table of Contents page
- Cover page 
- Non-tabular text overflow 
- Overlapping columns of tables (owing to small widths of the pipe tables in the markdown files)
  - Markdown files were edited (one-time edit) to increase the column widths of pipe tables
  - Not sure how this would affect the read the docs version of specs
- Automating header to include the latest release version number and date 

Issues that need to be fixed: 
- Internal links referencing either the same markdown file or other markdown files break during conversion from markdown to pdf. Currently, all of the internal links are removed and replaced by the accompanying text. 
- Superscripts are lost in conversion as well 
- Tried a couple different family of fonts to fix the missing character warnings for emojis used in the markdowns with no success. It would be ideal to fix this to match the Read The Docs version more closely
- Entity table in Appendix IV has overlapping unreadable columns since it's markdown equivalent has more than 10 columns. Refer to issue for more discussion

PR #422, opened 2020-02-19
[FIX] Unify section titles and table-of-contents entries

nicholst:
Work on #400 revealed that there were inconsistencies between the mkdocs navigation titles and the section titles in the markdown files, a difference that becomes more evident when only a rendered PDF is examined and compared to the HTML version.

These changes mainly comprise dropping some terms, removing unneeded hypens and removing abbreviations.  The abbreviations, so removed, then do need to be moved into the first line of the corresponding section.

See also #421 that discusses this issue.

### DIFF
--- a/.circleci/artifact_path
+++ b/.circleci/artifact_path
@@ -1,1 +1,1 @@
-0/home/circleci/project/site/01-introduction.html
+0/site/01-introduction.html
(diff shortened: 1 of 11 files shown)


nicholst:
> One additional (unrelated) request: Could you edit this file [...]

Done @sappelhoff !

sappelhoff:
Thanks @nicholst

PR #424, opened 2020-02-25
[ENH] Add part entity for complex-valued data

### LINKED ISSUE #367: Add "part" entity for magnitude and phase data
I would like to add support for magnitude and phase data across modalities (except func, as a new suffix was added there in #128). This new "part" entity is part of [BEP-001](https://github.com/bids-standard/bep001/blob/297e6534c5b8470a294c75ff09eae69dd491d867/ProposedChangesToSpecification.md#part-metadata-field), but it is applicable to more than the modalities covered in that extension, including SWI (see the "part" entity proposed in [BEP-004](https://docs.google.com/document/d/1kyw9mGgacNqeMbp4xZet3RnDhcMmf4_BmRgKaOkO2Sc/edit?pli=1#)).

I think it would be a good idea to add the entity to the main specification outside of any extension proposal, which will also make BEP-001 easier to review when the time comes.

tsalo:
Closes #367. This adds a new entity to distinguish magnitude and phase MRI data (except for BOLD and field map data, which have separate suffices for separating magnitude and phase).

The `part` entity is already proposed in [BEP-001](https://github.com/bids-standard/bep001/blob/297e6534c5b8470a294c75ff09eae69dd491d867/ProposedChangesToSpecification.md#part-metadata-field) and [BEP-004](https://docs.google.com/document/d/1kyw9mGgacNqeMbp4xZet3RnDhcMmf4_BmRgKaOkO2Sc/edit?pli=1#), but applies to more than multi-contrast MRI and SWI (e.g., single-band reference images).


sappelhoff:
Thanks for the proposel @tsalo. I am not too deep in the MRI related aspects of the spec, so let me summon @bids-standard/raw-mri for their opinions and reviews. :-)

effigies:
I'm not opposed to this in principle, but in #429 it looks like it's being used to split files that are too large. We should probably settle that...

I have some exposure to complex-valued images, but not very comprehensive. Is it the case that you always get it in phase/magnitude (which don't have a single-file representation) as opposed to complex-float values (which do)? Or that you always work in the phase/magnitude space so having to convert to/from complex is bound to accumulate arithmetic error?

Just for reference, the [types supported by NIfTI](https://github.com/NIFTI-Imaging/nifti_clib/blob/eedebca76263163b47a98be9b2e4f3fd75d251f7/niftilib/nifti1.h#L488-L518) are:

```C
#define DT_NONE                    0
#define DT_UNKNOWN                 0     /* what it says, dude           */
#define DT_BINARY                  1     /* binary (1 bit/voxel)         */
#define DT_UNSIGNED_CHAR           2     /* unsigned char (8 bits/voxel) */
#define DT_SIGNED_SHORT            4     /* signed short (16 bits/voxel) */
#define DT_SIGNED_INT              8     /* signed int (32 bits/voxel)   */
#define DT_FLOAT                  16     /* float (32 bits/voxel)        */
#define DT_COMPLEX                32     /* complex (64 bits/voxel)      */
#define DT_DOUBLE                 64     /* double (64 bits/voxel)       */
#define DT_RGB                   128     /* RGB triple (24 bits/voxel)   */
#define DT_ALL                   255     /* not very useful (?)          */


                            /*----- another set of names for the same ---*/
#define DT_UINT8                   2
#define DT_INT16                   4
#define DT_INT32                   8
#define DT_FLOAT32                16
#define DT_COMPLEX64              32
#define DT_FLOAT64                64
#define DT_RGB24                 128


                            /*------------------- new codes for NIFTI ---*/
#define DT_INT8                  256     /* signed char (8 bits)         */
#define DT_UINT16                512     /* unsigned short (16 bits)     */
#define DT_UINT32                768     /* unsigned int (32 bits)       */
#define DT_INT64                1024     /* long long (64 bits)          */
#define DT_UINT64               1280     /* unsigned long long (64 bits) */
#define DT_FLOAT128             1536     /* long double (128 bits)       */
#define DT_COMPLEX128           1792     /* double pair (128 bits)       */
#define DT_COMPLEX256           2048     /* long double pair (256 bits)  */
#define DT_RGBA32               2304     /* 4 byte RGBA (32 bits/voxel)  */
```

tsalo:
In my experience, you get separate magnitude and phase files. Siemens sequences have a "Magn./Phase" reconstruction option that generally outputs separate DICOM folders for the two types.

Also, the tools I'm aware of for working with magnitude+phase data (e.g., [phaseprep](https://github.com/ostanley/phaseprep), [t2star-mapping](https://github.com/neuropoly/t2star-mapping), and distortion correction in [sdcflows](https://github.com/nipreps/sdcflows)) all seem to use separate files.

effigies:
Yeah, I haven't seen anybody using complex values directly. If you haven't, it's probably best not to rock that particular boat.

src/99-appendices/04-entity-table.md
```diff
@@ -10,23 +10,24 @@ file type are denoted. Entity formats indicate whether the value is alphanumeric
...
+| Entity                                                                                         | Subject       | Session       | Task           | Acquisition   | Contrast Enhancing Agent | Reconstruction | Phase-Encoding Direction | Run           | Corresponding modality | Echo           | Recording           | Part               | Processed (on device) | Space           |
+| :--------------------------------------------------------------------------------------------- | :------------ | :------------ | :------------- | :------------ | :----------------------- | :------------- | :----------------------- | :------------ | :--------------------- | :------------- | :------------------ | :----------------- | :-------------------- | :---------------|
+| Format                                                                                         | `sub-<label>` | `ses-<label>` | `task-<label>` | `acq-<label>` | `ce-<label>`             | `rec-<label>`  | `dir-<label>`            | `run-<index>` | `mod-<label>`          | `echo-<index>` | `recording-<label>` | `part-<mag|phase>` | `proc-<label>`        | `space-<label>` |
```
yarikoptic, line 15: here should be no `<mag|phase>` should be just `<label>`.  Then specific modalities would come up with specific labels.

tsalo: Done!

yarikoptic: changes to table header widths need to propagate, otherwise leads to travis to puke on linting:
```
src/99-appendices/04-entity-table.md
  15:331-15:332  warning  Misaligned table fence  table-pipe-alignment  remark-lint
  15:355-15:356  warning  Misaligned table fence  table-pipe-alignment  remark-lint
  15:373-15:374  warning  Misaligned table fence  table-pipe-alignment  remark-lint
src/99-appendices/05-units.md: no issues found
```

tsalo: Ah, whoops. Thanks for catching that. Should be fixed now.

tsalo:
I can't figure out why linkchecker is failing. @yarikoptic do you have any ideas?

yarikoptic:
NB: copy/paste of the error would have expedited troubleshooting
```
URL        `https://www.w3schools.com/js/js_json_objects.asp'
Name       `JSON object'
Parent URL file:///root/build/site/02-common-principles.html, line 1136, col 18
Real URL   https://www.w3schools.com/js/js_json_objects.asp
Check time 0.062 seconds
Result     Error: SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:727)
```
and indeed:
```
$> wget https://www.w3schools.com/js/js_json_objects.asp
--2020-05-05 14:00:23--  https://www.w3schools.com/js/js_json_objects.asp
Resolving www.w3schools.com (www.w3schools.com)... 192.229.173.207
Connecting to www.w3schools.com (www.w3schools.com)|192.229.173.207|:443... connected.
ERROR: The certificate of ‘www.w3schools.com’ is not trusted.
ERROR: The certificate of ‘www.w3schools.com’ has expired.
The certificate has expired
```
that is the cost of the "security" ;) url is reachable via insecure http://www.w3schools.com/js/js_json_objects.asp as well, I guess worth a PR fixing it up.  Meanwhile twitted:
<blockquote class="twitter-tweet"><p lang="en" dir="ltr"><a href="https://twitter.com/wwwschools?ref_src=twsrc%5Etfw">@wwwschools</a> please address &quot;expired certificate&quot; for <a href="https://t.co/2jAeFoif1z">https://t.co/2jAeFoif1z</a> (I found no better means of contacting, sorry)</p>&mdash; Yaroslav Halchenko (@yarikoptic) <a href="https://twitter.com/yarikoptic/status/1257733110495686656?ref_src=twsrc%5Etfw">May 5, 2020</a></blockquote>

robertoostenveld:
I do not understand (from a standards- and consistency perspective) why we need a new entity to distinguish magnitude and phase for certain MR contrasts, whereas we already have it in the specification for others. See https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-2-two-phase-images-and-two-magnitude-images 

If we were to introduce `part-<mag|phase>`, is it subsequently also allowed to be used for BOLD contrast MR data? Would this ambiguity not lead to the risk of the specification becoming confusing? 

I believe the proposal to use it here reveals inconsistency in the use of modality either specified in the directory name (as in `sub-<label>/[ses-<label>/]<modality>/`) versus it being specified at the end of the filename (as in`_<modality>.nii.gz`).

yarikoptic:
Good observation @robertoostenveld !  I had similar reservation. And IMHO you are correct in:

> I believe the proposal to use it here reveals inconsistency in the use of modality either specified in the directory name (as in `sub-<label>/[ses-<label>/]<modality>/`) versus it being specified at the end of the filename (as in`_<modality>.nii.gz`).

just with a note that `<modality>/`, there is possibly an additional level of confusion since in BIDS we use "modality" as higher level construct (func, anat, dwi) and which is a superset of possible data file "types", and then the only `<modality` thing is 
```
$> git grep '<modality'                                                                             
src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md:        sub-<label>[_ses-<label>][_acq-<label>][_ce-<label>][_rec-<label>][_run-<index>]_<modality_label>.nii[.gz]
```
As you rightfully pointed out there are already `_magnitude{1,2}` and `_phase{1,2}` for `fmap` modality .  `fmap` could be an `_dir-..._epi.nii.gz` highlighting inconsistency.  IMHO it is just an artifact of the BIDS evolution ;)

tsalo:
> If we were to introduce part-<mag|phase>, is it subsequently also allowed to be used for BOLD contrast MR data? Would this ambiguity not lead to the risk of the specification becoming confusing?

I have to apologize for this. I added OPTIONAL to the entity table for BOLD contrast MR data in 6bcc2a5, but that was a mistake. This was corrected in 2f67010, after your comment. Using part with BOLD contrast MR data should **not** be allowed.

> I do not understand (from a standards- and consistency perspective) why we need a new entity to distinguish magnitude and phase for certain MR contrasts, whereas we already have it in the specification for others.

I agree that, from the perspective of a standard, this is not optimal. The existing suffices that distinguish magnitude and phase (i.e., `bold/phase` for func and `phase{1,2}/phasediff/magnitude{1,2}` for fmap) could probably have been consolidated if `part` and `echo` were used instead. However, I assume that field maps were incorporated into the specification before part was ever considered, and the choice to use `_<bold|phase>` instead of `_part-<mag|phase>_bold` was a conscious one made for practical reasons (see discussion in #128). We chose a separate suffix there because incorporating `part` for BOLD data would have caused problems for a lot of tools, which probably wasn't the best choice from the perspective of consistency.

Still, there are many other modalities that can have both magnitude and phase data, so I believe that there is a real need for a new entity to distinguish the two. If anything, I think that this would be an argument to _remove_ existing suffices that should be superseded by the part entity, rather than an argument against adding the entity. I don't think that's going to happen, though, since at least the fmap suffices are so widely used. I also don't know what process, if any, exists to remove things from the specification.

tsalo:
@yarikoptic Thank you for figuring out the linkchecker problem. It looks like it was fixed, since all tests are passing as of 2f67010.

yarikoptic:
With the #429 being resolved by #460 via renaming `_part` into `_split` for annotation of splits of MEEG files, IMHO the horizon is clear now for this to proceed forward.  Just some conflicts to be resolved first now (soon #475 will also make us to avoid such merge conflicts altogether ;) !)

tsalo:
@yarikoptic It should be good now. I'm really excited to see this one move forward!

robertoostenveld:
@tsalo thanks for clarifications. I agree.

> If anything, I think that this would be an argument to remove existing suffices that should be superseded by the part entity, rather than an argument against adding the entity. I don't think that's going to happen, though, since at least the fmap suffices are so widely used. I also don't know what process, if any, exists to remove things from the specification.

I propose you make a note of this in the [Suggestions for BIDS 2.0](https://docs.google.com/document/d/1LEgsMiisGDe1Gv-hBp1EcLmoz7AlKj6VYULUgDD3Zdw/edit#) document. Switching from 1.x to 2.0 would be the moment to make backward-incompatible changes.

tsalo:
@robertoostenveld Agreed. I've added a new suggestion to the Google Doc. It's wayyy down at the very, very bottom.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -204,8 +214,10 @@ Template:
 ```Text
 sub-<label>/[ses-<label>/]
     func/
-        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_<contrast_label>.nii[.gz]
-        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_sbref.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_bold.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_phase.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>][_part-<mag|phase>]_cbv.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>][_part-<mag|phase>]_sbref.nii[.gz]
```
yarikoptic, line 220: I am not sure if such detailed specification and expansion of different `contrast_label`s is needed here. 
`_bold` is indeed a special case though since now it has a dedicated `_phase`.  So now there cannot be another `func` contrast than `_bold` which would have its `_phase`. 
NB ASL BEP005 seems decided to avoid `func/` and went for a dedicted `perf/` (perfusion) although its purpose AFAIK would be `func/`.
Unfortunately I did not follow the discussion in #128.  I would have voted for any `contrast_label` to "assume" its default "composition" - either it is a mag or phase or combined if no `_part` is provided.  Then #128 could have resulted in adding `_part-phase` and having explicit `_part-phase_bold`, opening possibilities for any other `_part-phase_<contrast>`.

Sure thing `_part-` makes no sense with `_phase`, but I think that is just a corner case which might be gone in 2.0.
So why not just to keep `[_part-<mag|phase>]_<contrast_label>` and just add a note to `_phase` in the table that `_part` then should not be specified. Also it might be worth starting adding "deprecation" statements on elements which will be removed in 2.0, but that would be a separate issue I guess ;)

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -267,6 +279,25 @@ Please note that the `<index>` denotes the number/index (in a form of an
 integer) of the echo not the echo time value which needs to be stored in the
 field EchoTime of the separate JSON file.
 
+Complex-valued data MUST be split into one file for each data type.
+For BOLD data, there are separate suffices for magnitude (`_bold`) and phase
+(`_phase`) data.
```
yarikoptic, line 284: echoing on above I would have preferred `1.x` was "forward looking" and allowed `_part-<mag|phase>` for `_bold` while announcing `_phase` deprecated.  Similar to how such deprecations are done in software.  That would ease later migration to BIDS 2.0.  It would be only a slight burden for bids-apps to support both layouts IMHO, but would also simplify their support for BIDS 2.0.

yarikoptic (dismissed):
Well, may be a somewhat past-the-time idea/discussion point.  I am ok also with merging as is, and then moving discussion into another issue/PR since it could easily be incremental.

tsalo:
I've replaced the entity definition with the one from #508, to make integration with BEP001 as seamless as possible and to support real and imaginary data in addition to magnitude and phase data. I also merged from upstream and updated the schema, so I think this is ready for another round of review.

tsalo:
@agahkarakuzu GitHub won't let me request your review on this PR (and thus it may not even count toward the two required reviews), but I'd love to get your opinion.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -175,6 +175,35 @@ sequences using different contrast enhanced images. The label is the name of the
...
+```Text
+sub-01_part-mag_T1w.nii.gz
+sub-01_part-mag_T1w.json
+sub-01_part-phase_T1w.nii.gz
+sub-01_part-phase_T1w.json
+```
+
+Phase images MAY be in radians or in arbitrary units.
+The sidecar JSON file MUST include the units of the `phase` image.
+The possible options are: `radians` or `a.u.` for grayscale images.
```
tsalo, line 194: Related to #580, do we want "a.u." or "arbU"?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -175,6 +175,35 @@ sequences using different contrast enhanced images. The label is the name of the
...
+
+sub-01_part-phase.json
+
+```Text
+{
+   "Units": "radians"
+}
+```
+
+If the `part` key/value pair is absent from a filename, the image SHOULD be a magnitude image.
```
effigies, line 205: This feels like a strange directive. Does it permit me to have:

```
sub-01_T1w.nii.gz
sub-01_part-phase_T1w.nii.gz
```

I might instead treat this as `run-`, which we say may be omitted if only one run is provided.

```suggestion
When there is only a magnitude image of a given type, the `part` key MAY be omitted.
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
effigies: I agree with @yarikoptic. I would suggest that we say `_part-<mag|phase>`, and then say "For backwards compatibility, `_phase.nii.gz` is considered equivalent to `_part-phase_bold.nii.gz`."

tsalo: I didn't realize that was possible, but I certainly love the idea. By the way, for posterity's sake, I have an open issue in `bids-2-devel` about eliminating `_phase` (and other suffixes that could be better handled with this entity): https://github.com/bids-standard/bids-2-devel/issues/3.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: Agreed. I'd prefer to remove "grayscale", although I believe it's used fairly frequently in #508.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -267,7 +298,29 @@ Please note that the `<index>` denotes the number/index (in the form of a
...
+sub-01/
+   func/
+      sub-01_task-cuedSGT_part-mag_bold.nii.gz
+      sub-01_task-cuedSGT_part-mag_bold.json
+      sub-01_task-cuedSGT_part-phase_bold.nii.gz
+      sub-01_task-cuedSGT_part-phase_bold.json
+      sub-01_task-cuedSGT_part-mag_sbref.nii.gz
+      sub-01_task-cuedSGT_part-mag_sbref.nii.gz
+      sub-01_task-cuedSGT_part-phase_sbref.nii.gz
+      sub-01_task-cuedSGT_part-phase_sbref.nii.gz
```
effigies, line 320:
```suggestion
      sub-01_task-cuedSGT_part-mag_sbref.json
      sub-01_task-cuedSGT_part-phase_sbref.nii.gz
      sub-01_task-cuedSGT_part-phase_sbref.json
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: Should be good now (although the wording could use a review).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -173,6 +173,33 @@ sequences using different contrast enhanced images. The label is the name of the
...
+sub-01_part-phase_T1w.json
+
+```Text
+{
+   "Units": "rad"
+}
+```
+
+When there is only a magnitude image of a given type, the `part` key MAY be omitted.
+
```
effigies, line 202: Having removed section headers, the flow of this is now strange. We start with "This entity" and don't reference `part` until the next sentence. The `ContrastBolusIngredient` seems like it should be associated with `ce-`, which is defined above `part-`.

I've proposed #615 for promoting `Units` beyond this one context, which I think would also have the side effect of allowing this section to be a bit more compact and improve the flow.

tsalo: I think the text around `ContrastBolusIngredient` should be changed so it doesn't sound like it's talking about anatomical MRI-specific metadata in general, but for this PR I've just moved that paragraph up to be back with the `ce` entity definition.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -270,7 +302,29 @@ Please note that the `<index>` denotes the number/index (in the form of a
 nonnegative integer) of the echo not the echo time value which needs to be stored in the
 field EchoTime of the separate JSON file.
 
-Some meta information about the acquisition MUST be provided in an additional
+Complex-valued data MUST be split into one file for each data type.
+For BOLD data, there are separate suffixes for magnitude (`_bold`) and phase
+(`_phase`) data, but the `_phase` suffix is considered a legacy suffix
+(i.e., still supported in the specification, but to be deprecated later).
```
sappelhoff, line 308: good place to insert a link about our definition of deprecation here, see https://github.com/bids-standard/bids-specification/issues/611

tsalo: Agreed, although I was planning to wait until a corresponding PR was merged.

agahkarakuzu:
After long discussions about how to arrange suffixes needed for qMRI in #508 with @tsalo @emdupre @Gilles86 @tiborauer @lazaral and @ChristophePhillips we decided that introducing a concept of `logical grouping of images that are linked by varying acquisition parameters` serves as a common denominator for many applications. 

However, there is still a need to make the fundamental distinction between `parametric` (i.e. voxel values are in physical units) and `non-parametric` (i.e. voxel values are given in an arbitrary grayscale) `suffixes`. I think the schema solution is a great opportunity to do achieve this! 

Two examples for would look like this: 
 
#### parametric 
```diff
 suffixes:
   -  T1map 
   -  T2map 
   -  T2starmap
   -  T1rho
  extensions:
    - .nii.gz
    - .nii
    - .json
  entities:
    sub: required
    ses: optional
    run: optional
    acq: optional
    ce: optional
    rec: optional
    part: optional
+ category:
+   type: parametric
+   unit:  seconds 
``` 
#### non-parametric  
```diff
 suffixes:
   -  T1w
   -  T2w 
   -  PDw
   -  FLAIR
  extensions:
    - .nii.gz
    - .nii
    - .json
  entities:
    sub: required
    ses: optional
    run: optional
    acq: optional
    ce: optional
    rec: optional
    part: optional
+ category:
+   type: non-parametric
+   unit:  arbitrary 
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -194,6 +194,36 @@ fields specific to anatomical scans:
...
+```Text
+sub-01_part-mag_T1w.nii.gz
+sub-01_part-mag_T1w.json
+sub-01_part-phase_T1w.nii.gz
+sub-01_part-phase_T1w.json
+```
+
+Phase images MAY be in radians or in arbitrary units.
+The sidecar JSON file MUST include the units of the `phase` image.
+The possible options are `rad` or `a.u.`.
```
tsalo, line 214:
```suggestion
The possible options are `rad` or `arbitrary`.
```

tsalo:
@agahkarakuzu Thanks! We need to specify metadata fields in the schema at some point anyway, so I was thinking of specifying options for the `Units` field in that way, like the following:

```diff
suffixes:
   -  T1map 
   -  T2map 
   -  T2starmap
   -  T1rho
  extensions:
    - .nii.gz
    - .nii
    - .json
  entities:
    sub: required
    ses: optional
    run: optional
    acq: optional
    ce: optional
    rec: optional
    part: optional
+ metadata:
+   Units:
+       - seconds
```

I may not fully understand the different between parametric and nonparametric images, though. Is it enough if the required `Units` are something other than arbitrary? If so, it should be fairly easy to split up suffix tables in the specification text by that parameter.

NOTE: Adding metadata to the schema may or may not fall under #609 (I'm not entirely sure about the scope of that PR yet).

agahkarakuzu:
@tsalo indeed, as long as `units` is there, `parametric` vs `non-parametric` is kind of redundant. I will be following updates from these PRs, thank you!

tiborauer:
@agahkarakuzu, I think @Gilles86 pointed out that there is no unit e.g. in MTR?

agahkarakuzu:
@tiborauer I think we can come up with a reasonable solution for semi-quantitative methods with relative units. Such as `percentage` or `normalized percentage` etc.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
effigies: @tsalo #634 is merged. Want to update this? This PR overall LGTM, and this is the only thing left to change.

tsalo: Good catch. I've updated the text. It would be nice to link directly to the deprecation definition, but I did the next best thing and links to the Definitions section.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -236,7 +236,7 @@ Currently supported image contrasts include:
 |---------- | ---------------- | ---------------------------------------------------------------------------------------------------- |
 | BOLD      | bold             | Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)                                   |
 | CBV       | cbv              | Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images) |
-| Phase     | phase            | Phase information associated with magnitude information stored in BOLD contrast                      |
+| Phase     | phase            | Phase information associated with magnitude information stored in BOLD contrast. [Deprecated](../02-common-principles.md#definitions).         |
```
effigies: Let's specify what it should be replaced with, as well.

tsalo: I'm really looking forward to a later version of the work in #610 so I won't have to edit Markdown tables directly ever again.. But yes, will do.

effigies:
@tsalo What's the status of this? Are we waiting on something, or just need reviews?

tsalo:
We just need reviews. I think I've addressed all concerns from previous rounds.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -202,19 +232,21 @@ different reconstruction algorithms (for example ones using motion correction).
...
 
 Template:
 
 ```Text
 sub-<label>/[ses-<label>/]
     func/
-        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_<contrast_label>.nii[.gz]
-        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_sbref.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>][_part-<label>]_bold.nii[.gz]
+        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_phase.nii[.gz]
```
effigies, line 247: Since it's deprecated, I would suggest dropping this.

```suggestion
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -202,19 +232,21 @@ different reconstruction algorithms (for example ones using motion correction).
 
 Currently supported image contrasts include:
 
-| **Name**  | `contrast_label` | **Description**                                                                                      |
-|---------- | ---------------- | ---------------------------------------------------------------------------------------------------- |
-| BOLD      | bold             | Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)                                   |
-| CBV       | cbv              | Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images) |
-| Phase     | phase            | Phase information associated with magnitude information stored in BOLD contrast                      |
+| **Name**  | `contrast_label` | **Description**                                                                                                                                                                                                     |
```
effigies, line 235: Should we go ahead with #571 or leave it to another PR?

```suggestion
| **Name**  | `suffix` | **Description**                                                                                                                                                                                                     |
```

src/schema/entities.yaml
```diff
@@ -115,6 +115,23 @@ recording:
...
+    This entity is used to indicate which component of the complex
+    representation of the MRI signal is represented in voxel data.
+    The `part-<label>` key/value pair is associated with the DICOM tag
+    `0008,9208`.
+    Allowed label values for this entity are `phase`, `mag`, `real` and `imag`,
+    which are typically used in `mag/phase` or `real/imag` pairs.
+
+    Phase images MAY be in radians or in arbitrary units.
+    The sidecar JSON file MUST include the units of the `phase` image.
+    The possible options are `rad` or `a.u.`.
```
effigies, line 130:
```suggestion
    The possible options are `rad` or `arbitrary`.
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -202,19 +232,21 @@ different reconstruction algorithms (for example ones using motion correction).
...
+|---------- | ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| BOLD      | bold             | Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)                                                                                                                                                  |
+| CBV       | cbv              | Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images)                                                                                                                |
+| Phase     | phase            | Phase information associated with magnitude information stored in BOLD contrast. [Deprecated](../02-common-principles.md#definitions). [`part-phase`](../99-appendices/09-entities.md#part) should be used instead. |
```
effigies, line 239: I think we can be a little BOLDER about this, and more explicit about using the `bold` suffix.

```suggestion
| Phase     | phase            | [DEPRECATED](../02-common-principles.md#definitions). Phase information associated with magnitude information stored in BOLD contrast. This suffix should be replaced by the [`part-phase`](../99-appendices/09-entities.md#part) in conjunction with the `bold` suffix. |
```

effigies (commented):
Just about there, IMO.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: This will be handled automatically in #610, but I don't see any harm in doing it now as well.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: Hm... I think #610 will revert this, since the schema doesn't have any concept of "deprecated" just yet. Still, I think it's a reasonable practice to adopt.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -203,19 +233,20 @@ different reconstruction algorithms (for example ones using motion correction).
...
+| BOLD      | bold             | Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)                                                                                                                                                                                                       |
+| CBV       | cbv              | Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images)                                                                                                                                                                     |
+| Phase     | phase            | [DEPRECATED](../02-common-principles.md#definitions). Phase information associated with magnitude information stored in BOLD contrast. This suffix should be replaced by the [`part-phase`](../99-appendices/09-entities.md#part) in conjunction with the `bold` suffix. |
```
effigies, line 240:
```suggestion
| **Name**  | `suffix` | **Description**                                                                                                                                                                                                                                                          |
|---------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| BOLD      | bold     | Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)                                                                                                                                                                                                       |
| CBV       | cbv      | Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images)                                                                                                                                                                     |
| Phase     | phase    | [DEPRECATED](../02-common-principles.md#definitions). Phase information associated with magnitude information stored in BOLD contrast. This suffix should be replaced by the [`part-phase`](../99-appendices/09-entities.md#part) in conjunction with the `bold` suffix. |
```

effigies (approved):
LGTM. Now it's a race with #672.

tsalo:
@yarikoptic's approval is fairly old now, so I'd like to get one more before initiating the five-day embargo period. Does that sound good to everyone?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -195,6 +195,36 @@ fields specific to anatomical scans:
 | ----------------------- | --------------------- | ------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
 | ContrastBolusIngredient | OPTIONAL              | [string][]    | Active ingredient of agent. Values MUST be one of: IODINE, GADOLINIUM, CARBON DIOXIDE, BARIUM, XENON Corresponds to DICOM Tag 0018,1048. |
 
+The [`part-<label>`](../99-appendices/09-entities.md#part) key/value pair is
+used to indicate which component of the complex representation of the MRI
+signal is represented in voxel data.
+This entity is associated with the DICOM tag `0008,9208`.
+Allowed label values for this entity are `phase`, `mag`, `real` and `imag`,
+which are typically used in `mag/phase` or `real/imag` pairs.
```
sappelhoff, line 203: >  which are typically used in `mag/phase` or `real/imag` pairs.

this may make some people believe they have to specify `*_part-mag/phase*`, whereas the example shows that it should be rather a pair of files - one for mag, one for phase (vice versa for real, imag).

src/99-appendices/09-entities.md
```diff
@@ -151,6 +154,26 @@ sampling frequencies).
...
+Full name: Part
+
+Format: `part-<label>`
+
+Definition: This entity is used to indicate which component of the complex
+representation of the MRI signal is represented in voxel data.
+The `part-<label>` key/value pair is associated with the DICOM tag
+`0008,9208`.
+Allowed label values for this entity are `phase`, `mag`, `real` and `imag`,
+which are typically used in `mag/phase` or `real/imag` pairs.
```
sappelhoff, line 168: see comment above on that I find the `mag/phase` suggestive of being some kind of syntax we support (but we don't)

src/schema/entities.yaml
```diff
@@ -118,6 +118,24 @@ recording:
...
+part:
+  name: Part
+  entity: part
+  description: |
+    This entity is used to indicate which component of the complex
+    representation of the MRI signal is represented in voxel data.
+    The `part-<label>` key/value pair is associated with the DICOM tag
+    `0008,9208`.
+    Allowed label values for this entity are `phase`, `mag`, `real` and `imag`,
+    which are typically used in `mag/phase` or `real/imag` pairs.
```
sappelhoff, line 130: see comment above on that I find the `mag/phase` suggestive of being some kind of syntax we support (but we don't)

sappelhoff (commented):
LGTM overall, except a minor comment.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: What about the following?

> which are typically used in `part-mag`/`part-phase` or `part-real`/`part-imag` pairs of files.

sappelhoff: Much better, thanks!

sappelhoff (approved):

tsalo:
:tada: It looks like we can start the embargo period. If there are no concerns before next Tuesday (11/24), I think we can merge this, right?

sappelhoff:
It's been 6 days since the last interaction with this PR, and it has all approving reviews it needs, so I am merging now.

Thanks @tsalo

oesteban:
Question: does this PR not affect fieldmaps at all? (I'm looking at https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#fieldmap-data)

Admittedly, the spec for fieldmaps is pretty unfortunate: we have eight suffixes (magnitude, magnitude1/2, phase1/2, phasediff, fieldmap, and epi) and most of them overlap with the new entity. Trying to apply the entity would end up producing pretty weird namings, because `fieldmap` and `phase[1,2,diff]` are not MR sequence names (and `epi` is actually the only one that will typically need only the magnitude).

This PR kind of pushes towards something like:
```
fmap/
   sub-1_part-mag_echo-1_gre.nii.gz       # equiv. magnitude1
   sub-1_part-mag_echo-2_gre.nii.gz       # equiv. magnitude2
   sub-1_part-phase_echo-1_gre.nii.gz     # equiv. phase1
   sub-1_part-phase_echo-2_gre.nii.gz     # equiv. phase2
   sub-1_part-mag_sei.nii.gz              # sei = spiral-echo imaging; equiv. to magnitude
   sub-1_part-phase_sei.nii.gz            # equiv. to _fieldmap
   sub-1_part-mag_gre.nii.gz              # equiv. to _magnitude1 corresponding to phasediff
   sub-1_part-phase_gre.nii.gz            # equiv. to phasediff
```

Also, the units of these `part-phase` should be allowed to be given in rad, rad/s, or Hz (and could be an argument for ppm).

Another argument could be had to unify `_sei` and `_gre` as a unique suffix, if there were a metadata to indicate whether the image's range is clipped between -pi .. pi.

But this is backwards incompatible, so I guess (coming back to the beginning) that fieldmaps are unaffected by this PR?

oesteban:
although, not breaking compatibility so much, we could imagine something like:

```
fmap/
   sub-1_part-mag_echo-1_phasediff.nii.gz       # equiv. magnitude1
   sub-1_part-mag_echo-2_phasediff.nii.gz       # equiv. magnitude2
   sub-1_part-phase_echo-1_phasediff.nii.gz     # equiv. phase1
   sub-1_part-phase_echo-2_phasediff.nii.gz     # equiv. phase2
   sub-1_part-mag_phasediff.nii.gz              # equiv. to _magnitude1 corresponding to phasediff
   sub-1_part-phase_phasediff.nii.gz            # equiv. to phasediff
   sub-1_part-mag_fieldmap.nii.gz               # equiv. to magnitude
   sub-1_part-phase_fieldmap.nii.gz             # equiv. to _fieldmap
```

tsalo:
I have an open issue about that for BIDS 2.0: https://github.com/bids-standard/bids-2-devel/issues/3

I had assumed that the new suffix would just be `epi`, but `phasediff` sounds good too.

oesteban:
Feels a bit awkward to modify all other MRI instances that almost always use just the magnitude (80/20) and keep the subset where the 80/20 is reversed (i.e., 80% would be better off with part-).

agahkarakuzu:
@oesteban although we were lucky to start from a clean slate and have no concerns about retro-compatibility, I'd like to leave my two cents about how we approached this for RF fieldmaps (B1+ and B1-): 

To get rid of the ambiguity caused by the use of suffixes attaining common sequence names (e.g. `gre`) or terms (e.g. `phase`), we introduced the common concept of [entity-linked file collections](https://bids-specification.readthedocs.io/en/latest/99-appendices/10-file-collections.html). These collections make use of suffixes that can clearly describe the intended application of a set of files (e.g. `DEB0` instead of generic `gre` for double-echo GRE based B0 mapping) when coupled with linking entities (`echo`, `part`, `flip` etc.) relevant to the process.

Similar to that in B0 mapping, there are various B1 mapping methods revolving around alternating certain acquisition parameters. The file-collection approach enabled describing [several B1 mapping methods](https://bids-specification.readthedocs.io/en/latest/99-appendices/10-file-collections.html#fieldmap-data). 

When descriptive suffixes are available in the context of file collections, the use of `part` (and other linking entities) ease the definition of quantitative MRI inputs with clarity. I know retro-compatibility is a stumbling block for such changes at the moment, but if it is no longer an issue from `v2.0` on, I'd say `part` entity would be a great help for organizing B0 map inputs with file collections in mind.

effigies:
That fieldmaps are the most common case where `part-` would be most clarifying actually makes the transition harder. If we do that, then tools are going to need to support old- and new-style fieldmap names. So sdcflows, which aims to support the four types of fieldmaps would need to support seven naming conventions.

With `_phase.nii`, it was a pretty niche case and recently added, so deprecating it for `part-phase_bold.nii` affects quite few datasets and (probably) fewer tools.

Still, I think it might be a good idea to put this to the community. Deprecating 3 out of 4 existing fieldmap conventions would cause overhead for tools (that now have to support 7 conventions), but it would prod dataset curators toward a more coherent convention for their fieldmaps and ease the transition to BIDS 2.

oesteban:
@tsalo @agahkarakuzu @effigies thanks very much for your explanations, it makes sense - just wanted to be explicit as regards that leaving the fieldmaps part of the spec untouched. As Chris mentioned, I see the compromise we need to achieve between keeping tools updated within BIDS 1 but making coherent steps for a smooth transition towards BIDS 2.

Question for @agahkarakuzu, the B1 mapping methods you referenced, are those artifacts stored under `fmap/`?

PR #430, opened 2020-03-10
bids spec pdf version

Arshitha:
This is a PR almost identical to PR [#400 updated-pdf-version-specs](https://github.com/bids-standard/bids-specification/pull/400) except that this has much cleaner commit history and was created to address the issue of unwanted commits in the above referenced PR. 

To gain a better understanding of why certain decisions were made while fixing the issue of generating pdf version of the specs, discussions in the following issues would be useful references: 
- pdf version of spec bids-standard/bids-specification#135
- updated-pdf-version-specs bids-standard/bids-specification#400

### Bugs Fixed: 
- Stable Table of Contents page
- Cover page 
- Non-tabular text overflow 
- Overlapping columns of tables (owing to small widths of the pipe tables in the markdown files)
  - Markdown files were edited (one-time edit) to increase the column widths of pipe tables
  - Does not affect the read the docs version of specs. However, if any tables were added to the specs in future, there needs to a standard that works for both pdf AND read the docs version of the specs. 
- Automating header to include the latest release version number and build date (pulled from `mkdocs.yml`)

### Enhancements required (non-exhaustive list): 
- Internal links referencing either the same markdown file or other markdown files break during conversion from markdown to pdf. Currently, all of the internal links are removed and replaced by the accompanying text. 
- Superscripts are lost in conversion as well 
- Tried a couple different family of fonts to fix the missing character warnings for emojis used in the markdowns with no success. It would be ideal to fix this to match the Read The Docs version more closely
- Entity table in Appendix IV has overlapping unreadable columns since it's markdown equivalent has more than 10 columns. Refer to issue for more discussion


Arshitha:
@sappelhoff Except the last two commits the rest are perhaps from the master branch of my forked repo. would this work?

PR #427, opened 2020-02-25
pdf version specs

Arshitha:
This is a PR almost identical to PR [#400 updated-pdf-version-specs](https://github.com/bids-standard/bids-specification/pull/400) except that this has much cleaner commit history and was created to address the issue of unwanted commits in the above referenced PR. 

To gain a better understanding of why certain decisions were made while fixing the issue of generating pdf version of the specs, discussions in the following issues would be useful references: 
- pdf version of spec bids-standard/bids-specification#135
- updated-pdf-version-specs bids-standard/bids-specification#400

### Bugs Fixed: 
- Stable Table of Contents page
- Cover page 
- Non-tabular text overflow 
- Overlapping columns of tables (owing to small widths of the pipe tables in the markdown files)
  - Markdown files were edited (one-time edit) to increase the column widths of pipe tables
  - Does not affect the read the docs version of specs. However, if any tables were added to the specs in future, there needs to a standard that works for both pdf AND read the docs version of the specs. 
- Automating header to include the latest release version number and build date (pulled from `mkdocs.yml`)

### Enhancements required (non-exhaustive list): 
- Internal links referencing either the same markdown file or other markdown files break during conversion from markdown to pdf. Currently, all of the internal links are removed and replaced by the accompanying text. 
- Superscripts are lost in conversion as well 
- Tried a couple different family of fonts to fix the missing character warnings for emojis used in the markdowns with no success. It would be ideal to fix this to match the Read The Docs version more closely
- Entity table in Appendix IV has overlapping unreadable columns since it's markdown equivalent has more than 10 columns. Refer to issue for more discussion

PR #431, opened 2020-03-10
[INFRA] add support for building PDF versions of the spec

Arshitha:
This is a PR almost identical to PR [#400 updated-pdf-version-specs](https://github.com/bids-standard/bids-specification/pull/400) except that this has much cleaner commit history and was created to address the issue of unwanted commits in the above referenced PR. 

To gain a better understanding of why certain decisions were made while fixing the issue of generating pdf version of the specs, discussions in the following issues would be useful references: 
- pdf version of spec bids-standard/bids-specification#135
- updated-pdf-version-specs bids-standard/bids-specification#400

### Bugs Fixed: 
- Stable Table of Contents page
- Cover page 
- Non-tabular text overflow 
- Overlapping columns of tables (owing to small widths of the pipe tables in the markdown files)
  - Markdown files were edited (one-time edit) to increase the column widths of pipe tables
  - Does not affect the read the docs version of specs. However, if any tables were added to the specs in future, there needs to a standard that works for both pdf AND read the docs version of the specs. 
- Automating header to include the latest release version number and build date (pulled from `mkdocs.yml`)

### Enhancements required (non-exhaustive list): 
- Internal links referencing either the same markdown file or other markdown files break during conversion from markdown to pdf. Currently, all of the internal links are removed and replaced by the accompanying text. 
- Superscripts are lost in conversion as well 
- Tried a couple different family of fonts to fix the missing character warnings for emojis used in the markdowns with no success. It would be ideal to fix this to match the Read The Docs version more closely
- Entity table in Appendix IV has overlapping unreadable columns since it's markdown equivalent has more than 10 columns. Refer to issue for more discussion


Arshitha:
@sappelhoff done! Should we delete #427 ?

sappelhoff:
> @sappelhoff done! #427 and #430 are redundant at this point and can be confusing. How do we deal with this? Apparently, deleting a PR isn't straightforward?

I just closed the PRs, that should suffice.

src/05-longitudinal-and-multi-site-studies.md
```diff
@@ -1,4 +1,4 @@
-# Longitudinal and multi-site studies
+# Longitudinal studies and multi-site studies
```
sappelhoff: Why this change? It's `Longitudinal and multi-site studies` everywhere else, see: https://github.com/bids-standard/bids-specification/blob/ece9e8737d15a4ca9e5453dc8c473abbd3e4dd91/mkdocs.yml#L34

I think this change should be reverted

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -1,4 +1,4 @@
-# Magnetoencephalography
+# Magnetoencephalography
```
sappelhoff:
```suggestion
# Magnetoencephalography
```

unnecessary space at the end of line

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -1,4 +1,4 @@
-# Intracranial Electroencephalography
+# intracranial Electroencephalography
```
sappelhoff: one more unnecessary space at the end of the line: And Why did you turn the upper case i into a smaller case i? I think it should stay upper case, like in the TOC: https://github.com/bids-standard/bids-specification/blob/ece9e8737d15a4ca9e5453dc8c473abbd3e4dd91/mkdocs.yml#L30

Arshitha: when I added the pdf generation files after syncing my fork with the original repo, there were some changes in the markdown files that I had to fix manually so these were oversights. I've fixed them now.

sappelhoff:
PS: @Arshitha have you already added your contributions to our contributors list? You can just edit this wiki page and upon the next release, we'll add the information from the wiki to the repository:

https://github.com/bids-standard/bids-specification/wiki/Contributors

Arshitha:
@sappelhoff Thank you for all the guidance!

Issue #445, opened 2020-04-14
task "label" sometimes referred to as "task_name" or "task_label"


Remi-Gau:
There seems to be some cases where the `label` place holder for `task`changes a bit:

It seems that in most places it is just `label`. See fMRI section below.

```
sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_<contrast_label>.nii[.gz]
```

But sometimes it is `task_label`: see the ieeg section below.
```
sub-<label>[_ses-<label>]_task-<task_label>[_run-<index>]_ieeg.json
```

And sometimes it is `task_name`: see the behavioral section below.
```
sub-<label>/[ses-<label>/]
    beh/
        sub-<label>[_ses-<label>]_task-<task_name>_events.tsv
        sub-<label>[_ses-<label>]_task-<task_name>_events.json
```

I don't think this is a big issue but I am always afraid that inconsistencies might confuse people, especially new-comers.

Or maybe there is a good reason for these differences?

nicholst:
Good catch @Remi-Gau!

Seems like `task_label` and `task_names` are the anachronisms, and they could be changed to `label` as long as the surrounding text is clear about the meaning of the label.

Remi-Gau:
OK I will open a pull request to change those.

PR #446, opened 2020-04-14
[FIX] fix inconsistencies for task label between sections

Remi-Gau:
Related to issue #445 

- uniformize `task-<label>` in the filename template across sections
- make the `TaskName` description for the JSON sidecar exactly match that of fMRI
- fix some typos


sappelhoff:
Thanks for the fixes @Remi-Gau! :-)

Unfortunately the edits have misaligned a few table fences, Can you fix the linter, please?

Also I find this new phrasing a bit hard to understand. What was your intention with this change?

```diff
-The Task label is derived from this field ...
+Task label (`task-`) included in the file name is derived from this field ...
```

```
(`task-`) 
```

is not the task label, but rather the key of the "task key-value pair", with the label being the value. So that's why I am a bit confused.

Remi-Gau:
> Also I find this new phrasing a bit hard to understand. What was your intention with this change?
> 
> ```diff
> -The Task label is derived from this field ...
> +Task label (`task-`) included in the file name is derived from this field ...
> ```
> 
> ```
> (`task-`) 
> ```
> 
> is not the task label, but rather the key of the "task key-value pair", with the label being the value. So that's why I am a bit confused.

For this, I literally reused the wording from the fMRI section. So I suspect we might want to clarify this one to have all sections consistent. :-)

How about this?
```
The task label included in the file name is derived from this TaskName field by removing all non alphanumeric (`[a-zA-Z0-9]`) characters. For example TaskName `faces n-back` will correspond to task label `facesnback`.
```

FYI the `behavioral` section has no corresponding description for the task label but the explaining text says to refer to the other sections, so it should be enough IMHO.

sappelhoff:
> For this, I literally reused the wording from the fMRI section. 

I see :thinking: 

I like your new suggestion from https://github.com/bids-standard/bids-specification/pull/446#issuecomment-613522635 and agree that having this consistent across the spec is good to have :+1:

effigies:
Sorry, one annoying suggestion: "non alphanumeric" should be "non-alphanumeric".

And let's update the entry in MRI, to be consistent:

https://github.com/bids-standard/bids-specification/blob/42bcc6c6700777c0aff4ffc2e95d6ccbf70f0882/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md#L279

Remi-Gau:
> And let's update the entry in MRI, to be consistent:

Can't believe I did not update the MRI section. Silly me.

sappelhoff:
Thanks @Remi-Gau!

PR #448, opened 2020-04-14
REL: 1.3.1-dev

franklin-feingold:
step 8 of the [release protocol](https://github.com/bids-standard/bids-specification/blob/master/Release_Protocol.md) resetting the latest version name


effigies:
This doesn't need to be a PR. Just push it to `master`.

```
git push upstream 678df30:master
```

sappelhoff:
> This doesn't need to be a PR. Just push it to master.

probably not possible from the GUI (?) I'll rebase and merge, which will have about the same effect.

PR #452, opened 2020-04-16
[FIX] Add reference to PDF on front page of specification

nicholst:
The new PDF rendering isn't presently referenced within the specification.  The place to reference the PDF version is clearly the 'front matter' page `src/index.md`, but I'm not entirely pleased with my addition, tacking on the sentence...

> The entire specification can also be [downloaded as PDF](https://doi.org/10.5281/zenodo.3686061).

... but I couldn't see a more elegant edit.

This addresses #135.


sappelhoff:
Thanks @nicholst

PR #470, opened 2020-05-13
[ENH] Allow fractional seconds in scans file datetimes

### LINKED ISSUE #469: Allow microseconds in scans file datetimes
This is related to https://github.com/nipy/heudiconv/issues/447, https://github.com/nipy/heudiconv/pull/451, and https://github.com/physiopy/phys2bids/pull/219. Currently, the `_scans.tsv` file only allows `acq_time` with resolution down to the second. For the purposes of increased accuracy and to make it easier to synchronize scan data with simultaneously-acquired data from other modalities (e.g., physio data), it would be helpful if scans files' `acq_time` column allowed (but didn't require) microseconds as well. So `acq_time` could be `2009-06-15T13:45:30` or `2009-06-15T13:45:30.500000`.

tsalo:
Closes #469. Explicitly allows fractional seconds (i.e., sub-second resolution) in `acq_time` column of Scans file. Fractional seconds are (1) optional, (2) must follow a period, and (3) have no restriction on precision.

### DIFF
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -268,8 +268,10 @@ Datetime should be expressed in the following format
 `2009-06-15T13:45:30[.000000]` (year, month, day, hour (24h), minute, second,
 and optionally fractional second; this is equivalent to the RFC3339 "date-time"
 format, time zone is always assumed as local time).
+No specific precision is required for fractional seconds, but the precision
+SHOULD be consistent across the dataset
 For anonymization purposes all dates within one subject should be shifted by a
-randomly chosen (but common across all runs etc.) number of days.
+randomly chosen (but consistent across all runs etc.) number of days.
 This way relative timing would be preserved, but chances of identifying a
 person based on the date and time of their scan would be decreased.
 Dates that are shifted for anonymization purposes should be set to a year 1925
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -264,8 +264,8 @@ imaging acquisition sequence (each run `.nii[.gz]` file) within one session.
 Each `.nii[.gz]` file should be described by at most one row. Relative paths to
 files should be used under a compulsory `filename` header. If acquisition time
 is included it should be under `acq_time` header. Datetime should be expressed
-in the following format `2009-06-15T13:45:30` (year, month, day, hour (24h),
-minute, second; this is equivalent to the RFC3339 "date-time" format, time zone
+in the following format `2009-06-15T13:45:30[.000000]` (year, month, day, hour [24h],
+minute, second, and optionally fractional second; this is equivalent to the RFC3339 "date-time" format, time zone
 is always assumed as local time). For anonymization purposes all dates within
 one subject should be shifted by a randomly chosen (but common across all runs
 etc.) number of days. This way relative timing would be preserved, but chances
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -264,15 +264,16 @@ imaging acquisition sequence (each run `.nii[.gz]` file) within one session.
 Each `.nii[.gz]` file should be described by at most one row. Relative paths to
 files should be used under a compulsory `filename` header. If acquisition time
 is included it should be under `acq_time` header. Datetime should be expressed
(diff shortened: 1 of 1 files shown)


tsalo:
@sappelhoff It look like bids-validator does validate the scans file, and it should raise an error (number 84, specifically) if microseconds are included. See [here](https://github.com/bids-standard/bids-validator/blob/ecacf0d30f9bc803393c1bb9833179d47a5d5451/bids-validator/validators/tsv/checkAcqTimeFormat.js#L4) and [here](https://github.com/bids-standard/bids-validator/blob/920d1957c1788c96f3df5b0be3eb7c2be594f8d3/bids-validator/utils/issues/list.js#L465-L470).

I can edit list.js easily enough, since the part that needs to be changed is essentially just a json. However, I don't know JavaScript, so I have no idea how to update checkAcqTimeFormat.js to allow optional sections in the datetime format that is being checked.

src/03-modality-agnostic-files.md
```diff
@@ -264,15 +264,16 @@ imaging acquisition sequence (each run `.nii[.gz]` file) within one session.
...
-in the following format `2009-06-15T13:45:30` (year, month, day, hour (24h),
-minute, second; this is equivalent to the RFC3339 "date-time" format, time zone
-is always assumed as local time). For anonymization purposes all dates within
-one subject should be shifted by a randomly chosen (but common across all runs
-etc.) number of days. This way relative timing would be preserved, but chances
-of identifying a person based on the date and time of their scan would be
-decreased. Dates that are shifted for anonymization purposes should be set to a
-year 1925 or earlier to clearly distinguish them from unmodified data. Shifting
-dates is RECOMMENDED, but not required.
+in the following format `2009-06-15T13:45:30[.000000]` (year, month, day,
```
nicholst, line 267: Can we here or somewhere state that the number of fractional digits is up to the user?

tsalo: What do you think of "No specific precision is required for fractional seconds, but the precision should be consistent across the dataset."

sappelhoff:
ok, thanks for checking @tsalo 

> However, I don't know JavaScript, so I have no idea how to update checkAcqTimeFormat.js to allow optional sections in the datetime format that is being checked.

In that case, opening an issue on bids-validator would be a good idea and hopefully we find somebody to help us. Worst case we'll have to dig in ourselves (we'll see).

tsalo:
I just opened https://github.com/bids-standard/bids-validator/issues/957. We'll see if anyone has the time and bandwidth to help.

src/03-modality-agnostic-files.md
```diff
@@ -264,15 +264,16 @@ imaging acquisition sequence (each run `.nii[.gz]` file) within one session.
...
+hour (24h), minute, second, and optionally fractional second; this is
+equivalent to the RFC3339 "date-time" format, time zone is always assumed as
+local time). For anonymization purposes all dates within one subject should be
+shifted by a randomly chosen (but common across all runs etc.) number of days.
+This way relative timing would be preserved, but chances of identifying a
+person based on the date and time of their scan would be decreased. Dates that
+are shifted for anonymization purposes should be set to a year 1925 or earlier
+to clearly distinguish them from unmodified data. Shifting dates is
+RECOMMENDED, but not required.
```
effigies, line 276: Can I be annoying and ask that you move each new sentence to a new line? This is a good example of a small change that forces the entire paragraph to be in the diff because we don't break at sentence boundaries. If we're going to have a big diff, why not make it one that helps us have smaller diffs in the future?

effigies:
Should note I agree with @yarikoptic. The spec doesn't have to wait on the validator. The validator is obliged to follow the spec.

src/03-modality-agnostic-files.md
tsalo: Should be good now.

src/03-modality-agnostic-files.md
sappelhoff: > No specific precision is required for fractional seconds, but the precision should be consistent across the dataset

:+1:  with a capital SHOULD

tsalo: Done.

sappelhoff:
cool, we have 4 approvals (although they were dismissed by the last commit), so I am merging now. thanks @tsalo

Issue #289, opened 2019-08-01
Entity Table (Appendix) improvements: Custom CSS


sappelhoff:
Make the first row and first column "sticky" using custom CSS code so that when browsing the relatively large table, a reader does not have to keep in mind which column or row corresponds to which entity / data modality

for reference: https://github.com/bids-standard/bids-specification/pull/272#issuecomment-510967427

PR #152, opened 2019-02-15
[ENH] Add Glossary of terms/abbreviations used in the specification

yarikoptic:
Just an idea which came up while fighting over the `suffix` in PyBIDS: https://github.com/bids-standard/pybids/issues/380

BTW -- does anyone know about a proper way to define glossary and refer to the terms in this markdown/mkdocs, like [e.g. `:term:` in sphinx](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/glossary.html)?

PR #475, opened 2020-05-17
[INFRA] Convert entity table to yaml

### LINKED ISSUE #290: Split up the entity table into smaller sections
The entity table in the appendix is currently one large table for all data modalities .

It might make sense to split the table into several sub-tables (e.g., MRI, MEG, EEG).

1. This would make each table more readable
1. Users are typically interested in one modality at a time, thus having modality-separated entity tables would decrease noise while reading
1. Having entities split up by modalities would make it easier to insert links from each entity to an in-text description of the usage of that entity in the context of the specific modality

Note: This issue can be worked on independently from #289 ... solving both issues would be a great improvement.

### LINKED ISSUE #343: Update entity table
The entity table does not include, for example, `run-` for anatomical images, which are [allowed](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data) in the original.

We should go through the entity table and verify that it's consistent with the data it's intended to summarize. I think it would also be useful to link back to the original, to make future verification easier.

### LINKED ISSUE #466: Convert entity table to machine-readable format
I'm not sure if it's feasible, but it would be nice if the entity table was stored as a json file, in order to make it both programmatically accessible and centralized. I know that there is an equivalent file in [bids-validator](https://github.com/bids-standard/bids-validator/blob/6a937a1ba1d08d3e14d3f0c984ab6ba8a7f7a9fd/bids-validator/bids_validator/rules/file_level_rules.json) and [pybids](https://github.com/bids-standard/pybids/blob/5fab1dee770babffa7aef668b24cb59a90ba37ef/bids/layout/config/bids.json), but if the filename construction rules were centralized under the actual specification, then it would be much easier to update the specification across the ecosystem without having to update a range of other packages as well.

tsalo:
Closes #466, closes #343, and closes #290. Can be used, once an entity-table rendering script is written, to deal with #290.

The yaml/json structure will need to be have sufficient information for tools like `bids-validator`, `heudiconv`, and `pybids` to be able to use them. I've somewhat based the files on the `bids-validator` rules files.

Changes proposed:
- Entities and datatypes have been described with yaml files, stored in the new `schema/` folder.
    - This structure reflects more a proof of concept than a final version for the schema.
- Datatypes are broadly split into standard datatypes and "auxiliary" datatypes (i.e., datatypes that are associated with multiple modalities, and which don't contain imaging data directly).
- @yarikoptic wrote code to generate the Entity Table from the schema files, which I have modified to some extent.
- The Entity Table differs in some respects:
    - Run is optional for anat data. See #343.
    - Rows are in alphabetical order, based on the `Entity` string.
    - Under DWI, `bvec` and `bval` have been dropped and `sbref` has been added, since `bvec` and `bval` are extensions, not suffixes. The addition of `sbref` is related to #343. 
    - Under beh, the suffix `beh` has been added. Related to #343. 
    - Suffixes may be listed in slightly different orders, because in the new version they're listed based on subgroups within datatypes.
    - Single-suffix datatypes include the suffix in parentheses.
    - Auxiliary datatypes have the suffix followed by associated datatypes in parentheses (like the current table), but the associated datatypes are space-separated instead of slash-separated.
    - Headshape and markers are labeled like standard suffixes, rather than auxiliary datatypes.

edit @sappelhoff 2020-06-07: add to do list:

# to do

- [x] make sure #343  is closed by this PR
- [x] think about whether #290 can be addressed with this PR as well
- [ ] will this PR make https://github.com/bids-standard/bids-specification/issues/289 irrelevant? If yes, close it.
- [x] Update CONTRIBUTING.md with information about the schema.
- [ ] Updates pending #556.


src/schema/top_level_files.yml
```diff
@@ -1,32 +1,41 @@
...
 dataset_description:
+  required: True
   extensions:
     - json
 genetic_info:
+  required: False
   extensions:
     - json
 participants:
+  required: True
```
yarikoptic, line 22: I have got "burnt" on this one recently... apparently at least how it is worded now in the spec -- it is not required! See https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#participants-file which says "Optional: Yes" although it might have meant to be for `phenotype/`?

src/schema/datatypes/anat.yml
```diff
@@ -0,0 +1,44 @@
+group1:
```
yarikoptic, line 1: it feels we would just want a list here instead of coming up with some `groupX` key.

src/schema/associated_data.yml
```diff
@@ -0,0 +1,4 @@
+- code/
```
yarikoptic, line 1: YAML group recommends to use `.yaml` in favor of `.yml` so let's use `.yaml` extension

src/schema/datatypes/anat.yml
```diff
@@ -0,0 +1,44 @@
...
+  datatypes:
+    - anat
```
yarikoptic, line 3: May be a folder name should be changed to not conflict with `datatypes` which are listed within files -- it should be reflective of the filenames in the folder which do not necessarily correspond to datatypes (which are listed within).

yarikoptic: Also "datatype" term is not defined nicely in BIDS. Related #152 ...
may be this field is actually the `modalities`? ;-)

yarikoptic:
re "inclusion" -- I guess terminology is not clear.  The recent commit is adding `required` field -- so it is about stating the requirement to include any given file in BIDS dataset.  I meant "inclusion" of one .yaml file into another.  I.e. we need to establish a hierarchy with a single YAML file as an entry point, so loading that one would load all of them.  I would just use https://github.com/tanbro/pyyaml-include which is inline with other ad-hoc solutions proposed on https://stackoverflow.com/questions/528281/how-can-i-include-a-yaml-file-inside-another . 

I could take care about it whenever I get to work on rendering the entity table.

src/schema/datatypes/anat.yaml
```diff
@@ -0,0 +1,45 @@
+groups:
```
tsalo, line 1: To continue from https://github.com/bids-standard/bids-specification/pull/475#discussion_r428050851

Does this work?

src/schema/datatypes/anat.yml
tsalo: Did pybids switch from modality to datatype for this? I was just planning to go with what pybids has if it's not explicit in the specification.

tsalo:
> re "inclusion" -- I guess terminology is not clear. The recent commit is adding required field -- so it is about stating the requirement to include any given file in BIDS dataset. I meant "inclusion" of one .yaml file into another. I.e. we need to establish a hierarchy with a single YAML file as an entry point, so loading that one would load all of them. I would just use https://github.com/tanbro/pyyaml-include which is inline with other ad-hoc solutions proposed on https://stackoverflow.com/questions/528281/how-can-i-include-a-yaml-file-inside-another .
>
> I could take care about it whenever I get to work on rendering the entity table.

That would be great, thank you. I'm just not that familiar with yaml.

src/schema/datatypes/anat.yml
tsalo: Continued in new comment thread tied to `anat.yaml` at https://github.com/bids-standard/bids-specification/pull/475#discussion_r428977646.

src/schema/datatypes/anat.yml
yarikoptic: TL;DR: I was wrong - `modalities` would be wrong and `datatypes` indeed correct:
```
src/05-derivatives/02-common-data-types.md-```Text
src/05-derivatives/02-common-data-types.md-<pipeline_name>/
src/05-derivatives/02-common-data-types.md-    sub-<participant_label>/
src/05-derivatives/02-common-data-types.md:        <datatype>/
src/05-derivatives/02-common-data-types.md-            <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.<ext>
```
I will add it now to #152

yarikoptic: doh -- just now mentioned it was from common-derivatives branch/PR, but I think it holds generally.

yarikoptic: I must stop for today for the sake of humanity and myself: "Data type" is defined within that glossary already ;-) Sorry for all the noise

src/schema/datatypes/events.yaml
```diff
@@ -0,0 +1,15 @@
+group1:
+  datatypes:
+    - meg
+    - eeg
+    - ieeg
+  suffices:
```
yarikoptic, line 6: shouldn't it be `suffixes`?

yarikoptic: I feel like I have not done enough harm today, so now I am on it -- I will fix this and also rename `datatypes/` into `modalities/` since that is what it is! ;)

yarikoptic: no - I will stop doing harm -- I think it is a bigger question here, since it is neither datatypes nor modalities ATM. as for why it is not `datatypes/` -- because datatype shouldn't include another datatype (or itself), which it does now. I have pushed `suffices` -> `suffixes' though

src/schema/datatypes/anat.yaml
yarikoptic: I don't think this top level `groups:` is needed unless you are planing to come up with another key in addition to `groups` at this level.

src/schema/datatypes/meg.yaml
```diff
@@ -0,0 +1,50 @@
+group1:
+  datatypes:
+    - meg
+  suffices:
+    - meg
+  extensions:
+    - json
+    - ctf/
+    - fif
+    - 4d/
```
yarikoptic, line 10: `<rant>` wow, I didn't know that (sub)folders are allowed here.   I truly think this was a step backward for meeg proposal to just allow all those formats -- there is no 'standardization' now, tools will just support some but not the other formats etc. `</rant>`

src/schema/datatypes/events.yaml
tsalo: 🤦 yes suffixes my bad. thanks for catching it!

src/schema/datatypes/photo.yaml
```diff
@@ -0,0 +1,13 @@
+group1:
+  datatypes:
+    - meg
+    - eeg
+    - ieeg
+  suffixes:
+    - photo
```
yarikoptic, line 7: Dear @tsalo -- **Thank you** for leading this effort!  It is really fun to look at BIDS in such as structured way!

So within this `datatypes/` we indeed have datatypes which gained their `folder/` within BIDS hierarchy (they also have themselves duplicated now in `datatypes`) such as `anat/`, `func/`, `fmap/`, etc. (+ `beh/` which isn't imaging), and then we have another group -- those which could have indeed been a modality of their own, but they are not "imaging data". They became auxiliary files which could go along with base imaging (and `beh/`) `datatypes`.  I think that 
- all the `photo`, `channels` etc should get their own "folder" in this schema, e.g. `auxiliary` or  `auxdatatypes`, and be moved there (instead of being under `datatypes`)
- `datatypes/` entries can get rid of the `datatypes` field in their .yaml files -- they should then have 1-to-1 correspondence from the filename to corresponding datatype.

On the top level (for inclusion) I see smth like
```yaml
datatypes:
  anat: !include datatypes/anat.yaml
  func: !include datatypes/func.yaml
  ...
auxdatatypes:
  channels: !include auxdatatypes/channels.yaml
  event: !include auxdatatypes/events.yaml
```
This would make it easier to tell one (which gets its own folder in BIDS) from another (just provides additional suffixed files).

Alternative could be - keep as is, and then add additional field  (in a yet to be created higher level "inclusion" file) to signal which datatype is `auxiliary`, but then it would be a bit less structured IMHO and most likely we would need to maintain that redundant self mentioning within `datatypes` for the base data types.

src/schema/datatypes/anat.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - FLAIR
+      - FLASH
+      - PD
+      - PDmap
+      - PDT2
+      - inplaneT1
+      - inplaneT2
+      - angio
+    extensions:
+      - nii.gz
```
yarikoptic, line 21: uff... the difficult life of standardization folks.  In the light of the discussion within #152 I wonder if we should make them all explicitly start with `.` solely for consistency with terminology.  I do appreciate the fact that we do not do that for entities or suffixes (prefix with `_`) but oh well, what can I say -- standards are standards and it is beneficial to be consistent. What do you think @tsalo?

vsoch:
@tsalo is there something I can help with here?

tsalo:
@vsoch Absolutely! At this point, I think any input would be helpful, to ensure that the yaml/json files properly reflect the specification. Once the yaml/json files are more finalized, though, we will need scripts to build the tables for the specification, and bids-validator and pybids will also need interfaces. I know that @yarikoptic can help with that, but I won't be much use at that point, so your help would be amazing.

@yarikoptic I didn't realize I hadn't touched this in two weeks. Sorry about that! I will try to start responding more over the next couple of days.

src/schema/datatypes/anat.yaml
tsalo: Good catch. All done now.

src/schema/datatypes/anat.yaml
tsalo: Would it be better to have it as a list of dictionaries _or_ as separate "documents" in the same file (see [here](https://gettaurus.org/docs/YAMLTutorial/#YAML-Multi-Documents))?

src/schema/datatypes/photo.yaml
tsalo: I split `auxdatatypes` from `datatypes` and dropped the `datatypes` key from the `datatype` yamls. I have to say, it looks much cleaner.

vsoch:
@tsalo I helped with the original development of BIDS long ago and far away, but haven't really looked at or used it in years, so I probably am not one to give feedback on that. Once that is done and you need to build tables, however, that might be something I can help with, as long as it's possible to scope out what exactly is needed (e.g., an example output table to produce).

yarikoptic:
woohoo, update:
- dropped using that library for inclusion of yaml.  We can maintain original simple hierarchy of .yaml files and it works wonderfully with a little ad-hoc loader!
- in c41a0dcee1d34788e0ebdda23c25fc6286f3fcbf I added generation of the entity table, and it works!  Thanks @tsalo for entering all the specs!  

Next I will compare to the original and report back

yarikoptic:
@tsalo - could you take it from here for the Entity Table?  I would have
- analyzed differences (I reordered lines in the original, committed in a branch, and then used `git diff -w` to see what really is different) - may be something in schema needs fixing
- decide on may be splitting those `auxdatatypes` into a separate table.  I also thought to annotate the datatypes with trailing `/` since they have their directories and the aux with leading `_` since that is how they appear throughout BIDS

tsalo:
I think so. Thank you for writing it and getting it working!

I also started on metadata terms. I don't want to duplicate effort in #423, but it seems like most of the work there is on _extracting_ the terms from the specification and _then_ converting them to something more machine-readable in another repository. Hopefully translating terms over to yaml here will just make it easier for them in the long run.

vsoch:
Question - does bids have something in Python to extract some kind of json / other manifest for an entire dataset of files? I’m working on a data container idea and I thought that bods would be a good contender.

tsalo:
I'm pretty sure `pybids` has the full file list and that it can be exported to a dataframe.

vsoch:
Awesome! Thanks @tsalo

yarikoptic:
@vsoch yes, pybids. We use it in datalad to do exactly that - get a json with metadata. See https://github.com/datalad/datalad-neuroimaging/blob/master/datalad_neuroimaging/extractors/bids.py . If needed, can point to example json files we have for many bids datasets on datasets.datalad.org

yarikoptic:
@tsalo I think it would be better to not overwhelm the PR and reactor it into two. First to make the baby step for entity table. For that have only minimal sufficient amount of metadata defined. Then on top of that branch/pr do another one with terms. That would make it more feasible to get review done and accept the baby one first, while continuing work on a bigger piece.

yarikoptic:
As for the terms, we should merge effort with @dbkeator

tsalo:
@yarikoptic Okay, I can revert those changes and move them over to another branch.

yarikoptic:
Since it is a draft pr you/we could even go wild and just rewrite history to make it into a minimal set of nicely documented relevant commits, with the next ones need to be in a separate branch ;-) I do not care much about my authorship on those which I did if you happen to rewrite so my name disappears ;-)

tsalo:
The main thing we needed to do was to merge groups of suffixes within datatypes with the same entities. There are a number of remaining differences that I can handle, but I think how I fix them will depend on more substantial decisions. Still, here's a list:
- Rows are in alphabetical order.
- Suffixes may be listed in slightly different orders, because in the new version they're listed based on subgroups within datatypes.
- Single-suffix datatypes include the suffix in parentheses.
- Auxiliary datatypes have the suffix followed by associated datatypes in parentheses (like the current table), but the associated datatypes are space-separated instead of slash-separated.
- Headshape and markers are labeled like standard suffixes, rather than auxiliary datatypes.

@sappelhoff, per #290, the current idea is to split based on modality. Is that still preferred? I ask because the auxiliary datatypes are shared across modalities (`meg`, `eeg`, and `ieeg` mostly), so I'm not sure how best to do that.

Secondly, I don't love how the pattern switches from `datatype (suffixes)` to `suffix (datatypes)`. I think it could be confusing to users. At minimum, if we do split the table, then we should either standardize that or denote it in the text.

sappelhoff:
> @sappelhoff, per #290, the current idea is to split based on modality. Is that still preferred?

I think we can use a discussion of different alternatives and get views on this from different people. So there is currently no strict preference (at least not from my side).

tsalo:
Makes sense. Here are some options I was thinking of:

- Change auxiliary datatypes to standard, datatype-specific suffixes. This would result in duplication within the schema and in the entity table, since rows like "photo<br>(meg/eeg/ieeg)" would end up being stored as "meg<br>(photo)", "eeg<br>(photo)", and "ieeg<br>(photo)". I think this would fit well with splitting up the table by modality, since most modalities would have 2-7 rows.
- Split the table into "datatypes" and "auxiliary datatypes". Things like "channels", "headshape", "markers", "photos", "events" and "electrodes" would go in the second table.
- Split by datatype. This might result in too many tables though.

I think I prefer the first one, personally. It would impact the schema, but make things more consistent.

yarikoptic:
"duplication" is the grand father of a "bug", so I would try to stay away from any approach needing duplication.

tsalo:
@yarikoptic, that's definitely true, but the more I think about it, the more I see that the auxiliary datatypes are really suffixes that are shared across modalities. An alternative would be to restructure the schema so that there's one file per _suffix_, but that would massively expand the number of files.

dbkeator:
Hi,
This is really cool.  Just an FYI, we're working on storing all terms/concepts used in BIDS specifications (see [https://github.com/NIDM-Terms/terms/tree/master/terms/BIDS_Terms](https://github.com/NIDM-Terms/terms/tree/master/terms/BIDS_Terms)) and BEPs with relations to external terms (e.g. DICOM terms) and relations to BIDS internal concepts.  These relations sort of mirror what I see in the schema documents you've created (e.g. 'T1w' is a child of (we're using the isPartOf relationship) 'anat',  'T2w' isPartOf 'anat', etc.).  

In looking at the BIDS specs entity tables and your schema we find we're missing some of the BIDS reserved words like 'anat' and in other places in the BIDS spec we found a slight mismatch where we mislabeled 'T1w' as 'T1Weighted' (I think because T1Weighted was in the BIDS specification document) so we're working to fix up our terminology using your schema docs as references to be sure we have representations for each entity.

Would be great to work with you to keep things in sync.  For example, our BrainHack project was to create a tool which provides BIDS BEP developers a means of adding new BIDS specification terms (and checking that they are not re-purposing an existing term and giving it a different definition).  Once a new term is added we annotate them with additional ontologic properties linking them to other terms within the current BIDS ecosystem (here we might eventually use your schema documents for new BEPS to automagically set these relationships or vice-versa, using our relationships we could send PRs to update the schema documents)....

cheers!
Dave

sappelhoff:
just a short attention note (not sure if everybody was aware): The order of entities matters and should be preserved with the new representation that is being cooked up in this nice PR :-)

for more background, see: https://github.com/bids-standard/bids-validator/issues/767

satra:
noting this comment on the bids-validator thread here as well:

> the order does not seem to be consistently followed throughout the specification (for example: https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#preprocessed-coregistered-andor-resampled-volumes)

tsalo:
> The order of entities matters and should be preserved with the new representation that is being cooked up in this nice PR :-)

Absolutely! The order is specified by the order of the keys in [entities.yaml](https://github.com/bids-standard/bids-specification/blob/8d0b612bcce2e3a1a31d481e870d5b7a289a1a8b/src/schema/entities.yaml). Any tools which use the specification _will_ need to use ordered dictionaries though.

soichih:
Is there a way to include "required sidecar key" information? For example, "TaskName" fields must be set for func (and the "task" entity should match the value for "TaskName"..). Or is that out-of-scope for this information?

tsalo:
@soichih Converting the entire specification to a machine-parseable schema is definitely a long-term goal (at least for me), but we decided that metadata fields were out of scope for this particular PR (see https://github.com/bids-standard/bids-specification/pull/475#issuecomment-643651123).

I would like to open up a discussion about converting the whole specification (including metadata fields), but wasn't sure whether to just comment on #423 (since BIDS_terms has a lot of overlap with what we want to do) or to open a new issue. @dbkeator do you have a preference, since #423 is your issue?

yarikoptic:
yeap, and other PRs based on this PR could come to life introducing those additional aspects... to minimize possible conflicts though, would be nice to finalize and merge this PR first.

tsalo:
@yarikoptic That's a fair point. I think this PR is ready for review at this point, so I'll mark it as such.

To summarize the current state of the PR:
- Entities and datatypes have been described with yaml files, stored in the new `schema/` folder.
    - This structure reflects more a proof of concept than a final version for the schema.
- Datatypes are broadly split into standard datatypes and "auxiliary" datatypes (i.e., datatypes that are associated with multiple modalities, and which don't contain imaging data directly).
- @yarikoptic wrote code to generate the Entity Table from the schema files, which I have modified to some extent.
- The Entity Table differs in some respects:
    - Run is optional for anat data. I think this was an oversight in the previous version.
    - Rows are in alphabetical order.
    - Suffixes may be listed in slightly different orders, because in the new version they're listed based on subgroups within datatypes.
    - Single-suffix datatypes include the suffix in parentheses.
    - Auxiliary datatypes have the suffix followed by associated datatypes in parentheses (like the current table), but the associated datatypes are space-separated instead of slash-separated.
    - Headshape and markers are labeled like standard suffixes, rather than auxiliary datatypes.

sappelhoff:
Some weeks back I sneakily added a to do list to your PR description @tsalo, I repeat it here:

- [ ] make sure #343  is closed by this PR
- [ ] think about whether #290 can be addressed with this PR as well
- [ ] will this PR make https://github.com/bids-standard/bids-specification/issues/289 irrelevant? If yes, close it.

can you please address these three points? And with "address", I also include the option to say "this is out of scope, let's do it another time".

tsalo:
@sappelhoff Sorry about that! Totally missed the checklist.

In my most recent commit, I split up MRI, EEG/iEEG/MEG (just called Encephalography for now), and pure behavioral data, so I'm going to check off the item for #290 and update the first comment to close it. 

I also updated the entity table with certain fixes, as outlined in the first comment, so I'll check off the #343 item. I think I've caught all of the missing items, but I can't be sure. The other idea in that issue, to link back to the original entity definitions and datatypes in the Entity Table, is something I think should be put off until we start converting the whole specification to YAML.

Finally, for the last item, I haven't done any CSS or anything, and I think it's a little outside the scope right now. I _did_ reduce the entities in the entity tables to only those that are used in the modality each table covers, so there are fewer columns in each and that makes it easier to read them. So it's something at least.

src/99-appendices/04-entity-table.md
```diff
@@ -1,32 +1,48 @@
...
+| events<br>(meg eeg ieeg)   | REQUIRED      | OPTIONAL      | REQUIRED       |               | OPTIONAL      |                         |                 |                 |
+| photo<br>(meg eeg ieeg)    | REQUIRED      | OPTIONAL      |                | OPTIONAL      |               |                         |                 |                 |
+
+## Behavioral Data
+
+| Entity                          | Subject       | Session       | Task           |
+|---------------------------------|---------------|---------------|----------------|
+| Format                          | `sub-<label>` | `ses-<label>` | `task-<label>` |
+| beh<br>(stim physio events beh) | REQUIRED      | OPTIONAL      | REQUIRED       |
+
```
yarikoptic: You are so beautiful to me... ;-)

sappelhoff:
@nicholst I haven't tagged you in a while but we can use every reviewer we have for this PR --> the more eyes the better (also for checking the linked issues). so if you have time, a review would be welcome!

src/schema/entities.yaml
```diff
@@ -0,0 +1,127 @@
...
+dir:
+  name: Phase-Encoding Direction
+  description: "Similarly the OPTIONAL dir-<label> key/value can be used to
+                distinguish different phase-encoding directions."
+  format: label
+run:
+  name: Run
+  description: "If several scans of the same modality are acquired they MUST be
+                indexed with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc.
+                (only integers are allowed as run labels).
```
nicholst, line 72: Do we require that run indices start at 1?  Should we say that explicitly?

```suggestion
                (only natural numbers are allowed as run labels).
```

nicholst:
Wow... had no idea this was underway!  I made on little comment, but I'm not able to provide much review of the code `bids_scheme.py`.

I think the high level issue I have is: How does this work!?  On two levels:

1. When a trivial change is made to a schema file, how does the markdown entities table get rebuilt?
1. As BIDS is extended and maintained, how does the random contributor that knows nothing about this infrastructure know what to do... i.e. I assume if I propose a new modality as part of a BEP I'll need to change a schema somewhere, but how will I know to do that?  Where do we document the instructions for keeping these schema up to date and consistent with the body text of the standard?

tsalo:
> 1. When a trivial change is made to a schema file, how does the markdown entities table get rebuilt?

Currently, someone has to run the `bids_schema.py` script that @yarikoptic wrote and I edited in order to generate an updated entity table page. Hopefully, this could become automated at some point in the future, like how the PDF version is built in the continuous integration.

> 2. As BIDS is extended and maintained, how does the random contributor that knows nothing about this infrastructure know what to do... i.e. I assume if I propose a new modality as part of a BEP I'll need to change a schema somewhere, but how will I know to do that? Where do we document the instructions for keeping these schema up to date and consistent with the body text of the standard?

Documentation is definitely going to be an issue. The schema will have to be relatively stable, with extensive information about the different elements and their associated fields. I think that a page in Appendix would be good for this, plus YAML has the benefit of supporting comments.

As for the typical contributor, I guess I would assume that any failures in the documentation would be resolved by having knowledgeable reviewers. I wouldn't expect most folks to be familiar with YAML, but it's a fairly forgiving standard compared to JSON, and as long as the updates are rendered automatically in the CI and the reviewers know how the schema is set up, I think it should be fine. 

Regarding keeping the schema up-to-date with the text of the standard, my long-term hope is that the whole specification will end up being schema-fied, so that all (or almost all) of the text actually comes directly from the schema. I don't know how others feel about that idea, given that I haven't opened an issue for it yet, but doing so would make everything a lot cleaner, in my opinion.

src/schema/entities.yaml
tsalo: Since this is a change to the specification, should it be a separate issue? On the other hand, it's such a small change I don't know if it needs to be separated out.

nicholst:
OK. Perhaps the best place to document is maybe a new section [CONTRIBUTING.md](https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md)?  But mostly I guess it'll be up to the community and maintainers to ensure that there is no mismatch between the YAML and free text.

src/schema/entities.yaml
nicholst: Fair point. I'll lodge a comment elsewhere.

effigies: Unfortunately, natural numbers doesn't have a universally agreed meaning (some include 0, some don't) so this would require us to define it.

It might be better to say "only numeric values", as this is what is actually validated. It's equivalent to non-negative integers with optional zero-padding, but that seems like a cumbersome phrasing.

src/schema/top_level_files.yaml
```diff
@@ -0,0 +1,41 @@
+README:
+  required: True
+  extensions:
+    - .md
+    - .rst
```
effigies, line 5: This seems new. Is this encoding current validator practice, or are we proposing that Markdown and RST extensions should be accepted?

src/schema/top_level_files.yaml
```diff
@@ -0,0 +1,41 @@
...
+  extensions:
+    - .json
+phase2:
+  required: False
+  extensions:
+    - .json
+fieldmap:
+  required: False
+  extensions:
+    - .json
```
effigies, line 41: Unclear why fieldmap JSONs are being explicitly listed at the top level. Any sidecar that applies to all participants can be specified at this level.

src/schema/entities.yaml
```diff
@@ -0,0 +1,127 @@
...
+                and then be re-positioned on the scanner bed, the set of MRI
+                acquisitions will still be considered as a session and match
+                sessions acquired in other subjects.
+                Similarly, in situations where different data types are obtained
+                over several visits (for example fMRI on one day followed by DWI
+                the day after) those can be grouped in one session.
+                Defining multiple sessions is appropriate when several identical
+                or similar data acquisitions are planned and performed on all
+                -or most- subjects, often in the case of some intervention between
+                sessions (e.g., training)."
```
effigies, line 22: I believe YAML supports multi-line strings a bit more naturally: For your consideration:

```suggestion
  description: |
    A logical grouping of neuroimaging and behavioral data consistent across subjects.
    Session can (but doesn't have to) be synonymous to a visit in a longitudinal study.
    In general, subjects will stay in the scanner during one session.
    However, for example, if a subject has to leave the scanner room and then be
    re-positioned on the scanner bed, the set of MRI acquisitions will still be considered
    as a session and match sessions acquired in other subjects.
    Similarly, in situations where different data types are obtained over several visits
    (for example fMRI on one day followed by DWI the day after) those can be grouped in one
    session.
    Defining multiple sessions is appropriate when several identical or similar data
    acquisitions are planned and performed on all -or most- subjects, often in the case of
    some intervention between sessions (e.g., training)."
```

src/schema/entities.yaml
nicholst: Good point.  Please see #535.

src/schema/top_level_files.yaml
tsalo: This was something I adapted from the [top_level_rules in bids-validator](https://github.com/bids-standard/bids-validator/blob/029e37e2d77702a9c130a365b5480c3a41461297/bids-validator/bids_validator/rules/top_level_rules.json#L13-L16). I wasn't sure why they were included there but decided to err on the side of caution.

I can remove them.

src/schema/datatypes/beh.yaml
```diff
@@ -0,0 +1,20 @@
+- suffixes:
+    - stim
+    - physio
+  extensions:
+    - .tsv.gz
+    - .json
+  entities:
+    sub: required
+    ses: optional
+    task: required
```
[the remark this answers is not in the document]
tsalo: Currently `run` is not included for behavioral data, either in [the text](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/07-behavioral-experiments.html) or in the entity table. 

Is this something I should correct here or does it require a separate issue?

src/schema/datatypes/meg.yaml
```diff
@@ -0,0 +1,42 @@
+- suffixes:
+    - meg
+  extensions:
+    - 4d/
```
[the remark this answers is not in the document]
tsalo: Now that I look at it, it looks like the [4d format](https://bids-specification.readthedocs.io/en/stable/99-appendices/06-meg-file-formats.html#bti4d-neuroimaging) just has a folder with no suffix. Here's what it looks like:

```
sub-control01/
    ses-001/
        sub-control01_ses-001_scans.tsv
        meg/
            sub-control01_ses-001_coordsystem.json
            sub-control01_ses-001_headshape.pos
            sub-control01_ses-001_task-rest_run-01_meg/
                config
                hs_file
                e,rfhp1.0Hz.COH
                c,rfDC
            sub-control01_ses-001_task-rest_run-01_meg.json
            sub-control01_ses-001_task-rest_run-01_channels.tsv
```

I guess I should just put a `/` there, instead of `4d/` or `.4d/`?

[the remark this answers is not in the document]
sappelhoff: yeah, this format is the biggest pain ever ... we should ask some MEG people what to do (in particular, Robert O. once he's back from vacation)

Can we have `/` with a comment that this is for BTi/4D data?

src/schema/top_level_files.yaml
tsalo: I can't find a basis for this in the specification, pybids, or the validator, so I'm not sure what I got it from. I will remove these.

src/schema/top_level_files.yaml
tsalo: Removed in bfac102.

src/schema/top_level_files.yaml
tsalo: Removed in 6ce57e1.

src/schema/datatypes/meg.yaml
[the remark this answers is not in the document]
tsalo: Done in 4e291e2, but I'll hold off on resolving this in case further discussion is needed.

src/schema/datatypes/beh.yaml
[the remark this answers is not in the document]
yarikoptic: Ideally I would have made a separate issue (or PR on top of this one) so to not conflate this "refactoring" PR with introducing changes to BIDS itself. but if other bids people would bless it as is - good with me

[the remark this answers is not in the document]
effigies: Yeah, if `run` isn't in the text, let's make it a separate PR.

src/schema/entities.yaml
```diff
@@ -0,0 +1,124 @@
...
+    In general, subjects will stay in the scanner during one session.
+    However, for example, if a subject has to leave the scanner room and then be
+    re-positioned on the scanner bed, the set of MRI acquisitions will still be considered
+    as a session and match sessions acquired in other subjects.
+    Similarly, in situations where different data types are obtained over several visits
+    (for example fMRI on one day followed by DWI the day after) those can be grouped in one
+    session.
+    Defining multiple sessions is appropriate when several identical or similar data
+    acquisitions are planned and performed on all -or most- subjects, often in the case of
+    some intervention between sessions (e.g., training)."
```
effigies, line 19:
```suggestion
    some intervention between sessions (e.g., training).
```

src/schema/datatypes/beh.yaml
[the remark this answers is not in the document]
tsalo: I've opened #539, in case anyone wants to open a PR for it.

tsalo:
I've opened #540 for the overall specification conversion. It might be best considered as a project with specific issues and PRs targeting parts of it, but that can be discussed there.

yarikoptic:
> Overall LGTM. Some comments. I didn't review the Python script.

I am thinking we might want to move it e.g. to https://github.com/bids-standard/bids-schema and just reused here as a dependency (thus avoiding ad-hoc statement on dependencies etc)... While at that we could review it more specifically.  My worry though is that we do not have **any** (ad-hoc or not) statement on it requiring numpy and pandas.  Should we may be add pandas to `requirements.txt`?

soichih:
A *feature* request. :) It would be super nice if we could add "descriptions" for each entites

```
  entities:
    sub: required
        desc: Subject
    ses: optional
        desc: Session
    task: required
        desc: Task Name
```

etc..

tsalo:
@yarikoptic So the schema itself would be here, and scripts to build the specification in markdown and pdf forms would be in `bids-schema`? As long as the code is called in CI to build the readable version of the specification, I'm cool with it, but the names would be a little confusing IMHO.

@soichih We actually do have descriptions for entities, added fairly recently. They're in the `entities.yaml` file. Ultimately we should be able to link back from the entity table and the in-text descriptions of associated entities to the entity definitions, but I think that'll end up being part of #540.

yarikoptic:
filed https://github.com/bids-standard/bids-specification/issues/543 where we could continue discussion about (possible) code move

tsalo:
> OK. Perhaps the best place to document is maybe a new section [CONTRIBUTING.md](https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md)? But mostly I guess it'll be up to the community and maintainers to ensure that there is no mismatch between the YAML and free text.

@nicholst I've drafted some instructions in CONTRIBUTING.md, along with a brief description of the schema as a whole and each of the files within it. It's just a first pass, but it's something.

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
 
+## Updating the schema
+
+Portions of the BIDS specification are defined using YAML files, in order to make the specification machine-readable.
+Currently, the only portion of the specification that relies on this schema is the Entity Table, but any changes to the specification should be mirrored in the schema.
+
+### The format of the schema
+
+The schema reflects the files and objects in the specification, as well as associations between these objects.
+Here is a list of the files and subfolders of the schema, with a description for each:
```
nicholst, line 236: What is the order here?  At first I thought it was alphabetical, but it is not, and the first item listed is (as far as I understand) one of the least important YAML files.

Can I suggest reordering so that they are in some order that reflects how you might have to sift through them if you are a BIDS contributor making a change.

Then maybe change this:

```suggestion
Here is a list of the files and subfolders of the schema, roughly in order of importance:
```

(No need to say that you're going to give a description... that's clear).

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
+
+- `associated_data.yaml`:
+    Folders that contain files falling outside the jurisdiction of the specification, but which are commonly contained within the same folder as a BIDS dataset.
+    The schema specifies which folders are accepted and whether they are required or optional.
```
nicholst, line 240: I might just be being dense here, but are are folders "outside the jurisdiction" of the spec but which still can be considered "required"?  Isn't that a contradiction?

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
+- `top_level_files.yaml`:
+    Modality-agnostic files stored at the top level of a BIDS dataset.
+    The schema specifies whether these files are required or optional, as well as acceptable extensions for each.
+- `modalities.yaml`:
+    Modalities supported by the specification, along with a list of associated datatypes.
```
nicholst, line 245: AFAIK, "datatype" is not a thing in the spec. (e.g. I just searched the spec and found no reference to "datatype".  Can you give an explanation here or, at the top of the list, what is meant by "datatype"?

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
+- `entities.yaml`:
+    A list of entities (key/value pairs in folder and filenames) with associated descriptions and formatting rules.
+    The order of the entities in the file determines the order in which entities must appear in filenames.
+- `datatypes/*.yaml`:
```
nicholst, line 249: Oh... "datatypes".  This seems important.

Can this be the first item in the list?  And, also, from this description, I still don't see what a "datatype" is? I can only guess that you *don't* mean binary data representation (char, float, double), nor text encoding scheme, nor PET vs MRI... maybe `nii.gz` vs `csv`?  

Anyway, you see you really need some example to clarify this.

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
+    Datatypes supported by the specification.
+    Each datatype may support many suffixes.
+    These suffixes are divided into groups based on what extensions and entities are allowed for each.
+- `auxdatatypes/*.yaml`:
```
nicholst, line 253: Auxiliary data types also seems like it needs some examples... can't quite guess what this is supposed to be.

CONTRIBUTING.md
```diff
@@ -224,6 +225,54 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
+
+The schema formalizes the rules described in the specification text, so you must ensure that any changes which impact the rules of the specification (including, but not limited to, adding new entities, suffixes, datatypes, modalities, etc.) are reflected in the schema as well.
+
+#### 2. Generate an updated entity table before pushing your changes
+
+Run the Python script `tools/bids_schema.py`:
+
+```bash
+cd tools/
+python bids_schema.py entity ../src/schema/ ../src/99-appendices/04-entity-table.md
```
nicholst, line 269: FYI, just can confirm that I tried running this command on my Mac... had to get the right version of Python and install a few packages, but then it ran no problem.  The created MD table looks fine.

CONTRIBUTING.md
tsalo: Data types fall under the common principles. They correspond to the folders (`anat`, `func`, `meg`, etc.). I can split the word in the text to make it clearer, and maybe link back to the relevant part of common principles?

CONTRIBUTING.md
yarikoptic: FWIW, I see such need for manual run if the tool as temporary. Filed https://github.com/bids-standard/bids-specification/issues/544 to automate, and will look into it unless someone beats me to it

CONTRIBUTING.md
nicholst: Ah!  "data type" two words!  Yes, of course.

CONTRIBUTING.md
tsalo: That's a good point. I included a `required` field for the items in this file mostly to match my convention in other files, but we can assume anything in the file is not required. Rather than abandon that right now, I'll try to make the text a little less confusing. If that doesn't resolve it, I can drop the `required` field in the file.

CONTRIBUTING.md
tsalo: This, and your other comments, should be addressed now. Thanks!

nicholst: Super!  Looks much clearer now!

.gitignore
```diff
@@ -1,4 +1,6 @@
 site/
 .DS_Store
 src/.DS_Store
-src/04-modality-specific-files/.DS_Store
\ No newline at end of file
+src/04-modality-specific-files/.DS_Store
```
sappelhoff, line 4: could you change the several `.DS_Store` lines into a single line that covers `.DS_Store` everywhere, please?

tsalo: Should be good now.

tsalo:
@sappelhoff I've added a test in Travis to just check that the schema loads. It's very basic, but I think it's a reasonable smoke test. WDYT?

sappelhoff:
> I've added a test in Travis to just check that the schema loads. It's very basic, but I think it's a reasonable smoke test. WDYT?

I actually thought more about something like https://github.com/adrienverge/yamllint

because with the current test, we are depending on our own scripts, which we may or may not want to modify/migrate in the future :thinking:

tsalo:
Ah, I get it. `yamllint` looks good. Running it actually caught a bunch of small issues in the schema files, so that's good. I switched out the test too, and it's passing!

yarikoptic:
overall, are waiting on https://github.com/bids-standard/bids-specification/pull/556 first?  I am just afraid that keeping delaying it would cause more and more tune ups needed to be done to get it again in the mergeable state as more "manual" changes to entity table get introduced by outstanding PRs.

tsalo:
I'm happy to get this merged and then update #556 instead of the opposite, but it just seemed like #556 would probably be merged before this PR. I don't really know what the timeline is on this one.

effigies:
Merging this into #556 or vice versa shouldn't be too hard, so I would say we should keep this one moving. Are there any hold-ups I can help with? Does it need a review, or are there outstanding issues to discuss?

tsalo:
I think that all of the issues have been resolved, so I think a review would be helpful.

effigies:
First thing on Monday, if I don't end up working this weekend. :-)

CONTRIBUTING.md
```diff
@@ -224,6 +225,57 @@ GitHub has a [nice introduction](https://help.github.com/articles/github-flow/)
...
```
effigies, line 272: `chdir`ing and passing `../`paths seems odd.

```suggestion
python tools/bids_schema.py entity src/schema/ src/99-appendices/04-entity-table.md
```

src/schema/datatypes/func.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - .json
+  entities:
+    sub: required
+    ses: optional
+    task: required
+    acq: optional
+    rec: optional
+    run: optional
+    recording: optional
+    proc: optional
```
effigies, line 53: Can't find any support for this in the spec.

```suggestion
```

src/schema/datatypes/beh.yaml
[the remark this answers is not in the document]
effigies: #556 merged. Let's update `beh/`.

tsalo:
Thanks @effigies! I've merged your changes, updated from master, and added in the associated changes from #556.

src/schema/entities.yaml
```diff
@@ -0,0 +1,135 @@
...
+  format: label
+  description: |
+    Each task has a unique label that MUST only consist of letters and/or
+    numbers (other characters, including spaces and underscores, are not
+    allowed).
+    Those labels MUST be consistent across subjects and sessions.
+acq:
+  name: Acquisition
+  description: |
+    The OPTIONAL acq-<label> key/value pair corresponds to a custom label the
```
yarikoptic, line 36: if ``` `TXT` ``` is allowed in yaml, we better make those `acq-<label>` into ``` `acq-<label>` ``` right away (as well as sample filenames which could have some characters which markdown could interpret for formatting) to simplify any possible future rendering...  Since we do not render them currently though -- I think it would be ok to proceed as is (unless some other really needed changes are suggested) and just defray it to a separate PR

src/schema/entities.yaml
```diff
@@ -0,0 +1,135 @@
...
+    full brain low resolution and and one restricted field of view but high
+    resolution.
+    In such case two files could have the following names:
+    sub-01_acq-highres_T1w.nii.gz and sub-01_acq-lowres_T1w.nii.gz, however the
+    user is free to choose any other label than highres and lowres as long as
+    they are consistent across subjects and sessions.
+    In case different sequences are used to record the same modality (e.g. RARE
+    and FLASH for T1w) this field can also be used to make that distinction.
+    At what level of detail to make the distinction (e.g. just between RARE and
+    FLASH, or between RARE, FLASH, and FLASHsubsampled) remains at the
```
yarikoptic, line 49: note: we do have FLASH but no RARE or FLASHsubsampled defined in supported modalities... Filed separate https://github.com/bids-standard/bids-specification/issues/569 .

src/schema/entities.yaml
```diff
@@ -0,0 +1,135 @@
...
+    discretion of the researcher.
+  format: label
+ce:
+  name: Contrast Enhancing Agent
+  description: |
+    Similarly the OPTIONAL ce-<label> key/value can be used to distinguish
```
yarikoptic, line 55: optional or not is a property per specific modality etc.

"Similarly ..." is the paragraph opening .

`ce-<label>` is duplicate with the record itself.

"can be used to distinguish" is also somewhat "to make it a sentence".

Here and in others I would have just kept it to the point, and start from the 2nd sentence: "The label is the name of the contrast agent.".

src/schema/entities.yaml
```diff
@@ -0,0 +1,135 @@
...
+    If several scans of the same modality are acquired they MUST be indexed
+    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only integers
+    are allowed as run labels).
+    When there is only one scan of a given type the run key MAY be omitted.
+    Please note that diffusion imaging data is stored elsewhere (see below).
+  format: index
+mod:
+  name: Corresponding Modality
+  description: |
+    In such cases the OPTIONAL `mod-<label>` key/value pair corresponds to
```
yarikoptic, line 86: in what "Such cases"?

yarikoptic:
Thank you @tsalo !  It looks great!
While approving I have clicked too fast though -- I want to note that ideally descriptions in `entities.yaml` should be adjusted.  But that could be done even after this PR since they aren't used ATM anywhere yet.

src/schema/entities.yaml
tsalo: Do you mind if we push updating entity definitions until a PR for #567? I'm planning to make a version of #568 that is compatible with this PR's changes.

yarikoptic: I would not mind at all!  Just wanted to leave  a note ;-)

src/schema/entities.yaml
yarikoptic: nevermind it -- I somewhat missed the point that it is for `acq-`. I will mark this resolved

tsalo:
Now that there are two approvals, should we wait five days before merging (pending any requested changes, of course)?

effigies:
We don't really count cleanup changes as restarting the clock, usually. I would say this can be merged at @sappelhoff's convenience.

yarikoptic:
One quick click for @sappelhoff , one giant leap for BIDS!

src/schema/entities.yaml
```diff
@@ -0,0 +1,135 @@
...
+  description: |
+    The optional space label (`*[_space-<label>]_electrodes.tsv`) can be used
+    to indicate the way in which electrode positions are interpreted.
+    The space label needs to be taken from the list in Appendix VIII.
+  format: label
+split:
+  name: Split
+  description: |
+    In the case of long data recordings that exceed a file size of 2Gb, the
+    .fif files are conventionally split into multiple parts.
```
sappelhoff, line 126: for the future, the `split` entity could be used by any file format IMHO

src/schema/top_level_files.yaml
```diff
@@ -0,0 +1,22 @@
+---
+README:
+  required: true
+  extensions:
+    - None
+CHANGES:
+  required: true
+  extensions:
+    - None
```
sappelhoff, line 9:
```suggestion
    - None
LICENSE:
  required: false
  extensions:
    - None
```

PR #477, opened 2020-05-18
[INFRA] linkchecker - ignore github pull and tree URLs

### LINKED ISSUE #474: Link Checker broken due to excessive requests to GitHub
GitHub seems to be rate-limiting requests, and we're now getting a lot of results like:

```
URL        `https://github.com/bids-standard/bids-specification/pull/13'
Name       `#13'
Parent URL file:///root/build/site/CHANGES.html, line 1067, col 41
Real URL   https://github.com/bids-standard/bids-specification/pull/13
Check time 0.397 seconds
Size       1KB
Result     Error: 429 too many requests
```

See, for example, https://app.circleci.com/pipelines/github/bids-standard/bids-specification/1076/workflows/7acda833-9ecf-4c1a-a6f7-51f45641fbc1/jobs/2434.

We're running it twice, though I don't understand the explanatory comment:

https://github.com/bids-standard/bids-specification/blob/8a174714ed8751daf1b7b26628f441481042836b/.circleci/config.yml#L42-L45

@yarikoptic Any thoughts? Maybe we can adjust the commands to ensure there's no duplication? If linkchecker could output a list of targets that could be combined and deduplicated, that might be a useful approach.

yarikoptic:
It causes github to spit out "429 too many requests" HTTP error (Closes #474).
With this change we should skip ATM about 200 github URLs and be left
with only about 30. Hopefully that would suffice

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ jobs:
               --ignore-url 'file:///.*' \
               --ignore-url https://fonts.gstatic.com \
               --ignore-url "https://github.com/bids-standard/bids-specification/(pull|tree)/.*" \
+              --ignore-url "https://github.com/[^/]*" \
               ~/build/site/*html ~/build/site/*/*.html
             else
             echo "Release PR - do nothing"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
             linkchecker -t 1 --check-extern \
               --ignore-url 'file:///.*' \
               --ignore-url https://fonts.gstatic.com \
-              --ignore-url https://github.com/bids-standard/bids-specification/(pull|tree)/.* \
+              --ignore-url "https://github.com/bids-standard/bids-specification/(pull|tree)/.*" \
               ~/build/site/*html ~/build/site/*/*.html
             else
             echo "Release PR - do nothing"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,11 @@ jobs:
             linkchecker -t 1 ~/build/site/
             # check external separately by pointing to all *html so no
             # failures for local file:/// -- yoh found no better way,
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -42,7 +42,11 @@ jobs:
             linkchecker -t 1 ~/build/site/
             # check external separately by pointing to all *html so no
             # failures for local file:/// -- yoh found no better way,
-            linkchecker -t 1 --check-extern --ignore-url 'file:///.*' --ignore-url https://fonts.gstatic.com ~/build/site/*html ~/build/site/*/*.html
+            linkchecker -t 1 --check-extern \
+              --ignore-url 'file:///.*' \
+              --ignore-url https://fonts.gstatic.com \
+              --ignore-url https://github.com/bids-standard/bids-specification/(pull|tree)/.* \
```
effigies, line 48: Looks like bash is trying to interpret this.

```suggestion
              --ignore-url "https://github.com/bids-standard/bids-specification/(pull|tree)/.*" \
```

yarikoptic:
Thank you @effigies for commenting with the suggestion - could accept or even while on the phone!

.circleci/config.yml
```diff
@@ -42,7 +42,11 @@ jobs:
             linkchecker -t 1 ~/build/site/
             # check external separately by pointing to all *html so no
             # failures for local file:/// -- yoh found no better way,
-            linkchecker -t 1 --check-extern --ignore-url 'file:///.*' --ignore-url https://fonts.gstatic.com ~/build/site/*html ~/build/site/*/*.html
+            linkchecker -t 1 --check-extern \
+              --ignore-url 'file:///.*' \
+              --ignore-url https://fonts.gstatic.com \
+              --ignore-url "https://github.com/bids-standard/bids-specification/(pull|tree)/.*" \
```
effigies, line 48: Ignore users:

```suggestion
              --ignore-url "https://github.com/bids-standard/bids-specification/(pull|tree)/.*" \
              --ignore-url "https://github.com/[^/]*" \
```

yarikoptic:
woohoo -- all green now.  I guess we just need one more approval now @effigies and could have it merged

effigies:
Thanks @yarikoptic!

PR #494, opened 2020-06-04
ENH: Add RECOMMENDED DatasetType key to dataset description

effigies:
Last-minute addition to #265 (Common Derivatives).

This allows for unambiguous identification of raw or derivatives datasets by parsing a single file. This is needed because any feature of derivatives has the potential to be incorporated back into raw, unless it breaks backwards compatibility. In the extreme, it may only be possible to distinguish the two by determining whether a dataset validates as raw or as derivative.

cc @bids-standard/derivatives


effigies:
Hey, it's 4pm and there's a thumbs up! Thanks @emdupre! In it goes, and may history judge us justly.

PR #495, opened 2020-06-04
[FIX] Clarify language on unsetting a key/value pair

### LINKED ISSUE #492: Inheritance principle: unsetting of key-value pairs
Came across while reviewing #488, but listing separately since the discussion is not specific to the changes in that PR, and indeed isn't even specific to derivatives.

From "Common Principles", "The Inheritance Principle":

> There is no notion of "unsetting" a key/value pair. For example if there is a JSON file corresponding to particular participant/run defining a key/value and there is a JSON file on the root level of the dataset that does not define this key/value it will not be "unset" for all subjects/runs.

This might require elaboration, as I would suggest that *in the other direction*, unsetting a key/value pair *is* possible.

The situation is probably best understood as a sequential loading of JSON data from the root directory up to the file currently being processed. If a key is absent in a root JSON, but present in a sub-directory JSON, then that key/value pair gets loaded into the JSON object regardless of the fact that the key was not present in the root JSON. This seems to be what the current text is describing, although it is seemingly describing the effect from the perspective of the opposite direction, i.e. the absence of that key/value in the root JSON does not preclude it from being loaded when the sub-directory JSON is read.

Where the current text may be contradicted is the case where a key/value pair is present in *both* JSONs, but the sub-directory JSON assigns to that key the value `null`. This *could* arguably be interpreted as "unsetting" that key-value pair.

I'm happy to have a go at rewording, but I'd prefer to make sure that there's some agreement with my thinking on this before proceeding with such.

nicholst:
closes #492 where @Lestropie observed that the current language on inheritance is very confusing.  

This small PR simply replaces this (common principles) text

> There is no notion of "unsetting" a key/value pair. For example if there is a JSON file corresponding to particular participant/run defining a key/value and there is a JSON file on the root level of the dataset that does not define this key/value it will not be "unset" for all subjects/runs.

with

> There is no notion of "unsetting" a key/value pair. Once a key/value pair is set in a given level in the dataset, lower down in the hierarchy that key/value pair will always have some assigned value.

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -187,9 +187,10 @@ levels unless they are overridden by a file at the lower level. For example,
 TR to a specific value. If one of the runs has a different TR than the one
 specified in that file, another `sub-*_task-rest_bold.json` file can be placed
 within that specific series directory specifying the TR for that specific run.
-There is no notion of "unsetting" a key/value pair. Once a key/value pair is 
-set in a given level in the dataset, lower down in the hierarchy that 
-key/value pair will always have some assigned value. Files for a particular 
+There is no notion of "unsetting" a key/value pair.
+Once a key/value pair is set in a given level in the dataset, lower down in
+the hierarchy that key/value pair will always have some assigned value.
+Files for a particular 
 participant can exist only at participant level directory, i.e
 `/dataset/sub-*[/ses-*]/sub-*_T1w.json`. Similarly, any file that is not
 specific to a participant is to be declared only at top level of dataset for eg:
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -187,10 +187,9 @@ levels unless they are overridden by a file at the lower level. For example,
 TR to a specific value. If one of the runs has a different TR than the one
 specified in that file, another `sub-*_task-rest_bold.json` file can be placed
 within that specific series directory specifying the TR for that specific run.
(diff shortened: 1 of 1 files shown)


src/02-common-principles.md
```diff
@@ -187,10 +187,9 @@ levels unless they are overridden by a file at the lower level. For example,
 TR to a specific value. If one of the runs has a different TR than the one
 specified in that file, another `sub-*_task-rest_bold.json` file can be placed
 within that specific series directory specifying the TR for that specific run.
-There is no notion of "unsetting" a key/value pair. For example if there is a
-JSON file corresponding to particular participant/run defining a key/value and
-there is a JSON file on the root level of the dataset that does not define this
-key/value it will not be "unset" for all subjects/runs. Files for a particular
+There is no notion of "unsetting" a key/value pair. Once a key/value pair is 
+set in a given level in the dataset, lower down in the hierarchy that 
+key/value pair will always have some assigned value. Files for a particular
```
effigies, line 192: While we're here, let's add newlines after sentences, so that any future edits to the sentence can have a reasonably confined diff:

```suggestion
There is no notion of "unsetting" a key/value pair.
Once a key/value pair is set in a given level in the dataset, lower down in
the hierarchy that key/value pair will always have some assigned value.
Files for a particular 
```

effigies:
LGTM.

PR #505, opened 2020-06-18
[INFRA] make circle artifact link a GH action, point to pdf

### LINKED ISSUE #484: Use RTD builds for PRs instead of CircleCI
RTD people have an experimental feature of providing builds for PRs.  See e.g. https://github.com/con/open-brain-consent/pull/50 where it was already enabled and tested.

I do not have admin privileges for RTD for this project, so the most I could do is to pester RTD people to enable the feature for this repository, which I was told they have done!  So it is now for people with the super powers to actually make it work.  Should be quite straightforward, see https://docs.readthedocs.io/en/latest/guides/autobuild-docs-for-pull-requests.html for the documentation and common gotchas. In my case it was the need "to disconnect and reconnect to GitHub service" as now described in that document.

sappelhoff:
closes #484 

Some changes, because the HTML preview for PRs is now provided by Readthedocs.

- [x] turn circleici artifact redirector app into a GitHub action
- [x] make the app point to PDF build instead of HTML build

Now the big question is whether or not to remove the CircleCI build of the HTML version. Removing it would mean:

- remove the [`build_docs` job](https://github.com/bids-standard/bids-specification/blob/2284a705bcc2d08e85a704bf5dcf5ff6ff60c596/.circleci/config.yml#L3-L30)
- remove [Pipfile](https://github.com/bids-standard/bids-specification/blob/2284a705bcc2d08e85a704bf5dcf5ff6ff60c596/Pipfile#L1) and [Pipfile.lock](https://github.com/bids-standard/bids-specification/blob/2284a705bcc2d08e85a704bf5dcf5ff6ff60c596/Pipfile.lock#L1)

The readthedocs build apparently only uses [`requirements.txt`](https://github.com/bids-standard/bids-specification/blob/2284a705bcc2d08e85a704bf5dcf5ff6ff60c596/requirements.txt#L1) and the "RTD defaults" which they install for every build anyhow. (~interestingly, `pymdown-extensions` are not being installed by RTD, but are a requirement for the circleci build :shrug:~ edit: this pkg IS being installed, because it's a [dependency of "mkdocs-material"](https://github.com/squidfunk/mkdocs-material/blob/6992b3cd733699619bd3b197b400ff51436d9b54/requirements.txt#L25) )

I am not sure whether to remove or not. It seems fair to get rid of unneeded things ... especially if they are NOT an exact mirror of the Readthedocs build environment and thus cannot be of help for debugging.

What are your opinions?


sappelhoff:
> What are your opinions?

pinging @yarikoptic and @effigies who have some experience with this.

I just don't want to merge this without prior discussion / approval.

effigies:
Seems fine.

sappelhoff:
Okay, I removed pipfile and pipfile.lock and made the build_docs job in circleci leaner ... but couldn't get rid of it because it's needed for the linkchecker job.

PR #506, opened 2020-06-19
[FIX] Clarify that `acq_time` in `scans.json` refers to first data point acquired

### LINKED ISSUE #503: Clarify "acquisition time" in scans file
In the description for the `scans` file, the specification says:

> If acquisition time is included it should be under `acq_time` header.

It would helpful to specify that the acquisition time should correspond to the beginning of acquisition (i.e., when the first data point for the run was acquired).

NOTE: This issue stems from https://github.com/nipy/heudiconv/issues/450#issuecomment-644923590.

tsalo:
Closes #503.

Changes proposed:
- Clarify that acquisition time corresponds to start of acquisition (first data point).
- Add missing period in `03-modality-agnostic-files.md`.


tsalo:
@sappelhoff Thank you! I guess I will tag @yarikoptic and @effigies (since you :+1:ed the original issue).

effigies:
I might just say "refers" instead of "SHOULD correspond". I see it as a definition, not a validatable constraint.

PR #515, opened 2020-06-25
[ENH] Update `beh/` specification to contrast with any neural recordings

effigies:
Reread the behavioral section in light of #513, and realized it's still written in reference to MRI and not electrophysiological or other recordings that we might end up specifying with an alternative data type directory.

Also tried to apply something like a semantic line-breaking scheme, but it wasn't 100% principled and can probably be improved if it seems important.


sappelhoff:
+1 for renaming the section, but let's do line breaks and text (or everything) in #513, wdyt?

effigies:
If you want to cherry-pick this into #513, that's fine with me.

PR #519, opened 2020-06-30
BEP 012: Functional derivatives

effigies:
This PR has been extracted from #207, following the incorporation of BEP 003 - Common Derivatives (#265).

I've gone through a first pass of bringing it into conformation with BEP 003. I think there are no direct conflicts, so I would like to open this up to community comment.

I think the @bids-standard/derivatives-electrophys team should have a look at the `_timeseries.<ext>` components. I suspect that at least some of this could be written more generally to ensure that we make BEP021 a natural extension rather than an awkward re-architecting.

Moderators: @effigies @cmaumet

### DIFF
--- a/src/05-derivatives/05-functional-derivatives.md
+++ b/src/05-derivatives/05-functional-derivatives.md
@@ -1,0 +1,386 @@
+# Functional derivatives
+
+## Functional derivatives maps
+
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<participant_label>/
+        func/
+            <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.nii[.gz]
+            <source_keywords>[_space-<space>][_desc-<label>]_<suffix>.json
+```
+
+for example:
+
+```Text
+pipeline1/
+    sub-001/
+        func/
+            sub-001_task-rest_run-1_space-MNI305_mean.nii.gz
+            sub-001_task-rest_run-1_space-MNI305_mean.json
+```
+
+The following table lists allowed suffixes and their corresponding measures:
+
+| `<suffix>`       | Measure                                                          |
+| ---------------- | ---------------------------------------------------------------- |
+| `mean`           | Mean across the temporal/4th dimension of the data               |
+| `std`            | Standard deviation across the temporal/4th dimension of the data |
+| `tsnr`           | Temporal SNR (i.e. mean / std)                                   |
+| `sfs`            | Signal fluctuation sensitivity                                   |
+| `alff`           | Amplitude low frequency fluctuations                             |
+| `falff`          | Fractional amplitude of low frequency fluctuations               |
+| `reho`           | Regional homogeneity (voxelwise only)                            |
+| `dcb`, `dcw`     | Voxelwise degree centrality binary and weighted                  |
+| `ecb`, `ecw`     | Voxelwise eigenvector centrality binary and weighted             |
+| `lfcdb`, `lfcdw` | Local functional connectivity density                            |
+| `vmhc`           | Voxel mirrored homotopic connectivity                            |
+
+The following metadata JSON fields are valid for derivative maps:
+
+| Key name       | Description                                                               | Required for suffix        |
+| -------------- | ------------------------------------------------------------------------- | -------------------------- |
+| BandpassFilter | String describing all relevant parameters of the applied bandpass filter. | `alff`, `falff`            |
+| Neighborhood   | String describing neighborhood for regional measures.                     | `reho`                     |
+| Threshold      | String describing threshold used for determining graph edges.             | `dcb`, `dcw`, `ecb`, `ecw` |
+| Method         | String describing method used to calculate measure.                       | `dcb`, `dcw`, `ecb`, `ecw` |
+
+## Time series
+
+A time series is a chronologically ordered series of numeric values.
+Time series will generally be stored as tables, with a row of column headers
+indicating the name of the series.
+In the case where every voxel has a time series, then the data should be stored
+in a 4D NIfTI file.
+
+All time series files MUST be accompanied by a data dictionary in JSON format,
+consistent with the format described in
+[Common principles](../02-common-principles.md#tabular-files), which describes
+metadata for each column name.
+In the case of NIfTI time series files, the notion of column name does not
+apply, so column-level metadata may be applied to the entire file.
+In addition, the following fields apply to the entire file in all cases:
+
+| Field name        | Definition                                                                                                                                                   |
+| :---------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| SamplingFrequency | REQUIRED. Sampling frequency (in Hz) of all columns in the file. Special value `"TR"` indicates one sample per volume of a corresponding BOLD series.        |
+| StartTime         | OPTIONAL. Start time in seconds in relation to the start of acquisition of the first volume in the corresponding imaging file (negative values are allowed). |
+
+Note that there are several differences with these fields in
+[Physiological and other continuous recordings](../04-modality-specific-files/06-physiological-and-other-continuous-recordings.md).
+The `"TR"` sampling frequency serves to indicate that no resampling is needed
+to use the series as a regressor for BOLD data, including BOLD series with
+non-uniform sampling, such as clustered sparse acquisition.
+`StartTime` is assumed to be 0 and therefore not mandatory.
+Additionally, because time series TSV files have column headers, the `Columns`
+field is omitted.
+
+### General time series
+
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<participant_label>/
+        func/
+	        <source_keywords>[_desc-<label>]_timeseries.tsv
+	        <source_keywords>[_desc-<label>]_timeseries.json
+```
+
+for example:
+
+```Text
+sub-001/
+    func/
+        sub-001_task-rest_run-1_desc-confounds_timeseries.tsv
+        sub-001_task-rest_run-1_desc-confounds_timeseries.json
+```
+
+Any time series with common timing information may be stored in a `timeseries`
+file.
+If a time series is specified in another sub-section, then any additional
+required metadata MUST be stored in the JSON description of the column.
+
+#### Column names
+
+Column names are unique alphanumeric values that are defined in a relevant JSON
+sidecar file. For a series of related columns, it is RECOMMENDED to use
+zero-based indexing as a suffix. For example, the first six aCompCor components
+may be named `a_comp_cor_00` ... `a_comp_cor_05`. Custom column names must not
+conflict with reserved names specified in this document.
+
+Reserved names are specified in the remaining subsections.
+
+#### Time series transformations
+
+Any column may have the following suffixes appended to indicate transformations
+applied to the original time series data:
+
+| Transformation | Description                                                                 |
+| -------------- | --------------------------------------------------------------------------- |
+| `_shift_back`  | The time series has been lagged by one TR. `x_shift_back[i] = x[i-1]`       |
+| `_dt`          | The discrete first derivative of the time series. `x_dt[i] = x[i+1] - x[i]` |
+| `_sq`          | The square of the time series                                               |
+| `_var_norm`    | The time series has been variance normalized                                |
+| `_centered`    | The time series has had its mean subtracted                                 |
+
+For example, `rot_z_shift_back_sq` means the square of the lagged version of the
+Z rotation (see [Motion-related time series](#motion-related-time-series)).
+
+### ROI-based time series extraction
+
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<participant_label>/
+        func/
+            <source_keywords>[_atlas-<atlas_label>][_desc-<label>]_timeseries.<tsv|nii[.gz]>
+            <source_keywords>[_atlas-<atlas_label>][_desc-<label>]_timeseries.json
+```
+
+for example:
+
+```Text
+sub-001/
+    func/
+        sub-001_task-rest_run-1_atlas-AAL_timeseries.tsv
+        sub-001_task-rest_run-1_atlas-AAL_timeseries.json
+        sub-001_task-rest_run-1_desc-anaticor_timeseries.nii.gz
+        sub-001_task-rest_run-1_desc-anaticor_timeseries.json
+```
+
+ROI-based time series will generally be stored as tables, with a row of column
+headers indicating the name of the time series.
+In the case where every voxel has a time series (i.e., voxel-wise regressors,
+as in ANATICOR), then the time series should be saved as a NIfTI file.
+
+#### Column metadata special fields
+
+Atlas (optional) - A label indicating an atlas that defines a region or set of
+regions in the volume. An atlas may be three- or four-dimensional, which affects
+the interpretation of the roi index.
+
+ROI (optional) - For 3D atlases, the ROI label should be numeric, corresponding
+to the value of the voxels in the ROI. For 4D atlases, the ROI label should be
+numeric, corresponding to the volume index containing the ROI mask. The
+following special labels correspond to common ROIs that may be defined by many
+atlases or segmentation algorithms:
+
+| ROI Value      | Description                                                                            |
+| -------------- | -------------------------------------------------------------------------------------- |
+| WhiteMatter    | Signal derived from white matter ROI.                                                  |
+| CSF            | Signal derived from cerebro-spinal fluid ROI.                                          |
+| Background     | Signal derived from background (out of brain) ROI.                                     |
+| GrayMatter     | Signal derived from gray matter ROI.                                                   |
+| Ventricles     | Signal derived from ventricles ROI.                                                    |
+| CircleOfWillis | Signal derived from circle of Willis ROI.                                              |
+| GlobalSignal   | Vector of mean values within the brain mask. Can be used for Global Signal Regression. |
+
+#### Column names
+
+Column names are unique alphanumeric values that are defined in a relevant JSON
+sidecar file. A naming convention might be to concatenate the atlas, ROI index,
+summarization method (defined below) and transformations (see [Time series
+transformations](#time-series-transformations)) using snake case. For example,
+`harvard_oxford_cortical_4_PC` could indicate ROI 4 of the Harvard-Oxford
+cortical atlas, summarized by taking the first principal component.
+
+#### Summarization methods
+
+To indicate the summarization method applied to construct a single time series
+for an ROI, the following column suffixes are defined:
+
+| Column suffix | Description of the transformation                                                                                                                       |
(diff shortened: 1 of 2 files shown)


src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+Any column may have the following suffixes appended to indicate transformations
+applied to the original time series data:
+
+| Transformation | Description                                                                 |
+| -------------- | --------------------------------------------------------------------------- |
+| `_shift_back`  | The time series has been lagged by one TR. `x_shift_back[i] = x[i-1]`       |
+| `_dt`          | The discrete first derivative of the time series. `x_dt[i] = x[i+1] - x[i]` |
+| `_sq`          | The square of the time series                                               |
+| `_var_norm`    | The time series has been variance normalized                                |
+| `_centered`    | The time series has had its mean subtracted                                 |
```
tsalo, line 131: Would `X_centered_var_norm` then be a substitute for something like `_z` when z-scoring values? Or would users be more likely to create a new suffix for additional standard transformations like z-scoring?

Also, if the values are underscore-delimited, could they be converted to camel-case to make it easier to separate them (especially as new values are added)?

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+Example:
+
+```Text
+sub-001/
+    func/
+        sub-001_task-rest_run-1_timeseries.tsv
+```
+
+```Text
+white_matter_mean global_signal_mean ventricles_mean
```
tsalo, line 216: Given that these columns correspond to established ROI names, should those ROI names be used instead? E.g., `white_matter_mean` --> `WhiteMatter_mean`.

I see above that snake case is preferred, so I guess it's a choice between name consistency or case consistency. Unless ROI names were changed from camel case to snake case by default?

effigies: There was a decision around Derivatives RC1 to make all specified columns snake case instead of camel case, to be [consistent with BIDS Raw](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#tabular-files):

>  It is RECOMMENDED that the column names in the header of the TSV file are written in snake_case with the first letter in lower case (e.g., variable_name, not Variable_name)

I'm not sure if this is up for revisiting.

tsalo: Ah, I see. And metadata fields are established as camel case, so it's just a difference that will have to remain. That's fine then. Thanks for clarifying.

tsalo:
To continue the conversation from https://github.com/poldracklab/fmriprep/issues/2232#issuecomment-671653585:

@effigies said:

> Regarding the decomposition.json, it looks like a table serialized to JSON, so I wonder why we wouldn't target TSVs. The derivatives spec, while a useful tool for standardizing common things, is inherently incomplete, and tools are allowed to produce unspecified derivatives. It makes more sense to me to spend a year or two with things in an unstable but useful state than wedging data into a specified but awkward structure.

@emdupre said:

> I think we've generally aimed to build off of the fMRI derivatives BEP, which [currently recommends a `decomposition.json` file](https://github.com/bids-standard/bids-specification/pull/519/files#diff-b440766386de990996b8b9e5f9cd02a5R333-R348) specifically. Though I may have misunderstood how that file was originally envisioned !
>
> Personally, I like having the metadata file in a JSON rather than a TSV, but that's only because it feels more "BIDS-y" to me. Though I'm certainly not the expert in this discussion on that 😸 All that to say, and seconding @\oesteban: Maybe we should have this conversation on that specific part of the derivatives BEP ?

Personally, I'd like to see these decomposition metrics and classifications supported in as BIDS Derivatives-compatible a manner as possible, since adding metadata to decompositions is very common, especially if you want to retain as much information as possible (e.g., variance explained for PCA components in CompCor, AROMA classifications, etc.). If a TSV targeting the decomposition JSON, with its own JSON describing the TSV's columns, would be the best way to do it, then could that be supported here?

tsalo:
Per maintainers call today, two possible routes forward for decomposition statistics are (1) house everything in jsons (as `tedana` currently does) or (2) add a new suffix for TSVs containing these stats (e.g., `_decompStats.tsv`). The former is completely valid under the current rules, and the latter can always be adopted later, so the consensus was to stick with a json file for the immediate future, for both fMRIPrep and the Functional Derivatives BEP.

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
+# Functional derivatives
+
+## Functional derivatives maps
+
+A derivative map is a measure derived from a functional series, mapped onto spatial
+locations as defined by voxels in a volume or vertices on a surface.
```
guiomar, line 6: Are these the only ways to define a location? If we have an electrode, for example, its localization can't be Cz, must be its coordinates, right? Are coordinates also included in this definition?

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+[Common principles](../02-common-principles.md#tabular-files), which describes
+metadata for each column name.
+In the case of spatial time series files, the notion of column name does not
+apply, so column-level metadata may be applied to the entire file.
+In addition, the following fields apply to the entire file in all cases:
+
+| Field name        | Definition                                                                                                                                                   |
+| :---------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| SamplingFrequency | REQUIRED. Sampling frequency (in Hz) of all columns in the file. Special value `"TR"` indicates one sample per volume of a corresponding BOLD series.        |
+| StartTime         | OPTIONAL. Start time in seconds in relation to the start of acquisition of the first volume in the corresponding imaging file (negative values are allowed). |
```
guiomar, line 73:
```suggestion
| StartTime         | OPTIONAL. Start time (in seconds) in relation to the start of acquisition of the first volume in the corresponding imaging file (negative values are allowed). |
```

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+
+```Text
+pipeline1/
+    sub-001/
+        func/
+            sub-001_task-rest_run-1_space-MNI305_mean.nii.gz
+            sub-001_task-rest_run-1_space-MNI305_mean.json
+```
+
+The following table lists allowed suffixes and their corresponding measures:
```
guiomar, line 28:
```suggestion
The following table lists allowed suffixes in <suffix> and their corresponding measures:
```

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<label>/
+        func/
+            <source_entities>[_space-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.{nii[.gz],gii,dscalar.nii}
+            <source_entities>[_space-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.json
+```
+
```
guiomar, line 17: I always tend to forget what all these labels stand for, is there a place where they are listed? Maybe it could be good to include a link somewhere, or a short reminder

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+        sub-001_task-rest_run-1_atlas-AAL_timeseries.json
+        sub-001_task-rest_run-1_desc-anaticor_timeseries.nii.gz
+        sub-001_task-rest_run-1_desc-anaticor_timeseries.json
+```
+
+ROI-based time series will generally be stored as tables, with a row of column
+headers indicating the name of the time series.
+In the case where every voxel has a time series (i.e., voxel-wise regressors,
+as in ANATICOR), then the time series should be saved as a NIfTI file.
+
```
guiomar, line 163: Reminder link of available atlases

src/05-derivatives/05-functional-derivatives.md
tsalo: The derivatives entities are going to be summarized in #645. Does that work?

guiomar: Yes! Very useful! Thanks :)

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
+# Functional derivatives
+
```
Remi-Gau, line 2: There probably should some recommendations on how to name the `desc` label in the case that people have several `func` contrasts (for example suffixe `cbv`).

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+In the case where every voxel has a time series (i.e., voxel-wise regressors,
+as in ANATICOR), then the time series should be saved as a NIfTI file.
+
+#### Column metadata special fields
+
+Atlas (optional) - A label indicating an atlas that defines a region or set of
+regions in the volume. An atlas may be three- or four-dimensional, which affects
+the interpretation of the roi index.
+
+ROI (optional) - For 3D atlases, the ROI label should be numeric, corresponding
```
rwblair, line 170: All the examples encode the camel cased roi value in the column name. Does it make sense to duplicate this information in the ROI key in the sidecar?

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+#### Summarization methods
+
+To indicate the summarization method applied to construct a single time series
+for an ROI, the following column suffixes are defined:
+
+| Column suffix | Description of the transformation                                                                                                                       |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `_mean`       | The mean of voxel time series                                                                                                                           |
+| `_median`     | The median of voxel time series                                                                                                                         |
+| `_pc[_<x>]`   | The ith eigenvariate from principal component analysis, where `x` is 0 indexed. If `x` is not specified, the first component is implied (i.e., `pc_0`). |
```
rwblair, line 204: `harvard_oxford_cortical_4_PC` in example above, should this be `harvard_oxford_cortical_pc_4`?

Remi-Gau:
maintenance note: added a link to the rendered BEP in the top message of this PR.

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+| `std`            | Standard deviation across the temporal/4th dimension of the data |
+| `tsnr`           | Temporal SNR (i.e. mean / std)                                   |
+| `sfs`            | Signal fluctuation sensitivity                                   |
+| `alff`           | Amplitude low frequency fluctuations                             |
+| `falff`          | Fractional amplitude of low frequency fluctuations               |
+| `reho`           | Regional homogeneity (voxelwise only)                            |
+| `dcb`, `dcw`     | Voxelwise degree centrality binary and weighted                  |
+| `ecb`, `ecw`     | Voxelwise eigenvector centrality binary and weighted             |
+| `lfcdb`, `lfcdw` | Local functional connectivity density                            |
+| `vmhc`           | Voxel mirrored homotopic connectivity                            |
```
effigies, line 42: These should become `stat-<label>`s for consistency with BEP 16 and BEP 29. We need a new suffix. I'm a little inclined to `boldmap`, which makes clear that it's a statistic based on a bold series. @Remi-Gau @cmaumet @tsalo (and anybody else) any opinions?

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+
+The following metadata JSON fields are valid for derivative maps:
+
+| Key name       | Description                                                               | Required for suffix        |
+| -------------- | ------------------------------------------------------------------------- | -------------------------- |
+| BandpassFilter | String describing all relevant parameters of the applied bandpass filter. | `alff`, `falff`            |
+| Neighborhood   | String describing neighborhood for regional measures.                     | `reho`                     |
+| Threshold      | String describing threshold used for determining graph edges.             | `dcb`, `dcw`, `ecb`, `ecw` |
+| Method         | String describing method used to calculate measure.                       | `dcb`, `dcw`, `ecb`, `ecw` |
```
effigies, line 51: `BandpassFilter` should be conformed to electrophysiological derivatives. cc @sappelhoff for guidance.

Not sure if there are equivalents for these others.

src/05-derivatives/05-functional-derivatives.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Time series
+
+A time series is a chronologically ordered series of numeric values.
+Time series will generally be stored as tables, with a row of column headers
+indicating the name of the series.
+In the case where the time series varies spatially, the data SHOULD be stored in
+an appropriate file, such as a 4D NIfTI file, a `.time.gii` GIFTI file,
```
effigies, line 59: Per https://github.com/nipreps/fmriprep/issues/2888:

```suggestion
an appropriate file, such as a 4D NIfTI file, a `.func.gii` GIFTI file,
```

src/05-derivatives/05-functional-derivatives.md
sappelhoff: we currently specify filtering properties in an object of objects: https://bids-specification.readthedocs.io/en/latest/glossary.html#softwarefilters-metadata

example:

```JSON
{
    "SoftwareFilters":{
        "Anti-aliasing filter":{
            "half-amplitude cutoff (Hz)":500,
            "Roll-off":"6dB/Octave"
        },
        "Some other filter":{
            "some other parameter":"some value"
        }
    }
}
```

src/05-derivatives/05-functional-derivatives.md
Remi-Gau: And the more general rule for the derivative suffix would be `{rawsuffix}map` given that functional data has several potential suffixes for raw?

https://github.com/bids-standard/bids-specification/blob/6d7eb0fa8e40e481440976a353bb02d451935db5/src/schema/rules/files/raw/func.yaml#L5C1-L6C12

effigies: That would work for me. On the other hand, since it's not a quantitative map, not sure if it's violating an implicit principle started by qmri.

src/05-derivatives/05-functional-derivatives.md
effigies: https://bids-specification--519.org.readthedocs.build/en/519/derivatives/functional-derivatives.html#sidecar-json-_boldmapjson_cbvmapjson

image

src/05-derivatives/05-functional-derivatives.md
effigies: No. In this case the `4` is indicating an ROI, and `pc` indicates the first principal component. It should be lowercase, though.

src/05-derivatives/05-functional-derivatives.md
effigies: Currently covered by `boldmap` and `cbvmap`.

src/derivatives/functional-derivatives.md
```diff
@@ -0,0 +1,404 @@
...
+consistent with the format described in
+[Common principles](../common-principles.md#tabular-files), which describes
+metadata for each column name.
+In the case of spatial time series files, the notion of column name does not
+apply, so column-level metadata may be applied to the entire file.
+In addition, the following fields apply to the entire file in all cases:
+
+| Field name        | Definition                                                                                                                                                     |
+| :---------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| SamplingFrequency | REQUIRED. Sampling frequency (in Hz) of all columns in the file. Special value `"TR"` indicates one sample per volume of a corresponding BOLD series.          |
```
htwangtw, line 87: I think the actual TR value should be saved in the metadata for distributing the derivative. Otherwise users will need to find the original dataset to get the actual TR for analysis.

effigies: The BOLD timing information should be associated with the BOLD file, whether raw or derived. I'm not sure the advantage of duplicating it in `timeseries.tsv`. What is your use case?

htwangtw: Extracting time series from ROIs from multiple datasets of different TR, and we potentially need to resample all the data to the same sampling rate for time series model (e.g. autoregressive model).

effigies: Good point, as there may not be an associated BOLD files. But in that case, would you not just set `SamplingFrequency` to `1/RepetitionTime`? Along with `StartTime`, that's a complete definition.

We should consider the `VolumeTiming` case for BOLD. We could see if `VolumeTiming`'s definition can accommodate an expansion along the lines of "In the case of ROI time series, the VolumeTiming is the timing information for the source BOLD series." Or we could create a `SampleTiming` array that makes it explicit. Or we follow PET's lead with blood recordings and add a `time` column in seconds.

I'm a little inclined toward the `time` column, as that would be a reasonable addition to any `timeseries.tsv`, and cover even and unevenly-spaced time series.

It would probably be good to have an explicit rule where timeseries with no associated BOLD series and no `time` column SHOULD NOT set `"SamplingFrequency": "TR"`.

htwangtw: `time` column is the clearest way to me to include this piece of information. I like this idea.

shnizzedy:
In light of https://github.com/bids-standard/bids-specification/pull/981, specifically the `_motion` suffix, https://github.com/bids-standard/bids-specification/blob/843f4508018a2a8a3aecef3b26cfee399c3e39c8/src/schema/objects/suffixes.yaml#L672-L676 should https://github.com/bids-standard/bids-specification/blob/1c79cbc19f7ab9b99f38376978b37d78c8ce5759/src/derivatives/functional-derivatives.md?plain=1#L258-L269 revert to `_timeseries`, change to something unambiguous, or will the sidecar and context be sufficient to differentiate the different `_motion` files?

effigies:
I guess we need to see if registration-derived motion parameters are too different to use the same suffix. I assumed we had a case not too dissimilar and we could treat these as the same type of data, kind of like how you could create synthetic bold series from a forward model, but I have not considered this in detail.

src/derivatives/functional-derivatives.md
```diff
@@ -0,0 +1,404 @@
...
+
+#### Column names
+
+The six basic motion parameters derived from motion correction have the
+following names and units:
+
+| Column name                     | Units   | Description            |
+| ------------------------------- | ------- | ---------------------- |
+| `trans_x`, `trans_y`, `trans_z` | mm      | Translation parameters |
+| `rot_x`, `rot_y`, `rot_z`       | radians | Rotation parameters    |
```
Remi-Gau, line 278: Would make sense to also make recommendations for names in raw "physio motion" data for MRI

https://bids-specification.readthedocs.io/en/stable/modality-specific-files/physiological-and-other-continuous-recordings.html#recommendations-for-specific-use-cases

image

Issue #530, opened 2020-07-20
Group entities into "within-acquisition" and "between-acquisition"


tsalo:
Currently, at least within functional MRI, the specification has only one entity that typically distinguishes files from the same acquisition- echo. However, @tiborauer recently noted that this entity can be used for multi-echo fMRI (the standard use) and to separate single-echo acquisitions with different echoes, and the resulting files would have exactly the same names.

Part of the problem is that “run” is defined as “an uninterrupted repetition of data acquisition that has the same acquisition parameters and task”. Thus, it would be incorrect to increment the run entity across acquisitions of the single-echo BOLD scan, since the echo time is changing. This also highlights how “run” and “acquisition” aren’t quite synonyms.

Another part of the problem is that the echo entity doesn’t really have its own definition:
> Multi-echo data MUST be split into one file per echo. Each file shares the same name with the exception of the _echo-<index> key/value.

This says how users must label multi-echo data, but not how echo should be used in general. I think we should define echo as “within-acquisition”, so that, if the only thing that varies between files is that entity, those files can be assumed to come from the same acquisition. In cases of single-echo acquisitions with varying echo times, we should clarify that some other entity should distinguish the acquisitions, such as acq or run.

Given that #508 and #425 propose a number of new entities that refer to the same acquisition (or grouped scan collection, depending on how we define things), I think we should discuss the idea of a broader concept of “within-acquisition” and “between-acquisition” entities.

BTW, I know that I repeatedly said “acquisition” throughout this issue, but please see #529 about the idea of “grouped scan collections” and how that’s different from “acquisition”. I think that the within- and between- entities here would apply to those groups instead of specifically being limited to an “acquisition”.

Tagging @agahkarakuzu and @emdupre, with whom I talked about this in a call.

tsalo:
I think that this rule is implicitly in effect for `echo` based on how associated files (e.g., events files) are handled. See the [Task events](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/05-task-events.html) page, which says the following:

> In case of multi-echo task run, a single `_events.tsv` file will suffice for all echoes.

and

> For multi-echo files, the `*_events.tsv` file is applicable to all echos of particular run:
>
> ```
> sub-01_task-cuedSGT_run-1_events.tsv
> sub-01_task-cuedSGT_run-1_echo-1_bold.nii.gz
> sub-01_task-cuedSGT_run-1_echo-2_bold.nii.gz
[...]

PR #532, opened 2020-07-21
[FIX] Clarify use of session entity in file names

### LINKED ISSUE #531: Clarifying when ses-<label> is optional
It is clear that having a session level in the hierarchy is optional, but it is unclear to me whether it is optional in file names when a session folder is present.  For example, is this valid? `/sub-01/ses-01/anat/sub-01_T1w.nii.gz`

I would recommend clarifying this sentence under File name structure:

> For a data file that was collected in a given session from a given subject, the file name will begin with the string sub-\<label\>_ses-\<label\>.

Cheers

Moo-Marc:
Clarified that the session entity is required in file names when the session level is present in the folder structure.
Fixes #531.


satra:
ping @yarikoptic  and @mgxd - can't remember what we do in heudiconv, but this may need to be addressed.

src/99-appendices/04-entity-table.md
```diff
@@ -13,20 +13,22 @@ A general introduction to entities is given in the section on
...
-| electrodes<br>(eeg/ieeg)                                                                       | REQUIRED      | OPTIONAL      |                | OPTIONAL      |                          |                |                          |               |                        |                |                     |                       | OPTIONAL        |                 |
-| events<br>(meg/eeg/ieeg)                                                                       | REQUIRED      | OPTIONAL      | REQUIRED       |               |                          |                |                          | OPTIONAL      |                        |                |                     |                       |                 |                 |        
+| anat<br>(T1w T2w T1rho T1map T2map T2star FLAIR FLASH PD PDmap PDT2 inplaneT1 inplaneT2 angio) | REQUIRED      | REQUIRED*     |                | OPTIONAL      | OPTIONAL                 | OPTIONAL       |                          |               |                        |                |                     |                       |                 |                 |
```
effigies, line 16: This is going to conflict with #475. @tsalo Is this something that can be accommodated in the new structure?

I think it might be worth just having a sentence making clear that, although OPTIONAL, if the Session entity has any value, it MUST appear in both the directory structure and file name.

Moo-Marc: Fine with me.  Neither "REQUIRED" nor "OPTIONAL" is really clear in this case.  I also considered putting the asterisk on the column name instead of each cell.  Or if we don't like the asterisk, moving the sentence before the table is fine too.

tsalo: I was under the assumption that "OPTIONAL" generally referred to entities that would, in specific cases, be required. E.g., `echo` is an optional entity, but in cases where there are multiple echoes it is definitely required, so I agree that this should be reflected in the definition. The only one that immediately comes to mind as truly optional in all cases is `acq` (although I would like to perhaps change that in #530).

For #475, I could add a third label for entities easily enough, although I'd prefer to use something without special characters. Perhaps something like "CONDITIONAL"? The actual logic, where conditions under which an entity is required are defined in the specification, would have to wait for a larger conversion to a machine-parseable format.

Moo-Marc: I second "CONDITIONAL" given that rfc2119 defines "OPTIONAL" as being "truly optional".  Not a great definition, but I think we agree on what it means.  

For this PR however, how about I just remove my edits to the table?

sappelhoff: CONDITIONAL sounds nice to me --> we'd just have to "extend" the rfc2119 for our purposes.

re: this PR: +1 to remove edits to entity table and merge this PR as a small clarification. Could you @Moo-Marc then open a new issue to track the discussion on that some entities which are currently OPTIONAL, are actually CONDITIONAL (our new potential extension of rfc2119)?

Moo-Marc: Done!

src/02-common-principles.md
```diff
@@ -99,7 +99,9 @@ A file name consists of a chain of *entities*, or key-value pairs, a *suffix* an
 Two prominent examples of entities are `subject` and `session`.
 
 For a data file that was collected in a given `session` from a given
-`subject`, the file name will begin with the string `sub-<label>_ses-<label>`.
+`subject`, the file name MUST begin with the string `sub-<label>_ses-<label>`. 
+If the `session` level is omitted in the folder structure, the file name MUST begin 
+with the string `sub-<label>`.
```
effigies, line 104: Just to be pedantic. Feel free to reject if it's obvious:

```suggestion
with the string `sub-<label>`, without `ses-<label>`.
```

sappelhoff:
Thank you @Moo-Marc

yarikoptic:
> ping @yarikoptic and @mgxd - can't remember what we do in heudiconv, but this may need to be addressed.

I think we are ok with it (at least as far as base code + `reproin` heuristic goes, since other heuristics could potentially force adding `_ses-` on their whim)

PR #536, opened 2020-07-22
[FIX] Replace all non-breaking spaces with vanilla spaces

nicholst:
This PR replaces all non-breaking spaces with spaces (or deletes them if they were adjacent to a space).

As far as I could tell, none of these non-breaking spaces occurred with purpose, but were rather accidental. (Only purpose I can think of is to avoid a number getting separated from a unit, but they have not been used in this way so far in the BIDS spec.)

I realise it doesn't really matter, but my editor highlights them and, in the spirit of keeping clean and simple ascii markdown, I thought they should go.

(For reference these were found with `grep $'\xc2\xa0' *.md`)

### DIFF
--- a/src/01-introduction.md
+++ b/src/01-introduction.md
@@ -64,5 +64,5 @@ as well as other papers describing specific BIDS extensions (see below).
 
 BIDS has also a
 [Research Resource Identifier (RRID)](https://www.force11.org/group/resource-identification-initiative)
--  `RRID:SCR_016124` - which you can also include in your manuscript in addition
+-   `RRID:SCR_016124` - which you can also include in your manuscript in addition
 to citing the paper.
--- a/src/01-introduction.md
+++ b/src/01-introduction.md
@@ -7,7 +7,7 @@ different ways. So far there is no consensus how to organize and share data
 obtained in neuroimaging experiments. Even two researchers working in the same
 lab can opt to arrange their data in a different way. Lack of consensus (or a
 standard) leads to misunderstandings and time wasted on rearranging data or
-rewriting scripts expecting certain structure. Here we describe a simple and
+rewriting scripts expecting certain structure. Here we describe a simple and
 easy-to-adopt way of organising neuroimaging and behavioral data. By using this
 standard you will benefit in the following ways:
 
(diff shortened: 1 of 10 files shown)


src/99-appendices/06-meg-file-formats.md
```diff
@@ -6,7 +6,7 @@ RECOMMENDED values for `manufacturer_specific_extensions`:
 | Value                                                 | Definition                                                                            |
 | ----------------------------------------------------- | ------------------------------------------------------------------------------------- |
 | [`ctf`](06-meg-file-formats.md#ctf)                   | CTF (folder with `.ds` extension)                                                     |
-| [`fif`](06-meg-file-formats.md#neuromagelektamegin)   | Neuromag / Elekta / MEGIN  and BabyMEG (file with extension `.fif`)                   |
+| [`fif`](06-meg-file-formats.md#neuromagelektamegin)   | Neuromag / Elekta / MEGIN and BabyMEG (file with extension `.fif`)                   |
```
sappelhoff, line 9:
```suggestion
| [`fif`](06-meg-file-formats.md#neuromagelektamegin)   | Neuromag / Elekta / MEGIN and BabyMEG (file with extension `.fif`)                    |
```

src/01-introduction.md
```diff
@@ -64,5 +64,5 @@ as well as other papers describing specific BIDS extensions (see below).
 
 BIDS has also a
 [Research Resource Identifier (RRID)](https://www.force11.org/group/resource-identification-initiative)
-- `RRID:SCR_016124` - which you can also include in your manuscript in addition
+- `RRID:SCR_016124` - which you can also include in your manuscript in addition
```
sappelhoff, line 67: and here, one more space is needed for our MD "style":

```suggestion
-  `RRID:SCR_016124` - which you can also include in your manuscript in addition
```

src/01-introduction.md
```diff
@@ -64,5 +64,5 @@ as well as other papers describing specific BIDS extensions (see below).
 
 BIDS has also a
 [Research Resource Identifier (RRID)](https://www.force11.org/group/resource-identification-initiative)
-- `RRID:SCR_016124` - which you can also include in your manuscript in addition
+-  `RRID:SCR_016124` - which you can also include in your manuscript in addition
```
sappelhoff, line 67:
```suggestion
-   `RRID:SCR_016124` - which you can also include in your manuscript in addition
```

PR #568, opened 2020-08-10
[INFRA] Move entity definitions to a separate page

### LINKED ISSUE #567: Move entity definitions to separate page(s)
Currently, entities are defined in one section where they're used with a heading (which helps find the definition in the search bar), and are then referenced in all other sections where they're used. The problem is that this spreads those entity definitions out across various sections, with no interlinking between them, and with variable text at each location.

For example, the [`run` entity](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#the-run-entity) is defined under "Anatomy imaging data" with the following:

> If several scans of the same modality are acquired they MUST be indexed with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only integers are allowed as run labels). When there is only one scan of a given type the run key MAY be omitted. Please note that diffusion imaging data is stored elsewhere (see below).

However, in the MEG and EEG sections, at least, the `run` entity is not defined. It is just used in the filename templates. I should note, though, that in most cases where a definition is included, it's the same as the main definition with the heading. This is good for the specification, but it requires good maintenance instead of being automated as it could be.

I propose that we move the "canonical" definitions to either one page for all of them or one page for each, and then include links to the entity definitions in specification text. This will (1) standardize the definitions and (2) prepare us for when entities must be defined only once, in a canonical manner, as in #540.

## Alternatives to consider
- Link to entity definitions in the filename templates instead of, or in addition to, the specification text. Personally, I think this would be beautiful, but I don't think hyperlinks can be placed in code snippets.
- Use `mdinclude` or something similar to simply include the full definitions in all of the relevant sections of the specification text. I don't know how this would interact with `pandoc` or the search-bar, but it would be more readable.

tsalo:
Closes #567.

Changes proposed:
1. Add a new appendix for entity definitions, using `bids_schema.py`.
1. Link to appendix entity definitions in the entity table, using `bids_schema.py`.
1. Remove section headers for entity definitions in main text, to ensure that the search bar will point to the entities appendix page.
1. Add links from main text to appendix definitions.
1. Add a new function and command line subparser to `bids_schema.py` to generate the new appendix page.

Notes:
- I didn't include Derivatives in this, but that might be something to do in the future. It will involve (1) adding derivatives-specific entities to the appendix (or a separate appendix), (2) supporting derivatives in the schema, and (3) adding links to the appendix in the Derivatives sections of the text.
- The specification text is largely unchanged, but we will want to start removing entity definitions in the text, while leaving anything that provides necessary context for different modalities, such as examples.


tsalo:
It doesn't look like internal links work in the PDF, so we may need a workaround for that (even once we have the schema).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -138,56 +138,15 @@ modalities include:
...
-different reconstruction algorithms (for example ones using motion correction).
+Currently supported entities include:
+
+-   [`sub-<label>`](../99-appendices/09-entities.md#sub)
+-   [`ses-<label>`](../99-appendices/09-entities.md#ses)
+-   [`acq-<label>`](../99-appendices/09-entities.md#acq)
+-   [`ce-<label>`](../99-appendices/09-entities.md#ce)
+-   [`rec-<label>`](../99-appendices/09-entities.md#rec)
+-   [`run-<index>`](../99-appendices/09-entities.md#run)
+-   [`mod-<label>`](../99-appendices/09-entities.md#mod)
```
tsalo, line 149: Does anyone have any thoughts on this approach to linking entity definitions in the datatype sections?

sappelhoff: It looks quite clean and prevents duplication of information, which is nice.

Yet it also disturbs the reading flow slightly.

Furthermore, when reading a part of the spec, for example on MRI, a given entity can be described with respect to that modality and tune the examples towards that use case. If we describe the entities exclusively in the centralized location, these descriptions would have to have either only very general descriptions/examples, or be very long to make sure all use cases are covered.

tsalo: I definitely want the entity _definitions_ to be general across their applications, but I do see why section-specific examples are important. Part of the problem is that we want as much of the specification to be auto-generated from the schema as possible. This means limiting section-specific text as much as possible. On the other hand, we still need that textual glue to connect all of the auto-generated text within sections, and I'm not sure what the best way is to manage that.

tsalo:
@sappelhoff I switched to just linking from the text to the new page and dropping the section headers in the text. What do you think of the new structure? 

Ultimately I'd like to minimize that "glue" (i.e., the section-specific entity descriptions) while maintaining readability and useful section-specific information (like examples), but I think this is a good start.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: Okay, I dropped the awful lists and took a much more limited approach, so I'll resolve this thread.

sappelhoff (commented):
nice, this looks like a less invasive start to me, which we can iterate upon in the future.

> dropping the section headers in the text

why did you drop those?

tsalo:
I dropped the section headers for two reasons. First, the section headers only appeared for the first definition of each entity, setting that first definition up as the canonical one. If every data type section had subsections (with headers) for all of its applicable entities, I would have left it. Now, however, the main definition for each entity is in the Entities page, and each entity-specific text throughout the rest of the specification should slowly be downsized to just reflect additional information for that data type specifically. The second reason is that the section headers in the Entities page should be the ones that show up in the search bar.

sappelhoff (approved):
I think the proposed changes are a valuable change to the specification. Thanks for preparing it @tsalo - please tag some more people you'd like to review this (or write them an email, as some may not check their notifications)

tsalo:
Thanks @sappelhoff! Pinging @yarikoptic and @effigies.

yarikoptic (approved):
This is just **wonderful**, thank you @tsalo !

One little comment:  I would have added a comment (probably just HTMLs') on top of entities and entities table markdown files:

```html
<!-- 
  This file is produced by XXX script based on the src/schema.  DO NOT EDIT DIRECTLY. 
  Follow https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md#updating-the-schema 
-->
```


if

yarikoptic:
> EDIT: Per maintainers call, this can be considered a web-only feature. It's not a big deal if the PDF cannot support these links, even though it should be possible with some work.

IMHO worth an issue (if not there already) so someone eager could just make it resolved!  Would be a great addition to PDF rendering.

tsalo:
> This is just **wonderful**, thank you @tsalo !
> 
> One little comment: I would have added a comment (probably just HTMLs') on top of entities and entities table markdown files:
> 
> ```
> 
> ```

That is a great idea! I'll add this to the entities and entity table pages. I'm just putting it in the intro text for now, but it could be  built into its own function at some point.

I'll open an issue about the internal linking.

tsalo:
Thanks @sappelhoff and @yarikoptic. I'll wait the five days (starting yesterday), in case anyone wants to raise any concerns, before merging.


@yarikoptic Absolutely, the text could be moved out into a separate function, although I want to hold off on actually implementing that because I think the current code is going to be substantially restructured in the not-too-distant future.

tsalo:
The five day embargo period is over with no concerns raised, so I'm merging now. Thank you @sappelhoff and @yarikoptic!

PR #570, opened 2020-08-12
[ENH] added PPG as an accepted channel type for EEG, MEG and iEEG

### LINKED ISSUE #561: [Ephys] Add "ACCEL" as channel type for accelerometers
- [x] PPG (plethysmography) should be added as an accepted channel type in https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html. It is a common channel recorded with EEG (in the Muse for example) but is different from ECG (heart as well but records blood flow instead of electro-cardiography). --> done in https://github.com/bids-standard/bids-specification/pull/570

- [ ] "ACCEL" for accelerometers should also be an accepted type as it is even more common than PPG (Muse, OpenBCI, etc...)

robertoostenveld:
See #561 and https://github.com/bids-standard/bids-examples/pull/197#discussion_r469153788

This does not close ... #561 since I think that the ACCEL requires some more thought and discussion with other stakeholders.

### DIFF
--- a/src/04-modality-specific-files/03-electroencephalography.md
+++ b/src/04-modality-specific-files/03-electroencephalography.md
@@ -246,6 +246,7 @@ Note that upper-case is REQUIRED:
 | GSR      | Galvanic skin response                                       |
 | HEOG     | Horizontal EOG (eye)                                         |
 | MISC     | Miscellaneous                                                |
+| PPG      | Photoplethysmography                                         |
 | PUPIL    | Eye tracker pupil diameter                                   |
 | REF      | Reference channel                                            |
 | RESP     | Respiration                                                  |


robertoostenveld:
> would we have to add this entry to the tables in the MEG and iEEG sections as well?

the way I read the specification is that each of the MEG, EEG and iEEG chapters lists the channel types that are common for that specific modality, but also allows for the specification of channel types from the others. I.e. the lists are shared between MEG, EEG and iEEG, and a complete list would be the union of the three. 

I think better (more maintainable, and more clear for people who read it) would be to have the channel types in an appendix. That would also be good for upcoming modalities (NIRS, motion).

Issue #552, opened 2020-07-31
add optional fields to some of the JSONs


Remi-Gau:
Some suggestions for some optional fields that could be added in json files.

The idea is that some of those data should (must?) be provided in the papers we write (cf COBIDAS) so it could make sense if they were included in the dataset metadata.

### dataset_description.json

- URL to pre-registration
- Information about eventual statistical power analysis

~- Financial grants info~
~- ethics: information about the IRB that approved for the data acquisition~

### Add in json files related to task files

- for all modality with tasks (func, eeg, meg, ieeg) 
  - Presentation software used
    - OS
    - Software
    - Version
    - URL to code of that was run to present the software
- for task fMRI 
  - if some design efficiency optimization

Let me know how you feel. If there is some enthusiasm for some of those I could open separate PRs to keep things and discussions modular.

sappelhoff:
thanks for the suggestions @Remi-Gau 

I particularly like OS, software, Version, Url for better descriptions of tasks.

> Information about eventual statistical power analysis

could you go into a bit more detail and perhaps provide an example of what you would think such a field would look like?

VisLab:
Just to mention.  The current BIDS specification of dataset-description has
Funding and EthicsApprovals fields.  However, the EthicsApprovals field was
not  in the JSON schema in the bids-validator.  Hence pull request #1031 to
the bids-validator, which was just merged.

Remi-Gau:
@VisLab yup I did see your PR which made me realize I was out of date on that front. :smile: 

I crossed out the things that are already in the specs in my initial message.

Remi-Gau:
> thanks for the suggestions @Remi-Gau
> 
> I particularly like OS, software, Version, Url for better descriptions of tasks.
> 
> > Information about eventual statistical power analysis
> 
> could you go into a bit more detail and perhaps provide an example of what you would think such a field would look like?

Will open separate issues for both of those

sappelhoff:
I am closing this now, because @Remi-Gau started tracking the two proposals in two distinct issues with two distinct PRs each (I love it :+1: ) thanks Remi!

Issue #557, opened 2020-08-04
adding stastistical power analysis description to raw data


Remi-Gau:
> Information about eventual statistical power analysis

could you go into a bit more detail and perhaps provide an example of what you would think such a field would look like?

_Originally posted by @sappelhoff in https://github.com/bids-standard/bids-specification/issues/552#issuecomment-667657999_

This is pretty much the info you have about this in the [spreadhseet COBIDAS for MRI](https://docs.google.com/spreadsheets/d/1dCXP0MTK3DjY09ZFd7FXgv0Ngx16_YJwVBiXOeQbTho/edit?usp=sharing) 

| info                                                                                 | description or example |
|-----------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------|
| Source of predicted effect size                                                                                  | previous literature with citation; pilot data with description |
| Significance level                                                                                               | uncorrected alpha 0.05 for an ROI | FWE­corrected alpha 0.05   |
| Target power                                                                                                     |                                                                |
| Any other parameters set (e.g. for spatial methods a brain volume and smoothness may be needed to be specified)  |                                                                |

As far as I know the info in the EEG / MEG COBIDAS is lighter than this but I would prefer to make them consistent across modalities.

Remi-Gau:
closing this for now
not so sure anymore this is within the scope of BIDS

PR #572, opened 2020-08-15
[ENH] add preregistration URL to dataset description 

Remi-Gau:
This was mentioned in #552 and could constitute a small / easy addition to the dataset description.

- [x] It there is some interest I can add a line in the JSON example. 

to do:

- [x] open PR to bids-validator to add the field, see https://github.com/bids-standard/bids-specification/pull/572#issuecomment-732167563


sappelhoff (dismissed):

CPernet (approved):

emdupre:
We should probably add this field to the example JSON immediately below as well.

More generally, I suppose I'm a bit confused about this field. I usually see pre-registration to describe a specific effect, rather than collecting certain kinds of data. That is, we otherwise include optional fields that are focused on the data itself (like the data DOI rather than the paper DOI, or the ethics approval), rather than optional fields about the study the data come from.

Was there a conversation about what this field adds ? Sorry to be behind on this !

Remi-Gau:
This was one of the suggestions I made in #552 and is somewhat related to #557, though I admit that there has not been a lengthy discussion about this. So you not behind in any way.

Remi-Gau:
> I usually see pre-registration to describe a specific effect, rather than collecting certain kinds of data. The reason that's an important distinction for me here is that -- from my read -- we otherwise include optional fields that are focused on the data itself (like the data DOI rather than the paper DOI, or the ethics approval), rather than optional fields about the study where the data come from.

Well, unless I am mistaken, pre-registration should ideally describe (amongst other things) many of the decisions that will constrain what the data will be and how much of it there will be and hence what effect size one can hopefully expect to detect in this dataset in general and not just for the effect of interest.

Ideally, this should be taken care of by having info about the power analysis and sample size estimation (see #557) but more generally pre-reg contain a lot of potential metadata about the dataset that doesn't fit (yet but might never fit) into BIDS, and having a pointer to the pre-reg might be a way to "add" that extra information.

Also ethics approval and funding are already part of dataset_description.json so I think that it is not that clear if we should completely avoid items that are not "just on the data itself" but might also pertain to the study or set of studies the dataset is part of.

Though I agree that this could get us onto the slippery slopes of mount scope creep...    :mountain: :roller_coaster:

emdupre:
> Also ethics approval and funding are already part of dataset_description.json so I think that it is not that clear if we should completely avoid items that are not "just on the data itself" but might also pertain to the study or set of studies the dataset is part of.

Personally, I see these as more directly related to the data itself, but that might just be because they're so much more common (or even necessary) to collect data, whereas pre-reg is still relatively rare in neuroimaging. 

> more generally pre-reg contain a lot of potential metadata about the dataset that doesn't fit (yet but might never fit) into BIDS, and having a pointer to the pre-reg might be a way to "add" that extra information.

I'm not against more information, for sure ! I'm fine with this if you think it will be useful :+1:

Remi-Gau:
For more context, I think the idea partly came from a mix of reflection on the COBIDAS front and also some of the [guidelines for PET data](https://journals.sagepub.com/doi/10.1177/0271678X20905433) that are at the core of the PET BEP (though I don't think pre-reg is IN the BEP).

Maybe we should not "rush" on this one just because **I** think it will be useful.

Tagging @melanieganz who maybe has an opinion on this.

src/03-modality-agnostic-files.md
```diff
@@ -28,6 +28,7 @@ Every dataset MUST include this file with the following fields:
 | EthicsApprovals    | OPTIONAL. List of ethics committee approvals of the research protocols and/or protocol identifiers.                                                                                                                                                                |
 | ReferencesAndLinks | OPTIONAL. List of references to publication that contain information on the dataset, or links.                                                                                                                                                                     |
 | DatasetDOI         | OPTIONAL. The Document Object Identifier of the dataset (not the corresponding paper).                                                                                                                                                                             |
+| Preregistration    | OPTIONAL.  URL or Document Object Identifier of the time-stamped pre-registration of the study that produced the dataset.                                                                                                                                          |
```
effigies, line 31: DOIs can be made URIs by prepending `doi:` (or URLs by prepending `https://doi.org/`), so this puts the address of one URI type on par with a full URI. An implementing tool would basically have to say:

```
try:
    val = parseurl(preregistration)
except:
    val = parseuri("doi:" + preregistration)
```

It seems conceptually cleaner to permit any URI:

```suggestion
| Preregistration    | OPTIONAL. URI of a time-stamped preregistration of the study that produced the dataset.                                                                                                                                                                            |
```

I'm not sure if there's an argument for narrowing to URL.

I am also :+1: for adding an example URI to add to the `dataset_description.json`.

CPernet: a `study that produced the dataset` sounds weird, people produce datasets
suggest
`URL or DOI of a time-stamped preregistration of the study from which the dataset originates.`

Remi-Gau:
Looking through an [unofficial list of pre-reg](https://docs.google.com/spreadsheets/d/1_R3UBKvT4vEDOAPXxR8W98QGvZMhpuyxlvYzIGekf0w/edit?usp=sharing) I spotted some that could be interesting:
- pre-reg of th [NARPS dataset](https://aspredicted.org/ru8ua.pdf)
- an EEG [pre-reg](https://osf.io/uf9gh)
- an fMRI [pre-reg](https://osf.io/8hwdv)

Any preference? Or better suggestions?

effigies:
No preference for prereg.

I think I would rather have the URI definition in the spec, rather than directing readers to the starter kit. That is supposed to be informative but not authoritative. I will post more to #601.

effigies:
This needs to be rebased on master, and use the common URI definition.

Remi-Gau:
Finally took 5 minutes to make those changes and add and a link to an example pre-reg...

effigies:
Looks like the failing link is https://www.ncbi.nlm.nih.gov/pubmed/24474423 which redirects to https://pubmed.ncbi.nlm.nih.gov/24474423/. I guess we need to update PubMed links.

effigies (approved):

Remi-Gau:
> Looks like the failing link is https://www.ncbi.nlm.nih.gov/pubmed/24474423 which redirects to https://pubmed.ncbi.nlm.nih.gov/24474423/. I guess we need to update PubMed links.

If you give me a :+1: I can smuggle that change into the PR, if you want. :wink:

sappelhoff:
Why not do the link changes in the link PR we have open? #676 

:-)

sappelhoff (approved):

sappelhoff:
oh @Remi-Gau could you please open a PR to add this field to bids-validator here: https://github.com/bids-standard/bids-validator/blob/master/bids-validator/validators/json/schemas/dataset_description.json

then we can merge the validator PR as soon as we merge this one.

All given that no-one objects with valid arguments.

src/03-modality-agnostic-files.md
```diff
@@ -55,7 +56,8 @@ Example:
     "Alzheimer A., & Kraepelin, E. (2015). Neural correlates of presenile dementia in humans. Journal of Neuroscientific Data, 2, 234001. doi:1920.8/jndata.2015.7"
   ],
   "DatasetDOI": "doi:10.0.2.3/dfjj.10",
-  "HEDVersion": "7.1.1"
+  "HEDVersion": "7.1.1",
+  "Preregistration": "doi:10.1038/s41598-018-30701-5"
```
emdupre, line 60: This is the eventual article, it seems ? I'd expect that under "ReferencesAndLinks". I'd think to use this preregistration: https://osf.io/uf9gh

emdupre: I notice that the prereg itself doesn't have a DOI, which is maybe what influenced the decision ?

Still, I'd think we'd want to point to the original prereg in this field, if at all possible...

Remi-Gau: Oh dear. :facepalm: 
I did rush too much on this one. 

Thanks @emdupre for catching it. And yeah you are right this is exactly how I mixed things up: I assumed the DOI on the pre-reg page was that of the pre-reg and not the article.

effigies:
@emdupre Does this look good to you?

emdupre:
> @emdupre Does this look good to you?

I still have the same concerns that I expressed before, but the PR itself looks fine. Thanks for patching that link, @Remi-Gau !

effigies:
Ah, I didn't scroll up very far and had forgotten that context.

It does seem like ReferencesAndLinks would be an appropriate place to put the pre-registration that already exists. If the goal is to indicate how to link to a pre-registration, would an example where we do cite one there work? Or is there an advantage to a new field?

Remi-Gau:
> It does seem like ReferencesAndLinks would be an appropriate place to put the pre-registration that already exists. If the goal is to indicate how to link to a pre-registration, would an example where we do cite one there work? Or is there an advantage to a new field?

Hum... Gut feeling is that a link in the `ReferencesAndLinks` seems a bit "underspecifiied" for this. For a user you then would have to follow each link to know if it points to a paper or a pre-reg. Similarly the distinction would be lost for a machine, no?

NB: if we decide to put it in the `ReferenceAndLinks` remind me to revert the PR in the validator:
https://github.com/bids-standard/bids-validator/pull/1120

effigies:
ReferencesAndLinks is entirely open, so you could say:

```
"ReferencesAndLinks": ["Study preregistration at https://blahblah"]
```

The question is whether having a specific preregistration field so that it could be directly parsed (as opposed to read by a human and understood) is that useful. I really don't have a strong opinion here.

Remi-Gau:
> ReferencesAndLinks is entirely open, so you could say:
> 
> ```
> "ReferencesAndLinks": ["Study preregistration at https://blahblah"]
> ```

Oh. I was unclear about this. Good to know.


> The question is whether having a specific preregistration field so that it could be directly parsed (as opposed to read by a human and understood) is that useful. I really don't have a strong opinion here.

Hum... Part of me wants to "keep it simple" and say "Pre-reg are not THAT widespread in neuroiamging, so let's put in ReferenceAndLinks and cross that pre-reg bridge when we get to it".

The cautious part of me wants to have it in there early to prevent potentially having to change that later...

emdupre:
> cautious part of me wants to have it in there early to prevent potentially having to change that later...

I'm not sure that I see how it would be backwards incompatible? Of course, I hope that by the time we get to BIDS 2.0 pre-registrations will be well-established in neuroimaging ! But I don't think we'd have to wait quite that long if it becomes a clear need in the community.

Maybe two questions: (1) do we know (approximately) how many datasets would use this now, and (2) is there a use case in mind for it being a unique field ?

I think those are the two doubts I still have. But I'd agree with @effigies that I don't have a really strong opinion, here !

effigies:
> > cautious part of me wants to have it in there early to prevent potentially having to change that later...
> 
> I'm not sure that I see how it would be backwards incompatible?

Yeah, it wouldn't be backwards incompatible so much as something you need to know and care enough about to promote an entry in a references list to a preregistration field. Which is a concern, but not a change from the status quo.

In my estimation, for it to be actually useful, we would need tools that can make use of links from datasets to preregistrations OR a database of preregistered studies such that generic database-linking tools might connect the dots. In the absence of either, perhaps this is premature, since we also don't know what form of link would be most useful for such tools/databases. Maybe a URI would be less useful than a more structured object.

Remi-Gau:
Yeah I tend to agree and the more I think about it the more it feels like this is a 20% case at the moment.

Will think about it a bit more but will most likely end up closing this PR.

> In my estimation, for it to be actually useful, we would need tools that can make use of links from datasets to preregistrations OR a database of preregistered studies such that generic database-linking tools might connect the dots.

For the record, OSF could provide the infrastructure to help create and host a [registry](https://osf.io/registries) to centralize all the pre-reg of certain "scientific community", but it needs the backing of a science organization. So as long as we don't get the higher spheres involved, this avenue might be complicated.

Remi-Gau:
Will open a PR on the validator to revert the change this introduced.

sappelhoff:
Thanks @Remi-Gau

PR #576, opened 2020-08-19
Updating the bep009 with the latest master branch updates

melanieganz:
[MISC] I would just like to update the bep-009 branch that was made while ago with the latest updates from master before we add our actual changes and make a pull request.


effigies:
I pushed master to bep-009. There were no changes, so it went in clean.

PR #578, opened 2020-08-24
[ENH] Specify echo and run indices are nonnegative integers in schema

tsalo:
This incorporates the changes in #535 to the definitions for the `echo` and `run` entities to the schema. I should have commented directly on #535 about this, but I didn't think of it until this morning.

### DIFF
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -75,8 +75,8 @@ run:
   name: Run
   description: |
     If several scans of the same modality are acquired they MUST be indexed
-    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only nonnegative
-    integers are allowed as run labels).
+    with a key-value pair: `_run-1`, `_run-2`, ..., `_run-<index>`
+    (only nonnegative integers are allowed for the `<index>`).
     When there is only one scan of a given type the run key MAY be omitted.
   format: index
 mod:
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -94,7 +94,7 @@ echo:
     key/value.
     Please note that the `<index>` denotes the number/index (in the form of a
     nonnegative integer) of the echo not the echo time value which needs to be
-    stored in the field EchoTime of the separate JSON file.
+    stored in the field `EchoTime` of the separate JSON file.
   format: index
 recording:
   name: Recording
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -75,10 +75,9 @@ run:
   name: Run
   description: |
     If several scans of the same modality are acquired they MUST be indexed
-    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only integers
-    are allowed as run labels).
+    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only nonnegative
+    integers are allowed as run labels).
     When there is only one scan of a given type the run key MAY be omitted.
-    Please note that diffusion imaging data is stored elsewhere (see below).
   format: index
 mod:
   name: Corresponding Modality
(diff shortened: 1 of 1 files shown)


src/schema/entities.yaml
```diff
@@ -75,10 +75,9 @@ run:
   name: Run
   description: |
     If several scans of the same modality are acquired they MUST be indexed
-    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only integers
-    are allowed as run labels).
+    with a key-value pair: `_run-1`, `_run-2`, `_run-3` etc. (only nonnegative
+    integers are allowed as run labels).
```
sappelhoff, line 79:
```suggestion
    with a key-value pair: `_run-1`, `_run-2`, ..., `_run-<index>`
    (only nonnegative integers are allowed for the `<index>`).
```

sappelhoff (commented):
good catch @tsalo

src/schema/entities.yaml
tsalo: This puts the schema in conflict with the in-text definition. Should I add these changes to the in-text definition too?

nicholst: Didn't appreciate this would be a problem.  Happy to take this modification.

src/schema/entities.yaml
```diff
@@ -93,6 +92,9 @@ echo:
     Multi-echo data MUST be split into one file per echo.
     Each file shares the same name with the exception of the `_echo-<index>`
     key/value.
+    Please note that the `<index>` denotes the number/index (in the form of a
+    nonnegative integer) of the echo not the echo time value which needs to be
+    stored in the field EchoTime of the separate JSON file.
```
nicholst, line 97: I am a bit lost as to how we typeset keywords... should they be monspaced or not?  In [01-magnetic-resonance-imaging-data.md](https://github.com/bids-standard/bids-specification/tree/master/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md) you'll find
* `Echo Time` (monospaced, two words clearly referencing the DICOM Tag)
* Echo Times (not monospaced, two words)
* EchoTime (one word)
* `EchoTime` (one word, monospaced)

... anyway, if this "EchoTime" should be "`EchoTime`", then change it.

nicholst (approved):
Good catch... does illustrate the attentiveness needed to keep the schema up to date.

Aside from the anal comment about typesetting keywords, this looks good to go.

tsalo:
> Good catch... does illustrate the attentiveness needed to keep the schema up to date.

Very true, but once we are able to render the specification text from the schema (e.g., #568), then we'll be back to having to only update definitions in one location. Until then, I'll try to pay closer attention to any PRs that impact entities.

src/schema/entities.yaml
tsalo: I don't know the preference here either. It seems reasonable to monospace metadata fields though, so I'll do that.

tsalo:
```suggestion
    stored in the field `EchoTime` of the separate JSON file.
```

effigies:
We could add a quick test that just greps through the entire file for these description texts.

sappelhoff (approved):

nicholst (approved):

tsalo:
@effigies I'll open a new issue for that and merge this. Thanks all!

sappelhoff:
@nicholst I just realized that we probably missed a definition in the "common principles" section.

See: https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#definitions

under point 9:

> <index> - a numeric value, possibly prefixed with arbitrary number of 0s for consistent indentation, e.g., it is 01 in run-01 following run-<index> specification.

should this become `a nonnegative numeric value` ? cc @tsalo @effigies

nicholst:
Yes @sappelhoff !  Better yet, specify that it's an integer: `<index> - a nonnegative integer` ... right? We don't want decimals, right?

sappelhoff:
> Better yet, specify that it's an integer: <index> - a nonnegative integer ... right? We don't want decimals, right?

agreed!

Issue #585, opened 2020-08-28
Add altname or preferred variable name field to entities.yaml in schema


tsalo:
This new field would specify the preferred variable name by which the entities are referred in other tools, like `pybids`. One example is the `ce` entity. The full name for the entity is "Contrast enhancing agent", the entity itself is `ce`, and the pybids variable is `ceagent`. To specify the preferred variable name (or shortened name) would make tools that use the schema more useable to outsiders.

The exact variable names could be taken directly from pybids at this point. I don't know if any other elements to the specification need this field, so please let me know if I'm missing something.

This came up in today's NIDM-Terms call. Tagging @jbpoline, @satra, and @yarikoptic.

EDIT: Perhaps a top-level YAML file should be added that describes the other YAML files/folders, in which case we have the YAML file/folder name (e.g., `datatypes`), the actual name ("Data type"), and some preferred variable name, as used in tools like pybids (`datatype`).

effigies:
The rule we came up with for PyBIDS was that the entity's queryable name should have the entity key as an initial substring, and within that constraint it should be as readable as possible.

https://github.com/bids-standard/pybids/issues/489#issuecomment-531381408

tsalo:
To be honest, the only one that's not easily inferred at the moment is `ceagent`, although I get why you didn't go with `contrast`. In the rest of the cases, I think we could just grab the first word in the entity name. Still, I think moving that field (and naming logic) here would be good.

sappelhoff:
fyi @hoechenberger @adam2392 @agramfort @jasmainak

PR #588, opened 2020-09-01
altnames in BIDS yml schema 

jbpoline:
--- PLEASE READ AND DELETE THE TEXT BELOW BEFORE OPENING THE PULL REQUEST ---

See the [CONTRIBUTING](https://github.com/bids-standard/bids-specification/blob/master/CONTRIBUTING.md) guide. Specifically:

- Please keep the title of your pull request (PR) short but informative - it will
  appear in the changelog.
- Please ensure your name is credited on our [Contributors appendix](https://github.com/bids-standard/bids-specification/blob/master/src/99-appendices/01-contributors.md).
  To add your name, please edit our [Contributors wiki](https://github.com/bids-standard/bids-specification/wiki/Contributors) and add your name with the type of contribution.
  For assistance, please contact @franklin-feingold or @sappelhoff.
- Use one of the following prefixes in the title of your PR:
  - `[ENH]` - enhancement of the specification that adds a new feature or
    support for a new data type
  - `[FIX]` - fix of a typo or language clarification
  - `[INFRA]` - changes to the infrastructure automating the specification
    release (for example building HTML docs etc.)
  - `[MISC]` - everything else including changes to the file listing
    contributors
- If you are opening a PR to obtain early feedback, but the changes
  are not ready to be merged (a.k.a. Work in Progress PR) please
  use a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/).
- After opening the PR, our continuous integration services will automatically check your contribution  for formatting errors and render a preview of the BIDS specification with your changes.
  To see the checks and preview, scroll down and click on the `show all checks` link.
  From the list, select the `Details` link of the `ci/circleci: build_docs artifact` check to see the preview of the BIDS specification.

--- PLEASE READ AND DELETE THE TEXT ABOVE BEFORE OPENING THE PULL REQUEST ---

### DIFF
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -1,13 +1,13 @@
 ---
 sub:
   name: Subject
-  pybids_name: subject
+  altname: subject
   description: |
     A person or animal participating in the study.
   format: label
 ses:
   name: Session
-  pybids_name: session
+  altname: session
   description: |
     A logical grouping of neuroimaging and behavioral data consistent across
     subjects.
@@ -26,7 +26,7 @@ ses:
   format: label
 task:
   name: Task
-  pybids_name: task
+  altname: task
   format: label
   description: |
     Each task has a unique label that MUST only consist of letters and/or
@@ -35,7 +35,7 @@ task:
     Those labels MUST be consistent across subjects and sessions.
 acq:
   name: Acquisition
-  pybids_name: acquisition
+  altname: acquisition
   description: |
     The OPTIONAL acq-<label> key/value pair corresponds to a custom label the
     user MAY use to distinguish a different set of parameters used for
@@ -55,7 +55,7 @@ acq:
   format: label
 ce:
   name: Contrast Enhancing Agent
-  pybids_name: ceagent
+  altname: ceagent
   description: |
     Similarly the OPTIONAL ce-<label> key/value can be used to distinguish
     sequences using different contrast enhanced images.
@@ -65,7 +65,7 @@ ce:
   format: label
 rec:
   name: Reconstruction
-  pybids_name: reconstruction
+  altname: reconstruction
   description: |
     Similarly the OPTIONAL rec-<label> key/value can be used to distinguish
     different reconstruction algorithms (for example ones using motion
@@ -73,14 +73,14 @@ rec:
   format: label
 dir:
   name: Phase-Encoding Direction
-  pybids_name: direction
+  altname: direction
   description: |
     Similarly the OPTIONAL dir-<label> key/value can be used to distinguish
     different phase-encoding directions.
   format: label
 run:
   name: Run
-  pybids_name: run
+  altname: run
   description: |
     If several scans of the same modality are acquired they MUST be indexed
     with a key-value pair: `_run-1`, `_run-2`, ..., `_run-<index>`
@@ -89,7 +89,7 @@ run:
   format: index
 mod:
   name: Corresponding Modality
-  pybids_name: modality
+  altname: modality
   description: |
     In such cases the OPTIONAL `mod-<label>` key/value pair corresponds to
     modality label for eg: T1w, inplaneT1, referenced by a defacemask image.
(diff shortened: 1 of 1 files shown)


jbpoline:
Also referencing https://github.com/bids-standard/bids-specification/issues/585

sappelhoff (commented):
I like the idea, but we should come up with an altname for each entity, for example in the current proposal `split` still has  `''`as altname

tsalo:
@sappelhoff In the call that led to this PR, one idea was to essentially "inherit" the variable name from the entity name (i.e., convert it to lower-case and then use as-is) when the variable name isn't explicitly set. That said, I agree with you. I'd prefer to be explicit about it, rather than add a new rule.

src/schema/entities.yaml
```diff
@@ -116,13 +128,15 @@ proc:
...
   name: Space
+  altname: space
   description: |
     The optional space label (`*[_space-<label>]_electrodes.tsv`) can be used
     to indicate the way in which electrode positions are interpreted.
     The space label needs to be taken from the list in Appendix VIII.
   format: label
 split:
   name: Split
+  altname: ''
```
tsalo, line 139:
```suggestion
  altname: split
```

src/schema/entities.yaml
```diff
@@ -106,6 +117,7 @@ recording:
   format: label
 proc:
   name: Processed (on device)
+  altname: ''
```
tsalo, line 120:
```suggestion
  altname: processed
```

src/schema/entities.yaml
```diff
@@ -1,11 +1,13 @@
 ---
 sub:
   name: Subject
+  altname: subject
```
yarikoptic, line 4: I dislike duplication. Why not to make `altname` to be considered `name.lower()` if undefined, and define it **only** if it does not follow that rule?

sappelhoff: > Why not to make altname to be considered name.lower() if undefined, and define it only if it does not follow that rule?

no strong feeling, but looking at this it seems easier to simply add an `altname` everywhere (guarantee there is one). The alternative would be to codify the name.lower() rule somewhere and link to it - but then people have to find that rule, and always make a check "is there an altname?" --- if not, take "name" and apply .lower()

I am aware that the latter of what I said is still pretty easy, hence why I started my sentence with "no strong feeling".

What is your opinion @tsalo ?

tsalo: I generally prefer explicit over implicit. I think there is room for new rules to reduce duplication, but I think we should hold off on determining and implementing those rules until the schema is more stable. Once we have a complete and stable version of the schema, we'll be able to better see what duplication is necessary and what duplication can be removed.

jbpoline: I agree with this - sounds reasonable to me

sappelhoff: sounds like we are 3 people in favor of "duplication" here, @yarikoptic do you maybe want to tag some more reviewers that you want to provide their opinion?

Let me tag @effigies

effigies: I'll annoy everybody and say I'm fine with explicitness at the cost of duplication, but say that I find "altname" ambiguous in intent.

What we really have are three contexts: 1) Programmatic; 2) File name; 3) Spec text. Given that this document is tooling-centric, maybe the key should be the programmatic name:

```YAML
ceagent:
    name: Contrast Enhancing Agent
    entity: ce
    description: ...
reconstruction:
    name: Reconstruction
    entity: rec
    description: ...
```

We could be more explicit and say `display-name` instead of `name` and `entity-key` instead of `entity`.

WDYT?

tsalo: I think that's a reasonable change. It definitely makes including a "preferred variable name" less awkward.

sappelhoff: +1 also from my side

effigies:
I've proposed #616 as an alternative to this, based on recent discussion.

PR #606, opened 2020-09-14
[ENH] Describe arbitrary units in Common Principles

### LINKED ISSUE #580: Define arbitrary units (or synonym) in specification
The SI doesn't have a term for arbitrary/unknown units, as far as I can tell, which means that whatever term BIDS uses must be explicitly defined in the specification. BEP001 (#508) has adopted "arbitrary units (a.u.)", while [BEP009](https://docs.google.com/document/d/1mqMLnxVdLwZjDd4ZiWFqjEAmOmfcModA_R535v3eQs0) uses "unitless". According to [Wikipedia](https://en.wikipedia.org/wiki/Arbitrary_unit), alternatives include "unknown unit" or "procedure defined unit". Also, "a.u." can apparently be confused with "astronomical unit" or "atomic unit". Astronomical units probably won't come up in BIDS, but atomic units could. The abbreviation recommendations on wikipedia are "arb. unit" and "[arb'U]".

Whatever our decision is, the Units page of the specification should be updated and the choice in terminology should be fed back into the BEPs.

Note: This stems from [this Mattermost conversation](https://mattermost.brainhack.org/brainhack/pl/utu4itcjn3d7dysozb5sq36afe).

tsalo:
Closes #580. Although I considered adding something to Appendix 5 (Units), it didn't seem to fit well with the contents of the Appendix, so I've limited this change to a single list item in Common Principles#Units.

Changes proposed:

- Add an item to the additional rules list in Common Principes#Units describing how arbitrary units/unitless data should be handled.

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -544,7 +544,7 @@ For additional rules, see below:
 
 -   Frequency SHOULD be expressed in Hertz.
 
--   Arbitrary/unitless units SHOULD be represented with the format `arbitrary`.
+-   Arbitrary units SHOULD be indicated with the string `"arbitrary"`.
 
 Describing dates and timestamps:
 
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -544,6 +544,8 @@ For additional rules, see below:
 
 -   Frequency SHOULD be expressed in Hertz.
 
+-   Arbitrary/unitless units SHOULD be represented with the format `arbitrary`.
+
 Describing dates and timestamps:
 
 -   Date time information MUST be expressed in the following format


src/02-common-principles.md
```diff
@@ -544,6 +544,8 @@ For additional rules, see below:
 
 -   Frequency SHOULD be expressed in Hertz.
 
+-   Arbitrary/unitless units SHOULD be represented with the format `arbitrary`.
```
effigies, line 547: "unitless units" feels odd. What about:

```suggestion
-   Arbitrary units SHOULD be indicated with the string `"arbitrary"`.
```

tsalo: That sounds fine to me, although for the folks who were originally using "unitless" (e.g., in BEP009), I'd like to make sure that it's clear that they'd use "arbitrary" in that case as well. @melanieganz WDYT?

sappelhoff (approved):

effigies (approved):

sappelhoff:
@tsalo thanks for the PR. Can you go one more step (two actually :wink: ) and check that this new BIDS definition is observed in BEP009 and BEP001?

tsalo:
@sappelhoff There's a call this week for BEP001 where I plan to bring this up. I will also comment on BEP009 directly.

PR #616, opened 2020-09-18
SCHEMA: Declare entities by concept names, add entity field for filename components

### LINKED ISSUE #585: Add altname or preferred variable name field to entities.yaml in schema

effigies:
As proposed in https://github.com/bids-standard/bids-specification/pull/588#discussion_r487266873, roughly agreed on, and not (yet) argued against.

This uses the names settled on in https://github.com/bids-standard/pybids/issues/489#issuecomment-531381408 to allow pybids to query in a somewhat consistent manner.

Closes #585.
Closes #588.

### DIFF
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -1,11 +1,13 @@
 ---
 sub:
   name: Subject
+  pybids_name: subject
   description: |
     A person or animal participating in the study.
   format: label
 ses:
   name: Session
+  pybids_name: session
   description: |
     A logical grouping of neuroimaging and behavioral data consistent across
     subjects.
@@ -24,6 +26,7 @@ ses:
   format: label
 task:
   name: Task
+  pybids_name: task
   format: label
   description: |
     Each task has a unique label that MUST only consist of letters and/or
@@ -32,6 +35,7 @@ task:
     Those labels MUST be consistent across subjects and sessions.
 acq:
   name: Acquisition
+  pybids_name: acquisition
   description: |
     The `acq-<label>` key/value pair corresponds to a custom label the
     user MAY use to distinguish a different set of parameters used for
@@ -51,6 +55,7 @@ acq:
   format: label
 ce:
   name: Contrast Enhancing Agent
+  pybids_name: ceagent
   description: |
     The `ce-<label>` key/value can be used to distinguish
     sequences using different contrast enhanced images.
@@ -60,6 +65,7 @@ ce:
   format: label
 rec:
   name: Reconstruction
+  pybids_name: reconstruction
   description: |
     The `rec-<label>` key/value can be used to distinguish
     different reconstruction algorithms (for example ones using motion
@@ -67,12 +73,14 @@ rec:
   format: label
 dir:
   name: Phase-Encoding Direction
+  pybids_name: direction
   description: |
     The `dir-<label>` key/value can be used to distinguish
     different phase-encoding directions.
   format: label
 run:
   name: Run
+  pybids_name: run
   description: |
     If several scans of the same modality are acquired they MUST be indexed
     with a key-value pair: `_run-1`, `_run-2`, ..., `_run-<index>`
@@ -81,6 +89,7 @@ run:
   format: index
 mod:
   name: Corresponding Modality
+  pybids_name: modality
   description: |
     The `mod-<label>` key/value pair corresponds to modality label for defacing
     masks, e.g., T1w, inplaneT1, referenced by a defacemask image.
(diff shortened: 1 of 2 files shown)


satra:
@effigies - this really depends on how you think about the relation between pieces of information. i fear that the current formulation may handcuff future changes.

current proposal

```
pybids_variable: 
    bids_key: bids_name
```

as opposed to:

```
common_concept:
   preferred_variable_name: variable_name
   bids_key: bids_spec_name
```

so the current proposal ties the schema between two very specific things. everything can be reworked especially since the current yaml files don't form a formal general spec of any kind.

effigies:
Could you make your suggestion concrete?

src/schema/entities.yaml
```diff
@@ -1,11 +1,13 @@
 ---
-sub:
+subject:
   name: Subject
+  entity: sub
```
satra, line 4: > Could you make your suggestion concrete?

don't treat the key as the pybids variable.

```suggestion
subject:
  name: Subject
  entity: sub
  variable_name: subject
```

this results in a duplicate name but leaves subject the key to be an enumerated concept/term in bids.

effigies: Is there a case where this isn't duplicative? If the key will always be a valid variable name, then pybids might be able to move toward whatever is decided as the key. If not, then being explicit makes sense.

While this is phrased as a way to standardize variable names, the pybids variable names were chosen to solve the problem of referencing the concepts more clearly, although constrained by Python syntax.

satra: as you get to other concepts like metadata you may run into variable names being different or shorter again. we may consider that this should be RepetitionTime even in pybids, but just noting that there may be cases.

```
RepetitionTime:
  name: Repetition Time
  variable_name: TR
```

tsalo: I kind of liked @effigies' restructuring primarily because it avoided a "variable_name" or "altname" field, while still retaining the same information. I can see metadata fields where a variable name might need to be shorter than a meaningful concept name though.

effigies:
I'm skeptical that we're going to find this case. We have four variable sources: entities (alphanumeric), TSV column headers (snake_case), metadata fields (CamelCase when not matching columns) and a couple special structural variables (datatype, suffix). The entities are the only ones shortened enough to need renaming to reduce confusion. For columns and metadata fields, I can't really see a case as a layout consumer for the name not to match the standard.

satra:
just to clarify. my issue is not really about the key, it's about saying that the key is a pybids variable name and no other variable names could be possible in the future. if pybids choses to use the key as a variable name, that's fine, but let's not call the key a pybids variable name.

tsalo:
@satra I agree with that. I prefer to consider that top-level term a "common concept" (in single-word format), as you said above. Once we reorganize the schema in #609, I think these top-level names will instead by the filenames for split term-specific YAML files anyway, so it will be fairly distinct from a "variable name", even though the names will be good choices for variable names.

effigies:
Modified the title, as I think that's the only actual change required? Please let me know if there's a more substantial change needed.

tsalo (approved):

sappelhoff (approved):

tsalo:
@effigies do you know why the RTD CI is still registering as pending?

effigies:
Probably a temporary glitch on GitHub or RTD. Closing and reopening sometimes restarts CI...

effigies:
Passed this time.

tsalo:
I'm not sure if we want to apply the five-day rule to schema-related changes, but if so then we can merge this on Monday, I think. Pending any new concerns from the community, of course.

effigies:
I would consider this more on the order of infrastructure, and provisional at that. Also, the last commit was 13 days ago, so I'd say go ahead.

tsalo:
Sounds good to me! I'll merge now.

sappelhoff:
> I would consider this more on the order of infrastructure, and provisional at that. Also, the last commit was 13 days ago, so I'd say go ahead.

agreed.

PR #619, opened 2020-09-18
[FIX] improve HED documentation

VisLab:
This request makes some editorial improvements in HED.md such as adding subsections. It also adds the specification of how to document in machine-actionable form the meaning of columns in *_events.tsv files containing values rather than categorical values. There will be an accompanying pull request for the bids-validator incorporating the implementation of this feature.


src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
-Example:
+Example: An `*_events.tsv` annotated using HED tags for individual events.
 
 ```Text
-onset	duration	HED
-1.1	n/a	Event/Category/Experimental stimulus, Event/Label/CrossFix,  Sensory presentation/Visual, Item/Object/2D Shape/Cross
-1.3	n/a	Event/Category/Participant response, Event/Label/ButtonPress, Action/Button press
+onset  duration  HED
+1.1    n/a       Event/Category/Experimental stimulus, Event/Label/CrossFix,  Sensory presentation/Visual, Item/Object/2D Shape/Cross
+1.3    n/a       Event/Category/Participant response, Event/Label/ButtonPress, Action/Button press
```
sappelhoff, line 30: here you replaced the "tab" character `	` with a bunch of "space" characters. 

I know that we have both variants in the specficiation. The "tab" character style of formatting the TSV examples has the advantage that users can simply copy and paste it ... at the expense that sometimes the columns are not aligned.

I am not sure what kind of style we should enforce throughout the spec, but consistency would probably pay off. Opinions @tsalo @effigies ? I know that @yarikoptic always wants "tab" characters.

src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
-paradigm. HED was originally developed with EEG in mind, but is applicable to
-all behavioral experiments. Each level of a hierarchical tag is delimited with
-a forward slash (`/`). A HED string contains one or more HED tags separated by
-commas (`,`). Parentheses (brackets, `()`) group tags and enable specification
-of multiple items and their attributes in a single HED string (see section 2.4
-in [HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)).
+Hierarchical Event Descriptors (HED) are a controlled vocabulary of terms describing
+events in a machine-actionable form so that algorithms can use the information without
+manual recoding. HED was originally developed with EEG in mind, but is applicable to
+all behavioral experiments.
```
sappelhoff, line 6: adding a line break after each finished sentence:

```suggestion
manual recoding.
HED was originally developed with EEG in mind, but is applicable to all behavioral experiments.
```

src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
+
+Each level of a hierarchical tag is delimited with a forward slash (`/`). A HED string
+contains one or more HED tags separated by commas (`,`). Parentheses (brackets, `()`)
+group tags and enable specification of multiple items and their attributes in a
+single **HED string** (see section 2.4 in
+[HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)).
```
sappelhoff, line 12: again: line breaks after finished sentences.

```suggestion
Each level of a hierarchical tag is delimited with a forward slash (`/`).
A HED string contains one or more HED tags separated by commas (`,`).
Parentheses (brackets, `()`) group tags and enable specification of multiple items
and their attributes in a single **HED string** (see section 2.4 in
[HED Tagging Strategy Guide](http://www.hedtags.org/downloads/HED%20Tagging%20Strategy%20Guide.pdf)).
```

src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
-`*_events.tsv` file.
-
-Example:
+this method of annotation is not convenient unless the annotations are
+automatically generated. Usually annotations that appear in the `HED` column
+are specific to each individual event. Information that is common to groups
+of events can be annotated by category. Numerical values associated with each
+event can be annotated by value type.  Annotating by category and by value
+greatly reduces the effort required to HED tag data and improves the clarity
+for data users.
```
sappelhoff, line 42:
```suggestion
this method of annotation is not convenient unless the annotations are
automatically generated.
Usually annotations that appear in the `HED` column are specific to each individual event.
Information that is common to groups of events can be annotated by category.
Numerical values associated with each event can be annotated by value type. 
Annotating by category and by value greatly reduces the effort required to HED tag data
and improves the clarity for data users.
```

src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
 
 ```Text
-onset	duration	mycodes
-1.1	n/a	Fixation
-1.3	n/a	Button
-1.8	n/a	Target
+onset  duration  mycodes
+1.1    n/a       Fixation
+1.3    n/a       Button
+1.8    n/a       Target
```
sappelhoff, line 63: same comment as above --> not sure if in the whole BIDS spec, we should format the TSV examples using "tab" characters ... or a bunch of spaces.

tab seems cleaner and more consistent (it's always one character, ... not a varying number of space characters), but space characters currently make the examples look nicer :shrug:

src/99-appendices/03-hed.md
```diff
@@ -66,44 +74,114 @@ Example:
...
+## Best practices
+
+Most studies will have event categorical variables and value variables that are common across many
+of the datasets in the study. You should try to annotate these columns in a `*_events.json` sidecar
+as high in the study hierarchy as possible to avoid duplicate annotations. Annotations that can be
+placed in sidecars are preferred to those placed directly in the HED column, because they are simpler,
+more compact, and less prone to inconsistent annotation. Downstream tools should not distinguish
+between tags specified using the explicit HED column and the categorical specifications, but should
+form the union before analysis. Further, the [inheritance principle](../02-common-principles.md#the-inheritance-principle)
+applies, so the data dictionaries can appear higher in the BIDS hierarchy.
```
sappelhoff, line 166: line breaks:

```suggestion
Most studies will have event categorical variables and value variables that are common across many
of the datasets in the study.
You should try to annotate these columns in a `*_events.json` sidecar
as high in the study hierarchy as possible to avoid duplicate annotations.
Annotations that can be placed in sidecars are preferred to those placed directly in the HED column,
because they are simpler, more compact, and less prone to inconsistent annotation.
Downstream tools should not distinguish between tags specified using the explicit HED column and
the categorical specifications, but should form the union before analysis.
Further, the [inheritance principle](../02-common-principles.md#the-inheritance-principle) applies,
so the data dictionaries can appear higher in the BIDS hierarchy.
```

src/99-appendices/03-hed.md
```diff
@@ -66,44 +74,114 @@ Example:
...
+
+You can should try to annotate in as much detail as possible. The HED path structure makes it easy
+for analysis tools to extract tags at different levels of detail, for example only consider 2D
+shapes for stimuli, but not worry about color or shape details.
```
sappelhoff, line 170: corrected a typo ("can should"), added line breaks.

I think these two sentences (1) and (2) also need a more explicit logical connection --> more like: `(1) try to tag well ... because then (2)` ... currently it's just `(1) ... (2)`, without explicit connection

```suggestion
You should try to annotate your data in as much detail as possible.
The HED path structure makes it easy for analysis tools to extract tags at different levels of detail,
for example only consider 2D shapes for stimuli, but not worry about color or shape details.
```

src/99-appendices/03-hed.md
```diff
@@ -1,63 +1,71 @@
...
+In many experiments, the event instances fall into a much smaller number of
+categories, and often these categories are labeled with numerical codes or short names.
+This categorical information usually corresponds to one or more columns in `*_events.tsv`
+representing categorical values. Instead of tagging this information for each
+individual event, you can assign HED tags for each distinct categorical value
+in an accompanying `*_events.json` sidecar and allow the analysis tools to make
+the association with individual event instances during analysis.  The column
+name in the `*_events.tsv` identifies the type of categorical variable. The following
+`*_events.tsv` file has one categorical variable called `mycodes` that takes
+on three possible values: `Fixation`, `Button`, and `Target`.
```
sappelhoff, line 55: line breaks:

```suggestion
In many experiments, the event instances fall into a much smaller number of
categories, and often these categories are labeled with numerical codes or short names.
This categorical information usually corresponds to one or more columns in `*_events.tsv`
representing categorical values.
Instead of tagging this information for each individual event,
you can assign HED tags for each distinct categorical value
in an accompanying `*_events.json` sidecar and allow the analysis tools to make
the association with individual event instances during analysis.
The column name in the `*_events.tsv` identifies the type of categorical variable.
The following `*_events.tsv` file has one categorical variable called `mycodes` that takes
on three possible values: `Fixation`, `Button`, and `Target`.
```

sappelhoff (commented):
Thanks @VisLab and @happy5214

I left a few comments --> most of them related to formatting of the sentences. Remember that we try to start a new line for each new sentence, and wherever else it makes sense semantically, so that in the future the git diff will be more readable.

src/99-appendices/03-hed.md
VisLab: Apologies on the line breaks---I'll try to remember for next time.  I ended up changing the tabs to spaces because I found the table unreadable with the columns not aligned.  This is problematic for more complex tables.

sappelhoff (commented):
Thanks for addressing the fist round of reviews @VisLab 

regarding the "spaces" instead of "tabs" in the TSV examples, I am okay with them being spaces. We don't have a policy for that right now, and if we want one, we should discuss it in a separate, dedicated issue.

I left one more comment on what was unclear to me, it'd be nice if you could address that.

Finally, I think we can still include this PR in the 1.4.1 release that's upcoming, so I added a milestone to the PR.

sappelhoff (approved):

effigies (approved):
LGTM. I suggest a squash and merge when it's time.

sappelhoff:
Thanks @VisLab

yarikoptic:
> regarding the "spaces" instead of "tabs" in the TSV examples, I am okay with them being spaces. We don't have a policy for that right now, and if we want one, we should discuss it in a separate, dedicated issue.

sorry for coming late to the game:  by the definition of TSV (tab-separated values) **TAB must be used as a separator**, or it is not a TSV, and tools reading TSV file might not read it correctly if spaces are used.  There is IMHO no need to re-iterate that it must be TAB, like there is no need to provide a full description of JSON syntax while talking about JSON files.

sappelhoff:
We were talking only about how to format examples in the flow text of the spec. Markdown Codeblocks either containing spaces to look nice out tabs to be accurately reflecting tsv

VisLab:
I changed to spaces because the unaligned columns made the table very
difficult to understand the table. I don't think this implies that tsv
files shouldn't be tab-separated.  We could explicitly put an indication
that the table has been solumn-aligned to improve readability.  We could
also put a link to a file containing the actual tab separated information.
I don't think anyone actually cuts and pastes these examples into a file.
They should probably be looking at the examples repository for that.

yarikoptic:
Thank you for the explanation!  Indeed a larger issue. Filed: #644 (hopefully someone would get interested to harvest the kudos for the "large `git blame`" with such a contribution ;) ) and submitted #643 as a temporary measure ;)

Issue #239, opened 2019-05-29
[ENH] Allow images in dwi and func to have an IntendedFor field in their json sidecar


mattcieslak:
Currently EPI fieldmaps are stored in the `fmap` directory. The user specifies which scans in `func` or `dwi` the fieldmaps are supposed to correct via the `IntendedFor` field in the corresponding json sidecars.

Sometimes a whole DWI series is acquired with the reverse phase encoding direction of another DWI series and the low b images in each series are meant to correct the opposite PE direction series.

For this use case, would it be best to 

1. Manually extract some low b images from each series into a new nifti file in fmaps with `IntendedFor` specified in its sidecar or 
2. Add an `IntendedFor` field to the sidecars of the images in `dwi` and `func`?

Option 1 would go against keeping the images as raw as possible and would make automatic BIDS conversion very tricky. Option 2 might be confusing because images in `dwi` and `func` are not really "intended for" distortion correction.

I'd be very interested to hear opinions on this and happy to do a PR

effigies:
Ah, I hadn't understood that you were proposing an `IntendedFor` field for the actual func/dwi images. I had interpreted you as saying that `IntendedFor` only linked fmap to func, not fmap to dwi.

> Option 2 might be confusing because images in `dwi` and `func` are not really "intended for" distortion correction.

I think this is correct. If it will take modification to get them into a state where they can be used as fieldmaps, then it's misleading; if they're ready-to-use, then they're just fieldmaps. I think we need to consider the modified version derivatives (e.g., `<bids>/derivatives/dwi2fmap/sub-01/fmap/sub-01_acq-PA_epi.nii.gz` with an `IntendedFor` metadata field pointing to the correct DWI image), and work to make sure preprocessing pipelines can access derivatives.

In fMRIPrep, this is already something we need to do with regard to lesion masks and custom brain masks.

mattcieslak:
How would the `IntendedFor` fields in `<bids>/derivatives/dwi2fmap/sub-01/fmap/sub-01_acq-PA_epi.nii.gz` get specified? A different field in the original image's sidecar? Or maybe rely on the `ShimSettings` and `PhaseEncodingDirection` to determine if images can be used to correct each other?

francopestilli:
@mattcieslak @effigies are we expecting this to affect also how we save data related to the DWI modality?

effigies:
@mattcieslak To make it concrete, suppose we start with:

```
bids/
    sub-01/
        dwi/
[...]

Issue #263, opened 2019-07-03
Allow ``IntendedFor`` and make ``task-`` optional for ``_sbref.ext``


oesteban:
Related to #239.

Some fMRI studies will not acquire one SBRef for every multiband BOLD scan (this is probably true for DWI too, and thus #239).

Currently, the `task-Name` entity needs to be defined for SBRefs. Doing it optional, would allow the following:

```
func/
  sub-3010_sbref.nii.gz
  sub-3010_task-discountFix_run-1_bold.nii.gz
  sub-3010_task-manipulationTask_run-1_bold.nii.gz
  sub-3010_task-motorSelectiveStop_run-1_bold.nii.gz
  sub-3010_task-rest_run-1_bold.nii.gz
  sub-3010_task-rest_run-1_sbref.nii.gz
  sub-3010_task-stopSignal_run-1_bold.nii.gz
```

Where, by the inheritance principle, ``sub-3010_sbref.nii.gz`` is intended for all tasks but `rest` because a particular `sbref` exists for that one.

Additionally, I propose to allow defining the ``IntendedFor`` metadata for SBRefs, so the same behavior can be implemented without missing the provenance about what task was acquired closest to the sbref, e.g.:

```
func/
  sub-3010_task-discountFix_run-1_bold.nii.gz
  sub-3010_task-discountFix_run-1_sbref.json
  sub-3010_task-discountFix_run-1_sbref.nii.gz
  sub-3010_task-manipulationTask_run-1_bold.nii.gz
  sub-3010_task-motorSelectiveStop_run-1_bold.nii.gz
  sub-3010_task-rest_run-1_bold.nii.gz
  sub-3010_task-rest_run-1_sbref.nii.gz
  sub-3010_task-stopSignal_run-1_bold.nii.gz
```

where `sub-3010_task-discountFix_run-1_sbref.json` has an `IntendedFor` field:

```
{
  "IntendedFor": [
    "sub-3010_task-discountFix_run-1_bold.nii.gz",
    "sub-3010_task-manipulationTask_run-1_bold.nii.gz",
    "sub-3010_task-motorSelectiveStop_run-1_bold.nii.gz",
    "sub-3010_task-stopSignal_run-1_bold.nii.gz"
  ]
}
```

PR #622, opened 2020-09-28
[ENH] Allow encoding the fieldmapping intent of the protocol

oesteban:
This PR addresses the problem @mattcieslak spotted at in #239.

This enhancement (WIP) basically allows for researchers to encode the protocol's intent regarding fieldmaps.

As @satra introduced in https://github.com/bids-standard/bids-specification/issues/239#issuecomment-497329009, BIDS "*could encode intent and automation. Whether it should is a community decision.*"

This PR proposes a solution to encoding the intent. It doesn't modify anything to allow also encoding automation.

The PR attempts to be backward compatible.

I'm submitting this draft PR to open discussions and looking forward to feedback.

Resolves: #239.
References: #263, nipreps/dmriprep#43, bids-standard/bids-2-devel#39

### The gist of this PR...

... is the creation of two new metadata pairs `B0FieldIdentifier` and `B0FieldSource` to encode how the researcher originally designed the MR protocol regarding fieldmaps.

`B0FieldIdentifier` is set in all images that are to be used in estimating some fieldmap. Images with the same `B0FieldIdentifier` are expected to be used together (e.g., two `_epi` datasets under `fmap/` with opposed PE direction, two `_epi` datasets under `fmap/` plus two `_dwi` datasets under `dwi/`, or maybe two `_sbref` datasets under `func/` with opposed PEs, etc.)

`B0FieldSource` states identifiers intended for correcting the particular instance where it is defined.

### Currently unsupported example:

```Text
func/
  sub-01_task-rest_bold.json
  sub-01_task-rest_bold.nii.gz
  sub-01_task-rest_dir-ap_sbref.json
  sub-01_task-rest_dir-ap_sbref.nii.gz
  sub-01_task-rest_dir-pa_sbref.json
  sub-01_task-rest_dir-pa_sbref.nii.gz
```

would have the following metadata:

sub-01_task-rest_bold.json:
```JSON
{
  "B0FieldSource": "sbrefs_fmap",
  "PhaseEncodingDirection": "j",
  "RepetitionTime": 2.0,
  "TotalReadoutTime": 0.05
}
```

sub-01_task-rest_dir-ap_sbref.json
```JSON
{
  "B0FieldIdentifier": "sbrefs_fmap",
  "PhaseEncodingDirection": "j",
  "TotalReadoutTime": 0.05
}
```

and sub-01_task-rest_dir-pa_sbref.json
```JSON
{
  "B0FieldIdentifier": "sbrefs_fmap",
  "TotalReadoutTime": 0.05
}
```

### A currently supported example:

```Text
fmap/
  sub-01_phasediff.json
  sub-01_phasediff.nii.gz
  sub-01_magnitude1.nii.gz
func/
  sub-01_task-rest_bold.json
  sub-01_task-rest_bold.nii.gz
```

would have the following metadata:

sub-01_task-rest_bold.json:
```JSON
{
  "B0FieldSource": "phdiff_fmap",
  "PhaseEncodingDirection": "j",
  "RepetitionTime": 2.0,
  "TotalReadoutTime": 0.05
}
```

sub-01_phasediff.json
```JSON
{
  "B0FieldIdentifier": "phdiff_fmap",
  "EffectiveEchoSpacing": "0.00246",
  "IntendedFor": "func/sub-01_task-rest_bold.nii.gz"
}
```


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -70,8 +70,11 @@ reflected in a single DICOM tag for all possible aforementioned scan
 manipulations). See [here](https://lcni.uoregon.edu/kb-articles/kb-0003) and
 [here](https://github.com/neurolabusc/dcm_qa/tree/master/In/TotalReadoutTime)
 
-<sup>3</sup>We use the "FSL definition", i.e, the time between the center of the
```
oesteban, line 73: this was mentioned again under fieldmaps. I've moved that mention (after applying some improvements to the language) over here for completeness.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -104,6 +107,24 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, etc.
 | ----------------------------- | --------------------- | ------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | AnatomicalLandmarkCoordinates | RECOMMENDED           | [object][] of [arrays][] | Key:value pairs of any number of additional anatomical landmarks and their coordinates in voxel units (where first voxel has index 0,0,0) relative to the associated anatomical MRI, (e.g. `{"AC": [127,119,149], "PC": [128,93,141], "IH": [131,114,206]}, or {"NAS": [127,213,139], "LPA": [52,113,96], "RPA": [202,113,91]}`). |
 
+### Echo-Planar Imaging & B0 field-mapping
```
oesteban, line 110: this is the core of the proposal

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -440,45 +462,36 @@ JSON example:
 ```JSON
 {
-  "TotalReadoutTime": 0.095
+  "TotalReadoutTime": 0.095,
+  "B0FieldSource": ["phasediff_fmap0", "pepolar_fmap0"]
 }
 ```
 
 ## Fieldmap data
```
oesteban, line 470: I have refactored this section, as it clearly had different formatting. Now it should be more consistent with the above, and the coexistence of the new `B0FieldIdentifier` and `IntendedFor` should not flow unnaturally.

In terms of the spec, this is a refactor (i.e., no semantics about how to encode fieldmaps unrelated to `B0FieldIdentifier` should be found)

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -440,45 +462,36 @@ JSON example:
 ```JSON
 {
-  "TotalReadoutTime": 0.095
+  "TotalReadoutTime": 0.095,
+  "B0FieldSource": ["phasediff_fmap0", "pepolar_fmap0"]
```
tsalo, line 466: This should be a relative-ish path with the same format as `IntendedFor`, right? Can you replace these with realistic filenames?

tsalo: I misunderstood this proposal. Never mind!

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -104,6 +107,24 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, etc.
...
+| B0FieldIdentifier           | OPTIONAL              | [string][]                            | The presence of this key states that this particular 3D or 4D image may be used for fieldmap estimation purposes. The `B0FieldIdentifier` MUST be a unique string within one participant's tree, shared only by the images meant to be used as inputs for the estimation of a particular instance of the *B0 field* estimation.                                                                                                                     |
+| B0FieldSource               | OPTIONAL              | [string][] or [array][] of [string][] | At least one existing `B0FieldIdentifier` defined by other images in the participant's tree. This field states the *B0 field* estimation designated by the `B0FieldIdentifier` that may be used to correct the dataset for distortions caused by B0 inhomogeneities. `B0FieldSource` and `B0FieldIdentifier` are mutually exclusive. When a list of identifiers is given, the estimated fieldmap may be averaged across the different sources.      |
```
tsalo, line 126: Why use unique identifiers instead of paths to filenames, as in `IntendedFor`? I originally assumed that the identifier also indicated the type of B0 field map, but that doesn't seem to be the case. Does an identifier add anything over the filename?

oesteban: The key to these identifiers is that they allow the definition of higher-level data constructs.

> Does an identifier add anything over the filename?

See https://github.com/bids-standard/bids-specification/pull/622/files#diff-d668661006216379ab2a5c5ff7061b8aR384 

You could theoretically replace the two identifiers with lists-of-lists:
```
{
  "B0FieldSource": ["phasediff_fmap0", "pepolar_fmap0"]
}
```

But I think that is not:
- Readable
- Easy to encode
- Granular enough: the ID "phasediff_fmap0" defines much better how the estimation of the fieldmap must be done given a subset of the data having that `B0FieldIdentifier` (even more so for the pepolar fieldmap).

oesteban: Easy to encode as in separating concepts:
- What data should I use to estimate a particular fieldmap? and
- what of the estimated fieldmaps available in this dataset can/should be used to correct this particular image?

the ID creates more clear interface between producers and consumers.

oesteban: Another plus: IDs/tags are more reliable against missing data.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -577,35 +602,87 @@ sub-<label>/[ses-<label>/]
...
+
+or:
+
+```JSON
+{
+  "Units": "rad/s",
+  "B0FieldIdentifier": "se_fmap0",
+  "IntendedFor": [
+    "ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
+    "ses-post/func/sub-01_ses-post_task-motor_run-1_bold.nii.gz"
```
arokem, line 682: Just to see that I get this: In this case, there will be two separate files stored in the `fmap` folder, one with `sub-01_ses-pre` and one with `sub-01_ses-post` that are the field map files. What could the full filenames look like in this  case?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: That makes sense. I can see the utility of identifiers in that case, although I guess one could still use a list of lists to point to filenames. Thank you for the clarification.

tsalo: Two things I think are worth considering:

1. `B0FieldIdentifier` could be replaced by a column in the scans.tsv file (e.g., `phasediff_fmap0`) with a 0 for scans that aren't part of the group and a 1 for scans that are part of the group. I don't know if anyone else likes this idea, but it allows data providers to define arbitrary sets of files for their own purposes.
2. I don't think `B0FieldIdentifier` and `B0FieldSource` should be mutually exclusive. I'm thinking specifically of HCP style acquisitions, where you get separate runs with opposite encoding directions, so you can apply PEPOLAR-style unwarping to them. In that case the functional scan is part of its own B0 field source (along with the other run with the opposite direction) and it needs to have an identifier to that effect.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: hehehe, the PR needs to be much clearer...

So these two files are part of the `IntendedFor` field, which is what we already have in BIDS and says: the NIfTI to which this JSON file is associated is the fieldmap source for correcting these two other `_bold` files.

oesteban: Maybe the proposal is more readable here: https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#expressing-the-mr-protocol-intent-for-fieldmaps

oesteban:
Clean diff with #651 - https://github.com/oesteban/bids-specification/compare/rf/mri-fieldmaps...oesteban:enh/fieldmaps-refactor

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +730,66 @@ sub-<label>/[ses-<label>/]
...
+### Expressing the MR protocol intent for fieldmaps
+
+#### Using `B0FieldIdentifier` metadata
+
+The general purpose `B0FieldIdentifier` MRI metadata is RECOMMENDED for the
+prescription of the B<sub>0</sub> field estimation intent of the original acquisition
+protocol.
+This encoding fully overrides the `IntendedFor` approach (see below).
+However, the `IntendedFor` encoding is permitted and RECOMMENDED in combination
+with the `B0FieldIdentifier` to maintain compatibility with outdated software.
```
effigies, line 769: I think of "override" as more like "replaces", but we're recommending duplication. What about:

```suggestion
`B0FieldIdentifier` and `B0FieldSource` duplicate the capabilities of
the original `IntendedFor` approach while permitting more complex use cases.
It is RECOMMENDED to use both approaches to maintain compatibility with
tools that support older datasets.
```

effigies (commented):
Overall, looks good. I think the approach of maintaining both encodings is a good idea.

src/02-common-principles.md
```diff
@@ -149,6 +149,11 @@ data from subject `01`.
 The suffix `eeg` and the extension `.edf` depend on the imaging modality and
 the data format and indicate further details of the file's contents.
 
+Key-value pairs SHOULD NOT be used to encode scanning parameters of the
+corresponding data.
```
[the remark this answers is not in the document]
oesteban: Because a large part of these changes are just a refactor of current specification about fieldmaps, I stripped those changes out to #651, so that the actual changes of this new feature were more clear in a diff.

In particular, regarding this particular change:

- This message has been up-ranked to common principles: https://github.com/bids-standard/bids-specification/pull/651/files#diff-78a960934ce25feed2c29612b842d25131d0360556c2afc4dcaa9b60aceeb2baR152-R156
- A note has been at this point, linking to the common principles: https://github.com/bids-standard/bids-specification/pull/651/files#diff-e8c70f785c264706f1f453d744d53e74d7d4b77a5bbe25dc836adf6b96b36d50R719-R721

I believe it would be best following up this conversation in #651 (I'll rebase this from time to time).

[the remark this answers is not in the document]
oesteban: This is not to say I disagree with your suggestion, please let me know any changes you would suggest along what you pointed out, but in the context of the refactor PR.

src/02-common-principles.md
```diff
@@ -149,6 +149,11 @@ data from subject `01`.
...
+For instance, in echo-planar imaging MRI, the [`dir-<label>`](./99-appendices/09-entities.md#dir)
+MAY NOT be used as a replacement for the `PhaseEncodingDirection` metadata.
```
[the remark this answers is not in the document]
oesteban: I thought MAY NOT was standard - check https://github.com/bids-standard/bids-specification/pull/651/commits/ea830e5eea26af158a89b49e86f3627828c18a19 (under #651).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -107,6 +107,24 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, and so on.
...
+| B0FieldIdentifier           | OPTIONAL              | [string][]                            | The presence of this key states that this particular 3D or 4D image MAY be used for fieldmap estimation purposes. The `B0FieldIdentifier` MUST be a unique string within one participant's tree, shared only by the images meant to be used as inputs for the estimation of a particular instance of the *B<sub>0</sub> field* estimation.                                                                                                                           |
+| B0FieldSource               | OPTIONAL              | [string][] or [array][] of [string][] | At least one existing `B0FieldIdentifier` defined by other images in the participant's tree. This field states the *B<sub>0</sub> field* estimation designated by the `B0FieldIdentifier` that may be used to correct the dataset for distortions caused by B<sub>0</sub> inhomogeneities. `B0FieldSource` and `B0FieldIdentifier` are mutually exclusive. When a list of identifiers is given, the estimated fieldmap may be averaged across the different sources. |
```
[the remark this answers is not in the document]
oesteban: It is probably not clear in the current specs, but if two files share a `B0FieldIdentifier` that also means they can be corrected with the fieldmap derived from themselves, if that makes any sense (it makes sense, for instance, with EPIs and certain phasediff maps acquired with EPIs). Perhaps this is too implicit in the spec.

For the same reason you proposed to remove the averaging suggestion, it does not mean you can correct these images for distortions with other fieldmaps - that's a tooling decision (although a prior seems to make little sense). This is just to say that, if an image with fieldmapping potential has both "B0FieldSource" and "B0FieldIdentifier" then it is not compliant.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -561,45 +580,44 @@ JSON example:
...
 
-```JSON
-{
-   "IntendedFor": ["ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
-                   "ses-post/func/sub-01_ses-post_task-motor_run-1_bold.nii.gz"]
-}
-```
+| **Name**         | `suffix`         | **Description**                                                                                                                                                                                                      |
+| ---------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Magnitude        | magnitude\[1,2\] | Field-mapping MR schemes such as gradient-recalled echo (GRE) generate a Magnitude image to be used for anatomical reference. Requires the existence of Phase, Phase-difference or Fieldmap maps as described below. |
```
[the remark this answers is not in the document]
oesteban: There is a reason for this:

For Magnitude, you can have: `magnitude`, or `magnitude1` and not always mandatory `magnitude2`.

For phases you have to have `phase1` and `phase2` always (and there's no `phase`).

I'll add your suggestion for to remove the "as described below", but I'm not sure of how we should write the magnitude suffix though.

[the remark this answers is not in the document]
oesteban: (actually, moving your suggestion to #651).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -561,45 +580,44 @@ JSON example:
...
+ 1. [Phase-difference map](#case-1-phase-difference-map-and-at-least-one-magnitude-image)
+ 1. [Two phase maps](#case-2-two-phase-maps-and-two-magnitude-images)
+ 1. [Direct *field mapping*](#case-3-direct-field-mapping)
+ 1. ["*PEpolar*" fieldmaps](#case-4-multiple-phase-encoded-directions-pepolar)
 
-The IntendedFor field may contain one or more filenames with paths relative to
-the subject subfolder. The path needs to use forward slashes instead of backward
-slashes. Here's an example with multiple target scans:
+These four different types of field mapping strategies can be encoded
+as described [below](#types-of-fieldmaps), using the following image types:
```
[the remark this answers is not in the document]
oesteban: Moved to #651.

oesteban (commented):
Some stuff moved to #651

src/99-appendices/09-entities.md
```diff
@@ -103,8 +103,8 @@ Full name: Phase-Encoding Direction
 
 Format: `dir-<label>`
 
-Definition: The `dir-<label>` key/value can be used to distinguish
-different phase-encoding directions.
+Definition: The `dir-<label>` key/value can be set to an arbitrary alphanumeric label
+(for example, `dir-LR` or `dir-AP`) to distinguish different phase-encoding directions.
```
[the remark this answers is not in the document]
oesteban: #651.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +730,66 @@ sub-<label>/[ses-<label>/]
...
+| IntendedFor  | RECOMMENDED           | [string][] or [array][] of [string][] | Contains one or more filenames with paths relative to the participant subfolder. The path needs to use forward slashes instead of backward slashes. This field is OPTIONAL, and in case the fieldmaps do not correspond to any particular scans it does not have to be filled. |
+
+For example:
+
+```JSON
+{
+   "IntendedFor": [
+   	    "ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
+        "ses-pre/func/sub-01_ses-pre_task-motor_run-2_bold.nii.gz"
+    ]
```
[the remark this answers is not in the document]
oesteban: (#651)

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +730,66 @@ sub-<label>/[ses-<label>/]
...
+acquired for by filling the `IntendedFor` field in the corresponding JSON file.
+
+| **Key name** | **Requirement level** | **Data type**                         | **Description**                                                                                                                                                                                                                                                                |
+| ------------ | --------------------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| IntendedFor  | RECOMMENDED           | [string][] or [array][] of [string][] | Contains one or more filenames with paths relative to the participant subfolder. The path needs to use forward slashes instead of backward slashes. This field is OPTIONAL, and in case the fieldmaps do not correspond to any particular scans it does not have to be filled. |
```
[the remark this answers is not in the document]
oesteban: Added manually to latest co-authored commit 👍

effigies (approved):
LGTM.

effigies:
This needs another approval or critique... Pinging @arokem and @tsalo for another pass if you have some time.

More generally, this affects @bids-standard/raw-mri.

oesteban:
I have added a recommendation to derive `B0FieldIdentifier` from the Protocol Name DICOM tag of fieldmaps. Pinging @neurolabusc to ask about the appropriateness of the suggestion, as well as @mgxd / @yarikoptic (or anyone involved in Heudiconv) regarding whether that would make encoding easier.

neurolabusc:
1. Be aware that for GE MRI the user can not set the *Protocol Name* `(0018,1030)`. This explains why dcm2niix will substitute SequenceName (0018,0024) for GE MRI when the user [requests a protocol name in the filename](https://github.com/rordenlab/dcm2niix/blob/master/FILENAMING.md).
2. @drmclem  notes that for Philips the default B0 mapping protocols for fMRI produce a greyscale field map [calibrated in Hz](https://github.com/rordenlab/dcm2niix/issues/455). If anyone has Philips DICOMs data with these field maps, it would be great to know how we can detect this. A perfect dataset would be a Philips DICOM where the scanner is set to save the calibrated unwrapped field map for one series, and another identical series where the scanner saves the real/imaginary pairs. I do not have access to Philips hardware, so I can not resolve this without community help.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -631,7 +688,8 @@ For example:
 ```JSON
 {
    "EchoTime1": 0.00600,
-   "EchoTime2": 0.00746
+   "EchoTime2": 0.00746,
+   "B0FieldIdentifier": "phasediff_fmap0"
```
tsalo, line 692: I noticed that the example `B0FieldIdentifier` values in this PR are fairly standardized. @oesteban do you want a specific vocabulary to be put in place?

tsalo (commented):
This is a useful proposal, although it muddies the waters a bit until we can drop `IntendedFor` in 2.0. It also expands a new-ish approach to grouping files via metadata fields. I just have one question about how the fields should be named.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -737,7 +797,8 @@ For example:
 {
    "PhaseEncodingDirection": "j-",
    "TotalReadoutTime": 0.095,
-   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
+   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz",
+   "B0FieldIdentifier": "pepolar_fmap0"
```
sebastientourbier, line 801: In link with comment by @tsalo, it seems it used as prefix in this PR the fieldmap type as defined in section Fieldmap data (current line 607), so should not we use of `epi_fmap0` instead of `pepolar_fmap0`?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -107,6 +107,24 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, and so on.
...
+| --------------------------- | --------------------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| B0FieldIdentifier           | OPTIONAL              | [string][]                            | The presence of this key states that this particular 3D or 4D image MAY be used for fieldmap estimation purposes. The `B0FieldIdentifier` MUST be a unique string within one participant's tree, shared only by the images meant to be used as inputs for the estimation of a particular instance of the *B<sub>0</sub> field* estimation. It is RECOMMENDED to derive this identifier from DICOM tags, for example, the *Protocol Name* `(0018,1030)`.             |
```
tsalo, line 125: Something I forgot in my review- can't one image be a component of _multiple_ field map sets? Should this allow a list of strings? The same goes for `B0FieldSource`.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: I cannot think of a reason to establish a vocabulary for this. I think giving recommendations could be a possibility (along these lines I suggested somewhere using some DICOM field for inspiration - I should go back to that).

For instance, it could be convenient to avoid identifiers that would not be valid BIDS labels down the line (you can imagine a workflow making this field an entity inside BIDS-Derivatives). And yet, you can just strip disallowed chars out if you were to do that. So perhaps, do not use invalid BIDS-label chars for differentiating ids? E.g., one fieldmap set having `phasediff0` and another fieldmap set with `phasediff_0` (or maybe worse, `phasediff0_`). And still this feels a bit twisted for naming.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: Sure, the idea for these identifiers is that they must be unique within-subject. There's really no difference between `epi_fmap0` and `pepolar_fmap0`. However `epi_fmap0` could confuse some user that a valid raw suffix is necessary. I'm going to change that in line 692.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: Currently `B0FieldSource` admits a list of identifiers.

For `B0FieldIdentifier` - I don't think that's a common practice. By cases:

- Phases/phasediff/fieldmap - What would that mean? These are separate acquisitions so, from the workflow perspective, the only thing I would interpret is that the author want me to average them. You can achieve the same result by setting a list of identifiers to average in `B0FieldSource`.
- EPI (blip up/down) - Here it could have some consequence. But is there a reason to do this? Maybe if you want to fish for the best-looking correction exploring all the combinations possible of EPIs + DWI/BOLD/SBREF runs? Would you really need that from the encoder of a BIDS dataset? I can't really think of a use-case, but happy to be corrected.

oesteban:
Resuscitating this PR and attempting to bring it to the finish line.

> * Be aware that for GE MRI the user can not set the _Protocol Name_ `(0018,1030)`. This explains why dcm2niix will substitute SequenceName (0018,0024) for GE MRI when the user [requests a protocol name in the filename](https://github.com/rordenlab/dcm2niix/blob/master/FILENAMING.md).

Thanks for the tip! I've added Sequence Name as a potential alternative (5be9e5f)

> * @drmclem  notes that for Philips the default B0 mapping protocols for fMRI produce a greyscale field map [calibrated in Hz](https://github.com/rordenlab/dcm2niix/issues/455). If anyone has Philips DICOMs data with these field maps, it would be great to know how we can detect this. A perfect dataset would be a Philips DICOM where the scanner is set to save the calibrated unwrapped field map for one series, and another identical series where the scanner saves the real/imaginary pairs. I do not have access to Philips hardware, so I can not resolve this without community help.

I could be wrong, but I feel this type of fieldmaps from Philips should be encoded as instances of ["direct *field-mapping*"](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping). Happy to learn more, let me know if there's anything that is specific to this PR that interacts with those fieldmaps or whether we should open an issue for this Philips' artifacts.

@tsalo - Are my replies convincing?
- https://github.com/bids-standard/bids-specification/pull/622#discussion_r567053477
- https://github.com/bids-standard/bids-specification/pull/622#discussion_r567048675

@sebastientourbier I think I made a change as requested - please confirm it was satisfactory.

Feedback from HeudiConv gurus very much appreciated (@yarikoptic, @mgxd, feel free to cc)

Begging for green checkmarks that permit this PR to be merged - @arokem, @francopestilli, @Lestropie, @mattcieslak, @satra, @edickie, @bids-standard/raw-mri-func @bids-standard/raw-mri-dwi.

tsalo:
> @tsalo - Are my replies convincing?

They are, thank you. Most of my concern was with the introduction of a new grouping mechanism, but with #624 that mechanism is now part of the specification, so the feature itself makes a lot of sense. I will try to review again soon with an eye more toward wording.

drmclem:
Hi

Respinding to the above - I had some discussions over on dcm2niix on this issue - the B0 mapping sequence is a specific protocol option and produces the scan with the calculated maps and the source images. Note that B0 mapping has only become product recently and so in in prior software releases this was a WIP sequence and so the DICOM output for that may not be the same.

For the product version the maps will be 
0008,0008 ORIGINAL\PRIMARY\B0 MAP\B0\UNSPECIFIED
And source 
0008,0008 ORIGINAL\PRIMARY\M_FFE\M\FFE

In addition the scaling will specify RescaleType (0028,1054) [Hz] (or in the Dicom RealWorldUnit block).

Matthew

neurolabusc:
As @drmclem notes the latest release of dcm2niix [v1.0.20210317](https://github.com/rordenlab/dcm2niix/releases/tag/v1.0.20210317) will detect field maps that Philips estimated in [Hz](https://github.com/rordenlab/dcm2niix/issues/455). As @oesteban notes, this is an example of [direct field mapping](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping). dcm2niix gives these images the `_fieldmaphz` post fix the BIDS `ImageType` field reports `FIELDMAPHZ`. Here is an [example](https://github.com/neurolabusc/dcm_qa_enh/tree/master/Ref).

I have two questions:
 1. In future, dcm2niix should also generate the `"Units": "Hz",` tag for these [direct field maps](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping).
 2. The BIDS specification notes that GE scanners can create [direct field maps](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping). However, I have never seen an exemplar. Would it be possible for someone with access to a GE scanner to provide an example. Perhaps @mr-jaemin

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -121,6 +121,24 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, and so on.
...
+
+Echo-Planar Imaging (EPI) schemes typically used in the acquisition of
+diffusion and functional MRI may also be *intended for* estimating the
+*B<sub>0</sub>* field nonuniformity inside the scanner (in other words,
+*mapping the field*) without the acquisition of additional MRI schemes
+such as gradient-recalled echo (GRE) sequences that are stored under the
+`fmap/` folder of the BIDS structure.
+
+The modality labels `dwi` (under `dwi/`), `bold`, and `sbref` (under `func/`)
+and any modality under `fmap/` are allowed to encode the MR protocol intent for
```
mattcieslak, line 134: Should also include `asl` under `perf/`

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -828,6 +848,43 @@ are allowed across all the four scenarios:
 -   The OPTIONAL [`acq-<label>`](../99-appendices/09-entities.md#acq) key/value pair corresponds to a custom label
     the user may use to distinguish different set of parameters.
 
+### Expressing the MR protocol intent for fieldmaps
+
+Fieldmaps are typically acquired with the purpose of correcting one or more EPI
+scans under `func/` or `dwi/` for distortions derived from *B<sub>0</sub>*
```
mattcieslak, line 854: `perf/` here too

mattcieslak (approved):
Looks great, very excited to transition to this system.

oesteban:
> * In future, dcm2niix should also generate the `"Units": "Hz",` tag for these [direct field maps](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping).

I think this would be what the specs intend at this point.

> * The BIDS specification notes that GE scanners can create [direct field maps](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping). However, I have never seen an exemplar. Would it be possible for someone with access to a GE scanner to provide an example. Perhaps @mr-jaemin

Yes, they use spiral echo imaging - I can provide examples of this (or maybe find them in some dataset up in openneuro).

neurolabusc:
@oesteban Jaemin and I have added [support for direct fieldmaps](https://github.com/rordenlab/dcm2niix/issues/501) to dcm2niix (v1.0.20210410 from development branch). We also provide [GE and Philips validation datasets](https://github.com/neurolabusc/dcm_qa_fmap).
 - Since we do not have GE or Philips equipment at my center, I would be grateful if users of these systems could test the development release of dcm2niix to ensure it does not fail in edge cases. 
 - I think the BIDS standard should be internally consistent. The draft for direct field mapping introduces the field [Units](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping), but BEP009 uses [Unit](https://bids-specification.readthedocs.io/en/bep-009/04-modality-specific-files/09-positron-emission-tomography.html) for the same DICOM field. @effigies perhaps you have thoughts on this.
 - @oesteban with regards to my prior comments regarding GE and Protocol Name (0018,1030), note that the behavior of [v1.0.20210317](https://github.com/rordenlab/dcm2niix/releases) and later have changed, so GE is not treated as special:  `%d` = description (from 0008,103E),  `%p` = protocol name (0018,1030), `-z` = SequenceName (0018,0024). However, GE MR users should be aware that users may not have flexibility to set protocol name on the scanner console, to sequence name may be a more meaningful attribute.

effigies:
> * I think the BIDS standard should be internally consistent. The draft for direct field mapping introduces the field [Units](https://bids-specification--622.org.readthedocs.build/en/622/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#case-3-direct-field-mapping), but BEP009 uses [Unit](https://bids-specification.readthedocs.io/en/bep-009/04-modality-specific-files/09-positron-emission-tomography.html) for the same DICOM field. @effigies perhaps you have thoughts on this.

Good catch. It looks like `Units` in fieldmaps pre-dates `Unit`, and we haven't released PET, so we have a chance to fix without breaking backwards compatiblity.

oesteban:
Are any code owners willing to leave a green tick to get this merged? We will need two more (and/or give Matt rights, how does that process work?)

@bids-standard/raw-mri-dwi @bids-standard/raw-mri

francopestilli:
✅

tsalo (approved):
Everything looks good to me.

tsalo:
This PR has the necessary approvals at this point, I think, but since we're in the embargo period before a release, we can't merge this until Friday, at least.

effigies:
@oesteban Can you check that merge? I think I may have done it backwards.

oesteban:
The merge looks good, with the latest additions to `master` my latest commit was outdated. I quickly checked and I don't see any difference w.r.t. before the merge.

Should I just merge?

effigies:
@oesteban I think I restored something you intended to remove. I'm going to re-do the merge.

effigies:
Here's the change that I accidentally dropped: https://github.com/bids-standard/bids-specification/compare/b0312f8a6d1daa2d3683bc45bbd9e9f65d232c43..8d862e79d42390094e8cfa3ca75cedba2d4df5a2

PR #624, opened 2020-09-28
[ENH] Add "multipart DWI" acquisitions and refactor DWI specifications

oesteban:
Addresses the use case of DWI sequences that need to be split into several runs because, e.g., the scanner would exceed temperature specifications due to fast gradient-switching.

This PR also refactors the text of DWIs to better accommodate the addition and make DWIs more consistent with the rest of the file.

The approach to solve this is similar to that of #622 for fieldmaps.

References: nipreps/dmriprep#43.

### DIFF
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -378,10 +378,10 @@ participant, task and run takes precedence.
 
 Currently supported image modalities include:
 
-| **Name**              | `modality_label` | **Description**                                                                                                                                     |
-|---------------------- | ---------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
-| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting)                                                                                     |
-| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images, only when no gradient informations needs to be stored (i.e., all volumes are *b = 0*) |
+| **Name**              | `modality_label` | **Description**                                                                                                                                       |
+|---------------------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting).                                                                                     |
+| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images, only when no gradient informations needs to be stored (i.e., all volumes are *b = 0*). |
 
 Template:
 
@@ -450,7 +450,11 @@ bval example:
 and thus define those values for all sessions and/or subjects in one place (see
 Inheritance principle).
 
-## Multipart (split) DWI schemes
+As an exception to the [common principle](../02-common-principles.md#definitions)
+that parameters are constant across runs, the gradient table information (stored
+within the `.bval` and `.bvec` files) MAY change across DWI runs.
+
+### Multipart (split) DWI schemes
 
 Some MR schemes cannot be acquired directly by some scanner devices,
 requiring to generate several DWI runs that were originally meant to belong
@@ -459,9 +463,9 @@ For instance, some GE scanners cannot collect more than &asymp;160 volumes
 in a single run under fast-changing gradients, so acquiring *HCP-style*
 diffusion images will require splitting the DWI scheme in several runs.
 
-| **Key name**    | **Requirement level** | **Data type** | **Description**                                                                                                                                                                               |
-| --------------- | --------------------- | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| MultipartID     | REQUIRED              | [string][]    | Unique identifier to prescribe DWI segments belonging in a single multipart run.                                                                                                              |
+| **Key name**    | **Requirement level** | **Data type** | **Description**                                                                                                                                                                                                      |
+| --------------- | --------------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| MultipartID     | REQUIRED              | [string][]    | Unique identifier (participant's level) to prescribe DWI segments belonging in a single dataset. Segments can be designated by any combination of `dir-<label>`, `acq-<label>` and/or `run-<index>` key/value pairs. |
 
 JSON example:
 
@@ -495,7 +499,7 @@ sub-<label>/[ses-<label>/]
        sub-<label>_dwi.json
 ```
 
-## Other RECOMMENDED metadata
+### Other RECOMMENDED metadata
 
 See [Common metadata fields](#common-metadata-fields) for a list of
 additional terms that can be included in the corresponding JSON file.
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -376,6 +376,13 @@ participant, task and run takes precedence.
 
 ## Diffusion imaging data
 
+Currently supported image modalities include:
+
+| **Name**              | `modality_label` | **Description**                                                                                                                                     |
+|---------------------- | ---------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting)                                                                                     |
+| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images, only when no gradient informations needs to be stored (i.e., all volumes are *b = 0*) |
+
 Template:
 
 ```Text
@@ -389,28 +396,36 @@ sub-<label>/[ses-<label>/]
        sub-<label>[_ses-<label>][_acq-<label>][_dir-<label>][_run-<index>]_sbref.json
 ```
 
-Diffusion-weighted imaging data acquired for that participant. The OPTIONAL
-[`acq-<label>`](../99-appendices/09-entities.md#acq)
+Diffusion-weighted imaging data acquired for that participant.
+
+The OPTIONAL [`acq-<label>`](../99-appendices/09-entities.md#acq)
 key/value pair corresponds to a custom label the user may use to
-distinguish different set of parameters. For example this should be used when a
-study includes two diffusion images - one single band and one multiband. In such
-case two files could have the following names:
+distinguish different sets of parameters.
+For example, this should be used when a study includes two diffusion
+images -- one single band and one multiband.
+In such case two files could have the following names:
 `sub-01_acq-singleband_dwi.nii.gz` and `sub-01_acq-multiband_dwi.nii.gz`,
 however the user is free to choose any other label than `singleband` and
-`multiband` as long as they are consistent across subjects and sessions. For
-multiband acquisitions, one can also save the single-band reference image as
-type `sbref` (for example, `dwi/sub-control01_sbref.nii[.gz]`) The bvec and bval files
-are in the [FSL format](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FDT/UserGuide#DTIFIT):
+`multiband` as long as they are consistent across subjects and sessions.
+
+For multiband acquisitions, one can also save the single-band reference image as
+type `sbref` (e.g. `dwi/sub-control01_sbref.nii[.gz]`.)
+
+
+### REQUIRED gradient orientation information
+
+The bvec and bval files are in the
+[FSL format](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FDT/UserGuide#DTIFIT).
 The bvec files contain 3 rows with n space-delimited floating-point numbers
-(corresponding to the n volumes in the relevant NIfTI file). The first row
-contains the x elements, the second row contains the y elements and third row
-contains the z elements of a unit vector in the direction of the applied
-diffusion gradient, where the i-th elements in each row correspond together to
-the i-th volume with `[0,0,0]` for non-diffusion-weighted volumes. Inherent to
-the FSL format for bvec specification is the fact that the coordinate system of
-the bvecs is with respect to the participant, defined by the axes of the
-corresponding `dwi.nii` file, and not the magnet's coordinate system.
-Thus, any rotations applied to `dwi.nii` also need to be applied to the
+(corresponding to the n volumes in the relevant NIfTI file).
+The first row contains the *x* elements, the second row contains the *y* elements and
+the third row contains the *z* elements of a unit vector in the direction of the applied
+diffusion gradient, where the *i*-th elements in each row correspond together to
+the *i*-th volume, with `[0,0,0]` for non-diffusion-weighted volumes.
+Inherent to the FSL format for bvec specification is the fact that the coordinate system of
+the bvecs is with respect to the participant (i.e., defined by the axes of the
+corresponding dwi.nii file) and not the magnet's coordinate system, which means
+that any rotations applied to dwi.nii also need to be applied to the
 corresponding bvec file.
 
 bvec example:
@@ -434,6 +449,53 @@ bval example:
 `.bval` and `.bvec` files can be saved on any level of the directory structure
 and thus define those values for all sessions and/or subjects in one place (see
 Inheritance principle).
+
+## Multipart (split) DWI schemes
+
+Some MR schemes cannot be acquired directly by some scanner devices,
+requiring to generate several DWI runs that were originally meant to belong
+in a single one.
+For instance, some GE scanners cannot collect more than &asymp;160 volumes
+in a single run under fast-changing gradients, so acquiring *HCP-style*
+diffusion images will require splitting the DWI scheme in several runs.
+
+| **Key name**    | **Requirement level** | **Data type** | **Description**                                                                                                                                                                               |
+| --------------- | --------------------- | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| MultipartID     | REQUIRED              | [string][]    | Unique identifier to prescribe DWI segments belonging in a single multipart run.                                                                                                              |
+
+JSON example:
+
+```JSON
+{
+  "PhaseEncodingDirection": "j-",
+  "TotalReadoutTime": 0.095,
+  "MultipartID": "dwi_protocol1"
+}
+```
+
+For instance, the above JSON sidecar defined for a set of files via the
+`sub-<label>_dwi.json` file would indicate that all the runs 1 through 4
+belong in the same unique image:
+
+```Text
+sub-<label>/[ses-<label>/]
+    dwi/
+       sub-<label>_run-1_dwi.nii.gz
+       sub-<label>_run-1_dwi.bval
+       sub-<label>_run-1_dwi.bvec
+       sub-<label>_run-2_dwi.nii.gz
+       sub-<label>_run-2_dwi.bval
+       sub-<label>_run-2_dwi.bvec
+       sub-<label>_run-3_dwi.nii.gz
+       sub-<label>_run-3_dwi.bval
+       sub-<label>_run-3_dwi.bvec
+       sub-<label>_run-4_dwi.nii.gz
+       sub-<label>_run-4_dwi.bval
+       sub-<label>_run-4_dwi.bvec
+       sub-<label>_dwi.json
+```
+
+## Other RECOMMENDED metadata
 
 See [Common metadata fields](#common-metadata-fields) for a list of
 additional terms that can be included in the corresponding JSON file.
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -548,6 +548,9 @@ sub-<label>/[ses-<label>/]             # MultipartID
 
 ### Other RECOMMENDED metadata
 
+At minimum, the `PhaseEncodingDirection` and `TotalReadoutTime` metadata
+fields are RECOMMENDED to enable the correction of geometrical distortions
+with [fieldmap information](#fieldmap-data).
 See [Common metadata fields](#common-metadata-fields) for a list of
 additional terms that can be included in the corresponding JSON file.
 
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -415,7 +415,7 @@ distinguish different sets of phase-encoding directions.
 
 **Combining multi- and single-band acquisitions**.
 The single-band reference image MAY be stored with suffix `sbref` (for example,
-`dwi/sub-control01_sbref.nii[.gz]`,) as long as the image has no corresponding
+`dwi/sub-control01_sbref.nii[.gz]`) as long as the image has no corresponding
 [gradient information (`.bval` and `.bvec` sidecar files)](#required-gradient-orientation-information)
 to be stored.
 
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -413,11 +413,11 @@ The OPTIONAL [`dir-<label>`](../99-appendices/09-entities.md#dir)
 key/value pair corresponds to a custom label the user may use to
 distinguish different sets of phase-encoding directions.
 
-### Combining multi- and single-band acquisitions
-
-The single-band reference image MAY be stored as type `sbref` (e.g.,
+**Combining multi- and single-band acquisitions**.
+The single-band reference image MAY be stored with suffix `sbref` (e.g.,
 `dwi/sub-control01_sbref.nii[.gz]`,) as long as the image has no corresponding
-gradient information (`.bval` and `.bvec` sidecar files) to be stored.
+[gradient information (`.bval` and `.bvec` sidecar files)](#required-gradient-orientation-information)
+to be stored.
 
 Otherwise, if some gradient information is associated to the single-band diffusion
 image and a multi-band diffusion image also exists, the `acq-<label>` key/value pair
@@ -502,48 +502,48 @@ part of a unique multipart scan, then they will tag all four runs with the
 same `MultipartID` (shown at the right-hand side of the file listing):
 
 ```Text
-sub-<label>/[ses-<label>/]     # MultipartID
+sub-<label>/[ses-<label>/]         # MultipartID
   dwi/
-    sub-1_dir-AP_run-1.nii.gz  # dwi_1
-    sub-1_dir-AP_run-2.nii.gz  # dwi_1
-    sub-1_dir-PA_run-1.nii.gz  # dwi_1
-    sub-1_dir-PA_run-2.nii.gz  # dwi_1
+    sub-1_dir-AP_run-1_dwi.nii.gz  # dwi_1
+    sub-1_dir-AP_run-2_dwi.nii.gz  # dwi_1
+    sub-1_dir-PA_run-1_dwi.nii.gz  # dwi_1
+    sub-1_dir-PA_run-2_dwi.nii.gz  # dwi_1
 ```
 
 If, conversely, the resarcher wanted to store two multipart scans, one possibility
 is to combine matching phase-encoding directions:
 
 ```Text
-sub-<label>/[ses-<label>/]     # MultipartID
+sub-<label>/[ses-<label>/]         # MultipartID
   dwi/
-    sub-1_dir-AP_run-1.nii.gz  # dwi_1
-    sub-1_dir-AP_run-2.nii.gz  # dwi_1
-    sub-1_dir-PA_run-1.nii.gz  # dwi_2
-    sub-1_dir-PA_run-2.nii.gz  # dwi_2
+    sub-1_dir-AP_run-1_dwi.nii.gz  # dwi_1
+    sub-1_dir-AP_run-2_dwi.nii.gz  # dwi_1
+    sub-1_dir-PA_run-1_dwi.nii.gz  # dwi_2
+    sub-1_dir-PA_run-2_dwi.nii.gz  # dwi_2
 ```
 
 Alternatively, the researcher's intent could be combining opposed phase-encoding
 runs instead:
 
 ```Text
-sub-<label>/[ses-<label>/]     # MultipartID
+sub-<label>/[ses-<label>/]         # MultipartID
   dwi/
-    sub-1_dir-AP_run-1.nii.gz  # dwi_1
-    sub-1_dir-AP_run-2.nii.gz  # dwi_2
-    sub-1_dir-PA_run-1.nii.gz  # dwi_1
-    sub-1_dir-PA_run-2.nii.gz  # dwi_2
+    sub-1_dir-AP_run-1_dwi.nii.gz  # dwi_1
+    sub-1_dir-AP_run-2_dwi.nii.gz  # dwi_2
+    sub-1_dir-PA_run-1_dwi.nii.gz  # dwi_1
+    sub-1_dir-PA_run-2_dwi.nii.gz  # dwi_2
 ```
 
 The `MultipartID` metadata MAY be used with the
 [`acq-<label>`](../99-appendices/09-entities.md#acq) key/value pair, for example:
 
 ```Text
-sub-<label>/[ses-<label>/]         # MultipartID
+sub-<label>/[ses-<label>/]             # MultipartID
   dwi/
-    sub-1_acq-shell1_run-1.nii.gz  # dwi_1
-    sub-1_acq-shell1_run-2.nii.gz  # dwi_2
-    sub-1_acq-shell2_run-1.nii.gz  # dwi_1
-    sub-1_acq-shell2_run-2.nii.gz  # dwi_2
+    sub-1_acq-shell1_run-1_dwi.nii.gz  # dwi_1
+    sub-1_acq-shell1_run-2_dwi.nii.gz  # dwi_2
+    sub-1_acq-shell2_run-1_dwi.nii.gz  # dwi_1
+    sub-1_acq-shell2_run-2_dwi.nii.gz  # dwi_2
 ```
 
 ### Other RECOMMENDED metadata
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -376,12 +376,13 @@ participant, task and run takes precedence.
 
 ## Diffusion imaging data
 
+Diffusion-weighted imaging data acquired for a participant.
 Currently supported image modalities include:
 
-| **Name**              | `modality_label` | **Description**                                                                                                                                       |
-|---------------------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
-| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting).                                                                                     |
-| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images, only when no gradient informations needs to be stored (i.e., all volumes are *b = 0*). |
+| **Name**              | `modality_label` | **Description**                                                   |
+|---------------------- | ---------------- | ----------------------------------------------------------------- |
+| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting). |
+| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images.    |
 
 Template:
 
@@ -396,20 +397,23 @@ sub-<label>/[ses-<label>/]
        sub-<label>[_ses-<label>][_acq-<label>][_dir-<label>][_run-<index>]_sbref.json
 ```
 
-Diffusion-weighted imaging data acquired for that participant.
-
 The OPTIONAL [`acq-<label>`](../99-appendices/09-entities.md#acq)
 key/value pair corresponds to a custom label the user may use to
 distinguish different sets of parameters.
-For example, this should be used when a study includes two diffusion
-images -- one single band and one multiband.
+
+### Combining multi- and single-band acquisitions
+
+The single-band reference image MAY be stored as type `sbref` (e.g.,
+`dwi/sub-control01_sbref.nii[.gz]`,) as long as the image has no corresponding
+gradient information (`.bval` and `.bvec` sidecar files) to be stored.
+
+Otherwise, if some gradient information is associated to the single-band diffusion
+image and a multi-band diffusion image also exists,  the `acq-<label>` key/value pair
+MUST be used to distinguish both images.
 In such case two files could have the following names:
 `sub-01_acq-singleband_dwi.nii.gz` and `sub-01_acq-multiband_dwi.nii.gz`,
 however the user is free to choose any other label than `singleband` and
 `multiband` as long as they are consistent across subjects and sessions.
-
-For multiband acquisitions, one can also save the single-band reference image as
-type `sbref` (e.g. `dwi/sub-control01_sbref.nii[.gz]`.)
 
 ### REQUIRED gradient orientation information
 
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -411,7 +411,6 @@ however the user is free to choose any other label than `singleband` and
 For multiband acquisitions, one can also save the single-band reference image as
 type `sbref` (e.g. `dwi/sub-control01_sbref.nii[.gz]`.)
 
-
 ### REQUIRED gradient orientation information
 
 The bvec and bval files are in the
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -377,12 +377,12 @@ participant, task and run takes precedence.
 ## Diffusion imaging data
 
 Diffusion-weighted imaging data acquired for a participant.
-Currently supported image modalities include:
+Currently supported image types include:
 
-| **Name**              | `modality_label` | **Description**                                                   |
-|---------------------- | ---------------- | ----------------------------------------------------------------- |
-| DWI                   | dwi              | Diffusion-weighted imaging contrast (specialized T2\* weighting). |
-| Single-Band Reference | sbref            | Single-band reference for one or more multi-band `dwi` images.    |
+| **Name**              | `suffix` | **Description**                                                   |
+|---------------------- | -------- | ----------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -432,6 +447,53 @@ bval example:
 and thus define those values for all sessions and/or subjects in one place (see
 Inheritance principle).
 
+## Multipart (split) DWI schemes
+
+Some MR schemes cannot be acquired directly by some scanner devices,
+requiring to generate several DWI runs that were originally meant to belong
+in a single one.
```
tsalo, line 454: I'm not sure if we would want to do this, but would it make sense to use `split` in this case instead of `run`? `run` implies that the parameters are the same, so based on the current definition under **Common principles**, I think you'd want to use `acq` to distinguish the two files (e.g., `acq-first160dirs`, `acq-second160dirs`).

oesteban: At the moment, two `_dwi` runs with all other BIDS entities being equal can differ in there bvecs/bvals specifications. Therefore, you can perfectly have `run-1` and `run-2` encoding the first 160 and the second 160 orientations respectively.

I agree that whether bvecs/bvals should be constant across runs is rather unclear at the moment (and the validator does not check this particular for bvecs/bvals) - I can send a separate PR to clarify this. One thing is sure - there's a lot of data out there with several `run-` indices each with a different gradient table (and I don't think it would be wise more explicitly disallowing this).

That said, the idea for this PR is not to limit the utilization of the proposed new key (`MultipartID`) across runs, also across different acquisitions and/or directions. If the proposal is not clear as regards this, please let me know.

oesteban: @tsalo I have added https://github.com/bids-standard/bids-specification/pull/624/files#diff-d668661006216379ab2a5c5ff7061b8aR450-R452 to refer to this issue more explicitly in the gradient-table section

mattcieslak:
I really like the MultipartID idea! This is an elegant solution and will save some headaches for pipeline software. It will cause some headaches for data curators, though.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -386,25 +394,35 @@ sub-<label>/[ses-<label>/]
...
-contains the z elements of a unit vector in the direction of the applied
-diffusion gradient, where the i-th elements in each row correspond together to
-the i-th volume with `[0,0,0]` for non-diffusion-weighted volumes. Inherent to
-the FSL format for bvec specification is the fact that the coordinate system of
+(corresponding to the n volumes in the relevant NIfTI file).
+The first row contains the *x* elements, the second row contains the *y* elements and
+the third row contains the *z* elements of a unit vector in the direction of the applied
+diffusion gradient, where the *i*-th elements in each row correspond together to
+the *i*-th volume, with `[0,0,0]` for non-diffusion-weighted volumes.
+Inherent to the FSL format for bvec specification is the fact that the coordinate system of
```
mattcieslak, line 425: The FSL bvec system is a more complicated than this. It is only true [for some image orientations](https://www.mrtrix.org/2016/04/15/bug-in-fsl-bvecs-handling/). BIDS will need to decide whether they really want to be 100% FSL-compatible or stick to the image axis convention.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -432,6 +450,57 @@ bval example:
...
+| --------------- | --------------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| MultipartID     | REQUIRED              | [string][]    | Unique identifier (participant's level) to prescribe DWI segments belonging in a single dataset. Segments can be designated by any combination of `dir-<label>`, `acq-<label>` and/or `run-<index>` key/value pairs. |
+
+JSON example:
+
+```JSON
+{
+  "PhaseEncodingDirection": "j-",
+  "TotalReadoutTime": 0.095,
+  "MultipartID": "dwi_protocol1"
```
mattcieslak, line 476: great idea

effigies:
Does it make sense to use the split entity?

oesteban:
> Does it make sense to use the split entity?

I honestly don't know whether that would be backward compatible, would it?
If so, I'd be fine with that (without thinking a lot about it) -- although `split` would then almost completely void `run` of semantics for DWI: if you do the exercise of imaging `split` elevated to the definitions in common principles after "Run" then their definitions will be almost the same (differing just in particularities of f/dMRI), don't you think?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: Totally agree. My plan is to rebase your #352 on this after this is accepted (I believe this refactor should be less controversial), and then push that PR to the endline.

tsalo:
> If so, I'd be fine with that (without thinking a lot about it) -- although split would then almost completely void run of semantics for DWI: if you do the exercise of imaging split elevated to the definitions in common principles after "Run" then their definitions will be almost the same (differing just in particularities of f/dMRI), don't you think?

I'd say that the difference also comes in in how you analyze the data. In the case of separate DWI runs, would one just concatenate them in the fourth dimension and preprocess like that? In my (extremely limited) experience, most of the time when you have multiple DWI runs you're going to change the phase encoding direction and you'd undistort them differentially. In the case of having to break the protocol in half for the sake of the scanner, you'd just concatenate and analyze as one run, right?

oesteban:
> I'd say that the difference also comes in in how you analyze the data. In the case of separate DWI runs, would one just concatenate them in the fourth dimension and preprocess like that? In my (extremely limited) experience, most of the time when you have multiple DWI runs you're going to change the phase encoding direction and you'd undistort them differentially. In the case of having to break the protocol in half for the sake of the scanner, you'd just concatenate and analyze as one run, right?

Not sure this is how runs should be interpreted. I think BIDS should provide the researcher with the resources to encode their intent with the acquisition protocol (the what). I'm not convinced it should describe the processing (the how) in this particular instance. 

> EDIT: To clarify, if so then in that sense `split` and `run` would be different. `split` is more like "multiple parts to the same original parameters" while run is separate samples. I hope that makes sense...

I was alluding at this before: is it worth increasing the complexity to gain such little granularity of the semantics?

mattcieslak:
The two levels of groups are 1 the images share the same B0 field (were run with the same shim setting) and 2 they have the same phase encoding direction/TRT. In HCP protocols the scans are supposed to share the same B0 field but can have opposite phase encoding directions. Because TOPUP/eddy correct everything at once, everything under the same level 1 group goes in to the algorithm. For simpler motion correction methods you might only want to look at images that are in the same level 2 group. If we add a B0GroupID, the user can specify which scans are in the same level 1 group, and you can split them into level 2 groups based in PhaseEncodingDirection/TRT. The B0GroupID tag would also be super useful for fieldmaps in general

francopestilli:
@tsalo @oesteban nice discussion here. In general, concatenating the two DWI scans is an accepted and frequent approach. After motion compensation and other artifacts removal, the directions originally measured in the BVECS can be assumed to be slightly different in the two scans (each scan get's preprocessed slightly differently so the reference frame in the BVECS can be assumed to be slightly different). Note that this approach assumes that the BVECS are also changed (not just the DWI images) as part of artifacts removal to account for the changes in the image. 

Another common approach is to keep the data separated for test-retest reliability and cross-validation for the final model solutions.

Now, I am not sure this comment is helpful. Because it does really address the issue of whether split or run should be preferred. Does split assumes they MUT be the united (they are part of a single data set that is split)? If the answer is yes perhaps run would cover more options.

Finally, there the chance for a scary future where there will be a BVEC file per voxel. This is in theory already required every time we apply a non-linear transform the DWI data. But currently, we avoid doing that, because all solutions sounds like a nightmare.

oesteban:
> The two levels of groups are 1 the images share the same B0 field (were run with the same shim setting) and 2 they have the same phase encoding direction/TRT. In HCP protocols the scans are supposed to share the same B0 field but can have opposite phase encoding directions. Because TOPUP/eddy correct everything at once, everything under the same level 1 group goes in to the algorithm. For simpler motion correction methods you might only want to look at images that are in the same level 2 group. If we add a B0GroupID, the user can specify which scans are in the same level 1 group, and you can split them into level 2 groups based in PhaseEncodingDirection/TRT. The B0GroupID tag would also be super useful for fieldmaps in general

Please check for `B0FieldIdentifier` in #622. I think both PRs should take different rules.

oesteban:
> @tsalo @oesteban nice discussion here. In general, concatenating the two DWI scans is an accepted and frequent approach. After motion compensation and other artifacts removal, the directions originally measured in the BVECS can be assumed to be slightly different in the two scans (each scan get's preprocessed slightly differently so the reference frame in the BVECS can be assumed to be slightly different). Note that this approach assumes that the BVECS are also changed (not just the DWI images) as part of artifacts removal to account for the changes in the image.
> 
> Another common approach is to keep the data separated for test-retest reliability and cross-validation for the final model solutions.
> 
> Now, I am not sure this comment is helpful. Because it does really address the issue of whether split or run should be preferred. Does split assumes they MUT be the united (they are part of a single data set that is split)? If the answer is yes perhaps run would cover more options.

I think this is the key of the proposals here and in #622, and basically touches upon @satra's comment in https://github.com/bids-standard/bids-specification/issues/239#issuecomment-497329009 regarding whether BIDS should encode intent and/or automation.

This proposal is thought to allow encoding intent but not automation. IMHO, your comment touches upon the *automation* concept - prescribing not just what is the motivation for acquiring these particular elements of the data, but also giving instructions on how they should be processed and analyzed.

I'm confident that the resistance against allowing BIDS to encode intent will be pretty low. In a way, not doing so would go against the very principles of BIDS.

But I'm not so sure about automation, I think that requires a higher level discussion by the community as a whole.

At this moment, I guess the main question for the fate of this PR is whether we want to unfold the concept of `split` from the concept of `run`. To me, they are to faces of the same coin, and I don't mind keeping `run` (which could be somewhat biased towards the fMRI community) as long as the specs state explicitly that bvals/bvecs can change across runs. I have added a sentence to state that in this PR, and it is kind of implicit in the specs right now. This is basically the conversation @tsalo wanted to bring up, right?

effigies (commented):
The point of `split-` in ephys data is to say "For technical reasons, these files that would otherwise be one file have had to be split." In that case, it's mostly due to a maximum file size, but I don't see a sequence having to be split over multiple acquisitions for temperature reasons as fundamentally different. The idea is that "These data are meant to be analyzed as a unit."

If I understand this correctly, a set of files with the same `MultipartID` are considered the same? Or is it that they will have a sequence of `MultipartIDs` that impose an ordering? Relatedly, does the order matter, or is that up to software to determine an order? This might be all in my head, but I would interpret `split-1 ... split-4` as unambiguously declaring an intended order of concatenation while `run-1 ... run-4` declares an order of acquisition.

On balance, the question between `run` and `split` for me would be: Can the data be meaningfully analyzed in isolation? If so, `run-` makes sense. If they must be analyzed together, and could just as well be concatenated, then making it clear within the filename with `split-` makes sense to me.

oesteban:
> On balance, the question between `run` and `split` for me would be: Can the data be meaningfully analyzed in isolation? If so, `run-` makes sense. If they must be analyzed together, and could just as well be concatenated, then making it clear within the filename with `split-` makes sense to me.

So here there is a prescription of how you should process the data (just flagging it). That said, the problem with these kinds of DWI datasets is that, in theory, you could just stitch the splits together. In practice, they are typically preprocessed separately (mostly head motion correction) and then merged together.

I think here we want to preserve the order of acquisition. That is currently doable with `run-`. The addition of `MultipartID` just would tag runs to signal the notion that they can *potentially* be directly merged - but perhaps more sophisticated processing would be better before merging (i.e., some hierarchical head-motion correction/between-run-alignment).

I also proposed the new metadata key because I assume there is less friction in adding metadata affecting only DWI. If we were to accept `split-` for DWI then I don't see much of a point in having `run-` for them. Keeping the protocol's provenance (i.e., a less opinionated track of the ordering in which each run was acquired) seems a simpler change.

That all said, sorry for my ignorance -- I didn't know `split-` already existed for ephys (meaning, I looked at it as a bigger change before learning the keyword and semantics are already somewhere in BIDS).

Will have a look into your comments later, thanks for taking the time 👍

francopestilli:
> > On balance, the question between `run` and `split` for me would be: Can the data be meaningfully analyzed in isolation? If so, `run-` makes sense. If they must be analyzed together, and could just as well be concatenated, then making it clear within the filename with `split-` makes sense to me.
> 
> So here there is a prescription of how you should process the data (just flagging it). That said, the problem with these kinds of DWI datasets is that, in theory, you could just stitch the splits together. In practice, they are typically preprocessed separately (mostly head motion correction) and then merged together.
> 
> I think here we want to preserve the order of acquisition. That is currently doable with `run-`. The addition of `MultipartID` just would tag runs to signal the notion that they can _potentially_ be directly merged - but perhaps more sophisticated processing would be better before merging (i.e., some hierarchical head-motion correction/between-run-alignment).
> 
> I also proposed the new metadata key because I assume there is less friction in adding metadata affecting only DWI. If we were to accept `split-` for DWI then I don't see much of a point in having `run-` for them. Keeping the protocol's provenance (i.e., a less opinionated track of the ordering in which each run was acquired) seems a simpler change.
> 
> That all said, sorry for my ignorance -- I didn't know `split-` already existed for ephys (meaning, I looked at it as a bigger change before learning the keyword and semantics are already somewhere in BIDS).
> 
> Will have a look into your comments later, thanks for taking the time 👍

@oesteban I am supportive of using `-run` instead of adding `-split` there is sufficient overlap in the usage here that we do not need to add a new term. My 2 cents.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -432,6 +450,57 @@ bval example:
...
+       sub-<label>_run-2_dwi.nii.gz
+       sub-<label>_run-2_dwi.bval
+       sub-<label>_run-2_dwi.bvec
+       sub-<label>_run-3_dwi.nii.gz
+       sub-<label>_run-3_dwi.bval
+       sub-<label>_run-3_dwi.bvec
+       sub-<label>_run-4_dwi.nii.gz
+       sub-<label>_run-4_dwi.bval
+       sub-<label>_run-4_dwi.bvec
+       sub-<label>_dwi.json
```
josephmje, line 499: Would a json file for each of the individual runs no longer exist? Or would `sub-<label>_dwi.json` contain the unique tags that would trickle down and supersede the tags in the individual run jsons?

oesteban: This is just to make the example shorter, but the rationale of this new tag `MultipartID` still works with multiple jsons.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -432,6 +450,57 @@ bval example:
...
+requiring to generate several DWI runs that were originally meant to belong
+in a single one.
+For instance, some GE scanners cannot collect more than &asymp;160 volumes
+in a single run under fast-changing gradients, so acquiring *HCP-style*
+diffusion images will require splitting the DWI scheme in several runs.
+
+| **Key name**    | **Requirement level** | **Data type** | **Description**                                                                                                                                                                                                      |
+| --------------- | --------------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| MultipartID     | REQUIRED              | [string][]    | Unique identifier (participant's level) to prescribe DWI segments belonging in a single dataset. Segments can be designated by any combination of `dir-<label>`, `acq-<label>` and/or `run-<index>` key/value pairs. |
```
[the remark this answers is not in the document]
oesteban: will rewrite, thanks for pointing out.

effigies:
> That said, the problem with these kinds of DWI datasets is that, in theory, you could just stitch the splits together. In practice, they are typically preprocessed separately (mostly head motion correction) and then merged together.

It reads like you're interpreting `split-` as *requiring* concatenation before any processing, which I don't mean to imply. I can take a series apart and process each volume individually if I want; that's an algorithmic decision. The point of `split` is to clearly mark to tools and users that a set of files cannot be conceptually separated.

> I am supportive of using -run instead of adding -split there is sufficient overlap in the usage here that we do not need to add a new term.

`MultipartID` would be a new term. `split-` already exists within BIDS.

If the contention is that adding `split-` makes `run-` redundant, then:

1) Moving to `split-` or `MultipartID` will require all DWI datasets to be updated.
2) Neither is actually necessary, since you can always interpret `run-` as `split-` right now.

If they are not the same concept, then using `split-` involves renaming files in some datasets, rather than adding metadata in a way that respects the inheritance principle. This seems simpler to me and allows people to instantly distinguish multi-part sequences from multiple runs.

I think PyBIDS at least is robust enough to make the selection of related files trivial in either case. Not sure about BIDS-MATLAB (@Remi-Gau may be able to comment).

----

Following a maintainers meeting, I would say there is a strong hesitation to introducing an entirely new mechanism for a problem that already has a solution in BIDS. I think to add the metadata field, I'd like to see a use case that can't be satisfied by the entity.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
francopestilli: This seems good to me to.

oesteban:
> The point of `split` is to clearly mark to tools and users that a set of files cannot be conceptually separated.

Then I fully understood it, and that's why my impression is that `run` still better represents the case of DWI data. Perhaps we are misrepresenting the problem. 

Typically, when you have a GE scanner but you want to run an HCP-like sequence, you first take the gradient scheme of the HCP sequence and try to find a good split of it. However, if you break the continuity of the acquisition, then the sequence of gradients is not "compensated" anymore. Also, because you have to insert breaks, researchers typically acquire in a more resilient manner and perhaps place the b=0 volumes at different places. You may optimize the ordering of each split, with several criteria: minimize temperature increase, or permitting that a failed run (e.g., you run out of time, or the participant hit the panic button) does not turn the multipart acquisition unusable. I think the only detail that does not match well with the current `run-` definition is having constant gradients (`bvecs`/`bvals`) across runs - which is something that seems to be implicitly allowed by the specs examples at the moments (and, I haven't checked, but the validator probably allows).

> introducing an entirely new mechanism

introducing literally a piece of metadata of string type (i.e., a tag, a label, or a variable name if you will). The mechanism (sidecar JSON) is already there.

> for a problem that already has a solution in BIDS.

It hasn't:

```
dwi/
   sub-1_dir-ap_run-1.nii.gz
   sub-1_dir-ap_run-2.nii.gz
   sub-1_dir-pa_run-1.nii.gz
   sub-1_dir-pa_run-2.nii.gz
```
Currently, there's no way to say that these 4 are meant to be just one file (regardless it makes sense or not).

With the already existing `split-` (in ephys), this could go like:
```
dwi/
   sub-1_split-1_dir-ap_run-1.nii.gz
   sub-1_split-2_dir-ap_run-2.nii.gz
   sub-1_split-3_dir-pa_run-1.nii.gz
   sub-1_split-4_dir-pa_run-2.nii.gz
```
If the majority sees this clearer, easier, and already existing (IMHO all the adaptations you mentioned above that are necessary means it really doesn't exist for DWI), then I'm fine with it.

But I believe this doesn't really resolve the issue. Now say we have two multipart DWIs with 4 files each:
```
dwi/
   sub-1_split-1_dir-ap_run-1.nii.gz
   sub-1_split-1_dir-ap_run-3.nii.gz
   sub-1_split-2_dir-ap_run-2.nii.gz
   sub-1_split-2_dir-ap_run-4.nii.gz
   sub-1_split-3_dir-pa_run-1.nii.gz
   sub-1_split-3_dir-pa_run-3.nii.gz
   sub-1_split-4_dir-pa_run-2.nii.gz
   sub-1_split-4_dir-pa_run-4.nii.gz
```
you better have your `acq-` label still available (and use it as a sort of `MultipartID`) because otherwise, you can't untangle this mess.

To me, this seems easier:
```
dwi/                          # MultipartID
   sub-1_dir-ap_run-1.nii.gz  # dwi1
   sub-1_dir-ap_run-2.nii.gz  # dwi1
   sub-1_dir-ap_run-3.nii.gz  # dwi2
   sub-1_dir-ap_run-4.nii.gz  # dwi2
   sub-1_dir-pa_run-1.nii.gz  # dwi1
   sub-1_dir-pa_run-2.nii.gz  # dwi1
   sub-1_dir-pa_run-3.nii.gz  # dwi2
   sub-1_dir-pa_run-4.nii.gz  # dwi2
```

When I see this, I know that the designer of the protocol wants me to merge together runs 1 and 2 of both pe directions (which means more massaging than just head-motion correction), and runs 3 and 4 (both dir).


I realize that my first example could also show how `split-` runs short. Say that you want to combine by-2:
```
dwi/
   sub-1_split-1_dir-ap_run-1.nii.gz
   sub-1_split-1_dir-ap_run-2.nii.gz
   sub-1_split-2_dir-pa_run-1.nii.gz
   sub-1_split-2_dir-pa_run-2.nii.gz
```
Again, you need an `acq-label` to say if you should combine same or opposed `dir-`. On the contrary:
```
dwi/
   sub-1_dir-ap_run-1.nii.gz  # dwi1
   sub-1_dir-ap_run-2.nii.gz  # dwi2
   sub-1_dir-pa_run-1.nii.gz  # dwi1
   sub-1_dir-pa_run-2.nii.gz  # dwi2
```
seems pretty clear (regardless of whether this combination makes sense or not, nobody forces you to follow what the metadata says - e.g., skipping slice-timing correction when you have the times).

effigies:
Okay, so it looks like it isn't the case that you're grouping things that are just the same sequence, split because the scanner needs time to cool, but are doing arbitrary grouping. I definitely agree that you can't do arbitrary grouping with `split-`, which is an index that imposes a sequence on files with otherwise constant names.

You should update the examples to match what you've said here. It may be extremely obvious to DWI practitioners, but the existing examples look to me like a candidate for `split-`.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: please check if now it is clearer

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: Please let me know further concerns after the latest commits which sideways have contributed to clarify this.

edickie (approved):
Nice discussion. Sound good to me.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -388,52 +396,160 @@ sub-<label>/[ses-<label>/]
...
+
+The REQUIRED gradient orientation information corresponding to a DWI acquisition
+MUST be stored using `.bval` and `.bvec` pairs of files.
+The `.bval` and `.bvec` files can be saved on any level of the directory structure
+and thus define those values for all sessions and/or subjects in one place (see
+Inheritance principle).
+
+As an exception to the [common principles](../02-common-principles.md#definitions)
+that parameters are constant across runs, the gradient table information (stored
+within the `.bval` and `.bvec` files) MAY change across DWI runs.
```
oesteban, line 439: @tsalo please let us know if you think this addresses the issue you pointed out, by making explicit the exception (as I commented before, the exception is currently rather implicit.)

@bids-standard/raw-mri-dwi WDYT?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: I tagged you above, where the exception to the common principles is made explicit. The rest of the discussion I think can be followed with the bulk of the conversation (esp. Chris' comments). Marking resolved - please mark unresolved if you disagree.

oesteban:
@bids-standard/raw-mri can I ask for reviews?

francopestilli:
Hi Oscar

Yes! Can you point us to the location please?

Below what I see:

arokem:
@francopestilli : I find it easiest to cross-reference the diff view (https://github.com/bids-standard/bids-specification/pull/624/files) with the rendered version here: https://bids-specification--624.org.readthedocs.build/en/624/ and specifically, for this PR, this page: https://bids-specification--624.org.readthedocs.build/en/624/04-modality-specific-files/01-magnetic-resonance-imaging-data.html

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,160 @@ sub-<label>/[ses-<label>/]
...
-corresponding `dwi.nii` file, and not the magnet's coordinate system.
-Thus, any rotations applied to `dwi.nii` also need to be applied to the
-corresponding bvec file.
-
-bvec example:
+distinguish different sets of phase-encoding directions.
+
+**Combining multi- and single-band acquisitions**.
+The single-band reference image MAY be stored with suffix `sbref` (for example,
+`dwi/sub-control01_sbref.nii[.gz]`,) as long as the image has no corresponding
```
effigies, line 418:
```suggestion
`dwi/sub-control01_sbref.nii[.gz]`) as long as the image has no corresponding
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,160 @@ sub-<label>/[ses-<label>/]
...
+[gradient information (`.bval` and `.bvec` sidecar files)](#required-gradient-orientation-information)
+to be stored.
+
+Otherwise, if some gradient information is associated to the single-band diffusion
+image and a multi-band diffusion image also exists, the `acq-<label>` key/value pair
+MUST be used to distinguish both images.
```
effigies, line 424: I'm assuming so, but just to confirm, does this match current practice? This seems to be a new explicit requirement, although I see that there's never been `_sbref.bval`.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,160 @@ sub-<label>/[ses-<label>/]
...
+In such a case, two files could have the following names:
+`sub-01_acq-singleband_dwi.nii.gz` and `sub-01_acq-multiband_dwi.nii.gz`.
+The user is free to choose any other label than `singleband` and
+`multiband`, as long as they are consistent across subjects and sessions.
+
+### REQUIRED gradient orientation information
+
+The REQUIRED gradient orientation information corresponding to a DWI acquisition
+MUST be stored using `.bval` and `.bvec` pairs of files.
```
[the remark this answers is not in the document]
effigies: I can find no examples of `bvals`/`bvecs` in OpenNeuro, and I do not think the validator has ever recognized those extensions.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,160 @@ sub-<label>/[ses-<label>/]
...
+  dwi/
+    sub-1_acq-shell1_run-1_dwi.nii.gz  # dwi_1
+    sub-1_acq-shell1_run-2_dwi.nii.gz  # dwi_2
+    sub-1_acq-shell2_run-1_dwi.nii.gz  # dwi_1
+    sub-1_acq-shell2_run-2_dwi.nii.gz  # dwi_2
+```
+
+### Other RECOMMENDED metadata
 
+At minimum, the `PhaseEncodingDirection` and `TotalReadoutTime` metadata
```
effigies, line 551: "Minimum" + "RECOMMENDED" feels like a mismatch.

```suggestion
The `PhaseEncodingDirection` and `TotalReadoutTime` metadata
```

effigies (approved):
LGTM. Small comments.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: Related to this. Just wanted to add that after this is merged, we should push with @mattcieslak for #352. After merge I'll take care of rebasing that one.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: If you see a `_sbref.bval` file then it should be filled with zeros. There could be a use case for nonzero (but very low b-vals << 1000) but why don't use `_dwi` which is more appropriate?

This does match current practice, and as you were pointing, it was implicit in the spec. I just made it explicit, if that makes sense.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: This was meant to say that amongst all the metadata entries we could recommend, these two are really necessary if one wants to be able to perform certain processing (just in case someone wants to recommend something in regards, not just removing the awkward combo).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,160 @@ sub-<label>/[ses-<label>/]
...
+### Multipart (split) DWI schemes
+
+Some MR schemes cannot be acquired directly by some scanner devices,
+requiring to generate several DWI runs that were originally meant to belong
+in a single one.
+For instance, some GE scanners cannot collect more than &asymp;160 volumes
+in a single run under fast-changing gradients, so acquiring *HCP-style*
+diffusion images will require splitting the DWI scheme in several runs.
+Because researchers will generally optimize the data splits, these will likely
+not be directly concatenable anymore.
```
nicholst, line 483: "concatenable" is a mouthful, maybe:
```suggestion
not be able to be directly concatenated.
```

nicholst (approved):
I'm weak on diffusion but I read for clarity and general BIDS consistency... LGTM!

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
mattcieslak: There's one large case, which is siemens scanners can't make a b=0 only series without  b > 0 scan included. So the sbref will include a few b=0 images followed by a random b=1000. +1 for optional gradient files for sbrefs

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,161 @@ sub-<label>/[ses-<label>/]
...
+[gradient information (`[*_]dwi.bval` and `[*_]dwi.bvec` sidecar files)](#required-gradient-orientation-information)
+to be stored.
+
+Otherwise, if some gradient information is associated to the single-band diffusion
+image and a multi-band diffusion image also exists, the `acq-<label>` key/value pair
+MUST be used to distinguish both images.
+In such a case, two files could have the following names:
+`sub-01_acq-singleband_dwi.nii.gz` and `sub-01_acq-multiband_dwi.nii.gz`.
+The user is free to choose any other label than `singleband` and
+`multiband`, as long as they are consistent across subjects and sessions.
```
mattcieslak, line 428: How will the pipelines know that the sbref is intended to be used as the anatomical reference for the multiband sequence? Does this go in the sidecar?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,161 @@ sub-<label>/[ses-<label>/]
...
+The `[*_]dwi.bval` and `[*_]dwi.bvec` files MUST follow the
+[FSL format](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FDT/UserGuide#DTIFIT):
+The `[*_]dwi.bvec` file contains 3 rows with *N* space-delimited floating-point numbers
+(corresponding to the *N* volumes in the corresponding NIfTI file.)
+The first row contains the *x* elements, the second row contains the *y* elements and
+the third row contains the *z* elements of a unit vector in the direction of the applied
+diffusion gradient, where the *i*-th elements in each row correspond together to
+the *i*-th volume, with `[0,0,0]` for *non-diffusion-weighted* (also called *b*=0 or *low-b*)
+volumes.
+Following the FSL format for the `[*_]dwi.bvec` specification, the coordinate system of
```
mattcieslak, line 452: The official FSL format for bvecs is more complicated than that and if the bvecs are truly aligned with the image axes many software packages will misinterpret the bvecs. For example an LPS+ image axis will result in flipped y bvecs in FSL and mrtrix, but not DSI Studio or Dipy.

It makes more sense to either say it's "similar to the FSL format, with bvecs aligned to the image axes" or to make it clear that bvecs are *exactly* the FSL format and will intrinsically flip bvecs for some non LAS+ image orientations. The latter seems definitely less BIDSy because it's additional complexity to meet what is arguably a bug in a specific software package rather than to simply describe the data. This problem goes away with the RASB file format in the other PR

mattcieslak (changes requested):
Looks great, very excited to use these new features! The only concern is I would really appreciate some clarity on the bvec orientation requirements

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: This particular is out of the scope of the refactor - let's take this one to the finish line and then see if extending the `IntendedFor` to sbrefs could do.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: I think the BIDS curator has two options in that case:

- Store a `_dwi` file, although there are not enough orientations to fit a diffusion model this is indeed a *diffusion-weighted* MRI.
- During conversion, drop the b > 0 volume to store a `_sbref` file.

Allowing `_sbref.bval` adds quite a bit of complexity to PyBIDS and software and irregularity (w.r.t. sbrefs under `func/`), while IMHO misrepresenting the nature of the file.

Would you reconsider your +1 to having optional gradient files for sbrefs?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
nicholst: Actually! Yes, I wanted to flag this.   

I realise this is maybe out of scope of this PR, but note that Appendix 8 mentions 3 types of coordinate systems, "device coordinates" (scanner), "file format coordinates" (DICOM), annd "head" coordinates (for example, NIFTI).

Maybe I can offer a change that links up with the definitions in that appedix?  See next...

oesteban: Similarly, here I just wanted to:
- Add the `MultipartID` concept
- Refactor the specs of DWI to be more consistent with the functional MRI (a bit more developed than DWI for historical reasons) and making explicit some caveats very implicit in the current spec.

I think #352 provides a great opportunity to refine what is said about the FSL format, make a clear route for deprecation and resolve the problem it originally addressed.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: (btw, I'm with @effigies here and I don't think we should allow `.[bval|bvec][|s]`, the validator is really useful in indicating the user that those files are not part of the spec)

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,161 @@ sub-<label>/[ses-<label>/]
...
+the *b* vectors MUST be defined with respect to the participant (in other words, defined
+by the axes of the corresponding `_dwi` NIfTI file) and not the magnet's coordinate system.
```
nicholst, line 454:
```suggestion
the *b* vectors MUST be defined with respect to the head coordinantes in the corresponding `_dwi` NIfTI file and not the scanner's device coordinate system (see [Coordinate systems](08-coordinate-systems.md)).
```

nicholst: And... btw, is "scanner's device coordinate system" satisfactorily clear in `../08-coordinate-systems.md`?  All that is said is
>  all data acquisition starts with "device coordinates" (scanner), which does not have to be identical to the initial "file format coordinates" (DICOM), which are again different from the "head" coordinates (for example, NIFTI).

oesteban: Thanks for this suggestion. I've taken the liberty of editing your suggestion to fix a little typo ("coordinante") and introduce new lines (I had also fixed the link in parallel to your edit).

Regarding this: as I commented above, I think we should not make very deep changes regarding the gradients, as I think it will more clear for everyone to tackle them in full within #352.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: I have accepted the suggestion (adding a link to the coordinate systems is helpful and does not entail any big change to the gradients). That said, I still think we should make the deep revision of gradients in #352.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
mattcieslak: Dropping the b > 0 images would be a processing step, so the sbrefs would no longer be "raw" like the other images. This is actually the same issue that the reverse phase encoding direction images have - the `fmaps/*_epi` images require .bval and .bvec files to be interpreted correctly for distortion correction. 

These are major BIDS limitations that are prevent dmri workflows from using real BIDS and require users put the extra gradient information files into .bidsignore. I think there is another PR open for this, though. I'd be happy if there was even a sidecar field that has the bvals in it, since the gradient directions are irrelevant fir sbrefs and epi fieldmaps.

oesteban: > Dropping the b > 0 images would be a processing step, so the sbrefs would no longer be "raw" like the other images.

Dropping the b > 0 would be the same as people are actually doing with nonsteady states in functional MRI (at some places), for instance. Or if you intendedly leave out of conversion a whole bunch of acquisitions that are otherwise already in the DICOM structure. I'm not saying anyone should do this, but it is an option - this is outside of BIDS purview.

>  This is actually the same issue that the reverse phase encoding direction images have - the fmaps/*_epi images require .bval and .bvec files to be interpreted correctly for distortion correction.

I don't think this is the same issue. If you have diffusion weights (i.e., nonzero b-values) then you are indeed acquiring DWI - and these files should not be under `fmap/` regardless of whether you can fit a diffusion model or not. I *think* BIDS has sufficient specifications to write these files under `dwi/` as `_dwi.nii` files with corresponding gradient info. 

The problem of estimating a fieldmap using these DWIs is a completely different issue, which is actually addressed in #622.

Finally, accepting `sbref.bval` files would probably not be backward compatible (and a big change I would leave for another PR to keep this one with a focused scope).

> These are major BIDS limitations that are prevent dmri workflows from using real BIDS and require users put the extra gradient information files into .bidsignore.

Can you point at some publicly shared dataset?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
mattcieslak: sounds good. Let me know when you get to #352 and I am happy to help

oesteban:
What's the protocol for approved PRs? I don't think it would hurt to get more reviews and I'm happy to push for a bit longer, but I'm unclear as to when and who should hit merge on this one :)

effigies:
Five working days from last commit, so this can be merged on Thursday if there are no further changes needed.

francopestilli:
@oesteban @arokem Some of us collect B0's files in a separate scan. This in theory can be saved as fmap/epi but I wonder whether we should envision a tag to keep track of the fact that these files really pertain DWI. Does it make sense? I cc Dan here, working on a related project for brainlife.io @dlevitas

oesteban:
> @oesteban @arokem Some of us collect B0's files in a separate scan. This in theory can be saved as fmap/epi but I wonder whether we should envision a tag to keep track of the fact that these files really pertain DWI. Does it make sense? I cc Dan here, working on a related project for brainlife.io @dlevitas

Is the goal of these b0s to inform fieldmap estimation? If so, the fmap/epi should be fine. However, I feel that by _the fact that these files really pertain DWI_ you actually mean that there are other purposes and with current BIDS the intent for these files to also inform fieldmap estimation cannot be stated if they are stored under dwi.

I think that, with the clarifications to the bvecs/bvals table in this PR, you could store the b0s as `_dwi` and have both bvecs/bvals files be all zeros. If those b0s are matched with some dwi, you could add an `acq-b0` to disentangle the files in a very semantic way.

Then, with #622, you'd be able to indicate that these b0s are meant to be applied in fieldmap estimation.

Finally, allowing a `dwi/b0` suffix could also be interesting - but I think that's is beyond the scope of this PR. WDYT?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -389,52 +397,162 @@ sub-<label>/[ses-<label>/]
...
 0 0 0.021828 -0.015425 -0.70918 -0.2465
 0 0 0.80242 0.22098 -0.00063106 0.1043
 0 0 -0.59636 0.97516 -0.70503 -0.96351
 ```
 
-The bval file contains the b-values (in s/mm<sup>2</sup>) corresponding to the
-volumes in the relevant NIfTI file), with 0 designating non-diffusion-weighted
-volumes, space-delimited.
+The `[*_]dwi.bval` file contains the *b*-values (in s/mm<sup>2</sup>) corresponding to the
+volumes in the relevant NIfTI file), with 0 designating *b*=0 volumes, space-delimited.
```
francopestilli, line 468: depending on the scanner and sequence some of the BVALs might have very small values in locations where 0 is expected (Siemens sometimes adds 5, 10, 15 as an example). Similarly, values for non-zero BVALs can be not-rounded off say 990 or 995 instead of 1,000 or 2020, 2010 or 1995 instead of 2000. I think this is fine for 80% of the situations but it might require changes in the future.

francopestilli (approved):

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: I think @mattcieslak initiated the right place for this conversation with #352. For a cleaner diff that assumes this PR has been already merged, please use https://github.com/oesteban/bids-specification/compare/enh/dwi-combinations...mattcieslak:dwi_tsv

oesteban:
Thank you all.

Merge and split discussions over:
  - #352 - gradient information storage, all *b* = 0 dwi/epi/sbrefs, etc;
  - #622 - fieldmap information intent

?

effigies:
@sappelhoff If you have no objections, this is eligible to be merged.

sappelhoff:
No objections from my side :+1:

PR #626, opened 2020-09-29
[FIX] replace "e.g.," by "for example"

Remi-Gau:
- [x] replace "e.g." by "for example"
- [x] linting of tables
- [x] linting of line length
- [x] ensure that wihin a paragraph new sentences start on a new line


effigies:
@Remi-Gau Before we go too far, I think we should probably verify that this is an accessibility improvement. I believe I got this recommendation from [The Turing Way Style Guide](https://the-turing-way.netlify.app/community-handbook/style.html), but I'm having trouble finding the justification in their [reference](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style), or other accessibility websites. It's possible that `e.g.` and so forth are no longer considered inaccessible?

Maybe @KirstieJane (or some other representative) could clarify the situation?

arokem:
I think that the key sentence in that style guide is "eg can sometimes be read aloud as ‘egg’ by screen reading software"

effigies:
Ah, yup:

> #### eg, etc and ie
> eg can sometimes be read aloud as ‘egg’ by screen reading software. Instead use ‘for example’ or ‘such as’ or ‘like’ or ‘including’ - whichever works best in the specific context.
> 
> etc can usually be avoided. Try using ‘for example’ or ‘such as’ or ‘like’ or ‘including’. Never use etc at the end of a list starting with these words.
> 
> ie - used to clarify a sentence - is not always well understood. Try (re)writing sentences to avoid the need to use it. If that is not possible, use an alternative such as ‘meaning’ or ‘that is’.

I was searching for "e.g." and "latin", so missed that.

tsalo:
So is it enough to just make sure that we use "e.g." instead of "eg" (which is a typo I think we'd catch during PR review anyway)?

PeerHerholz:
> I think that for non-native speakers, avoiding e.g. and using 'for example' might be more transparent.

+1 on that, I experienced issues with that as well.

effigies (approved):

Release_Guideline.md
```diff
@@ -32,7 +32,7 @@ Patch releases will generally be more frequent, and indicate less significant ch
 The following is a non-exhaustive set of justifications for a patch release:
 
 - A modality field has changed and the [bids-validator](https://github.com/bids-standard/bids-validator) has been updated to reflect this change.
-- Links or information in the specification are no longer accurate, *e.g.* if a BEP document is added or moved
+- Links or information in the specification are no longer accurate, *for example* if a BEP document is added or moved
```
tsalo, line 35: Can the italics be dropped? I assume they were there before because "e.g." is Latin.

effigies: :+1: I pretty much ignored the non-spec bits.

Remi-Gau: They are a couple of italics in the non-specs part. 

I left them because they were there before indeed. :-)

Will remove them

sappelhoff (approved):
cool, thanks a bunch @Remi-Gau 

I think the Turing way style guide also has a point with "etc" --> it's really seldom necessary (if ever) to use "etc" :thinking: perhaps we can remove that from the spec in another PR.

Remi-Gau:
> I think the Turing way style guide also has a point with "etc" --> it's really seldom necessary (if ever) to use "etc" thinking perhaps we can remove that from the spec in another PR.

Oh no! that's going to be the source of more insomnia for me. :-p

Will take care of it. :-)

tsalo (approved):

sappelhoff (approved):

effigies:
I'd suggest going ahead and merging this sooner than later, since it's semantically neutral and is going to generate merge conflicts.

sappelhoff:
> I'd suggest going ahead and merging this sooner than later, since it's semantically neutral and is going to generate merge conflicts.

agreed.

Thanks a lot @Remi-Gau

PR #628, opened 2020-10-01
[FIX] change remaining latin expressions (etc and i.e.)

Remi-Gau:
## TO DO
- [x] fix remaining latin expression (etc and i.e)
- [x] lint tables
- [x] lint paragraphs
  - [x] enforce line length
  - [x] start sentences in a new line if the modified paragraph

## for reviewers

@effigies @sappelhoff 

This one is more complicated than for `e.g.`. I am not sure that some of the changes I made "read well": suggestions welcome.

I will get on to the linting once we agree on the content, if that is fine with you

This will most likely create merge conflicts after #626 is merged (because it seems Latin expressions tend to live in pack and cluster), so let me know whether we should:
- merge #626 first and then solve conflicts (my personal preference)
- "extend" #626 with the current changes and have one single PR for all the "Latin busting" changes.

Not sure that the specs have a position on the sempiternal "Oxford comma" debate... But this PR might raise the question. :stuck_out_tongue_closed_eyes:

Issue #403, opened 2020-02-03
Add DOI links and references section for cited papers?


alexandreroutier:
Hello everyone,

When reading the specifications of the upcoming BIDS Derivatives, I noticed that several papers were cited.

I was wondering if adding DOI links e.g. ([Tournier et al., 2007](https://dx.doi.org); [Descoteaux et al. 2009](https://dx.doi.org)) and a References section at the bottom of the page with full references e.g. :
- Tournier JD, Calamante F, Connelly A. *Robust determination of the fibre orientation distribution in diffusion MRI: non-negativity constrained super-resolved spherical deconvolution.* Neuroimage, 2007. [10.1016/j.neuroimage.2007.02.016](https://doi.org/10.1016/j.neuroimage.2007.02.016) (example of formatting)
- etc.

could enrich the document? For diffusion MRI, on top of specifying the derivates, it also makes a very good entry point on what are the possibilities to process this modality.

Besides, I noticed some discrepancies when citing a paper e.g. :
 - (Tournier et al. 2007; Descoteaux et al. 2009) or (Zhang et al. 2012, Daducci et al., 2015)
 - Conturo, T. et al. (1999); Mori, S. et al. (1999) or Garyfallidis et al. (2010); Mori, S. et al. (1999)

Is there any consensus about it?

Currently, I think these questions and propositions involve:
- intracranial Electroencephalography (iEEG)
- Derivatives fMRI
- Derivatives Diffusion MRI

Best,
Alexandre

PS : I take also this opportunity to congratulate you for these great specifications!

sappelhoff:
> I was wondering if adding DOI links e.g. (Tournier et al., 2007; Descoteaux et al. 2009) and a References section at the bottom of the page with full references could enrich the document?

> Besides, I noticed some discrepancies when citing a paper ... Is there any consensus about it

Currently most (all?) references to academic or non-academic sources are in the form of a brief identifier and an embedded link. As you noted, the brief identifier such as "Tournier et al. 2007" currently has no consistent form across the spec.

I agree that it may be worthwhile to decide for a consistent form for at least all academic references and update the spec accordingly. Perhaps APA style?

Regarding a reference section at the end of each page, spelling out all mentioned references in full form, I think the only benefit for this would be a scenario where somebody prints the spec and then cannot access the link anymore. 

[...]

PR #629, opened 2020-10-01
[FIX] URI "definition" and recommendation

Remi-Gau:
## TO DO
- [x] add URI definition after the "File Formation specification" section
- [x] make sure DOIs in the specs (in json examples or otherwise) have the format `doi:<path>`

## For reviewers
- is this the right "place" for that definition?
- anything else but the DOIs that should change?


src/02-common-principles.md
```diff
@@ -517,6 +517,18 @@ for more information.
 }
 ```
 
+## Uniform Resource Indicator 
+
+Uniform Resource Indicator (URI) are a string referring to a resource and SHOULD
```
sappelhoff, line 522:
```suggestion
A Uniform Resource Indicator (URI) is a string referring to a resource and SHOULD
```

src/02-common-principles.md
```diff
@@ -517,6 +517,18 @@ for more information.
...
+have the form `<scheme>:[//<authority>]<path>[?<query>][#<fragment>]`, as specified 
+in [RFC 3986](https://tools.ietf.org/html/rfc3986). 
+This applies tp URLs and other common URIs include Digital Object Identifiers (DOIs),
```
sappelhoff, line 525:
```suggestion
This applies to URLs and other common URIs, including Digital Object Identifiers (DOIs), 
```

sappelhoff (commented):
> anything else but the DOIs that should change?

IMHO all JSON fields that contain URLs or DOIs. For example `CogAtlasID`, which says:

image

source: table in https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson

the big benefit would be that we can validate for URIs, see: https://github.com/bids-standard/bids-validator/issues/1072

the drawback would be that probably so far people have been specifying DOIs either as a URL, or potentially using just the DOI without any prefixes.

But I could live with that drawback in order to make the spec better in the future. Could be considered a breaking change though :man_shrugging:

Remi-Gau:
This looks ready to go. Some changes are I think not problematic but some other I am much less sure about especially in the "higher level" files ("common principles" and "modality agnostic") as they could "break" things.

Will add comments through the review panel.

src/03-modality-agnostic-files.md
```diff
@@ -26,8 +26,8 @@ Every dataset MUST include this file with the following fields:
...
-| DatasetDOI         | OPTIONAL              | [string][]               | The Document Object Identifier of the dataset (not the corresponding paper).                                                                                                                                                                          |
+| ReferencesAndLinks | OPTIONAL              | [array][] of [strings][] | List of references to publication that contain information on the dataset, or [URI](../02-common-principles.md#uniform-resource indicator).                                                                                                                                                                  |
+| DatasetDOI         | OPTIONAL              | [string][]               | The Document Object Identifier of the dataset (not the corresponding paper) expressed as a valid [URI](../02-common-principles.md#uniform-resource indicator).                                                                                                                                                                          |
```
Remi-Gau, line 30: This could potentially break things for datasets out there that are not using a valid URI for this.

src/02-common-principles.md
```diff
@@ -449,7 +449,7 @@ using an object containing the following fields:
 | Description  | RECOMMENDED           | [string][]                | Description of the column.                                                                                      |
 | Levels       | RECOMMENDED           | [object][] of [strings][] | For categorical variables: An object of possible values (keys) and their descriptions (values).                 |
 | Units        | RECOMMENDED           | [string][]                | Measurement units. SI units in CMIXF formatting are RECOMMENDED (see [Units](./02-common-principles.md#units)). |
-| TermURL      | RECOMMENDED           | [string][]                | URL pointing to a formal definition of this type of data in an ontology available on the web.                   |
+| TermURL      | RECOMMENDED           | [string][]                | [URI](../02-common-principles.md#uniform-resource indicator) pointing to a formal definition of this type of data in an ontology available on the web.                   |
```
Remi-Gau, line 452: To me (as a user) it feels awkward o have the keyname mention "URL" but definition asking for an URI.

Remi-Gau:
One last question: that relates to the diversity of format we use to refer to papers in the specs themselves.

There are a couple of links to some papers in the specs that use a URL rather than a DOI. Should those be changed?

See here:
https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#introduction

For example, should this:
```
(see e.g.,
[https://doi.org/10.1016/j.neuroimage.2012.01.024](https://doi.org/10.1016/j.neuroimage.2012.01.024))
```

be turned into this:
```
(see e.g.,
[doi:j.neuroimage.2012.01.024](https://doi.org/10.1016/j.neuroimage.2012.01.024))
```

sappelhoff:
> There are a couple of links to some papers in the specs that use a URL rather than a DOI. Should those be changed?

I think that'd be reasonable. And I think @yarikoptic would agree as well. See his comment: https://github.com/bids-standard/bids-specification/issues/403#issuecomment-640243595

src/02-common-principles.md
sappelhoff: True, these fields should stay at URL, which is a specific subclass of URI. In the validator, we can still start to validate them as URI --> and it will not break datasets that have been previously valid.

src/03-modality-agnostic-files.md
sappelhoff: agreed, this will most likely break some datasets ... I know for sure that it'll break those that I prepared, see for example https://github.com/bids-standard/bids-examples/blob/8031430b2da07b7f2be58f92b0bd11e2f27a5645/eeg_matchingpennies/dataset_description.json#L15

yet I would be in favor of this breakage for these reasons:

1. it's *very* easy to fix
1. it will lead to a more clear specification (one way to specify the doi, instead of at least 3: `https...`, `doi:...`, or directly `10.17605/OSF.IO/CJ2DR`)
1. it will allow us to better validate this field using the json schemas ... validate a URI formatting instead of simply a string type

Drawbacks:

- Some datasets will remain broken, because their owners won't update them
- We might annoy some users

src/02-common-principles.md
Remi-Gau: okidoki will revert any of the changes where the keyname contains URL.

src/04-modality-specific-files/05-task-events.md
```diff
@@ -165,7 +165,7 @@ The object supplied for `StimulusPresentation` SHOULD include the following key-
...
 | SoftwareRRID    | RECOMMENDED           | [string][]    | [Research Resource Identifier](https://scicrunch.org/resources) of the software that was used to present the stimuli. Examples: The RRID for Psychtoolbox is 'SCR_002881', and that of PsychoPy is 'SCR_006571'. |
 | SoftwareVersion | RECOMMENDED           | [string][]    | Version of the software that was used to present the stimuli.                                                                                                                                                    |
-| Code            | RECOMMENDED           | [string][]    | [URI][uri] of the code used to present the stimuli. Persistent identifiers such as DOIs are preferred. If multiple versions of code may be hosted at the same location, revision-specific URIs are recommended.  |
+| Code            | RECOMMENDED           | [string][]    | [URI](../02-common-principles.md#uniform-resource indicator) of the code used to present the stimuli. Persistent identifiers such as DOIs are preferred. If multiple versions of code may be hosted at the same location, revision-specific URIs are recommended.  |
```
Remi-Gau, line 168: I suspect that I should actually use the `link definitions` section at the bottom of each document even when it comes to internal links, right?

sappelhoff: The `link definitions` section has two advantages: 

1. links that are repeating in the document only have to be defined once
1. the reading flow of the markdown source text is not interrupted by long links (this is especially useful for formatting markdown tables, where you don't want to have too long columns just due to links)

we currently don't have strict guidelines on when to use which kind of link.

Remi-Gau:
This one has been rebased and conflicts have been resolved. 

So this should be ready to merge before we can move on to #628 for more of the same (and potentially #636 ?)

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -8,7 +8,7 @@ context of the academic literature:
 > Gramfort, A., Henson, R.N., Jas, M., Litvak, V., Moreau, J., Oostenveld, R.,
 > Schoffelen, J.-M., Tadel, F., Wexler, J., Baillet, S. (2018). **MEG-BIDS, the
 > brain imaging data structure extended to magnetoencephalography**. Scientific
-> data, 5. doi: [10.1038/sdata.2018.110](https://doi.org/10.1038/sdata.2018.110)
+> data, 5. [doi:10.1038/sdata.2018.110](https://doi.org/10.1038/sdata.2018.110)
```
sappelhoff, line 11: can you please revert this change, so that it does not create a conflict for #630 ?

src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -7,7 +7,7 @@ context of the academic literature:
 > Pernet, C. R., Appelhoff, S., Gorgolewski, K.J., Flandin, G., Phillips, C.,
 > Delorme, A., Oostenveld, R. (2019). **EEG-BIDS, an extension to the brain
 > imaging data structure for electroencephalography**. Scientific data, 6.
-> doi: [10.1038/s41597-019-0104-8](https://doi.org/10.1038/s41597-019-0104-8)
+> [doi:10.1038/s41597-019-0104-8](https://doi.org/10.1038/s41597-019-0104-8)
```
sappelhoff, line 10: can you please revert this change, so that it does not create a conflict for #630 ?

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -13,7 +13,7 @@ context of the academic literature:
 > Ramsey, N. F., Stolk, A., Swann, N. C., Tadel, F., Voytek, B., Wandell, B. A.,
 > Winawer, J., Whitaker, K., Zehl, L., Hermes, D. (2019). **iEEG-BIDS,
 > extending the Brain Imaging Data Structure specification to human intracranial
-> electrophysiology**. Scientific data, 6. doi: [10.1038/s41597-019-0105-7](https://doi.org/10.1038/s41597-019-0105-7)
+> electrophysiology**. Scientific data, 6. [doi:10.1038/s41597-019-0105-7](https://doi.org/10.1038/s41597-019-0105-7)
```
sappelhoff, line 16: can you please revert this change, so that it does not create a conflict for #630 ?

sappelhoff (commented):
LGTM! 

However, we need other people to chime in about the potential breakage of now wanting our DOIs specified as URI in the from `doi:thedoi`

From a technical validation perspectives, only the datasets will break that specified the DOI as just `thedoi`, because those that used `https://doi.org/thedoi` still have a valid URI.

here's the link to previous discussion on this: https://github.com/bids-standard/bids-specification/pull/629#discussion_r498693352

effigies (commented):
This looks great! I have made some suggestions to maintain backwards compatibility. Let me know what you think.

src/03-modality-agnostic-files.md
```diff
@@ -26,8 +26,8 @@ Every dataset MUST include this file with the following fields:
...
-| ReferencesAndLinks | OPTIONAL              | [array][] of [strings][] | List of references to publication that contain information on the dataset, or links.                                                                                                                                                                  |
-| DatasetDOI         | OPTIONAL              | [string][]               | The Document Object Identifier of the dataset (not the corresponding paper).                                                                                                                                                                          |
+| ReferencesAndLinks | OPTIONAL              | [array][] of [strings][] | List of [URI][uri] references to publication(s) that contain information on the dataset.                                                                                                                                                              |
```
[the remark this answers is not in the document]
Remi-Gau: Fine with me.

src/03-modality-agnostic-files.md
```diff
@@ -26,8 +26,8 @@ Every dataset MUST include this file with the following fields:
...
+| DatasetDOI         | OPTIONAL              | [string][]               | The Digital Object Identifier of the dataset (not the corresponding paper) expressed as a valid [URI][uri].                                                                                                                                           |
```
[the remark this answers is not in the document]
Remi-Gau: Should we wait till #634 is merged?

src/02-common-principles.md
```diff
@@ -204,7 +204,7 @@ field in `dataset_description.json` of each subfolder of `derivatives` to:
 
 ```JSON
 {
-  "SourceDatasets": [ {"URL": "file://../../rawdata/"} ]
+  "SourceDatasets": [ {"URI": "file://../../rawdata/"} ]
```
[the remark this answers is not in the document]
Remi-Gau: Well... That's why we have BIDS 2.0 for ! :rocket:

src/02-common-principles.md
```diff
@@ -520,6 +520,18 @@ for more information.
...
+A Uniform Resource Indicator (URI) is a string referring to a resource and SHOULD
+have the form `<scheme>:[//<authority>]<path>[?<query>][#<fragment>]`, as specified 
+in [RFC 3986](https://tools.ietf.org/html/rfc3986). 
+This applies to URLs and other common URIs, including Digital Object Identifiers (DOIs), 
+which may be fully specified as `doi:<path>`, 
+for example, [doi:10.5281/zenodo.3686061](https://doi.org/10.5281/zenodo.3686061).
+A given resource may have multiple URIs.
+When selecting URIs to add to dataset metadata, it is important to consider 
+specificity and persistence.
+
```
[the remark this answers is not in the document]
Remi-Gau: Does it have to cover URLs specifying the DOI resolver? 

Still trying to wrap my around this: I thought that URLs would be a valid URI anyway.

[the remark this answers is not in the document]
Remi-Gau: Yeah I was really not sure about its proper place. Definitely makes sense to have it lower down this hierarchy.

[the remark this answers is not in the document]
Remi-Gau: > I'm okay with promoting this above `Units`, but I think `Participant names and other labels` are more fundamental to understanding BIDS concepts and so would push this down below that.

Done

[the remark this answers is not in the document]
satra: `doi:<thing>` and `https://doi.org/<thing>` are equivalent from an identifier perspective. this may be more of a preference as saying  `doi:<thing>` is preferred. bids may decide to not allow the full url formulation. but that seems a bit more draconian.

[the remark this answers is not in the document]
effigies: The question is "What do we mean when we say DOI". I would argue that we mean "doi:<DOI>", not an arbitrary URI that may or may not refer to the same object. Currently all the examples show `<DOI>`, but it seemed like a good idea to include `https://doi.org/<DOI>` as a thing that will be warned (which is all DEPRECATED means until BIDS 2.0) about by the validator.

If that feels like it's overstepping, then I'm okay with changing to: 

```Markdown
Bare DOIs such as `10.18112/openneuro.ds000001.v1.0.0` are DEPRECATED.
```

src/03-modality-agnostic-files.md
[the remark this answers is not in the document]
effigies: Oh damn, I thought it had been.

src/02-common-principles.md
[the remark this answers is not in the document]
Remi-Gau: I would say let's make go "small steps" and just have :

```Bare DOIs such as `10.18112/openneuro.ds000001.v1.0.0` are DEPRECATED.```

File: src/03-modality-agnostic-files.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (sappelhoff) [reply src/03-modality-agnostic-files.md]: I'll merge it now, the discussion (issue + following PR) about it was open long enough for people to chime in.

File: src/02-common-principles.md [hunk 13]
```diff
@@ -536,6 +536,18 @@ it refers to. For example `sub-01` refers to the `sub` entity (a
...
```
effigies, line 550:
```suggestion

Several fields are designated for DOIs, for example, `DatasetDOI` in `dataset_description.json`.
DOI values SHOULD be fully specified URIs such as `doi:10.18112/openneuro.ds000001.v1.0.0`.
Bare DOIs such as `10.18112/openneuro.ds000001.v1.0.0` are [DEPRECATED].
```

src/03-modality-agnostic-files.md
```diff
@@ -26,8 +26,8 @@ Every dataset MUST include this file with the following fields:
...
-| DatasetDOI         | OPTIONAL              | [string][]               | The Document Object Identifier of the dataset (not the corresponding paper).                                                                                                                                                                          |
+| ReferencesAndLinks | OPTIONAL              | [array][] of [strings][] | List of references to publications that contain information on the dataset. A reference may be textual or a [URI][uri].                                                                                                                               |
+| DatasetDOI         | OPTIONAL              | [string][]               | The Digital Object Identifier of the dataset (not the corresponding paper). DOIs SHOULD be expressed as a valid [URI][uri]; bare DOIs such as `10.0.2.3/dfjj.10` are [DEPRECATED][deprecated].                                                        |
```
effigies, line 30: Okay with leaving this separate deprecation notice, since this has been around so long.

effigies (commented):
I'd still like to have a general note with the URI definition, but I can be argued out of it if you're so inclined.

src/02-common-principles.md
Remi-Gau: I am completely OK with this formulation and only excluding bare ROI but allowing valid URI in general.

effigies (approved):

sappelhoff (approved):

sappelhoff:
Thanks @Remi-Gau

yarikoptic:
> > There are a couple of links to some papers in the specs that use a URL rather than a DOI. Should those be changed?
> 
> I think that'd be reasonable. And I think @yarikoptic would agree as well. See his comment: [#403 (comment)](https://github.com/bids-standard/bids-specification/issues/403#issuecomment-640243595)

yes, I do. Thank you! ;)

PR #643, opened 2020-10-08
[MISC] add brief note that TSV example in the spec may currently use either tab or space characters

yarikoptic:
Inspired by realization while discussing now merged https://github.com/bids-standard/bids-specification/pull/619 

See individual commits description for more information.  A satellite issue on harmonization of tsv examples ~~will be filed shortly~~ #644

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -455,6 +455,7 @@ are occasionally formatted using space characters instead of tabs to improve
 human readability.
 Directly copying and then pasting these examples from the specification
 for use in new BIDS datasets can lead to errors and is discouraged.
+
 Tabular files MAY be optionally accompanied by a simple data dictionary
 in the form of a JSON [object](https://www.json.org/json-en.html)
 within a JSON file.
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -446,8 +446,8 @@ exponent. TSV files MUST be in UTF-8 encoding.
 Example:
 
 ```Text
-onset duration  response_time correct stop_trial  go_trial
-200 200 0 n/a n/a n/a
+onset	duration	response_time	correct	stop_trial	go_trial
+200	200	0	n/a	n/a	n/a
 ```
 
 Tabular files MAY be optionally accompanied by a simple data dictionary
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -450,10 +450,11 @@ onset	duration	response_time	correct	stop_trial	go_trial
 200	200	0	n/a	n/a	n/a
 ```
 
-**Note**: TSV examples in the document can be re-formatted using
-spaces instead of tabs for human readability, and have tabs replaced
-with spaces by used for rendering HTML or PDF.
-
+**Note**: The TSV examples in this document (like the one above this note)
+are occasionally formatted using space characters instead of tabs to improve
+human readability.
+Directly copying and then pasting these examples from the specification
+for use in new BIDS datasets can lead to errors and is discouraged.
 Tabular files MAY be optionally accompanied by a simple data dictionary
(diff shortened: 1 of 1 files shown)


sappelhoff:
I feel like this should be an issue outlining the problem --> if all agree that all TSV examples in the spec should be formatted using tabs, then we would want a follow up PR that fixes all examples.

VisLab:
The tsv examples in the spec are only a few lines long and I think it would
be more  effective to have a link to an actual real tsv file wth tabs
rather than expecting that people are going to cut and paste the toy
examples.  I find it very difficult to understand what the various values
correspond to when the column values don't line up --- especially since
these values can have blanks in them.

yarikoptic:
> I feel like this should be an issue outlining the problem --> if all agree that all TSV examples in the spec should be formatted using tabs, then we would want a follow up PR that fixes all examples.

some issue beyond he one I have referenced? (#644)

> The tsv examples in the spec are only a few lines long and I think it would
> be more  effective to have a link to an actual real tsv file wth tabs
> rather than expecting that people are going to cut and paste the toy examples.

Good idea, but again a more "global one" -- to allow spec to point to good examples (not only for .tsv files but whole datasets, other types of files etc) from external links.  We already have references to various aspects, and have linkchecker validating their validity.  If  there is no issue yet, please start one?

> I find it very difficult to understand what the various values
> correspond to when the column values don't line up --- especially since
> these values can have blanks in them.

yeap, I am with you on this one. in #644 I outlined a possible solution at the level of "renderer" of the spec from source (which imho should use tabs) into visual form (html or pdf).  The problem now that there is no consistency.  This PR at least tries to bring some consistency within the "common principles" description (where fields iirc do not align visually anyways but also simply incorrect by using spaces instead of tabs).

sappelhoff (commented):
> some issue beyond he one I have referenced? (#644)

no that one is perfect - I just overlooked it, sorry :)

src/02-common-principles.md
```diff
@@ -446,10 +446,14 @@ exponent. TSV files MUST be in UTF-8 encoding.
...
-onset duration  response_time correct stop_trial  go_trial
-200 200 0 n/a n/a n/a
+onset	duration	response_time	correct	stop_trial	go_trial
+200	200	0	n/a	n/a	n/a
 ```
 
+**Note**: TSV examples in the document can be re-formatted using
+spaces instead of tabs for human readability, and have tabs replaced
+with spaces by used for rendering HTML or PDF.
+
```
sappelhoff, line 456: I would rephrase this note:

```suggestion
**Note**: The TSV examples in this document (like the one above this note)
are occasionally formatted using space characters instead of tabs to improve
human readability.
Directly copying and then pasting these examples from the specification
for use in new BIDS datasets can lead to errors and is discouraged.
```

yarikoptic: thanks! accepted

src/02-common-principles.md
```diff
@@ -446,10 +446,15 @@ exponent. TSV files MUST be in UTF-8 encoding.
...
-200 200 0 n/a n/a n/a
+onset	duration	response_time	correct	stop_trial	go_trial
+200	200	0	n/a	n/a	n/a
 ```
 
+**Note**: The TSV examples in this document (like the one above this note)
+are occasionally formatted using space characters instead of tabs to improve
+human readability.
+Directly copying and then pasting these examples from the specification
+for use in new BIDS datasets can lead to errors and is discouraged.
```
sappelhoff, line 457:
```suggestion
for use in new BIDS datasets can lead to errors and is discouraged.

```

sappelhoff (approved):

Remi-Gau (approved):
That looks good to me.

sappelhoff:
Thanks everyone.

PR #645, opened 2020-10-16
[SCHEMA] Add derivatives entities to the schema

tsalo:
This stems from https://github.com/mne-tools/mne-bids/issues/449#issuecomment-709893588.

Changes proposed:
- Add the derivatives entities (`res`, `den`, `label`, and `desc`) to the schema.
- Redefine the `space` entity so it works for EEG/MEG/iEEG and MRI data.
- Generate the updated Entities page.

### DIFF
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -138,6 +138,9 @@ space:
     (for EEG/MEG/iEEG data) or
     the spatial reference to which a file has been aligned (for MRI data).
     The space label needs to be taken from the list in Appendix VIII.
+
+    For EEG/MEG/iEEG data, this entity can be applied to raw data, but
+    for other data types, it is restricted to derivative data.
   format: label
 split:
   name: Split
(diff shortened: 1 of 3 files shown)


src/schema/entities.yaml
```diff
@@ -153,3 +155,28 @@ split:
     It is RECOMMENDED that .fif files with multiple parts use the
     `split-<index>` entity to indicate each part.
   format: index
+res:
+  name: Resolution
+  entity: res
+  description: |
+    Resolution of regularly sampled N-D data.
```
effigies, line 162:
```suggestion
    Resolution of regularly sampled N-dimensional data.
    MUST have a corresponding `Resolution` metadata field to provide interpretation.
```

src/schema/entities.yaml
```diff
@@ -153,3 +155,28 @@ split:
...
+  format: label
+den:
+  name: Density
+  entity: den
+  description: |
+    Density of non-parametric surfaces.
```
effigies, line 168:
```suggestion
    Density of non-parametric surfaces.
    MUST have a corresponding `Density` metadata field to provide interpretation.
```

src/schema/entities.yaml
```diff
@@ -153,3 +155,28 @@ split:
...
+  format: label
+label:
+  name: Label
+  entity: label
+  description: |
+    Tissue-type label, following a prescribed vocabulary.
```
effigies, line 174:
```suggestion
    Tissue-type label, following a prescribed vocabulary.
    Applies to binary masks and probabilistic/partial volume segmentations that describe
    a single tissue type.
```

effigies (commented):
Looks reasonable. I've added a little bit of extra content so they don't look quite so bare.

adam2392:
Is it possible to enumerate which ones are referred to as `source_entities` and then which entities are reserved for `derivatives`? So I would imagine a list of supported raw/source entities and then a running list of derivative entities?

E.g. `desc` from what I understand is only meant for `derivatives`, so it should never be present in a BIDS source dataset

tsalo:
@adam2392 We could add a field to the schema that describes scope, but I think that could quickly become confusing (e.g., `space` is raw for MEG/EEG/iEEG but derivatives-only for MRI). Would it be enough if I expanded the descriptions to clarify that they can only be used with Derivatives data?

adam2392:
> @adam2392 We could add a field to the schema that describes scope, but I think that could quickly become confusing (e.g., `space` is raw for MEG/EEG/iEEG but derivatives-only for MRI). Would it be enough if I expanded the descriptions to clarify that they can only be used with Derivatives data?

Oh okay yeah that sounds good! I suppose that's a pretty simple fix for now and makes sense given what you said. Is `desc` the only Derivatives-only entity?

adam2392 (approved):

effigies (approved):

sappelhoff (approved):

sappelhoff:
@tsalo @effigies unless you have final issues with this, I think one of you can merge this.

Issue #649, opened 2020-10-22
Templating issue: should electrodes.tsv and coordsystem.json writing per eeg/ieeg only have certain entities?


adam2392:
## Summary
This came up in https://github.com/mne-tools/mne-bids/issues/592

According to the spec: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#electrode-description-_electrodestsv

ieeg can have subject/session/space for electrodes.tsv.

while eeg can have subject/session/acquisition/run for electrodes.tsv (and maybe space?)

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv

Is this correct?

## Proposed Solution
My understanding is that you only need one `electrodes.tsv/coordsystem.json` per session directory because all the electrodes/channels are the same "within the same session" for iEEG. So you can have multiple tasks, runs, acquisitions, etc. 

But I suppose this might be different for say scalp EEG based on reading the above? Either way some clarification would help immensely.

Based on the discussion for the above, I can make a subsequent PR if it is deemed appropriate.

sappelhoff:
IMHO, "run" does not make sense for EEG electrodes.tsv --> there is typically just one per session, just like in iEEG (although in iEEG, it could potentially even be the same for several sessions, which is seldom to never true for EEG).

The acquisition entity makes sense to add something like `acq-pre` and `acq-post` to flag the digitization of electrode coordinates (using whatever method or device) prior to data recording and after data recording --> this is sometimes done. However for my own data I only record once.

Finally, the `space` entity does not seem relevant for EEG --> I have never encountered a dataset or workflow where the electrodes were digitized using different methods to have them in different spaces ... well apart from studies that are exactly about that topic: comparing different digitization methods such as in this paper where @robertoostenveld is coauthor: https://doi.org/10.1016/j.jneumeth.2019.108378 ("Using a structured-light 3D scanner to improve EEG source modeling with more accurate electrode positions")

adam2392:
So eeg: `subject/session/acq` if available
and ieeg: `subject/session/space` if available?

And no need to include extra entities because they just then imply there are multiple coords (when in reality they're all copies of the same file)?

Seems like this should produce a warning in the `bids-validator` if there are "extra" entities (but prolly not error).

sappelhoff:
> So eeg: subject/session/acq if available
and ieeg: subject/session/space if available?

[...]

PR #650, opened 2020-10-22
[FIX] Fixing template string on electrodes for eeg and ieeg.

adam2392:
# Summary

Closes: #649 

Fixes the template strings in the modality-specific spec to match that of https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg

### DIFF
--- a/src/04-modality-specific-files/03-electroencephalography.md
+++ b/src/04-modality-specific-files/03-electroencephalography.md
@@ -279,7 +279,7 @@ Template:
 sub-<label>/
     [ses-<label>]/
       eeg/
-        [sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_electrodes.tsv]
+        sub-<label>[_ses-<label>][_acq-<label>][_space-<label>]_electrodes.tsv]
 ```
 
 File that gives the location of EEG electrodes. Note that coordinates are
--- a/src/04-modality-specific-files/03-electroencephalography.md
+++ b/src/04-modality-specific-files/03-electroencephalography.md
@@ -279,7 +279,7 @@ Template:
 sub-<label>/
     [ses-<label>]/
       eeg/
-        sub-<label>[_ses-<label>][_acq-<label>][_space-<label>]_electrodes.tsv]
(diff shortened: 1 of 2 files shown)


src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -279,7 +279,7 @@ Template:
 sub-<label>/
     [ses-<label>]/
       eeg/
-        [sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_electrodes.tsv]
+        sub-<label>[_ses-<label>][_acq-<label>][_space-<label>]_electrodes.tsv]
```
sappelhoff, line 282:
```suggestion
        sub-<label>[_ses-<label>][_acq-<label>][_space-<label>]_electrodes.tsv
```

adam2392: Why is the extra bracket needed?

sappelhoff: It's not needed, it's there by mistake. The square brackets are meant to indicate optional versus required entities :)

adam2392: Ah I see, I think I confused myself ... :p

sappelhoff (approved):

sappelhoff:
This is in accordance with the entity table: https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html#encephalography-eeg-ieeg-and-meg ... so this is truly a fix.

unfortunately, as discussed [in this issue comment](https://github.com/bids-standard/bids-specification/issues/649#issuecomment-714607648), the validator is not precise enough to validate this. So the validator will still accept files like: `*_task-Nonsense_electrodes.tsv`. There is a similar issue here: https://github.com/bids-standard/bids-validator/issues/848

Still +1 to merge, then adjust mne-bids (https://github.com/mne-tools/mne-bids/pull/604), and then eventually improve the validator (better sooner than later)

sappelhoff:
Thanks @adam2392

PR #610, opened 2020-09-16
[SCHEMA] Render schema elements in text

tsalo:
References #544 and closes #591. Uses the [`mkdocs-macros`](https://mkdocs-macros-plugin.readthedocs.io/en/latest/) plugin to generate certain markdown elements (e.g., the filename patterns) from the schema. This should easily extend to tables and json examples, as well.

Changes proposed:

- Replace manual Entity Table and Entity description file creation scripts with functions that are called by the `mkdocs-macros` plugin within the markdown files.
- Move all auxdatatypes into new groups under the associated data types. This is primarily because I had so much trouble getting a generic filtering function working with the original structure.
- Add standard Python stuff to the gitignore.
- Reorganize Python code into a mini package.
- Add several Python libraries and `macros` as dependencies.
- Replace entities in datatypes files with the actual names instead of their shorthand versions.
- Split the iEEG/EEG/MEG entity table into a Biopotential Amplification (iEEG/EEG) table and an MEG table.

# TODO:

- [x] Reorganize Python code. Even though the `macros` documentation says that the `define_env` function can be in a module, I can't seem to get that working, so it's in its own file at the top level for now.
- [X] Generic filtering function. We need to be able to filter the schema down to specific classes or elements within classes.
    - Something on the order of `pybids` is probably overkill, but I imagine the `pybids` devs would have some insights on this.
    - I added something in 9ce6d7f that is fairly flexible. We should improve it at some point though.
- [X] Updated entity tables. Currently, because of the lack of a filtering function, only the full entity table is generated.
- [X] Combine `.nii` and `.nii.gz` as `.nii[.gz]` in file formats.
- [X] ~~Consider dropping `.json` from the list of extensions in the schema. JSON files are not alternatives to the other extensions. They are required (at some level).~~
    - I've decided to ignore this nuance for now. Will just treat json like any other extension and show it in the format snippets.
- [x] Deal with remark. TravisCI is failing because remark flags the code snippets like `{{ make_filename_template(datatypes=["anat"]) }}` as bad.
- [x] Update contributing information. Fewer steps are needed when updating the schema now, since rendering happens automatically.
- [x] Fix PDF rendering. Currently, it just shows the inline code without the text _produced_ by the code.
- [x] Remove old outputs (currently retained for comparison).

# More To Dos (added by @sappelhoff here for better visibility, but can be done by anybody)

- open issues:
    - [X] eventually turn schemacode into its own package and migrate to separate repo?
        - I've brought this up in #543.
[...]

PR #651, opened 2020-10-23
[FIX] Rewrite the MRI/fieldmaps subsection for consistency with the rest of specs

oesteban:
The fieldmaps specification was really particular, not abiding by the formatting and style of other MRI sections.

This PR is meant as a refactor, just clarifying some aspects and standardizing the section before #622 is finally addressed.


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +704,57 @@ sub-<label>/[ses-<label>/]
...
+| ------------ | --------------------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| IntendedFor  | RECOMMENDED           | [string][] or [array][] of [string][] | Contains one or more filenames with paths relative to the participant subfolder. The path needs to use forward slashes instead of backward slashes. This field is OPTIONAL, and in case the fieldmaps do not correspond to any particular scans it does not have to be filled. |
+
+For example:
+
+```JSON
+{
+   "IntendedFor": ["ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
+                   "ses-post/func/sub-01_ses-post_task-motor_run-1_bold.nii.gz"]
```
tsalo, line 752: This example seems unrealistic. Is there any real-world situation where people would acquire one field map in a longitudinal study and then apply it to fMRI scans from multiple sessions? I would lean toward a more common case.

```suggestion
   "IntendedFor": ["ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
                   "ses-pre/func/sub-01_ses-pre_task-motor_run-2_bold.nii.gz"]
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -609,32 +601,34 @@ sub-<label>/[ses-<label>/]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.nii[.gz]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.json
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude1.nii[.gz]
+        [sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]]
```
tsalo, line 604: Including brackets around the whole filename to indicate that it's an optional file isn't used consistently across the specification. I think it really only shows up in the MEG/EEG sections, so I'm not sure if we want to introduce it here.

Also, the schema isn't capable of storing that kind of information at the moment, so those brackets will be removed when/if #610 is merged.

oesteban: Not sure what the alternative would be. Any suggestions?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: Agreed. Sorry I saw your suggestion too late (and didn't commit your suggestion), but I came up with exactly the same example.

oesteban:
This is ready for review

cc/ @bids-standard/raw-mri-anat @bids-standard/raw-mri-dwi @bids-standard/raw-mri-func

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -567,39 +567,31 @@ JSON example:
 
 ## Fieldmap data
 
-Data acquired to correct for B0 inhomogeneities can come in different forms. The
-current version of this standard considers four different scenarios. Please note
-that in all cases fieldmap data can be linked to a specific scan(s) it was
-acquired for by filling the IntendedFor field in the corresponding JSON file.
-For example:
+Data acquired to correct for B0 inhomogeneities can come in different forms.
+The current version of this standard considers four different scenarios.
```
sappelhoff, line 571: I suggest to briefly list the four scenario names already after this sentence.

The table that follows contains 5 rows and is a bit of a clash unless we are already introduced to the 4 scenarios.

oesteban:
Particularly interested in @chrisgorgo's opinion on my commit (https://github.com/bids-standard/bids-specification/pull/651/commits/dd3c87fc4eb0db13824c9222a26e1cd75fb4302d), as I remember some debate when `dir-<label>` was added to the standard.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
sappelhoff: > Not sure what the alternative would be

I think it'd be fine not to communicate requirement level in the "template" section. After all, we are also not distinguishing between recommended and required files either. Users can see which files are optional from the spec text.

oesteban: I'm not particularly attached to this indication of an optional data object, but I think there are two different discussions here.

1. Whether the schema can store this information at the moment -- here the question is whether the schema should store this information, actually. The validator is going to know whether `magnitude2` must be found (phase1/2 fieldmaps) or is optional (phasediff fieldmaps).
2. Whether the template is the right place to state the indication

IMHO, 1) is harder to address, and outside the scope of this PR -- I guess that merging #610 will have to address this question, and if that means that the brackets in the template are gone because some other mechanism implements this not, I'm fine by that (that's how BIDS evolves, anyways).

Then 2) - I think that the Template is basically what most of the people check when preparing their data. Therefore, I think this is the most effective way of expressing that this element is optional. It also needs to be clearly stated within the text, but considering that magnitude2 may and may not be optional (phasediff and phase1/2, respectively), there is a larger chance this is lost in some future refactor. For clarity, I think having the template annotated with requirement levels, especially when irregular, is very useful.

@tsalo: would something like:

```
        sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]  # OPTIONAL
```
work out?

tsalo: There really isn't any mechanism for stating that some suffixes or extensions are optional in combination with others in the schema at the moment, so #610 will erase any indications of that from templates and users will need to rely on the text for that kind of information. I don't think it's that big a deal to address this issue in #610 though.

oesteban: What would you prefer then for this PR?

```
        [sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]]
```
or
```
        sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]  # OPTIONAL
```
I'm leaning towards the `# OPTIONAL` annotation which is more visible and obvious - WDYT?

tsalo: I think `# OPTIONAL` works well.

src/02-common-principles.md
```diff
@@ -149,6 +149,11 @@ data from subject `01`.
 The suffix `eeg` and the extension `.edf` depend on the imaging modality and
 the data format and indicate further details of the file's contents.
 
+Key-value pairs SHOULD NOT be used to encode scanning parameters of the
+corresponding data.
+For instance, in echo-planar imaging MRI, the [`dir-<label>`](./99-appendices/09-entities.md#dir)
+MAY NOT be used as a replacement for the `PhaseEncodingDirection` metadata.
```
tsalo, line 155:
```suggestion
Certain key-value pairs are used to distinguish files with different scanning parameters.
In these cases, the key-value pair SHOULD NOT be used as a replacement for a corresponding metadata field.
The associated field MUST still be present in the file's metadata.
For instance, in echo-planar imaging MRI, the [`dir-<label>`](./99-appendices/09-entities.md#dir) entity MAY be used to distinguish files with different phase-encoding directions, but the file's metadata sidecars MUST still include the `PhaseEncodingDirection` field.
```

This may not be any better, but I thought this deserved a bit more of an explanation.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -567,39 +567,37 @@ JSON example:
...
+| Phase-difference | phasediff        | Some scanners subtract the `phase1` from the `phase2` map and generate a unique `phasediff` file. For instance, this is a common output for the built-in fieldmap sequence of Siemens scanners.                      |
+| Fieldmap         | fieldmap         | Some MR schemes such as spiral-echo (SE) sequences are able to directly provide maps of the B0 field inhomogeneity.                                                                                                  |
+| EPI              | epi              | The phase-encoding polarity (PEpolar) technique combines two or more Spin Echo EPI scans with different phase encoding directions to estimate the underlying inhomogeneity/deformation map.                          |
+
+Two OPTIONAL key/value pairs following more general rules of the specification
+are allowed across all the four scenarios:
+
+  - The OPTIONAL [`run-<index>`](../99-appendices/09-entities.md#run) key/value pair correspond to a one-base index
+    to distinguish multiple fieldmaps with the same parameters.
```
tsalo, line 593:
```suggestion
  - The OPTIONAL [`run-<index>`](../99-appendices/09-entities.md#run) key/value pair corresponds to a one-based index
    to distinguish multiple fieldmaps with the same parameters.
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -609,32 +607,34 @@ sub-<label>/[ses-<label>/]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.nii[.gz]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.json
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude1.nii[.gz]
+        sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]  # OPTIONAL
```
tsalo, line 610: One thing that confuses me about this template (even before this PR)- wouldn't we still have jsons for the magnitude maps?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -672,22 +678,28 @@ sub-<label>/[ses-<label>/]
...
+   "Units": "rad/s"
 }
 ```
 
-### Case 4: Multiple phase encoded directions ("pepolar")
+#### Case 4: Multiple phase encoded directions ("pepolar")
+The phase-encoding polarity (PEpolar) technique combines two or more Spin Echo
+EPI scans with different phase encoding directions to estimate the underlying
+inhomogeneity/deformation map.
+These `_epi.nii[.gz]` files can be a 3D or 4D --
```
tsalo, line 699:
```suggestion
These `_epi.nii[.gz]` files can be 3D or 4D --
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +710,57 @@ sub-<label>/[ses-<label>/]
...
-inhomogeneity/deformation map. Examples of tools using this kind of images are
-FSL TOPUP, AFNI 3dqwarp and SPM. In such a case, the phase encoding direction is
-specified in the corresponding JSON file as one of: `i`, `j`, `k`, `i-`, `j-`,
-`k-`. For these differentially phase encoded sequences, one also needs to
-specify the Total Readout Time defined as the time (in seconds) from the center
-of the first echo to the center of the last echo (aka "FSL definition" - see
-[here](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/topup/Faq#How_do_I_know_what_phase-encode_vectors_to_put_into_my_--datain_text_file.3F) and
-[here](https://lcni.uoregon.edu/kb-articles/kb-0003) how to calculate it). For
-example
+The [`dir-<label>`](../99-appendices/09-entities.md#dir) key/value pair is
```
tsalo, line 713:
```suggestion
The [`dir-<label>`](../99-appendices/09-entities.md#dir) key/value pair used to
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +710,57 @@ sub-<label>/[ses-<label>/]
...
+distinguish between different phase-encoding directions is REQUIRED for these
+files, and SHOULD NOT be used to replace the required `PhaseEncodingDirection`
+(nor any other acquisition parameter, as prescribed by the
+[common principles](../02-common-principles.md#file-name-structure).)
```
tsalo, line 717:
```suggestion
[common principles](../02-common-principles.md#file-name-structure)).
```

tsalo (changes requested):
I have some minor changes, but this is looking really good. Thank you for doing it!

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: The magnitude maps are part of the acquisition (as the phase maps are), so all the metadata is probably best stored only on the phase information files. I'm not sure what metadata could possibly exist that only applies to the magnitude (and doesn't to the phase).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -567,39 +567,37 @@ JSON example:
...
 
-```JSON
-{
-   "IntendedFor": ["ses-pre/func/sub-01_ses-pre_task-motor_run-1_bold.nii.gz",
-                   "ses-post/func/sub-01_ses-post_task-motor_run-1_bold.nii.gz"]
-}
-```
+| **Name**         | `suffix`         | **Description**                                                                                                                                                                                                      |
+| ---------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Magnitude        | magnitude\[1,2\] | Field-mapping MR schemes such as gradient-recalled echo (GRE) generate a Magnitude image to be used for anatomical reference. Requires the existence of Phase, Phase-difference or Fieldmap maps.                    |
```
oesteban, line 583: @effigies, what should we do about the brackets? just spell out the three options?

effigies: Oh, I think I misunderstood what was going on here. I think as long as it's clear from the text how they're used, then having two different notations seems fine as a way of flagging "check the text for details".

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
tsalo: That's true, but I don't think that kind of relationship is supported in the specification. I guess we would call it something like horizontal inheritance?

I could be wrong though.

tsalo: For tools like `pybids`, I think you need sidecar files for the magnitude images in order to extract their metadata. @effigies is that still the case or can `pybids` infer metadata from linked files?

oesteban: I haven't encountered any case where PyBIDS could not correctly query the metadata of fieldmaps.

oesteban:
Care for a couple of green ticks?

cc/ @bids-standard/raw-mri-anat @bids-standard/raw-mri-dwi @bids-standard/raw-mri-func

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +714,60 @@ sub-<label>/[ses-<label>/]
...
-specify the Total Readout Time defined as the time (in seconds) from the center
-of the first echo to the center of the last echo (aka "FSL definition" - see
-[here](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/topup/Faq#How_do_I_know_what_phase-encode_vectors_to_put_into_my_--datain_text_file.3F) and
-[here](https://lcni.uoregon.edu/kb-articles/kb-0003) how to calculate it). For
-example
+The [`dir-<label>`](../99-appendices/09-entities.md#dir) entity is REQUIRED
+for these files.
+This key-value pair SHOULD NOT be used to replace the required `PhaseEncodingDirection`
+(nor any other acquisition parameter, as prescribed by the
+[common principles](../02-common-principles.md#file-name-structure)).
```
effigies, line 721:
```suggestion
(see [File name structure](../02-common-principles.md#file-name-structure)).
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +714,60 @@ sub-<label>/[ses-<label>/]
...
+
+### Expressing the MR protocol intent for fieldmaps
+
+Fieldmaps are typically acquired with the purpose of correcting one or more EPI
+scans under `func/` or `dwi/` for distortions derived from *B<sub>0</sub>*
+nonuniformity.
+This linking between fieldmaps and their targetted data MAY be encoded with the
+`IntendedFor` metadata.
+
+#### Using `IntendedFor` metadata
```
effigies, line 752: I guess the reason for dropping `IntendedFor` from earlier examples is because it's now introduced afterwards? I think this is generally confusing, as it breaks up fields that are going to appear together.

I'm not sure whether we should push this back up to the introductory section or just add:

```Markdown
See [Using `IntendedFor` metadata](#using-intendedfor-metadata)
for details on the `IntendedFor` field.
```

After each example.

effigies (changes requested):

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -609,32 +607,38 @@ sub-<label>/[ses-<label>/]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.nii[.gz]
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.json
         sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude1.nii[.gz]
+        sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_magnitude2.nii[.gz]  # OPTIONAL
```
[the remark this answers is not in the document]
oesteban: I think it is important to mark optional elements on the template. Following that argument all fieldmaps are essentially optional because a BIDS-valid dataset may or may not have them.

There's been some discussion about this annotation with @tsalo. Could you read through that and then let us know if you still think this change should be made (i.e., remove the OPTIONAL annotation in the template)?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -672,22 +682,28 @@ sub-<label>/[ses-<label>/]
...
+| **Key name**   | **Requirement level** | **Data type**            | **Description**                                                                    |
+| -------------- | --------------------- | ------------------------ | ---------------------------------------------------------------------------------- |
+| Units          | REQUIRED              | [string][]               | Units of the fieldmap: Hertz (`Hz`), Radians per second (`rad/s`), or Tesla (`T`). |
+
 For example:
 
 ```JSON
 {
-   "Units": "rad/s",
-   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
```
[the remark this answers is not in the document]
oesteban: Yup, sorry about this.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -672,22 +682,28 @@ sub-<label>/[ses-<label>/]
...
+   "Units": "rad/s"
```
oesteban, line 695:
```suggestion
   "Units": "rad/s",
   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -672,22 +682,28 @@ sub-<label>/[ses-<label>/]
...
 }
 ```
```
oesteban, line 698:
```suggestion

See [Using `IntendedFor` metadata](#using-intendedfor-metadata)
for details on the `IntendedFor` field.

```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +714,60 @@ sub-<label>/[ses-<label>/]
...
+| TotalReadoutTime       | REQUIRED              | [number][]    | See [in-plane spatial encoding](#in-plane-spatial-encoding) table of fields.       |
+
+For example:
 
 ```JSON
 {
-   "TotalReadoutTime": 0.095,
-   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
+   "TotalReadoutTime": 0.095
```
oesteban, line 735:
```suggestion
   "TotalReadoutTime": 0.095,
   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -698,35 +714,60 @@ sub-<label>/[ses-<label>/]
...
 }
 ```
```
oesteban, line 738:
```suggestion

See [Using `IntendedFor` metadata](#using-intendedfor-metadata)
for details on the `IntendedFor` field.

```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
oesteban: > I guess the reason for dropping `IntendedFor` from earlier examples is because it's now introduced afterwards?

Correct. I've edited as suggested.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: Also, I'm growing a feeling that `_sbref` should not exist altogether.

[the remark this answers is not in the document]
effigies: Agreed that all files are optional, so it seems strange to mark one out as super-extra optional in either form. (I did find and mark unresolved that comment thread, btw.)

Immediately below the template is: 

> where the REQUIRED _phasediff image corresponds to the phase-drift map between echo times, the REQUIRED _magnitude1 image corresponds to the shorter echo time, and the OPTIONAL _magnitude2 image to the longer echo time.

I don't understand why we need to add novel markup to the template when the text is right there.

Just as a final point, I would say that I'm not opposed to the concept of indicating in the template that one file in a group is optional, but I am opposed to introducing it in a one-off manner. I would rather we made a principled decision that takes into account all such groupings and applies it to the entire spec in one PR.

That said, I won't block this PR over this quibble.

effigies (approved):

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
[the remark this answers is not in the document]
oesteban: > I don't understand why we need to add novel markup to the template when the text is right there.

Most people will not read that paragraph, especially if they feel they know the specification and are making a quick confirmatory look-up.

> Just as a final point, I would say that I'm not opposed to the concept of indicating in the template that one file in a group is optional, but I am opposed to introducing it in a one-off manner. I would rather we made a principled decision that takes into account all such groupings and applies it to the entire spec in one PR.

I agree with this. For this PR I see three alternatives:

1. Leave it as is now and come back to the issue later (or within #610)
2. Roll it back to `[]` since that nomenclature, even though not formalized, is apparently present elsewhere in BIDS (and finally address it within #610).
3. Remove any annotations (and bring them up in a separate PR or #610).

I think 3 would be the best option, but I'm afraid that this will concept will be probably lost (I'm honestly not willing to go through the whole spec to propose something in a separate PR or in #610). So, after the discussion with @tsalo I think I prefer option 1.

oesteban:
Can I ask for one more blessing?

tsalo (approved):

effigies:
5 business days since last change and 2 positive reviews. Merging.

Thanks for this effort, @oesteban!

PR #666, opened 2020-11-09
[ENH] Add Defaced and SkullStripped boolean fields to raw MR images

### LINKED ISSUE #664: Add OPTIONAL SkullStripped metadata to any MR images
Derivatives has introduced a REQUIRED metadata field `SkullStripped` to imaging files.

https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#preprocessed-coregistered-andor-resampled-volumes

One situation we have in OpenNeuro is that some datasets were skull-stripped instead of merely defaced. This has led to fMRIPrep developing a heuristic to detect skull-stripped images. If datasets could specify in metadata that images have been skullstripped, it would bypass the need to continually refine the heuristic to account for false positives and negatives.

It violates a bit the notion of "raw" datasets as data-off-the-scanner, but if we consider skull-stripping as a kind of extreme defacing (which is how it seems to have been used), then it's not inconsistent with previous decisions.

I would like to make this an OPTIONAL field for raw MR data. How do people feel about this?

effigies:
I decided to add a Defaced field along with SkullStripped. And I made them RECOMMENDED so that the validator can eventually prod people to fill them in.

Closes #664.

### DIFF
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -107,6 +107,16 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, and so on.
 | ----------------------------- | --------------------- | ------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | AnatomicalLandmarkCoordinates | RECOMMENDED           | [object][] of [arrays][] | Key:value pairs of any number of additional anatomical landmarks and their coordinates in voxel units (where first voxel has index 0,0,0) relative to the associated anatomical MRI (for example, `{"AC": [127,119,149], "PC": [128,93,141], "IH": [131,114,206]}`, or `{"NAS": [127,213,139], "LPA": [52,113,96], "RPA": [202,113,91]}`). Each array MUST contain three numeric values corresponding to x, y, and z axis of the coordinate system in that exact order. |
 
+### Curation details
+
+These keys allow dataset curators to indicate modifications made to imaging data
+for the purposes of preserving privacy.
+
+| **Key name**  | **Requirement level** | **Data type** | **Description**                                                                         |
+| ------------- | --------------------- | ------------- | --------------------------------------------------------------------------------------- |
+| Defaced       | RECOMMENDED           | [boolean][]   | Indicates whether the image has been defaced as part of the dataset preparation.        |
+| SkullStripped | RECOMMENDED           | [boolean][]   | Indicates whether the image has been skull-stripped as part of the dataset preparation. |
+
 ### Institution information
 
 | **Key name**                | **Requirement level** | **Data type** | **Description**                                                                                                                                                          |


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -107,6 +107,16 @@ Useful for multimodal co-registration with MEG, (S)EEG, TMS, and so on.
 | ----------------------------- | --------------------- | ------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | AnatomicalLandmarkCoordinates | RECOMMENDED           | [object][] of [arrays][] | Key:value pairs of any number of additional anatomical landmarks and their coordinates in voxel units (where first voxel has index 0,0,0) relative to the associated anatomical MRI (for example, `{"AC": [127,119,149], "PC": [128,93,141], "IH": [131,114,206]}`, or `{"NAS": [127,213,139], "LPA": [52,113,96], "RPA": [202,113,91]}`). Each array MUST contain three numeric values corresponding to x, y, and z axis of the coordinate system in that exact order. |
 
+### Curation details
```
tsalo, line 110: Should this go under the Anatomical MRI section or is it generic across MRI data types?

satra:
if this is supposed to apply to a broader set of scans today, would this be useful to extend to a general set of things that scanners can currently do?  

for example (based on siemens sequences):
- motion corrected (PACE)
- distortion corrected (can't remember the sequence name)
- bias field corrected (normalize on)

or humans/setups may do: 
- motion corrected (external tracking)
- motion reduced (caseforge helmets, etc.,.)

i.e allow a more general set of terms associated with various corrections, knowing that a given scan may have multiple corrections applied to it. and that future sequences may do more things.

or does skullstripping stand out as one of the more drastic things and you want to limit to that.

also i agree with @tsalo to make this more generic than just anatomical scans.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
effigies: Skull-stripping can in principle be done to any MR image, since there is signal outside the brain that can show up in many sequences (e.g., the eyes show up quite clearly in EPI). Defacing is not well defined in non-anatomical modalities, but it seemed appropriate to keep these together.

What if I changed to `RECOMMENDED for anatomical images`?

tsalo: If it's feasible for all MR image types, then I'm good with it as-is. Thanks for the clarification.

effigies:
I'm not trying to limit to skull stripping in principle, but there are a couple factors that make this timely:

1) A practical use case. Detecting skull-stripping is hard, and most algorithms are not idempotent. A metadata field presents a straightforward alternative to a development rabbit hole.
2) Already defined in derivatives. The scope for breaking backwards compatibility now or in the future is low.

I think in the long run it does make sense to note steps taken by scanners. Process-wise, I think it makes sense to think about how to specify them in derivatives to cover both the scanner and preprocessing cases, and then see if we can move them back to BIDS-Raw in a backwards compatible way.

So we may want to look at BEPs 11 and 12 for whether we're defining things in a way they can be pulled back into Raw when appropriate.

----

Does this approach sound reasonable? And does the proposed implementation of these two terms make extending to these cases harder?

satra:
> Does this approach sound reasonable? 

yes

> And does the proposed implementation of these two terms make extending to these cases harder?

i am fine with the current approach. however, sometimes using a `boolean` property raises the question whether one would ever include a recommended property if it was `False`. if `False` is not required, then instead one could consider the term as a value target for a property. for example `"CorrectionsApplied":  [ "Defacing", "MotionCorrection"]`

effigies:
> i am fine with the current approach. however, sometimes using a `boolean` property raises the question whether one would ever include a recommended property if it was `False`. if `False` is not required, then instead one could consider the term as a value target for a property. for example `"CorrectionsApplied": [ "Defacing", "MotionCorrection"]`

Fair. `SkullStripped` is defined in derivatives, so we can either use it as-is and blur the distinction, or have a distinct way of marking raw data. FWIW, I think the list approach for derivatives was rejected as too difficult to standardize (at the time, anyway), but it might be a constrained enough space for raw.

oesteban:
Sorry to push back in the opposite direction.

I think that skull-stripped datasets should follow BIDS-Derivatives specifications.

Given the practical situation (datasets already being shared with skull-stripped images and software failing to detect this automatically), I think then adding a custom metadata key for OpenNeuro should be enough.

I would take this to some extreme - e.g., if your defacing tool resampled the dataset in any way (typically done to align the AC-PC line with axial), I would also push these to the BIDS-Derivatives specs, if possible.

My worry is that someone who would be allowed to share data just-defaced could feel incentivized to share skull-stripped brains for obscure reasons (or just because files take a lot less to upload, who knows) while they would not make the same decision if the dataset cannot be called BIDS-raw anymore.

That said, I agree with @satra that a lot of the parameters (of the scanner) about corrections and equalizations happening at scan time and reconstruction could have better representation on the specs.

effigies:
> adding a custom metadata key for OpenNeuro should be enough.

I don't understand what this means, practically. OpenNeuro absolutely does not care about whether an image is skull-stripped, so the metadata isn't for it. fMRIPrep does care because we don't know how to identify the case reliably.

oesteban:
> I don't understand what this means, practically.

As per #664, this PR stems from this problem:

> One situation we have in OpenNeuro is that some datasets were skull-stripped instead of merely defaced. This has led to fMRIPrep developing a heuristic to detect skull-stripped images. If datasets could specify in metadata that images have been skullstripped, it would bypass the need to continually refine the heuristic to account for false positives and negatives.

So these are the connections I made:

BIDS is updated -> OpenNeuro sets the new metadata for those known images where this applies -> fMRIPrep checks for this metadata instead of using an unreliable heuristic.

What I meant is that the first link of the chain is perfectly removable. And, if at some point OpenNeuro could validate and distribute derivatives, then move these anatomical images to corresponding derivatives trees.

oesteban:
For example, ds001 has some 10 subjects and only one is skull-stripped. If, this one subject does not have the T1w under the BIDS dataset, and has it under the derivatives folder instead it is very obvious what is going on (you don't know what is the exact source of this irregularity, but you do know that the original T1w of this subject is not available).

With a metadata key, this salient irregularity is washed over and you need to check into all files' metadata before doing anything. Trivial for software, yes, but prone to human errors.

effigies:
So the notion is to add an unspecified, hidden field that fMRIPrep will respect? That feels less satisfying than an incursion of "too much derivativation" into raw datasets.

effigies:
What if we add a line:

```
It is RECOMMENDED to modify the data as little as possible while complying with legal 
and ethical obligations.
```

And then the validator warns on `"SkullStripped": true` (or whatever the final version of this looks like) in a raw dataset?

oesteban:
I'm very much +1 on adding such a recommendation, regardless of the fate of this particular PR.

Then regarding the proposed new metadata, if the BIDS-Derivatives were settled in (i.e., validated as reliably as the BIDS-Validator gets today on raw data) - would you add this option here or just use BIDS-Derivatives for these anatomical MRI?

I think this change is pushed by the fact that BIDS-Derivatives have not reached yet first-class citizenship in the ecosystem, rather than an actual hole in the BIDS specs. By this I mean I understand the intention and it seems like a sensible temporal solution. However, _temporal solutions_ are a risky business within BIDS 1.0 as we will not be able to deprecate them when BIDS-Derivatives are handled seamlessly.

Adding these kinds of loopholes dilutes the (already low) clarity of _raw_ as _data spit by the scanner_, and adding it to anatomical images (where _raw_ is perhaps a bit clearer than for other sequences that need, e.g., the assistance of a compute node or GPU for reconstruction) makes me a bit nervous.

That said, I think I've had sufficient time/space to express my opinion - I will not block this PR, and please, others weigh in :)

sappelhoff:
I think that Oscar's point is generally valid, however in this specific case I think that Chris' proposal from https://github.com/bids-standard/bids-specification/pull/666#issuecomment-726761989 is fair.

- If providing this field in BIDS makes it much easier for downstream pipelines to work with the data, that's a win.
- If we appropriately warn users to please not abuse this feature, I would trust most users to not intentionally screw up their data.
- And those users that *do* screw up their data, usually find even worse ways.

oesteban:
> * If providing this field in BIDS makes it much easier for downstream pipelines to work with the data, that's a win.

I disagree with this - *fMRIPrep* (which is the example at hand) can suggest users modify their input BIDS structure to have this metadata without it being standard. We've done it in the past with, for instance, manually delineated masks of tumors to allow spatial normalization. Today, this can/should be done using BIDS-Derivatives, but *fMRIPrep* asked its users to include the mask files in the BIDS-raw tree when BIDS-Derivatives did not exist yet (*fMRIPrep* still does this because we haven't been able to update).

> * And those users that _do_ screw up their data, usually find even worse ways.

This is the problem - because an intentional degradation of the data using this feature would go through easily, as it is not one of those _worse ways_. The landscape of researchers is definitely nonuniform - some are deeply convinced about data sharing and will strive for distributing the most pristine possible data. However, some are just starting and full of concerns they only half-way understand. These researchers are at risk of misunderstanding this kind of "feature". If BIDS-Derivatives hadn't existed, I'd have a more flexible opinion about this - but IMHO, the appropriate tool is right there.

CGSchwarzMayo:
I came upon this old thread in looking for a .json field that I could population to indicate that images had been de-faced, and with what software/version they had been de-faced. I think the de-facing field has changed a bit since 2020 and I'm hoping the BIDS group might be willing to revisit it. 

Some major points of change in the past few years:
1) De-facing is becoming much more common in research areas it didn't used to be. In particular, most large brain aging and dementia imaging studies (e.g. ADNI, SCAN, A4, ALLFTD, MCSA) are de-facing where they did not before, and these, along with datasets that have been de-facing for longer periods such as HCP and UK-Biobank, make up a substantial portion of publicly available brain imaging datasets.
2) De-facing is no longer limited to structural MRI. Rather, it is also being applied in PET (FDG, amyloid, tau, and others), CT, and a growing number of brain MRI sequences (T1, T2, FLAIR, T2*/GRE, and even perfusion/ASL). Diffusion and functional/BOLD MRI are the most prominent exceptions that, for now, are not known to be identifiable without de-facing, but most other common brain images are, at this point.
3) De-faced images are increasingly stored and distributed in DICOM. This is the primary and only distribution method for most of the studies listed above. These DICOM are marked appropriately using DeidentificationMethod and DeidentificationMethodCodeSequence, but when they are converted to .nii, this information is lost because there is no .json field to transfer it to. 
4) A continued proliferation of de-facing software has led to a need to track not just whether images have been de-faced but with what software and with what software version, because data de-faced using different software may not be combinable.

One solution proposed here has been Bids Derivatives. De-facing is actually discussed as the first example on the BIDS-Derivatives page [https://bids-specification.readthedocs.io/en/stable/derivatives/introduction.html]() When de-facing is performed during the curation process, the image would be stored as if it were raw data e.g. sub-01/anat/sub-01_T1w.nii.gz. From what I can tell, this solution makes the image indistinguishable from unmodified raw data. While for most downstream consumers the de-faced data will be the only copy they ever see, so treating it as original primary is appropriate, I don't see that there are standard fields in BIDS-specified  .json files that would allow specifying the de-facing software or version used in a standardized and machine-readable format. 

Without a standardized .json field, there is no place to store this information that is critical to both users and database maintainers in maintaining and understanding the images. Adding such a field would also allow its preservation when de-faced DICOM are converted to .nii via dcm2niix and similar programs, vs. now the information is lost because no BIDS fields have been standardized. 

Would the BIDS maintainers be interested in re-visiting this idea with top-level fields for deface-software and deface-software-version? While I'm personally less invested in skull-stripping, I would envision the same argument for software+version would be appropriate for skull-strip related tags as well.

effigies:
The situation has definitely changed since this was opened. Last year, the following text was adopted:

> The following sections cover additions to and divergences from "raw" BIDS. Raw data are data that have been curated into BIDS from a non-BIDS source. If a dataset is derived from at least one other valid BIDS dataset, then it is a derivative dataset.

A dataset that is skull-stripped or defaced prior to or during curation into BIDS is raw, by definition.

> 3\. De-faced images are increasingly stored and distributed in DICOM. This is the primary and only distribution method for most of the studies listed above. These DICOM are marked appropriately using DeidentificationMethod and DeidentificationMethodCodeSequence, but when they are converted to .nii, this information is lost because there is no .json field to transfer it to.

I would be +1 for adding these DICOM fields to BIDS directly.

* https://dicom.innolitics.com/ciods/rt-dose/patient/00120063
* https://dicom.innolitics.com/ciods/rt-dose/patient/00120064

While this is useful for converting from de-identified DICOM, it would be good to have recommendations for how a curating tool should populate these fields. Something like:

````
A conversion or curation tool SHOULD append a description of the de-identification
process performed, along with the name and version of the tool.
For example:

```json
  "DeidentificationMethod": ["skull-strip; bids-curator-X (v1.0)"],
```
````

I don't have a strong opinion on what recommendation is to be made. Doing something similar to what is done in DICOM is probably best, but I haven't seen these.

effigies:
Actually, this PR is so old that this conversation will be buried. I'm going to close it and start a new one based on your comment.

PR #669, opened 2020-11-11
[ENH] Bep 005: Arterial Spin Labeling

sappelhoff:
follow up of https://github.com/bids-standard/bids-specification/pull/652

closes https://github.com/bids-standard/bids-specification/issues/41

# rendered draft: https://bids-specification.readthedocs.io/en/bep-005/

TCON agenda and notes: https://docs.google.com/document/d/1AVj8SvVnkNcInvYpljFQgc3mpkxnjVvKTbRVigMHKBs/edit

PR #652, opened 2020-10-23
[ENH] Bep 005: Arterial Spin Labeling

HenkMutsaerts:
These are the pre-existing BIDS keys for which ASL has additional requirements:

EchoTime 
FlipAngle
MagneticFieldStrength 
PulseSequenceType 
RepetitionTime 
SliceTiming
VolumeTiming

# Edit (10/28): [View the draft extension in the browser](https://bids-specification--652.org.readthedocs.build/en/652/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#arterial-spin-labeling-perfusion-data)

PR #671, opened 2020-11-16
[ENH] BEP001 - RepetitionTimeExcitation and RepetitionTimePreparation

agahkarakuzu:
*** 
### Headnote

A while ago we started #508, where you can find information on the overall purpose of `BEP001`. The PR #508 was closed as discussions called for some major revisions and to split `BEP001` pull requests into manageable parts.

*** 

Dear BIDS community, 

The current [definition](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#required-fields) of the `RepetitionTime` field was subject to a long [discussion](https://groups.google.com/g/bids-discussion/c/MLUqmcD1XSY/discussion), because:

1. The definition of the term is specific to the task imaging data; therefore, not applicable for other MRI modalities that are not designed to collect volumetric data within a single `TR`. 
2. If made to the current `RepetitionTime` field, the change required by (1) would break retro-compatibility for the main user-base of the BIDS. 

To have best of both worlds, we came up with a [solution](https://groups.google.com/g/bids-discussion/c/UXXOOEQ4t0M?pli=1) by introducing 2 more metadata fields: `RepetitionTimeExcitation` and `RepetitionTimePreparation`. These definitions i) do not interfere with the current definition of the `RepetitionTime`, retaining compatibility with massive amount of fMRI data and ii) cover the basis for simple (e.g. `T1w`) or more complex MRI acquisitions (e.g. `MP2RAGE`) with proper definitions. 

In this **Part-2** of BEP001 pull requests, we propose including these metadata fields to enable new anatomy imaging applications without affecting the existing datasets. In a recent meeting, we also realized that these definitions can be useful for the BEP005 team, hence pinging #669. 

Big thanks to @KirstieJane for carefully adjusting the wording in these suggested changes and to @mathieuboudreau for his help with the terms.

On behalf of the BEP001 core team:

 Gilles de Hollander (@Gilles86), Alberto Lazari (@lazaral), Christophe Phillips (@ChristophePhillips), Kirstie Whitaker (@KirstieJane), Tibor Auer (@tiborauer). 

Best regards.


tsalo:
I understand why this proposal is necessary for BIDS 1.X, but, to clarify, would this change be adequately resolved in BIDS 2.0 by the proposal in https://github.com/bids-standard/bids-2-devel/issues/18? I think it's important that we have a corresponding proposal for a cleaner, backwards-incompatible definition of RepetitionTime that works for all groups for 2.0, even if 2.0 won't be released any time soon.

agahkarakuzu:
@tsalo when backwards compatibility is no longer an issue, replacing `RepetitionTime` definition with that of `RepetitionTimeExcitation` (with minor edits to generalize) and declaring `RepetitionTimeExcitation` as deprecated would solve it. 

On the other hand, `RepetitionTimePreparation` would still be needed. It would only require it to be used together with `RepetitionTime` instead of `RepetitionTimePreparation`.

tsalo:
Thank you for the clarification. Assuming this is merged with minimal changes, we should make sure that plan is reflected in the `bids-2-devel` issue.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -194,6 +194,8 @@ fields specific to anatomical scans:
...
 | ContrastBolusIngredient | OPTIONAL              | [string][]    | Active ingredient of agent. Values MUST be one of: IODINE, GADOLINIUM, CARBON DIOXIDE, BARIUM, XENON Corresponds to DICOM Tag 0018,1048. |
+| RepetitionTimeExcitation | OPTIONAL. The time in seconds between successive excitation pulses that excite a spin system. The DICOM tag that best refers to this parameter is [(0018, 0080)](http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,0080)). This field may be used together with the `RepetitionTimePreparation` for certain use cases, such as [MP2RAGE](https://infoscience.epfl.ch/record/172927/files/mp2rage.pdf). Use `RepetitionTimeExcitation` (in combination with `RepetitionTimePreparation` if needed) for anatomy imaging data rather than `RepetitionTime` as it is already defined as the amount of time that it takes to acquire a single volume in the [task imaging data](#task-including-resting-state-imaging-data) section.|
+| RepetitionTimePreparation | OPTIONAL. The period of time in seconds that it takes a preparation pulse block to re-appear at the beginning of the succeeding (essentially identical) pulse sequence block. |
```
effigies, line 198:
```suggestion
| RepetitionTimeExcitation | OPTIONAL              | [number][]    | The time in seconds between successive excitation pulses that excite a spin system. The DICOM tag that best refers to this parameter is [(0018, 0080)](http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,0080)). This field may be used together with the `RepetitionTimePreparation` for certain use cases, such as [MP2RAGE](https://infoscience.epfl.ch/record/172927/files/mp2rage.pdf). Use `RepetitionTimeExcitation` (in combination with `RepetitionTimePreparation` if needed) for anatomy imaging data rather than `RepetitionTime` as it is already defined as the amount of time that it takes to acquire a single volume in the [task imaging data](#task-including-resting-state-imaging-data) section.|
| RepetitionTimePreparation | OPTIONAL              | [number][]    | The period of time in seconds that it takes a preparation pulse block to re-appear at the beginning of the succeeding (essentially identical) pulse sequence block. |
```

effigies (commented):
LGTM, just reformatting to specify the requirement level and data type in separate columns. We're also going to need to update the table fences in both tables to satisfy the linter.

Remi-Gau (approved):
Makes sense to me.

patsycle (commented):
Nice addition! I noticed that only a 'number' is allowed, why not an array of numbers? For example, that would be prefered for BEP005 (asl), but I assume this might be the case for other extensions?

agahkarakuzu:
@patsycle so far we discussed that for the metadata fields that will have an associated entity such as `flip` and `echo`. We discussed in the meeting that array option will be inserted somewhere in BEP001 PRs, where exactly that information is going to be inserted is also a question for BIDS maintainers. 

Should it be under `Data Type` column in the table, given as `number OR array of numbers` ? If this is the case, then we need clear descriptions of when either option is applicable.

patsycle:
@agahkarakuzu Indeed, I guess this is under DataType as this is already the case for SliceTiming and VolumeTiming. We will take a look at it already and we can further discuss on Friday.

Gilles86:
@patsycle, I would like to stress that both `SliceTiming` and `VolumeTiming` contain arrays that pertain to parameters that _necessarily_ have to be arrays. By definition, they *cannot* be a single number. This is very different for the `RepetitionTime`-fields, as well as all almost other fields in this table. By default, they are defined over all Volumes/Slices. This default easily reaches the 80/20-rule. 

I would like to warn against now generalizing _all_ `number`-fields in this table to `array of numbers` or even `number OR array of numbers`, just to accommodate ASL acquisitions. If we choose that road, what do we do with the `SliceTiming`? Do we put `array of numbers OR array of array of numbers` there, just in case the timing of the slices changes from volume to volume?

And what about the validator? Does it now have to check every single field in the json-sidecar against the number of volumes in the nifti-file?

We already have a mechanism for iterations over parameters with the `echo-<echo>`, `flip-<flip angle>`, etc key-value pairs. By making things so flexible as `number OR array-of-numbers`, we introduce a lot of ambiguity into the standard and this is exactly what we want to avoid.

I appreciate that the ASL-community would like to stick to their standard practice and put volumes with different acquisition parameters in the same 4D nifti-file and I think we should try to accommodate this within BIDS, to aid broad adoption. However, I don't think practices in ASL should dictate practices in the rest of the field of (q)MRI. Especially not if we just spend 3+ years working out that such a practice leads to problematic edge cases in multiple domains (e.g., when you iterate over more than one parameter like in MPM or MP2RAGE-ME) and hardens the life of BIDS-app developers (if there are multiple ways to store a MP2RAGE in BIDS-format, most of your development time is gonna be invested in very bug-sensitive bookkeeping code).

Can't we make special cases for the fields where this is absolutely necessary for ASL in the table you put under `*_asl.json file` in your [pull request](https://github.com/bids-standard/bids-specification/blob/60be75e9fa408d11487f1929ebca625276fdf87f/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md)?

effigies:
Created https://github.com/bids-bep001/bids-specification/pull/5 to suggest language for variable RTE/RTP. There was discussion about whether this would be enough to make it clear that qMRI sequences ought to parameterize their volumes by filename rather than metadata, so I would appreciate suggestions for additional language.

patsycle:
@agahkarakuzu @effigies Probably stupid question, but what are the next steps for this? I guess Agah has to move the suggestions from https://github.com/bids-bep001/bids-specification/pull/5 to here. And after that? Will this one be merged at the same time as BEP005 or will it first go into BEP005?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -194,6 +194,8 @@ fields specific to anatomical scans:
...
 | ----------------------- | --------------------- | ------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
 | ContrastBolusIngredient | OPTIONAL              | [string][]    | Active ingredient of agent. Values MUST be one of: IODINE, GADOLINIUM, CARBON DIOXIDE, BARIUM, XENON Corresponds to DICOM Tag 0018,1048. |
+| RepetitionTimeExcitation | OPTIONAL              | [number][]    | The time in seconds between successive excitation pulses that excite a spin system. The DICOM tag that best refers to this parameter is [(0018, 0080)](http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,0080)). This field may be used together with the `RepetitionTimePreparation` for certain use cases, such as [MP2RAGE](https://infoscience.epfl.ch/record/172927/files/mp2rage.pdf). Use `RepetitionTimeExcitation` (in combination with `RepetitionTimePreparation` if needed) for anatomy imaging data rather than `RepetitionTime` as it is already defined as the amount of time that it takes to acquire a single volume in the [task imaging data](#task-including-resting-state-imaging-data) section.|
```
Remi-Gau, line 197:
```
[MP2RAGE](https://infoscience.epfl.ch/record/172927/files/mp2rage.pdf)
```

Detail I am not sure about because I don't think it is consistent through the specs, but isn't it better to have a DOI for references? 

I think we sort of have in mind to automate the "bibliography" part of the specs, no?

effigies:
@patsycle I would suggest merging this into master, not into BEP-005, when it's ready. But BEP-005 should proceed as if this is merged, to avoid having to wait on it. (If it's useful to merge into BEP-005 so that the document can be read as a whole, that's fine, too.)

agahkarakuzu:
@patsycle I will move my commits from `bids-bep005` pull request to here soon. But I think the [changes](https://github.com/bids-bep001/bids-specification/pull/5#discussion_r528800173) you can make to the `FlipAngle` and `EchoTime` on BEP005 are independent from these changes time- and practice-wise. We are aiming that when both #671 and #652 are merged, there is a coherent language/convention.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
agahkarakuzu: Sure @Remi-Gau, I will replace them with DOI URLs.

agahkarakuzu: 1a9412c32bdb6aa14314b1e0df2cb63c70a81e59

patsycle:
@effigies @agahkarakuzu Good, thanks for the information. Agah, I added your suggestions for FlipAngle and EchoTime in BEP005.

agahkarakuzu:
Thank you @patsycle! Happy that now we have archetype definitions in the main text to clarify when and which convention applies.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -191,9 +191,11 @@ JSON file. See [Common metadata fields](#common-metadata-fields) for a
...
+| RepetitionTimeExcitation | OPTIONAL             | [number][] | The time elapses (in seconds) between two successive excitations. The DICOM tag that best refers to this parameter is [(0018, 0080)](http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,0080)). This field may be used together with the `RepetitionTimePreparation` for certain use cases, such as [MP2RAGE](https://doi.org/10.1016/j.neuroimage.2009.10.002). Use `RepetitionTimeExcitation` (in combination with `RepetitionTimePreparation` if needed) for anatomy imaging data rather than `RepetitionTime` as it is already defined as the amount of time that it takes to acquire a single volume in the [task imaging data](#task-including-resting-state-imaging-data) section. |
```
effigies, line 197: I think we can use "interval" to mean a duration of time. "interval, in seconds," should be doubly clear in case somebody has another notion for interval.

```suggestion
| RepetitionTimeExcitation | OPTIONAL             | [number][] | The interval, in seconds, between two successive excitations. The DICOM tag that best refers to this parameter is [(0018, 0080)](http://dicomlookup.com/lookup.asp?sw=Tnumber&q=(0018,0080)). This field may be used together with the `RepetitionTimePreparation` for certain use cases, such as [MP2RAGE](https://doi.org/10.1016/j.neuroimage.2009.10.002). Use `RepetitionTimeExcitation` (in combination with `RepetitionTimePreparation` if needed) for anatomy imaging data rather than `RepetitionTime` as it is already defined as the amount of time that it takes to acquire a single volume in the [task imaging data](#task-including-resting-state-imaging-data) section. |
```

@patsycle @HenkMutsaerts Is this also supposed to be an array? Or does ASL only need RTP to be variable?

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -191,9 +191,11 @@ JSON file. See [Common metadata fields](#common-metadata-fields) for a
...
+| RepetitionTimePreparation | OPTIONAL            | [number][] or [array][] of [numbers][] | The period of time in seconds that it takes a preparation pulse block to re-appear at the beginning of the succeeding (essentially identical) pulse sequence block. The data type number may apply to files from any MRI modality concerned with a single value for this field. The data type array provides a value for each volume in a 4D dataset and should only be used when the volume timing is critical for interpretation of the data, such as in [ASL](#arterial-spin-labeling-perfusion-data).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
effigies, line 198:
```suggestion
| RepetitionTimePreparation | OPTIONAL            | [number][] or [array][] of [numbers][] | The interval, in seconds, that it takes a preparation pulse block to re-appear at the beginning of the succeeding (essentially identical) pulse sequence block. The data type number may apply to files from any MRI modality concerned with a single value for this field. The data type array provides a value for each volume in a 4D dataset and should only be used when the volume timing is critical for interpretation of the data, such as in [ASL](#arterial-spin-labeling-perfusion-data).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```

effigies (commented):
A couple very quick thoughts and a question. I'll look again more thoroughly ASAP.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
patsycle: @effigies For ASL, we only need RTP to be an array. So RTE can stay a number. Thanks ;)

effigies (approved):
Since only RTP needs to have an array, I think this looks good as-is. Please consider the rewording suggestions for "interval", but otherwise I think this is very clear.

sappelhoff (approved):
The link check failure is due to the as of yet missing ASL link: `[ASL](#arterial-spin-labeling-perfusion-data)`

I'd be fine merging this, given that we are planning to merge ASL very soon and that apart from qMRI and ASL, not many other PRs are expected to be worked on these days.

PR #678, opened 2020-11-23
[MISC] use RFC 2119 language in legend of the "volume timing" table

Remi-Gau:
Small update to be consistent with the incoming legend on BEP005.


sappelhoff (approved):
good catch! Thanks

PR #679, opened 2020-11-23
[DOC] add contributing guidelines to add figures in the specs

Remi-Gau:
The need for guidelines regarding figures in the specifications was mentioned in a meeting related to BEP005 and was also pointed in the PET BEP.

This is a PR to get some sort of guidelines started.

### Questions

- Do we want to have specific format recommendations for non-vector graphics? png?
- Maybe some size limits on the images too to avoid having people send us 15 Mo images. :-)

I might be overlooking some aspects with the pdf conversion process too that I don't know too well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,7 @@ Jump to the following sections:
 -   [Understanding issues](#understanding-issues)
 -   [Writing in markdown](#writing-in-markdown)
 -   [Fixing Remark errors from Travis](#fixing-travis-remark-errors)
+-   [## Adding a figure to the specifications](#adding-a-figure-to-the-specifications)
 -   [Making a change with a pull request](#making-a-change-with-a-pull-request)
 -   [Example pull request](#example-pull-request)
 -   [Commenting on a pull request](#commenting-on-a-pull-request)
@@ -232,6 +233,29 @@ mv flagged_file_fixed.md flagged_file.md
 git add flagged_file.md
 git commit -m 'STY: Fixed Markdown style'
 ```
+
+## Adding a figure to the specifications
+
+> A figure is worth a 1000 words! 
+
+If you think that a figure or a picture can help summarize several aspects or notions of the 
+specifications, do not hesitate to make a suggestion by showing a draft in a Github issue.
+
+You can then submit your image in a pull request.
+
+Images should be added to an `images` folder that is at the same level as the Markdown file
+where your image will be added. For example if you want to add a figure `figure01.png` to 
+`src/05-derivatives/01-introduction.md` then your image should go 
+`src/05-derivatives/images/figure01.png`.
+
+Figures can be inserted in a Markdown like this:
+
+```markdown
+![ ](relative_path_to_file "alternative name")
+```
+
+If you are adding a figure (and not picture) make sure to also join a vector format 
+of that figure (ideally as an `.svg` file).
 
 ## Making a change with a pull request
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -257,14 +257,14 @@ Figures can be inserted in a Markdown like this (see also
 ```
 ### Recommendations for figures
 
-1. Try to keep the file size of your figure relatively small (inferior to 500 Ko) 
+1. Try to keep the file size of your figure relatively small (smaller than 500 Kb) 
 to keep the repository light and reduce the load time of the specs 
 for people who do not necessarily have broad-band internet.
 
-1. Figures in the main part of the specification should aim to be for very "comprehensive" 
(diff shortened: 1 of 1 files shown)


Remi-Gau:
### Other questions

Should we mention that the main specs should be for very "comprehensive" figures and that for "smaller" figures they could go in the appendices or the starter kit?

sappelhoff:
> Should we mention that the main specs should be for very "comprehensive" figures and that for "smaller" figures they could go in the appendices or the starter kit?

yes


Just an early comment that I'd like all people who upload an image to include a README file that details where the image came from and how it can be reproduced. Preferably with a link to the file that generated the figure.


Also, I think that several people may advocate for not having image files in this repository at all, and instead including them in a submodule or in some other way. :thinking: so that may be worth thinking about as well. The motivation behind that would be that image files are usually not well tracked with git, and tend to bloat up the history and repo size unnecessarily.

sappelhoff (commented):
to resolve:

- [ ] discuss if it makes more sense to have an "images" repo and include it as a [git submodule](https://git-scm.com/book/de/v2/Git-Tools-Submodule) in bids-specification, rather than adding images here in this repo directly
- [ ] add a part to the documentation that users should please add a README (or edit one if already there) that details how their figure was created, preferably with link to the source.

Remi-Gau:
>     * discuss if it makes more sense to have an "images" repo and include it as a [git submodule](https://git-scm.com/book/de/v2/Git-Tools-Submodule) in bids-specification, rather than adding images here in this repo directly

I am tempted to kick that one down the road and cross that bridge when we get to it. The reason is that I don't exactly expect the specification to start accumulating images like a "family album".

Another reason is that sub-module adds another layer of technological complexity that could make it harder for people to contribute. ( "A repo within a repo?? I just want to add an image, not shoot a remake of Inception!" :wink: )

So I would suggest clear guidelines on not submitting 15 mb images as a first line of defense. 

What do people think?

sappelhoff:
> What do people think?

I agree with that. We don't strictly need the technologically advanced method right now.

guiomar:
Thanks! I think that figures are always enriching, so would very much like to have illustrations in the specs to facilitate comprehension. I don't know about the technical details. But probably each bep may have max 2 or 3, is this too much? Maybe if in the future, the technical details are difficult to handle, we can explore these repos within repos, but for now we can proceed the simple way, with some indications :)

I also wonder if it could be useful to recommend each bep to have at least one explanatory figure.

CONTRIBUTING.md
```diff
@@ -233,6 +234,44 @@ git add flagged_file.md
...
+
+Figures can be inserted in a Markdown like this (see also
+[Markdown-Cheatsheet](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#images)):
+
+```markdown
+![text to show if image does not load](relative_path_to_file "text to show when hovering over image")
+```
+### Recommendations for figures
+
+1. Try to keep the file size of your figure relatively small (inferior to 500 Ko)
```
sappelhoff, line 260: did you mean this?

```suggestion
1. Try to keep the file size of your figure relatively small (smaller than 500 Kb) 
```

CONTRIBUTING.md
```diff
@@ -233,6 +234,44 @@ git add flagged_file.md
...
+![text to show if image does not load](relative_path_to_file "text to show when hovering over image")
+```
+### Recommendations for figures
+
+1. Try to keep the file size of your figure relatively small (inferior to 500 Ko) 
+to keep the repository light and reduce the load time of the specs 
+for people who do not necessarily have broad-band internet.
+
+1. Figures in the main part of the specification should aim to be for very "comprehensive" 
+but "smaller" figures can find their home in the appendices or the BIDS-starter-kit.
```
sappelhoff, line 265:
```suggestion
1. Figures in the main part of the specification should aim to be very "comprehensive" 
but "smaller" figures can find their home in the appendices or the BIDS-starter-kit.
```

CONTRIBUTING.md
```diff
@@ -233,6 +234,44 @@ git add flagged_file.md
...
+
+1. If you are adding a figure (and not picture) make sure to also join a vector format 
+of that figure (ideally as an `.svg` file) as this makes it easier to edit it in the
+future.
```
sappelhoff, line 269:
```suggestion
1. If you are adding a figure (and not picture) make sure to also supply a vector format 
of that figure (ideally as an `.svg` file) as this makes it easier to edit it in the
future.
```

sappelhoff (approved):
I think this is a good addition to our contributor documentation.

PS: The Ci failures are related to the MRI page which is going to be fixed from the PET PR.

effigies (approved):

sappelhoff:
Thanks @Remi-Gau

PR #684, opened 2020-11-30
[MAINT] Include master updates in bep009 branch in order to tackle all comments on bep009 until now.

melanieganz:
In order to tackle all comments so far on the bep009 branch, we need to update it with the master changes that were made while the bep 009 pull request has been open.


melanieganz:
Hi everyone, we want to tackle the remaining comments we received on bep009 and for some of them, e.g. comments on the references, we need to update the bep009 branch with the new changes implemented in master. We\d like to mere this once before ASL also gets merged.

sappelhoff (approved):

PR #692, opened 2020-12-07
[INFRA] Move part entity to before recording entity

tsalo:
Closes None, but addresses an issue raised in the maintainers Mattermost channel.

Changes proposed:
- Move part entity to before recording entity, so that it fits with other related entities (e.g., echo, inv, flip). Note that this shouldn't impact any filenames, as recording and part shouldn't be present in the same filename.
- Rebuild entity table and entity description appendices.


sappelhoff (approved):

effigies (approved):
`part-` wasn't in a release, and I also can't find any cases where they show up together, so I think it would be safe anyway.

agahkarakuzu:
@tsalo good catch! 
image

I was about to open an issue about this :)

tsalo:
It looks like the linkchecker failure is unrelated to the changes in this PR. Should I just merge as-is?

sappelhoff:
> It looks like the linkchecker failure is unrelated to the changes in this PR. Should I just merge as-is?

:+1:

PR #688, opened 2020-12-01
[ENH] BEP001 - Entity-linked file collections

effigies:
Trying again. I squashed #683 onto the latest `master` and pushed to `bids-bep001/file-collections-redux` so I didn't damage other history.

This replaces #668 and #683.

PR #694, opened 2020-12-08
[FIX] Clarify run indexing information for MRI acquisitions

dlevitas:
Adding fix, regarding this [NeuroStars issue](https://neurostars.org/t/specifying-run-number-s-in-bids/17530)

### DIFF
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -169,12 +169,15 @@ If several scans of the same modality are acquired they MUST be indexed with the
 [`run-<index>`](../99-appendices/09-entities.md#run) key-value pair:
 `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
 run labels).
-If however these scans of the same modality contain different labels
-(for example `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional,
-unless the labels are the same.
-When there is only one scan of a given type the run key MAY be
-omitted. Please note that diffusion imaging data is stored elsewhere (see
-below).
+If several scans with the same acquisition parameters are acquired in the same session,
+they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
+`_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
+run labels).
+
+If different entities apply,
+such as a different session indicated by [`ses-<label>`](../99-appendices/09-entities.md#ses),
+or different acquisition parameters indicated by [`acq-<label>`](../99-appendices/09-entities.md#acq),
+then `run` is not needed to distinguish the scans and MAY be omitted.
 
 The OPTIONAL [`acq-<label>`](../99-appendices/09-entities.md#acq)
 key/value pair corresponds to a custom label the user
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -168,7 +168,11 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
 If several scans of the same modality are acquired they MUST be indexed with the
 [`run-<index>`](../99-appendices/09-entities.md#run) key-value pair:
 `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
-run labels). If however these scans of the same modality contain different labels (e.g. `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional, unless the labels are the same. When there is only one scan of a given type the run key MAY be
+run labels).
+If however these scans of the same modality contain different labels
+(for example `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional,
+unless the labels are the same.
+When there is only one scan of a given type the run key MAY be
 omitted. Please note that diffusion imaging data is stored elsewhere (see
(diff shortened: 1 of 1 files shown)


src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -168,7 +168,7 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
 If several scans of the same modality are acquired they MUST be indexed with the
 [`run-<index>`](../99-appendices/09-entities.md#run) key-value pair:
 `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
-run labels). When there is only one scan of a given type the run key MAY be
+run labels). If however these scans of the same modality contain different labels (e.g. `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional, unless the labels are the same. When there is only one scan of a given type the run key MAY be
```
sappelhoff, line 171: Thanks for the PR @dlevitas I suggest to break lines at reasonable places to improve future git diff, and to replace "e.g." by more readable "for example", because we are trying to reduce the use of latin in the BIDS specification where possible.

```suggestion
run labels).
If however these scans of the same modality contain different labels
(for example `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional,
unless the labels are the same.
When there is only one scan of a given type the run key MAY be
```

tsalo:
I don't have time to review this just yet, but I did want to note that this speaks to the need to delineate which entities define a file collection (as in #688) and which ones differentiate runs (#530).

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -168,7 +168,11 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
 If several scans of the same modality are acquired they MUST be indexed with the
 [`run-<index>`](../99-appendices/09-entities.md#run) key-value pair:
 `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
-run labels). When there is only one scan of a given type the run key MAY be
+run labels).
+If however these scans of the same modality contain different labels
+(for example `_acq-`, `_ce-`, `_desc-`) then the run label and indexing are optional,
+unless the labels are the same.
+When there is only one scan of a given type the run key MAY be
```
effigies, line 175: What about:

```Markdown
If several scans with the same acquisition parameters are acquired in the same session,
they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
`_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
run labels).
If different entities apply,
such as a different session indicated by [`ses-<label>`](../99-appendices/09-entities.md#ses),
or different acquisition parameters indicated by [`acq-<label>`](../99-appendices/09-entities.md#acq),
then `run` is not needed to distinguish the scans and MAY be omitted.
```

This replaces the somewhat ambiguous concept of "modality" with acquisition parameters; in almost all cases identical acquisition parameters should imply the same modality.

This may also make the sentence "When there is only one scan of a given type the run key MAY be omitted." unnecessary, though I don't think that it harms clarity to keep it in.

dlevitas: This wording sounds fine with me. I've gone ahead and made the changes, sorry for the delayed response.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -168,9 +168,16 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
 If several scans of the same modality are acquired they MUST be indexed with the
 [`run-<index>`](../99-appendices/09-entities.md#run) key-value pair:
 `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
```
effigies, line 170:
```suggestion
```

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -168,9 +168,16 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
...
-run labels). When there is only one scan of a given type the run key MAY be
-omitted. Please note that diffusion imaging data is stored elsewhere (see
-below).
+run labels).
```
effigies, line 171:
```suggestion
```

effigies (commented):
I think there was some ambiguity about my last proposal. I was suggesting replacing the entire paragraph.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -168,9 +168,16 @@ For example, `sub-01_mod-T1w_defacemask.nii.gz`.
...
+If several scans with the same acquisition parameters are acquired in the same session,
+they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
+`_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
+run labels).
+
+If different entities apply,
+such as a different session indicated by [`ses-<label>`](../99-appendices/09-entities.md#ses),
+or different acquisition parameters indicated by [`acq-<label>`](../99-appendices/09-entities.md#acq),
+then `run` is not needed to distinguish the scans and MAY be omitted.
```
tsalo, line 180: I think this is crucial information that should be reflected in the run entity definition, rather than in a specific modality's section.

effigies:
Closing in favor of #719.

Issue #695, opened 2020-12-21
Make the possiblity to store stimuli files more "obvious"


Remi-Gau:
Eventhough storing stimuli files is supported by BIDS, the information is "a bit hard to find". 

For example, a search on the mkdocs does not bring up the section `task events` where the information is located (in the description of the `stim_file` column for events file).

Similarly, it is unclear at the moment if one can have stimuli in a BIDS dataset that are not referenced in any events file.

Would it make sense to take some of that content out of the table and possibly "combine" it with the sitmuli database section?

Remi-Gau:
fixed by #697

PR #697, opened 2020-12-28
[MISC] refactor stimuli mentioning sections in the events page

Remi-Gau:
PR related to #695 

- moves all sections about stimuli (whether about the stimuli folder or the database) together (including specific column names from the first table).

This is a draft looking for suggestions for improvements.

### Also included

- adding data types in the first table (leaving some `????` for values I am unsure about)

### Questions

- should the `identifier` and `database` columns be included in the tables of column names or were they there as just suggestions for the example?
- the `database` column is not described in the example JSON: should this be fixed?


src/04-modality-specific-files/05-task-events.md
```diff
@@ -33,16 +33,15 @@ Each task events file REQUIRES a corresponding task imaging data file
...
+| response_time   | OPTIONAL              | [number][]    | Response time measured in seconds. A negative response time can be used to represent preemptive responses and "n/a" denotes a missed response.                                                                                                                                                                                                                     |
+| value           | OPTIONAL              | ????          | Marker value associated with the event (for example, the value of a TTL trigger that was recorded at the onset of the event).                                                                                                                                                                                                                                      |
```
Remi-Gau, line 43: Not sure what Datatype this is.

src/04-modality-specific-files/05-task-events.md
```diff
@@ -33,16 +33,15 @@ Each task events file REQUIRES a corresponding task imaging data file
...
+| HED             | OPTIONAL              | ????          | Hierarchical Event Descriptor (HED) Tag. See [Appendix III](../99-appendices/03-hed.md) for details.                                                                                                                                                                                                                                                               |
```
Remi-Gau, line 44: Not sure if HED have a specific datatype that should be mentioned here.

tagging @VisLab

src/04-modality-specific-files/05-task-events.md
sappelhoff: I have seen strings and numbers. For example in BrainVision files you can find a marker called `S  1` (string) that gets written when a TTL byte of `1` is received by the recording hardware. Then again some other hardware may simply encode that as `1` (number).

We could also keep it unspecified, or put something like `string, or number, or any other datatype`, I am not sure at all.

src/04-modality-specific-files/05-task-events.md
sappelhoff: HED tags are usually strings, but I'll let Kay confirm just to be sure.

src/04-modality-specific-files/05-task-events.md
```diff
@@ -1,48 +1,47 @@
...
+are supported (in contrast to "block" designs) - each "block of events" can be
+represented by an individual row in the `_events.tsv` file (with a long
+duration).
+
 Template:
 
 ```Text
 sub-<label>/[ses-<label>]
-    func/
+    <data_type>/
```
Remi-Gau, line 17: Events do not only concern `func` (at least mentioned in `beh`) so I suggest making that more general.

**On that note**

There are several bids-examples datasets for "eeg", "meg", "ieeg" that have `events.tsv`. But this seems to not be mentioned in the specifications that you can have those files there.

Should we at least mention in the `events` page what data type allow `events.tsv`?

And update the schema accordingly?

src/04-modality-specific-files/05-task-events.md
Remi-Gau: @VisLab what would be the datatype for the HED column in an `event.tsv`

src/04-modality-specific-files/05-task-events.md
Remi-Gau: Forget the part of the schema: I did not realize that the latest version in `master` has the events suffix for `eeg`, `meg`, `ieeg` and that this is rendered automatically in the specification upon build.

VisLab:
The HED column type is a string.

src/04-modality-specific-files/05-task-events.md
sappelhoff: looking at this again, let's say 
```
[string][] or [number][]
```

src/04-modality-specific-files/05-task-events.md
sappelhoff: as reported here: https://github.com/bids-standard/bids-specification/pull/697#issuecomment-789128879, this should be a string

Remi-Gau:
@effigies @sappelhoff 

Except for this failing link in a file I did not touch, this is good to go.

effigies (approved):

sappelhoff (approved):
Thanks @Remi-Gau

PR #706, opened 2021-01-09
[ENH] - task label for ASL 

a3sha2:
- [x] the asl is  similar to bold, task is also performed during some of [ASL acquisitions](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6309706)   and it will be nice to include `_task-<label>` also for ASL and m0scan

- [x] it is very important to keep m0scan in `perf`   directory not in `fmap`


sappelhoff:
CC @patsycle @HenkMutsaerts

patsycle:
> * [x]  the asl is  similar to bold, task is also performed during some of [ASL acquisitions](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6309706)   and it will be nice to include `_task-<label>` also for ASL and m0scan
> * [x]  it is very important to keep m0scan in `perf`   directory not in `fmap`

About the task: this will be implemented in a later version for ASL. For the current version, we opted only to implement the approaches described from the White paper. Adding task in the specification is not a big job, however, it will be much more work for the validator.

About the m0scan: agree that m0scan should always be stored within the perf folder. However, in case of pe polar, the m0scan with reversed direction is a correction for the m0scan in the perf directory. Therefore, the one with reversed direction should go to the fmap directory, to be consistent with other MR modalities. So the original text was correct.

sappelhoff:
I agree with @patsycle that more "features" to ASL can and should be incorporated *after* the initial incorporation of ASL into BIDS.

At this stage of the ASL BEP, we only accept changes if they address either:

1. bugs or mistakes in the currently proposed feature set
2. incompatibilities with the existing BIDS

Everything else (backwards compatible features) will have to wait until after the initial release.

HenkMutsaerts (changes requested):
Thanks for the suggestion, we have discussed if we would want to include the same task key as with fMRI, but decided against it for this first ASL BIDS implementation, as this would make for a quite heavy new key. In 95% or 99% of ASL studies, there is no functional ASL. Adding this key would mean that all ASL files would require this key sign the rest value, since this key is required for func/fMRI. Therefore, we decided to not include functional ASL in this first ASL BIDS implementation. This also means that the pepolar fmap for ASL would not get this key, contrary to what you suggest here. Does this explain it nicely? @patsycle perhaps we can add this text to the paper. So @a3sha2 you can undo this suggestion. Thanks, Henk

a3sha2:
I just reverse back to original docs by removing `task-label`

PR #707, opened 2021-01-12
[FIX] Molar activity, timeZero, autosampler often recorded

mathesong:
I've just made a few very small fixes.

Molar activity was specified as specific activity in the text, so I've fixed it. Regarding units, it could be either GBq/umol or Bq/mol: these are equivalent (according to the consensus article).  Bq/mol feels more reasonable, since there are no multiplicative factors, but I've done it this way to match the example below.

I fixed one instance of `timeZero` to `TimeZero` to match the other instances

I modified "blood data is most often recorded using an autosampler" to "is often recorded using an autosampler".  I think more often than not, blood data is collected without an autosampler.


mathesong:
I realise I should have made this PR as suggestions to the open PR. I'll do it the correct way instead and close this.

PR #714, opened 2021-01-22
SCHEMA: Update entity YAML keys

effigies:
In keeping with #616, updating the keys to "common concepts" for inv, res, den and desc. I did not do mt.

### DIFF
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -132,7 +132,7 @@ inversion:
     denotes the number/index (in the form of a nonnegative integer), not the `InversionTime`
     value which needs to be stored in the field `InversionTime` of the separate JSON file.
   format: index
-mt:
+mtransfer:
   name: Magnetization Transfer
   entity: mt
   description: |
--- a/src/schema/entities.yaml
+++ b/src/schema/entities.yaml
@@ -121,7 +121,7 @@ flip:
     denotes the number/index (in the form of a nonnegative integer), not the `FlipAngle`
     value which needs to be stored in the field `FlipAngle` of the separate JSON file.
   format: index
-inv:
+inversion:
   name: Inversion Time
   entity: inv
   description: |
@@ -212,7 +212,7 @@ split:
     It is RECOMMENDED that .fif files with multiple parts use the
     `split-<index>` entity to indicate each part.
   format: index
-res:
+resolution:
   name: Resolution
   entity: res
   description: |
@@ -222,7 +222,7 @@ res:
 
     This entity is only applicable to derivative data.
   format: label
-den:
+density:
   name: Density
   entity: den
   description: |
@@ -242,7 +242,7 @@ label:
 
     This entity is only applicable to derivative data.
   format: label
-desc:
+description:
   name: Description
   entity: desc
   description: |


sappelhoff (approved):

tsalo (approved):
LGTM!

yarikoptic (approved):

effigies:
Sorry, threw this together immediately before a meeting, so didn't have time to write much, and was kind of hoping the discussion would happen in the absence of prompts. Did we want `inversion` to be `inversion_time` or some such? What about `mt`?

Since the schema is intended to be used by other tools, getting a pretty definite policy early would be good so we don't pull the rug out from under anybody in the future.

tsalo:
>Sorry, threw this together immediately before a meeting, so didn't have time to write much, and was kind of hoping the discussion would happen in the absence of prompts. Did we want inversion to be inversion_time or some such? What about mt?

I prefer inversion, since we have the `echo` pattern already. As for `mt`, I don't have a good answer, but @agahkarakuzu might be able to weigh in? My first though is `magtrans`, which isn't great.

agahkarakuzu:
@effigies mapping between entity and parameters are `inv` --> `InversionTime` and `mt` --> `MTState`. 

Instead of using `inv`, `inversion` sounds feasible if you are comfortable with longer entity names. 

As for `mt`, I cannot think of a single word that can represent it. Maybe something like `magntx` if `tx-->transfer` is not too encryptic, or `magntran`, not sure.

sappelhoff:
> I prefer inversion, since we have the echo pattern already.

agreed

as for mt, I think `mt` is fine. `magntran` isn't pretty but sounds okay as well.

agahkarakuzu:
@sappelhoff @tsalo @effigies how would you like to proceed with these entities? We'll need to convert some datasets according to the latest changes, that'd be great if we could stabilize the entity names soon.

tsalo:
@agahkarakuzu sorry for the confusion. The names we're discussing don't impact the actual entities as they're used in datasets, so you're all set. They mostly impact how we recommend that folks refer to the entities in code (e.g., with argument names and internal variable names).

agahkarakuzu:
Thank you for the clarification @tsalo ! I guess I'll consider these while extending the validator then?

tsalo:
I believe that the "full" names are relevant in the validator, although the actual regular expressions still use the entity keys (the "short" names). Sorry to say, but I do so little with the validator that I can't remember for sure.

effigies:
Thanks for clarifying, @tsalo. Yes, this is just a consideration that what names we use here are very likely to become the names that tools use internally for the variables that are populated by parsing entities, as they migrate to depending on the schema. The multi-word concepts just need a sensible rule that we can apply consistently.

If we follow the precedent of `ceagent` (entity `ce`, so must start with `ce`, use `agent` as a noun), then `mt` would be `mtransfer`. Would that be readable for people?

tsalo:
What about `mtstate`?

effigies:
That would be fine with me. Is there any chance of collision with the `MTState` metadata variable?

agahkarakuzu:
I think `mtransfer` is quite clear.

tsalo:
You're totally right. I forgot that it's a rule in the specification that entities and metadata fields not duplicate one another, regardless of capitalization. I don't think we've applied that to the variable names of the entities, but I don't see why we should complicate things. I vote for `mtransfer` then.

tsalo (approved):
Re-approving just to make my support clear 😄

sappelhoff (approved):

effigies:
@tsalo @sappelhoff I'll leave this to you to merge when you think appropriate. It doesn't fall under our usual 5 day rule, but I also don't see a need to do it quickly.

Issue #721, opened 2021-02-03
BEP-001 clarification thread


effigies:
I'm working on getting some [BEP-001 example datasets](https://osf.io/k4bs5/) in shape prior to the spec release. As I have questions I'll post them here. Some of them may require spec clarification, so I'm opening the issue here rather than in the examples repo.

1) Working on ds-mp2rage, the current MP2RAGE files have the `inv` entity, but not `flip`. This makes sense to me in that they co-vary, so either could have equally validly been chosen. That said, the template [here](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data) declares `inv` and `flip` are both required for MP2RAGE. That does not seem to match anything else in the spec, though. Is this a problem with the template?

2) ds-mtsat initially had `acq-T1w`, `acq-MTon` and `acq-MToff`. I changed these to `flip-2`, `flip-1_mt-on` and `flip-1_mt-off`. Was that correct? We'll need to add metadata, but that can go in examples, as long as we get the baseline issues done here.

---

More questions to come as I come up with them.

cc @agahkarakuzu @Gilles86 @ChristophePhillips (please add other BEP-001/qMRI people I've missed) @bids-standard/maintainers

---

Converted datasets for your approval:

* [ ] https://gin.g-node.org/markiewicz/ds-mp2rage
* [ ] https://gin.g-node.org/markiewicz/ds-mp2rageme
* [ ] https://gin.g-node.org/markiewicz/ds-mpm
* [ ] https://gin.g-node.org/markiewicz/ds-mtsat
* [ ] https://gin.g-node.org/markiewicz/ds-qsm
* [ ] https://gin.g-node.org/markiewicz/ds-sa2rage
* [ ] https://gin.g-node.org/markiewicz/ds-vfa

Gilles86:
Hey @effigies 

Great to hear that you are working on this!

The central idea of MP2RAGE is that you acquire the largely same image for two different _inversion times_ that are hence T1 and PD-weighted.

For effiency reasons, these two images _usually_ have different flip angles, but they absolutely don't have to (whereas by definition they have to have different inversion times). I would, thus, like in my example datasets on [osf](https://osf.io/k4bs5/), only use the `inv-` key/value pairs in the file names. I cannot think of any use case where it makes sense to me to use the `flip-` key/value pair in the file name of MP2RAGE files, but maybe one of the other cc'd people can!

[...]

PR #722, opened 2021-02-04
FIX: Make flip optional for MP2RAGE

tsalo:
References #721.

Changes proposed:
- Make the `flip` entity optional for the MP2RAGE suffix.

### DIFF
--- a/src/schema/datatypes/anat.yaml
+++ b/src/schema/datatypes/anat.yaml
@@ -133,7 +133,7 @@
     ceagent: optional
     reconstruction: optional
     echo: optional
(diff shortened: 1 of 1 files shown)


tsalo:
I'm not 100% sure if flip should be optional or prohibited.

effigies:
Makes sense to me. Will defer to Agah and Gilles about whether there's any reason it should be disallowed.

effigies (approved):
Let's go with this.

Gilles86:
Sorry for my late reply. LGTM.

PR #725, opened 2021-02-09
FIX: Add deprecated anatomical MRI suffixes back into schema

tsalo:
Closes None. Stems from https://github.com/bids-standard/bids-validator/pull/1161#discussion_r572940783 by @effigies.

Changes proposed:
- Add deprecated `T2star`, `PD`, and `FLASH` suffixes back into the schema.
- Rename suffix groups in `anat.yml` to match the variable names in the validator.

### DIFF
--- a/src/schema/datatypes/anat.yaml
+++ b/src/schema/datatypes/anat.yaml
@@ -10,6 +10,9 @@
     - inplaneT2
     - PDT2
     - angio
+    - T2star  # deprecated
+    - FLASH  # deprecated
+    - PD  # deprecated
   extensions:
     - .nii.gz
     - .nii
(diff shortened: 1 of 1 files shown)


effigies:
Do these have an impact on rendering anywhere?

tsalo:
Other than the entity table, no they shouldn't. The filename patterns will just generalize any patterns with more than 4 (I think) suffixes to just `suffix`.

sappelhoff (commented):
oh, good catch! Also, I like the new (more descriptive) names instead of "nth group"

effigies (approved):

sappelhoff (approved):

sappelhoff:
Thanks :+1:

PR #728, opened 2021-02-10
Update qMRI fieldmap schema

effigies:
This PR:

1) Adds a number of fieldmaps that are in the spec but not the schema:
    - TB1AFI
    - TB1TFL
    - TB1RMF
    - RB1COR
    - TB1SRGE
2) Allows `inv-` entities for all qMRI fieldmaps, in accordance with https://bids-specification.readthedocs.io/en/latest/99-appendices/10-file-collections.html#fieldmap-data
3) Renames group comments more informatively.

I notice https://bids-specification.readthedocs.io/en/latest/99-appendices/10-file-collections.html#fieldmap-data also permits `mt-`, but I don't see any examples or use cases that call for it. Is it supposed to be permitted?

Clarifying these issues here, and will update the validator according to any decisions made.

cc @ChristophePhillips @Gilles86 @agahkarakuzu @tsalo


tsalo (commented):
It looks like `TB1AFI` requires `acq` for the two TRs. `TB1TFL` and `TB1RMF` seem to require `acq` for something more complicated. `RB1COR` requires `acq` for coils, although I think that could be replaced with the `coil`/`ch` entity when we get that added.

I'm not sure if the `acq` entity is _required_ for all of them, since the text is a little ambiguous, but it does _seem_ required. Would be good to get the qMRI folks' input on that.

effigies:
Yes, the text says "should", not required, so I left it as optional for now.

tsalo (approved):
Assuming that the specification text is accurate w.r.t. `acq`'s requirement level, these changes look good to me.

effigies:
Added ab25456 so that all my questions/suggestions for the BEP-001 contributors about intent are in a single place.

agahkarakuzu:
Thank you @effigies I'll go through them tomorrow.

agahkarakuzu:
> I'm not sure if the acq entity is required for all of them, since the text is a little ambiguous, but it does seem required. Would be good to get the qMRI folks' input on that.

@tsalo the `acq` was the only option to describe some of the fieldmaps. They are `required` in practice, but I avoided it given that acq is freeform. So I used the requirement level `should`. There is explanation about this in the appendix text, so if `optional` is a better selection for the schema, that's all good. As you pointed out, we can update these when there are new entities. 

From the files changed, I see there was a confusion about between `TB1RMF` an `TB1RFM`. I also noticed that there was a typo about this int he appendix as well. The correct form is `TB1RFM`, so I can unify them to this similar to what @effigies did in 9672f91.

The other files changed looks good overall, now I'll review them one by one. Thank you!

src/99-appendices/11-qmri.md
```diff
@@ -529,7 +529,7 @@ values of the `acq` entity SHOULD begin with either `anat` or `famp` and MAY be
          └── sub-01_acq-famp_TB1TFL.json
 ```
 
-The example above applies to the `TB1RFM` suffix as well.
+The example above applies to the `TB1RMF` suffix as well.
```
agahkarakuzu: I'll change all the instances to `TB1RFM` if that's fine by @effigies

src/schema/datatypes/fmap.yaml
```diff
@@ -64,4 +65,59 @@
     reconstruction: optional
     echo: required
     flip: required
+    inversion: optional
     part: optional
+# RFFieldMaps
+- suffixes:
+    - TB1AFI
+    - TB1TFL
+    - TB1RMF
```
agahkarakuzu: Need to update respectively.

src/schema/datatypes/fmap.yaml
```diff
@@ -47,8 +47,9 @@
     ceagent: optional
     reconstruction: optional
     flip: required
+    inversion: optional
     part: optional
-# Fourth group
+# TB1EPI
```
agahkarakuzu, line 52: I found leftover `fa` in the appendix, will fix them.

effigies:
> There is explanation about this in the appendix text, so if `optional` is a better selection for the schema, that's all good.

Right now the schema only distinguishes required and optional, and doesn't have a notion of recommended. So we'll keep it as optional.

> The correct form is `TB1RFM`

Fixed in 3bfeabb.

agahkarakuzu:
Thanks @effigies! Can I push changes to this branch for other corrections in the appendix?

effigies:
> Can I push changes to this branch for other (relevant) corrections in the appendix?

Sure.

agahkarakuzu:
@effigies I noticed that I don't have write access. If that's how it should be, then I can open a PR from a fork.

effigies:
It's just mass change `fa-` to `flip-`? I can do that real quick.

agahkarakuzu:
image

image

image

Yes, these are the changes

src/99-appendices/11-qmri.md
```diff
@@ -62,14 +62,14 @@ Below is an example file collection for RF transmit field map `TB1EPI`:
...
      └── fmap/
-         ├── sub-01_fa-1_echo-1_TB1EPI.nii.gz
-         ├── sub-01_fa-1_echo-1_TB1EPI.json
-         ├── sub-01_fa-1_echo-2_TB1EPI.nii.gz
-         ├── sub-01_fa-1_echo-2_TB1EPI.json
-         ├── sub-01_fa-2_echo-1_TB1EPI.nii.gz
-         ├── sub-01_fa-2_echo-1_TB1EPI.json
-         ├── sub-01_fa-2_echo-2_TB1EPI.nii.gz
-         └── sub-01_fa-2_echo-2_TB1EPI.json
+         ├── sub-01_flip-1_echo-1_TB1EPI.nii.gz
```
tsalo, line 65: `flip` needs to come after `echo`, I think.

tsalo: oh but we can just render this with code if you want

agahkarakuzu: 🤦  sorry I missed that, you are right.

effigies: Done.

effigies:
Tests passing. Are we good to merge?

agahkarakuzu (approved):

PR #735, opened 2021-02-15
[FIX] split MEG files should be listed separately in scans.tsv

eort:
As discussed in [#692](https://github.com/mne-tools/mne-bids/issues/692) in the mne-bids repo, MEG recordings that consist of split files, all parts should be listed separately in the `scans.tsv` file.


src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -40,7 +40,7 @@ limit.
 For example Neuromag/Elekta/Megin, which can produce several files
 for a single recording.
 Both `some_file.fif` and `some_file-1.fif` would belong to a single recording.
-In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files.
+In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files. Split files SHOULD be listed as separate recordings in the `scans.tsv` file.
```
sappelhoff, line 43: just adding a line break here

```suggestion
In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files.
Split files SHOULD be listed as separate recordings in the `scans.tsv` file.
```

sappelhoff (approved):
LGTM, thanks @eort

sappelhoff:
@eort could you please add yourself to the contributors list here --> https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -40,7 +40,8 @@ limit.
 For example Neuromag/Elekta/Megin, which can produce several files
 for a single recording.
 Both `some_file.fif` and `some_file-1.fif` would belong to a single recording.
-In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files.
+In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files. 
+If there are split files, they SHOULD be listed as separate recordings in the `scans.tsv` file.
```
hoechenberger, line 44: Why make it a SHOULD when it could also be a MUST? :) Just saying…

sappelhoff: I was thinking for backward compatibility reasons. But one could make the case that because the case was unclear before (whether to put each split individually, or only a single split into scans), the usual BIDS rules must have applied (putting each scan individually), and thus this is a backward compatible clarification that removes an unfortunate ambiguity.

So +1 to change SHOULD to MUST

hoechenberger: Yes my POV is also: it was unclear / unspecified before, now let's not make the mistake to leave us in this miserable situation by making it a wobbly SHOULD. -> Let's go for the MUST.

ftadel:
I haven't read the context, but it should be clear that is only about FIF files.
For other CTF and 4D MEG recordings, the split files are within the folder referenced in the scans.tsv.
(for other manufacturers, I don't know...)

Note that the split FIF files are linked with hard-coded file names (tag FIFF.FIFF_REF_FILE_NAME). 
Ideally, renaming them should be done carefully by a dedicated program.
Does MNE-BIDS does this?

ftadel (approved):
SHOULD or MUST: 
preferably MUST, the least ambiguity the better

hoechenberger:
> Ideally, renaming them should be done carefully by a dedicated program.
> Does MNE-BIDS does this?

I'm actually not entirely sure how well this is currently being handled, but if we're not doing this properly in MNE-BIDS, then it's a bug we need to fix. :)

src/04-modality-specific-files/02-magnetoencephalography.md
eort: MUST it is then

eort:
> Ideally, renaming them should be done carefully by a dedicated program.
> Does MNE-BIDS does this?

I can't say anything about how consistently MNE-BIDS does that, but for my data it works fine.

hoechenberger (approved):

sappelhoff:
> I'm actually not entirely sure how well this is currently being handled, but if we're not doing this properly in MNE-BIDS, then it's a bug we need to fix. :)

pretty sure we are just using MNE-Python for that in mne-bids, and MNE-Python *does* properly rename multi-part FIF files. It's even mentioned as an example/recommended tool to use in the bids spec

image

adam2392 (approved):

effigies:
We should clarify:

1) Multi-part runs should be discussed in scans.tsv, not just in split, and the same acq_time is REQUIRED across parts of the same run (e.g., echo-, part-, split-)
2) Explicit note in split that acq_time should be the same for all splits, not `acq_time(split2) == acq_time(split1) + duration(split1)`
3) scans.tsv remains OPTIONAL

tsalo:
> 1. Multi-part runs should be discussed in scans.tsv, not just in split, and the same acq_time is REQUIRED across parts of the same run (e.g., echo-, part-, split-)

My personal opinion is that the idea that all files should be listed in `scan.tsv` (when provided), should specifically be brought up in the description for `scans.tsv` and not in the definition for `split`.

Something like:

> When the OPTIONAL `scans.tsv` file is provided, split files MUST have the same acquisition time (`acq_time`).

robertoostenveld:
@tsalo wrote

> ...all files should be listed in scan.tsv (when provided), should specifically be brought up in the description for scans.tsv ...

For FieldTrip data2bids and [examples](https://www.fieldtriptoolbox.org/example/bids/) I have also been pondering about what is to be listed in the `scans.tsv` file.  That is most explicit [here](https://www.fieldtriptoolbox.org/example/bids_pom/#final-_scanstsv-file). But I don't find it so clear what constitutes data (to be listed) and what metadata (not to be listed). 

E.g. for PET there are the `pet.nii` files (obviously), but should the `blood.tsv` files also be listed in the `scans.tsv`? And for fMRI the `bold.nii` files would be listed in case `scans.tsv is` provided, but should physio, stim and events also be listed in that case?

However, for the concrete case here (split/part/echo) I don't think this confusion applies. I do like the suggestion to require `acq_time` to be the same for all.

eort:
>  all files should be listed in scan.tsv (when provided), should specifically be brought up in the description for scans.tsv and not in the definition for split

Yes, I agree. But should these two things be only mentioned in the scans.tsv section or also in the split sections?

src/99-appendices/01-contributors.md
```diff
@@ -196,3 +196,4 @@ If you contributed to the BIDS ecosystem and your name is not listed, please add
 -   Lyuba Zehl 📖
 -   Melanie Ganz-Benjaminsen 📖🔣💻🤔📆🔍📢
 -   Martin Noergaard 📖🔣💻🤔📢
+-   Eduard Ort 📖
```
sappelhoff, line 199: I am removing this entry, because you'll be added to the list as part of this PR: #736 

```suggestion
```

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -40,7 +40,8 @@ limit.
 For example Neuromag/Elekta/Megin, which can produce several files
 for a single recording.
 Both `some_file.fif` and `some_file-1.fif` would belong to a single recording.
-In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files.
+In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files. 
+If there are split files, they MUST be listed as separate recordings in the `scans.tsv` file.
```
sappelhoff, line 44:
```suggestion
If there are split files and the optional `scans.tsv` is provided, remember to list all parts
of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
MUST be identical, as described in
[Scans file](../03-modality-agnostic-files.md#scans-file).
```

src/99-appendices/06-meg-file-formats.md
```diff
@@ -167,6 +167,8 @@ updated.
 It is RECOMMENDED that FIFF files with multiple parts use the `split-<index>`
 entity to indicate each part.
 
+Split files MUST be listed as separate recordings in the `scans.tsv` file.
+
```
sappelhoff, line 171:
```suggestion
If there are split files and the optional `scans.tsv` is provided, remember to list all parts
of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
MUST be identical, as described in
[Scans file](../03-modality-agnostic-files.md#scans-file).

```

src/schema/entities.yaml
```diff
@@ -218,6 +218,7 @@ split:
     pointers will be updated.
     It is RECOMMENDED that .fif files with multiple parts use the
     `split-<index>` entity to indicate each part.
+    Split files MUST be listed as separate recordings in the `scans.tsv` file.
```
sappelhoff, line 221:
```suggestion
    If there are split files and the optional `scans.tsv` is provided, remember to list all parts
    of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
    MUST be identical, as described in
    [Scans file](../03-modality-agnostic-files.md#scans-file).
```

sappelhoff:
> E.g. for PET there are the pet.nii files (obviously), but should the blood.tsv files also be listed in the scans.tsv? And for fMRI the bold.nii files would be listed in case scans.tsv is provided, but should physio, stim and events also be listed in that case?

That's a good point that we should clarify as part of a bigger issue/PR :+1: 


> Multi-part runs should be discussed in scans.tsv, not just in split, and the same acq_time is REQUIRED across parts of the same run (e.g., echo-, part-, split-)

This should be specified in this section: https://github.com/bids-standard/bids-specification/blame/master/src/03-modality-agnostic-files.md#L326

@eort can you make a proposal please?

src/03-modality-agnostic-files.md
```diff
@@ -339,8 +339,16 @@ The purpose of this file is to describe timing and other properties of each
 imaging acquisition sequence (each *run* file) within one session.
 Each neural recording file should be described by at most one row.
 Relative paths to files should be used under a compulsory `filename` header.
-If acquisition time is included it should be under `acq_time` header.
+Exceptions are runs that consist of multiple parts.
+For example runs that include the `echo-`, `part-`, or `split-` fields.
```
tsalo, line 343: I don't understand how runs with multiple parts are an exception. It says above that each file gets a line, and when runs have multiple parts (i.e., multiple files), each file gets a line.

src/03-modality-agnostic-files.md
```diff
@@ -364,6 +372,8 @@ Example `_scans.tsv`:
 filename	acq_time
 func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
 func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
+meg/sub-01_task-rest_split-01_meg.nii.gz 1877-01-10T12:15:27
+meg/sub-01_task-rest_split-02_meg.nii.gz 1877-01-10T12:15:27
```
tsalo, line 376: Given that these files are from 5 months apart, it seems like this should be structured as a longitudinal dataset.

tsalo (changes requested):
I just have a few minor things.

src/03-modality-agnostic-files.md
```diff
@@ -339,8 +339,16 @@ The purpose of this file is to describe timing and other properties of each
...
 Relative paths to files should be used under a compulsory `filename` header.
-If acquisition time is included it should be under `acq_time` header.
+Exceptions are runs that consist of multiple parts.
+For example runs that include the `echo-`, `part-`, or `split-` fields.
+All parts that appear as separate files in the respective subdirectory have to be
+separately listed on one row each.
+The relative path in the `filename` header should include the field that specifies the parts.
+If acquisition time is included it should be listed under the `acq_time` header.
 Acquisition time refers to when the first data point in each run was acquired.
+If a `_scans.tsv` file is provided and includes the `acq_time` header, the acqusition times
```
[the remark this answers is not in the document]
hoechenberger: There's still a typo here

src/03-modality-agnostic-files.md
sappelhoff: I cannot make a code suggestion for this whole part because GitHub doesn't support it, but this is what I would write:

```Markdown
...
The purpose of this file is to describe timing and other properties of each
imaging acquisition sequence (each *run* file) within one session.

Each neural recording *file* SHOULD be described by one row.
Note that some neural recordings consist of multiple parts that are distributed over several files,
for example through `echo-`, `part-`, or `split-` entities.
Such recordings are still to be documented with one row per file, and
their acquisition time as documented under the `acq_time` column (see below)
MUST be identical, if that column is supplied.

Relative paths to files should be used under a compulsory `filename` header.

If acquisition time is included it should be listed under the `acq_time` header.
Acquisition time refers to when the first data point in each run was acquired.
Datetime should be expressed as described in [Units](./02-common-principles.md#units).
...
```

It seems to me like this part of the spec was written before the strict rules about "MUST" and "SHOULD" were introduced, so these words are used very leniently here ...

cc @tsalo @eort

src/03-modality-agnostic-files.md
```diff
@@ -364,6 +372,8 @@ Example `_scans.tsv`:
 filename	acq_time
 func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
 func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
+meg/sub-01_task-rest_split-01_meg.nii.gz 1877-06-15T12:15:27
+meg/sub-01_task-rest_split-02_meg.nii.gz 1877-06-15T12:15:27
```
sappelhoff, line 376: change a space character (` `) to a "true *tab* character" (`	`)

```suggestion
meg/sub-01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27
meg/sub-01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27
```

src/03-modality-agnostic-files.md
eort: > It seems to me like this part of the spec was written before the strict rules about "MUST" and "SHOULD" were introduced, so these words are used very leniently here 

Should this be changed then? There are several lower case `should`s in the text. Are they different shoulds then the `SHOULD`'s or is that just inconsistent?

eort: @sappelhoff I like your version. But I think I'd prefer to mention the acq_time information for split files only after the acq_time information in general was described. So basically, not making this forward link (see below). But that's maybe style ;)

eort: > Such recordings are still to be documented

Does this imply MUST or SHOULD? (in the current commit I opted for MUST).

sappelhoff: > Should this be changed then? There are several lower case shoulds in the text. Are they different shoulds then the SHOULD's or is that just inconsistent?

The situation is unclear to me, let's not touch lines that we wouldn't have to touch anyways. But those that we *do* touch, we need to discuss

> Does this imply MUST or SHOULD? (in the current commit I opted for MUST).

"Such recordings are still to be documented"  implies a MUST :+1: 


> So basically, not making this forward link (see below). But that's maybe style preference ;)

I agree that that's a better style :+1:

sappelhoff (approved):
cool, I think this reads well now and addresses all initial criticisms and suggestions

eort:
Nice!

Once this is merged, does the validator need to be updated to test for that change?

sappelhoff:
> Once this is merged, does the validator need to be updated to test for that change?

Taylor has opened an issue for that: https://github.com/bids-standard/bids-validator/issues/1192

but it will be non-trivial to implement, unless you are well versed in JS :) (in the latter case it may or may not be easy)

src/03-modality-agnostic-files.md
```diff
@@ -337,11 +337,21 @@ Optional: Yes
...
+Some recordings consist of multiple parts, that span several files, 
+for example through `echo-`, `part-`, or `split-` entities.
+Such recordings MUST be documented with one row per file.
+
 Relative paths to files should be used under a compulsory `filename` header.
-If acquisition time is included it should be under `acq_time` header.
+
+If acquisition time is included it should be listed under the `acq_time` header.
 Acquisition time refers to when the first data point in each run was acquired.
+Furthermore, if this header is provided, the acqusition times of all files that
```
tsalo, line 350:
```suggestion
Furthermore, if this header is provided, the acquisition times of all files that 
```

src/99-appendices/06-meg-file-formats.md
```diff
@@ -167,6 +167,11 @@ updated.
 It is RECOMMENDED that FIFF files with multiple parts use the `split-<index>`
 entity to indicate each part.
 
+If there are split files and the optional `scans.tsv` is provided, remember to list all parts
+of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
```
tsalo, line 171:
```suggestion
If there are split files and the optional `scans.tsv` is provided, remember to list the
files separately in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
```

I think it's confusing to refer to a "split file".

tsalo (changes requested):
Thanks for the hard work on this. There is one minor typo, and I have a thought about wording. I'd prefer to avoid the term "split file," since it makes it harder to distinguish grouping elements (such as "run", "acquisition", and "file"). While a "run" or "acquisition" can correspond to multiple files, I think the singular word "file" should only apply to one actual file object.

src/schema/entities.yaml
```diff
@@ -218,6 +218,10 @@ split:
     pointers will be updated.
     It is RECOMMENDED that .fif files with multiple parts use the
     `split-<index>` entity to indicate each part.
+    If there are split files and the optional `scans.tsv` is provided, remember to list all parts
+    of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
```
sappelhoff, line 222: applying taylor's suggestion also here

```suggestion
    If there are split files and the optional `scans.tsv` is provided, remember to list the
    files separately in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
```

eort:
@tsalo Makes perfect sense! I adopted your suggestions.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -40,7 +40,11 @@ limit.
 For example Neuromag/Elekta/Megin, which can produce several files
 for a single recording.
 Both `some_file.fif` and `some_file-1.fif` would belong to a single recording.
-In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files.
+In BIDS, the [`split`](../99-appendices/09-entities.md#split) entity is RECOMMENDED to deal with split files. 
+If there are split files and the optional `scans.tsv` is provided, remember to list all parts
+of the split file in `scans.tsv` and that the entries for the `acq_time` column in `scans.tsv`
+MUST be identical, as described in
```
tsalo, line 46: @eort I think you just missed this one in the recent set of changes.

eort: yes, I have. Thanks for pointing it out!

eort: Yes, I did. Thanks for pointing out!

sappelhoff (approved):

src/04-modality-specific-files/02-magnetoencephalography.md
tsalo: Sorry for the delay in responding, but it looks like you _reverted_ one of the text changes in eaf6e62 instead of updating the last text snippet, so now only 1 of the 3 duplicate text snippets reflect the new changes.

eort: Sorry, should be fixed now.

eort:
According to ci/circlci, there seems to be a certificate issue for https://www.cognitiveatlas.org/task/id/trm_54e69c642d89b
I don't know what to do about it.

sappelhoff:
> According to ci/circlci, there seems to be a certificate issue for https://www.cognitiveatlas.org/task/id/trm_54e69c642d89b
I don't know what to do about it.

that's nothing we can solve - we'll ignore it for now and revisit if it doesn't fix itself within a couple of weeks.

rwblair:
Sorry about that, cert should be renewed now.

effigies:
Restarted the Circle job.

src/04-modality-specific-files/02-magnetoencephalography.md
tsalo: It looks like you reverted the last snippet to the old text instead of using the new text in the other snippets.

eort: It actually became a hybrid version. Given your initial remark on that particular issue, the current version should also be fine (the phrase `split file` is not used anymore), but if you think it is better I can also change 
> all parts separately

to 

> all files separately

tsalo: Ah, I see. Thanks! I think saying "files" instead of "parts" is key, since "part" is ambiguous, while "file" isn't, so if you could change that, that would be great.

tsalo (approved):
LGTM. Thanks!

sappelhoff (approved):

sappelhoff:
Thanks @eort

PR #742, opened 2021-02-23
Updated TotalAcquiredVolumes into TotalAcquiredPairs

effigies:
TotalAcquiredVolumes became TotalAcquiredPairs with the following changes:
- required: it appears that this should be required as it is crucial information for quantitative analysis
- small changes in description


effigies (commented):
@patsycle @HenkMutsaerts This looks good to me. Thanks for your diligence in working through all these issues before release.

PR #753, opened 2021-03-11
[INFRA] Use macros to render filename templates inline

tsalo:
This PR replaces manually-generated filename templates for PET with functions that will automatically generate them.

It also adds `trc` to the set of entities and adds PET to the entity table.


mnoergaard (approved):
LGTM! Thanks @tsalo!

PR #754, opened 2021-03-13
[FIX] Mix up (typo) between ficiduals and landmarks in EEG spec

rob-luke:
I think this is a typo. But I may also just be confused.


yarikoptic (approved):
Looks like a correct fix to me (like a left over from some renaming), but I am not familiar with eeg portion of the spec

sappelhoff (approved):
Lgtm

sappelhoff:
thanks!

Issue #471, opened 2020-05-14
Specification of relative or absolute path in IntendedFor and SpatialReference  is not consistent 


robertoostenveld:
While reading the [common derivatives draft](https://bids-specification.readthedocs.io/en/common-derivatives/05-derivatives/01-introduction.html) as the rendered version of #265, I noticed that for `SpatialReference` it uses an absolute path:

```
    "SpatialReference": "/sub-01/anat/sub-01_desc-combined_T1w.nii.gz"
```

whereas in the existing MRI examples there is
```
   "IntendedFor": "func/sub-01_task-motor_bold.nii.gz"
```

and in the EEG examples there is again
```
  "IntendedFor":"/sub-01/ses-01/anat/sub-01_T1w.nii",
```
 
This made me realize that the specification is not clear w.r.t. references to other files. E.g. in the `SpatialReference` case (which is in the derivative folder): would that point to a file in the derivative itself, or could it also point to a file at a higher level (e.g. the raw level)? In this case I know it points within the derivative dataset since it contains the `desc` keyword (which is not defined for raw). But had it been `/sub-01/anat/sub-01_T1w.nii.gz`, then I would have said that it must have been pointing to a file in the raw dataset. The specification of a relative path, and the specification of what specifically it is relative to, might avoid confusion.

robertoostenveld:
Reading a bit further, I see another `SpatialReference` that points towards a `.gii` file in the derivatives dataset, but also a `RawSources` that (with precisely the same path structure) points to a file in the higher level raw dataset.

effigies:
Thanks for bringing this up, @robertoostenveld. As another data point, `stim_file` in [Task events](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html#task-events) is a relative path (not starting with `/`), but relative to the `/stimuli` directory.

I can see why `RawSources` prefers to its paths to relative to the raw dataset instead of the current dataset, because the datasets may not have constant locations relative to one another. If my `dataset_description.json` says:

```
{
    ...
    "SourceDatasets": [
        {
            "DOI": "10.18112/openneuro.ds000114.v1.0.1"
        }
    }
}
```

Then I have enough information to find the files. Any more specific path I might give will be very fragile if I organize my derivatives anywhere besides in the `derivatives/` subdirectory of the sources.

[...]

PR #757, opened 2021-03-15
[WIP; ENH] Storage of iEEG electrodes in subject-specific FreeSurfer space

adam2392:
Closes: #747 

Currently, this is a draft to demonstrate exactly what would fix issues discussed in #747 .

What I'm not sure of is how to modify the following:

```
### Nonstandard coordinate system identifiers

The following template identifiers are RECOMMENDED for individual- and study-specific reference
spaces.
In order for these spaces to be interpretable, `SpatialReference` metadata MUST be provided, as
described in [Common file level metadata fields][common file level metadata fields].

In the case of multiple study templates, additional names may need to be defined.

| Coordinate System | Description                                                                                                                                                                                                                                                           |
| ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| individual        | Participant specific anatomical space (for example derived from T1w and/or T2w images). This coordinate system requires specifying an additional, participant-specific file to be fully defined. In context of surfaces this space has been refered to as `fsnative`. |
| study             | Custom space defined using a group/study-specific template. This coordinate system requires specifying an additional file to be fully defined.                                                                                                                        |
```

Since I'm not sure what exactly the `individual` coordinate system is used for. It seems like an umbrella term that also encompasses `fsnative`.

### DIFF
--- a/src/99-appendices/08-coordinate-systems.md
+++ b/src/99-appendices/08-coordinate-systems.md
@@ -151,6 +151,11 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     [ACPC site](https://www.fieldtriptoolbox.org/faq/acpc/) on the FieldTrip
     toolbox wiki.
 
+-   `fsnative`: The origin of the coordinate system is at the center of
+    isotropic 1 mm 256x256x256 volume in RAS orientation. This corresponds
+    to electrode coordinates localized on the subject-specific ``T1.mgz``
+    FreeSurfer file.
+
 -   Any keyword from the list of
     [Standard template identifiers](#standard-template-identifiers)
 


adam2392:
@sappelhoff not sure what's up w/ the link-checker, i'll take another look today.

Anything holding up this you think?

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,19 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     [ACPC site](https://www.fieldtriptoolbox.org/faq/acpc/) on the FieldTrip
     toolbox wiki.
 
+-   `fsnative`: The subject-specific FreeSurfer T1w space. This corresponds to
+    the original scanner image after it is re-sampled and re-scaled
+    to an isotropic 1 mm 256x256x256 volume in RAS orientation. The origin
+    is wherever the original origin of the ``orig.mgz`` file is. One SHOULD
+    align the origin to the ``ACPC`` before using this coordinate system.
```
sappelhoff, line 158: what is `orig.mgz`? 

>  One SHOULD align the origin to the ``ACPC`` before using this coordinate system

why SHOULD one do that?  ... and why not MUST?

sappelhoff:
So would the workflow for this solution be something like:

```
derivatives/
    freesurfer/
        sub-01/
            T1.mgz <-- based on sub-01_T1w.nii
sub-01/
    anat/
        sub-01_T1w.nii
    ieeg/
        sub-01_space-fsnative_electrodes.tsv
        sub-01_space-fsnative_coordsystem.json <-- IntendedFor contains path to freesurfer T1.mgz in derivatives?
```

... or am I misunderstanding this?

If I am correct, then we still need to figure out how to appropriately point to a derivatives file, given that they could be stored anywhere and simply using "relative paths from BIDS root" are not guaranteed to work.

cc @effigies

adam2392:
Yes that is correct!

src/99-appendices/08-coordinate-systems.md
adam2392: I suppose it can be a MUST, so that way fsnative origin is assumed to be AC of subjects brain but I think this might depart from what ppl consider fsnative generally. I don't think acpc alignment is required by all ppl? 

Cc: @robertoostenveld

adam2392:
> ```
> If the derivative data is stored as a subdirectory of `bids_root`, then `IntendedFor` must point to that file relative to `bids_root`. If not, then a DOI should be present that links to that dataset.
> ```

Is there any update with regards to this? Should I include this part, to specify that `fsnative` and `fsnativetkr` are derivative coordinate systems?

src/99-appendices/08-coordinate-systems.md
robertoostenveld: if it is not AC, but a random other location in the body (e.g. the left big toe), then I would say that it is not `fsnative`. In that case it should be specified with `other` and the details are to be specified in the CoordinateSystemDescription.

sappelhoff:
Thanks for the reminder, and sorry that there is so little progress - it seems like very few people are interested in solving this problem.

> If the derivative data is stored as a subdirectory of `bids_root`, then `IntendedFor` must point to that file relative to `bids_root`. If not, then a DOI should be present that links to that dataset.

so am I understanding your proposal correctly?

1. if `derivatives/` is part of a BIDS dir, interpret `IntendedFor` as a relative path from the BIDS dir root
2. if `derivatives/` is NOT part of a BIDS dir, a DOI "should be present" to point to the `derivatives/`, and then `IntendedFor` will be used "as if" `derivatives/` were nested in the BIDS dir?

>  then a DOI should be present that links to that dataset.

where should that DOI be present?

What about datasets that do not have a DOI?

robertoostenveld:
the proposed `fsnativetkr` coordinate system is specific to a particular anatomical MRI in FS format. E.g. consider that you start with a 1mm T1 scan and another 0.7 mm T1 scan from the patient. The number of slices and hence the alignment of the first (lower-resolution) scan within the 256x256x256 volume are likely to be different than the number of slices and hence the alignment  of the higher-resolution scan. If you run freesurfer on the low-res, the output will be shape-wise very similar to that of the high-res, but the data expressed in the `fsnativetkr` coordinate system will be shifted with respect to each other. 

I think that `fsnativetkr` therefore cannot be specified without the corresponding `IntendedFor` linking to an MRI that must represent the 256x256x256 volume in the FS file format.

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,20 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
...
 
+-   `fsnative`: The subject-specific FreeSurfer T1w space. This corresponds to
+    the original scanner image after it is re-sampled and re-scaled
+    to an isotropic 1 mm 256x256x256 volume in RAS orientation. The origin
+    is wherever the original origin of the ``orig.mgz`` file is. One MUST
+    align the origin to the ``ACPC`` before using this coordinate system.
+
+-   `fsnativetkr`: The subject-specific FreeSurfer T1w space. The origin of
+    the coordinate system is at the center of isotropic 1 mm 256x256x256
+    volume (that is the voxel center is at slices (128, 128, 128)) in RAS
```
robertoostenveld, line 162: are these indices 0-offset or 1-offset? 

Should the middle of a 256x256x256 volume not be represented as [127.5 127.5 127.5] (when 0-offset) or  [128.5 128.5 128.5] (where 1-offset)?

The middle of a 2x2x2 volume would be at voxel [0.5 0.5 0.5] if you start counting voxels at [0 0 0], or at  [1.5 1.5 1.5] when you start counting at [1 1 1].

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,20 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     [ACPC site](https://www.fieldtriptoolbox.org/faq/acpc/) on the FieldTrip
     toolbox wiki.
 
+-   `fsnative`: The subject-specific FreeSurfer T1w space. This corresponds to
+    the original scanner image after it is re-sampled and re-scaled
+    to an isotropic 1 mm 256x256x256 volume in RAS orientation. The origin
+    is wherever the original origin of the ``orig.mgz`` file is. One MUST
+    align the origin to the ``ACPC`` before using this coordinate system.
```
robertoostenveld, line 158: the origin must be aligned precisely to the anterior commissure (AC), right? Not to the PC, and also not to an arbitrary point along the AC-PC line.  

AFAIK the reason for referring to ACPC is that the origin is at AC and the y-axis is along the AC-PC line with positive y pointing to anterior (of the subject) and positive z pointing to a mid-hemisphere point that is superior to the AC.

src/99-appendices/08-coordinate-systems.md
adam2392: I'm pretty sure it is with 0-offset.

That's what I thought initially... but I think there is some offset, so for example, running the following

```
        mri_info --vox2ras-tkr <img>

    This will generally be the 4x4 matrix for FreeSurfer output.::

            [
                [-1.0, 0.0, 0.0, 128.0],
                [0.0, 0.0, 1.0, -128.0],
                [0.0, -1.0, 0.0, 128.0],
                [0.0, 0.0, 0.0, 1.0],
            ]
```

This essentially centers the coordinates at the "center" of the volume "almost".

Reference: https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg69541.html

src/99-appendices/08-coordinate-systems.md
adam2392: Yes you're right. Sorry ACPC alignment is more then what is needed, simply setting the origin at AC should be sufficient and then RAS orientation.

adam2392:
> I think that `fsnativetkr` therefore cannot be specified without the corresponding `IntendedFor` linking to an MRI that must represent the 256x256x256 volume in the FS file format.

Then perhaps what is needed to fully define this TKR coordinate system is:

"The ``IntendedFor`` also MUST be defined linking to an MRI that MUST represent the 256x256x256 volume in the FreeSurfer file format."?

adam2392:
https://bids-specification--757.org.readthedocs.build/en/757/99-appendices/08-coordinate-systems.html#ieeg-specific-coordinate-systems

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,20 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     [ACPC site](https://www.fieldtriptoolbox.org/faq/acpc/) on the FieldTrip
     toolbox wiki.
 
+-   `fsnative`: The subject-specific FreeSurfer T1w space. This corresponds to
+    the original scanner image after it is re-sampled and re-scaled
+    to an isotropic 1 mm 256x256x256 volume in RAS orientation. The origin
```
robertoostenveld, line 156: `fsnative` is not defined relative to the 1mm 256^3 MRI, but is defined relative to the AC of the individual participant.

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,20 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
...
+    is wherever the original origin of the ``orig.mgz`` file is. One MUST
+    align the origin to the ``AC`` before using this coordinate system. If the origin is a random other location in the body, then it MUST be specified with ``Other`` and further details descriped in the ``CoordinateSystemDescription``.
```
robertoostenveld, line 158: I suggest to not state what one must do (i.e. the action), but what the data should be like (the factual situation that results as outcome from the action). 

I also suggest not to mention in which cases `fsnative` is **not** to be used.

src/99-appendices/08-coordinate-systems.md
adam2392: I'm not sure I follow. Isn't `fsnative` just the xyz coordinates in RAS space  based on the `vox2ras` affine transformation?

src/99-appendices/08-coordinate-systems.md
adam2392: That's a good point. I adjusted accordingly and took out some language.

src/99-appendices/08-coordinate-systems.md
robertoostenveld: RAS is only the specification of the direction of the axes, the origin also needs to be defined. That origin is at the anterior commissure, or not?

adam2392: Ah yes you're right. I just noticed I had two incongruent sentences. Adjusted now to specify AC.

adam2392:
> 1. if `derivatives/` is part of a BIDS dir, interpret `IntendedFor` as a relative path from the BIDS dir root
> 2. if `derivatives/` is NOT part of a BIDS dir, a DOI "should be present" to point to the `derivatives/`, and then `IntendedFor` will be used "as if" `derivatives/` were nested in the BIDS dir?
> 
> > then a DOI should be present that links to that dataset.
> 
> where should that DOI be present?
> 
> What about datasets that do not have a DOI?

@sappelhoff is the best place to handle this within this PR? Perhaps adding a statement:

```
The `IntendedFor` key must be defined either with: i) a DOI or ii) a relative path from the BIDS directory
```

?

sappelhoff:
> is the best place to handle this within this PR? Perhaps adding a statement:

although it will probably delay the efforts here, I think it's best to first deal with that issue here: https://github.com/bids-standard/bids-specification/issues/471#issuecomment-815245258

and then come back to the present one.

sappelhoff:
@adam2392 -- just a note that #918 has been merged and BIDS URIs will be part of the 1.8 release (probably some time in September). Not sure if what you proposed in this PR is still needed and/or salvageable. If not, please close the PR :-)

Else. we can leave it open and keep working on it.

src/99-appendices/08-coordinate-systems.md
```diff
@@ -151,6 +151,16 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
...
+-   `fsnative`: The subject-specific FreeSurfer T1w space. This corresponds to
+    the original scanner image after it is re-sampled and re-scaled
+    to an isotropic 1 mm 256x256x256 volume in RAS orientation. The origin MUST be the Anterior Commissure.
+
+-   `fsnativetkr`: The subject-specific FreeSurfer T1w space. The origin of
+    the coordinate system is at the center of isotropic 1 mm 256x256x256
+    volume (that is the voxel center is at slices (128, 128, 128)) in RAS
+    orientation. This corresponds to FreeSurfer's surface RAS coordinates (TKR coordinates).
+    The origin MUST be the Anterior Commissure. The ``IntendedFor`` MUST link to an MRI that represents the 256x256x256 volume in the FreeSurfer file format. The difference between this coordinate system and ``fsnative`` is that TKR coordinates are defined on the surface of the brain based on FreeSurfer.
+
```
dorahermes, line 163: Freesurfer has several surfaces (pial, white, inflated... ), it is unclear which surface is meant

dorahermes (changes requested):
Nice, clarifying surface space would be helpful.

effigies:
Switching to draft to make PRs ready for review more identifiable.

PR #760, opened 2021-03-17
[FIX] run entity - "scans" -> "data samples"

yarikoptic:
As we have now other modalities, such as EEG and MEG, "scan" is not really the
most appropriate term here.  Having said that we do use _scans etc for EEG and
MEG as well ATM, as it was the "forward compatible" with existing MRI-centric
BIDS back then.

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -240,7 +240,7 @@ run:
   name: Run
   entity: run
   description: |
-    If several scans with the same acquisition parameters are acquired in the same session,
+    If several data samples (e.g., MRI scans) with the same acquisition parameters are acquired in the same session,
     they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
     `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
     run labels).
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -240,7 +240,7 @@ run:
   name: Run
   entity: run
   description: |
-    If several data samples (e.g., MRI scans) with the same acquisition parameters are acquired in the same session,
+    If several data acquisitions (for example, MRI scans or EEG recordings) with the same acquisition parameters are acquired in the same session,
     they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
     `_run-1`, `_run-2`, `_run-3`, and so on (only nonnegative integers are allowed as
     run labels).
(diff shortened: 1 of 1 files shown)


effigies:
My intuition would be that a "data sample" is a volume, while "recording" would cover a time series. Not sure how other modalities would think about it.

This isn't an objection, just trying to see if there's a clearer word. I'm okay with "data sample" if it's clear to most people.

yarikoptic:
FWIW, I am not entirely happy with "data sample" either for the reasons you have stated.  May be someone could suggest a better name...

sappelhoff:
I think I would find "recordings" most intuitive. But I don't think "scans" is a problem either - in particular because we do use `scans.tsv` for (structly speaking) non-scanning modalitites (EEG, ...)

```diff
-If several scans with the same acquisition parameters ...
+If several recordings (e.g., MRI scans) with the same acquisition parameters ...
```

sappelhoff (commented):
this is a relatively minor thing, but perhaps some more @bids-standard/maintainers have an opinion so that we can address this?

sappelhoff:
FYI, I rebased on `master` and force pushed.

src/schema/objects/entities.yaml
```diff
@@ -240,7 +240,7 @@ run:
   name: Run
   entity: run
   description: |
-    If several scans with the same acquisition parameters are acquired in the same session,
+    If several data samples (e.g., MRI scans) with the same acquisition parameters are acquired in the same session,
```
sappelhoff, line 243:
```suggestion
    If several data samples (e.g., MRI scans, EEG recordings) with the same acquisition parameters are acquired in the same session,
```

tsalo:
I agree that recording is a bit clearer than data sample, but the definition of run under common principles explicitly links it to "data acquisition". Should we use that instead?

sappelhoff (approved):

src/schema/objects/entities.yaml
```diff
@@ -240,7 +240,7 @@ run:
   name: Run
   entity: run
   description: |
-    If several scans with the same acquisition parameters are acquired in the same session,
+    If several data acquisitions (for example, MRI scans or EEG recordings) with the same acquisition parameters are acquired in the same session,
     they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
```
sappelhoff, line 244:
```suggestion
    If several data acquisitions (for example, MRI scans or EEG recordings)
    with the same acquisition parameters are acquired in the same session,
    they MUST be indexed with the [`run-<index>`](../99-appendices/09-entities.md#run) entity:
```

sappelhoff:
fyi @yarikoptic we discussed this in yesterday's maintainers meeting and as all suggestions to date were improvements, we just decided to just go with one of them.

PR #761, opened 2021-03-17
[ENH] Add a statement addressing metadata conflicts

### LINKED ISSUE #138: Precedence in JSON/NIfTI metadata mismatch
Over in bids-standard/pybids#357, we've been having a small discussion over the precedence of metadata when JSON sidecars and NIfTI disagree. Specifically the question was whether to take the 4th pixdim from the NIfTI or the RepetitionTime entry in the BIDS metadata, which only matters when there is disagreement.

It is the case that, at least for RepetitionTime, a mismatch is treated as an error by the validator. However, in practice, not every dataset is run through the validator before being exposed to pybids or pushed through a BIDS App... In user support for fMRIPrep, we've seen enough issues that would be solved by running through the validator that we now bundle the validator and run it before starting fMRIPrep.

There are other metadata entries that do not produce errors on disagreement, such as PhaseEncodingDirection and SliceEncodingDirection (packed in the dim_info field), SliceTiming (slice_code), which thus need to have a defined precedence (or begin producing errors).

fMRIPrep for one has made a policy of *only* querying JSON for pretty much everything besides voxel spatial dimensions, affines and data shape, and this seems like the correct approach for BIDS-aware applications. But it is also the case that we have a number of BIDS Apps that are just shims over existing non-BIDS-aware apps, which will thus only have NIfTI headers to rely upon, except where the shim writer also feeds in JSON metadata through another side channel.

I don't really have a satisfying proposal to make here, so this is mostly the start of a discussion. My inclination is that we should continue to prioritize JSON-encoded metadata, and increase the coverage of potential header/sidecar conflicts in the validator. In addition to the idea of respecting BIDS as a structure is the fact that JSON is easier to inspect and edit by hand, whereas modifying NIfTI headers requires more knowledge of some tool that can do the job. But making this a spec requirement may render some existing BIDS Apps non-compliant.

I would also suggest that a useful tool (possibly validator mode) would be to sync JSON metadata into NIfTI headers to resolve mismatches.

This may be somewhat related to #102.

### LINKED ISSUE #1133: Question: consistency of metadata in .nwb and bids filename/side-car - must be consistent, or allowed to be overloaded?
.nwb files are supported by BIDS since awhile.  NWB files are hdf5 files following quite elaborate [nwb schema](https://github.com/NeurodataWithoutBorders/nwb-schema/) which includes metadata which is also covered by BIDS itself, such as subject_id, session_id, etc.  I wonder, are we to allow such metadata be "overloaded" by BIDS (so it is `sub-001` in BIDS while .nwb file says "mouse3") or they must be consistent (and thus bids-validator should error out)?

effigies:
Closes #138.

Closes #1133

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -84,6 +84,25 @@ and/or files (like `events.tsv`) are fully omitted *when they are unavailable or
 instead of specified with an `n/a` value, or included as an empty file
 (for example an empty `events.tsv` file with only the headers included).
 
+### Metadata conflicts
+
+BIDS specifies that metadata be stored in human-readable formats such as
+TSV (see [Tabular files](#tabular-files)) or
+JSON (see [Key/value files](#keyvalue-files-dictionaries)).
+In some cases, this duplicates metadata contained in a data file.
+
+In cases of conflict, the BIDS metadata is considered authoritative.
+If BIDS metadata is defined,
+format-specific metadata MUST NOT conflict to the extent permitted by the format.
+
+If an exact representation of the BIDS metadata is possible in the format,
+then the metadata SHOULD be duplicated inside the data file.
+If an exact representation of the BIDS metadata is not possible in the format,
+then the closest possible approximation of the BIDS metadata SHOULD be used.
+Special null or undefined values MAY be used when available.
+If the format-specific metadata field is defined,
+the BIDS metadata SHOULD also be defined.
+
 ## Filesystem structure
 
 Data for each subject are placed in subdirectories named "`sub-<label>`",


sappelhoff (approved):

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
 be reviewed to include common data types in the future releases of the BIDS
 specification.
 
+### Metadata conflicts
+
+BIDS specifies that metadata be stored in human-readable formats such as
+TSV (see [Tabular files](#tabular-files)) or
+JSON (see [Key/value files](#keyvalue-files-dictionaries)).
+In some cases, this duplicates metadata internal to the imaging file.
```
yarikoptic, line 133: applicable not only to "imaging" data AFAIK thus (similar to #760 somewhat) might need a more generic wording here.  Meanwhile:
```suggestion
In some cases, this duplicates metadata internal to the data sample file(s).
```
and made it `(s)` since there could be those `_part-` and smth else which would correspond to the same "data sample"...

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
...
+
+In cases of conflict, the BIDS metadata is considered authoritative.
+If BIDS metadata is defined,
+format-specific metadata MUST NOT conflict to the extent permitted by the format.
+
+If an exact representation of the BIDS metadata is possible in the format,
+then the metadata SHOULD be duplicated inside the imaging file.
+If an exact representation of BIDS metadata is not possible in the format,
+then the imaging file metadata SHOULD be undefined
+or set to the closest possible approximation of the BIDS metadata.
```
yarikoptic, line 143:
```suggestion
If only possibly an exact representation of BIDS metadata is possible in the format,
then the data file metadata SHOULD be set to the closest possible approximation of the BIDS metadata.
```
I think would make it less ambiguous.  I do not think we should allow for completely undefined in such cases.  If "representation" is not possible at all  -- there is nothing to talk about.

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
...
+TSV (see [Tabular files](#tabular-files)) or
+JSON (see [Key/value files](#keyvalue-files-dictionaries)).
+In some cases, this duplicates metadata internal to the imaging file.
+
+In cases of conflict, the BIDS metadata is considered authoritative.
+If BIDS metadata is defined,
+format-specific metadata MUST NOT conflict to the extent permitted by the format.
+
+If an exact representation of the BIDS metadata is possible in the format,
+then the metadata SHOULD be duplicated inside the imaging file.
```
yarikoptic, line 140:
```suggestion
then the metadata SHOULD be duplicated inside the data file.
```

yarikoptic (changes requested):
some wording tune up

src/02-common-principles.md
effigies: The situation I was considering here is that someone might prefer to zero out a value rather than have an approximate but incorrect value in the header. For instance, suppose you have a slice timing sequence that doesn't fit any defined `NIFTI_SLICE_SEQ_*` field. I'd rather set it `NIFTI_SLICE_UNKNOWN` than the closest approximation.

I suspect this type of case is going to be relatively common for binary formats.

src/02-common-principles.md
tsalo: I like "data file" more than "data sample file" (as you proposed in https://github.com/bids-standard/bids-specification/pull/761#discussion_r596093857). The pairing of "data file" and "metadata file" is clearer IMHO.

src/02-common-principles.md
yarikoptic: I see. Yes, for `NIFTI1_SLICE_ORDER` indeed should be set to `NIFTI_SLICE_UNKNOWN` and thus indeed "undefined".

Majority of formats likely to not have `nil` or `None` or explicit "unknown" like in your example to specify the "undefined" value. So might be worth explicitly stating to set to "undefined" if explicit value is reserved for that (i.e. not `0` for some int or float field)?  and also make it as a measure of the last resort while still keeping "SHOULD" on approximate values to make those files usable with generic tools?

effigies: What do you think of 679b8e0?

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
...
+### Metadata conflicts
+
+BIDS specifies that metadata be stored in human-readable formats such as
+TSV (see [Tabular files](#tabular-files)) or
+JSON (see [Key/value files](#keyvalue-files-dictionaries)).
+In some cases, this duplicates metadata contained in a data file.
+
+In cases of conflict, the BIDS metadata is considered authoritative.
+If BIDS metadata is defined,
+format-specific metadata MUST NOT conflict to the extent permitted by the format.
```
yarikoptic, line 137: - how poor validator would be able to check that "approximately" the same, or this MUST NOT will not be validated, and thus not enforced?

- as much as I like consistency, it might be pragmatically prohibitive to achieve it.  E.g. if metadata is contained within 1GB (or 1TB?) data file and then extracted and fixed up in metadata file (might be just a minor fix) -- should 1G/TB be re-uploaded?

With that in mind I just wonder if we should keep it as SHOULD NOT (warning if mismatch; not an error if it was MUST NOT) level?

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
...
 specification.
 
+### Metadata conflicts
+
+BIDS specifies that metadata be stored in human-readable formats such as
+TSV (see [Tabular files](#tabular-files)) or
+JSON (see [Key/value files](#keyvalue-files-dictionaries)).
+In some cases, this duplicates metadata contained in a data file.
+
+In cases of conflict, the BIDS metadata is considered authoritative.
```
satra, line 135: could we clarify what a conflict is? for example, is a change of floating point precision a conflict? one reason we store things in binary is because floating point data maintains some veracity. if we see this outside the nifti lens for other modalities allowed by bids (the eeg/meg or upcoming ephys/microscopy formats, where precision starts mattering).

src/02-common-principles.md
effigies: I could go with SHOULD NOT. What about something like:

> Generally, if BIDS metadata is defined, format-specific metadata SHOULD NOT conflict, to the extent permitted by the format. Specific metadata fields may impose a strict requirement.

I'm thinking here of `RepetitionTime` and `pixdim[4]` which MUST agree.

src/02-common-principles.md
effigies: > could we clarify what a conflict is?

I can think of a few:

1) Equivalent fields, different values. `RepetitionTime` vs `pixdim[4]`
2) Disparate representations. `PhaseEncodingDirection` (enum) vs `dim_info` (bitfield)
3) Range of representations. `PhaseEncodingDirection` (axis + direction) vs `dim_info` (axis)

> for example, is a change of floating point precision a conflict?

I'm interpreting this as a question of tolerance? Yes, there should be some tolerance. I imagine the order of tolerance would probably be a case-by-case question.

> one reason we store things in binary is because floating point data maintains some veracity. if we see this outside the nifti lens for other modalities allowed by bids (the eeg/meg or upcoming ephys/microscopy formats, where precision starts mattering).

This seems to be arguing that we should say the highest-precision metadata should be authoritative, whether BIDS or format-internal. Or am I misunderstanding?

satra: > This seems to be arguing that we should say the highest-precision metadata should be authoritative, whether BIDS or format-internal.

in general yes, but we may want to clarify that for floating point values, the internal format always wins when close. this would perhaps avoid the situations where some has zeroed out something like TR.

effigies: This feels like a hard thing to write a rule for. Here's something:

> In general, the more precise metadata representation SHOULD be considered authoritative, with BIDS metadata being presumptively more precise. A notable exception to this presumption is if a value is represented as a floating point value in the internal metadata format and a decimal string in the JSON sidecar; to avoid accumulating error in float-decimal conversions, the floating point value is authoritative.

This actually contravenes existing practice in fMRIPrep and FitLins with regard to `RepetitionTime`/`pixdim[4]`, and may do so for MNE-Python as well, but I guess that's what standardization is for.

src/02-common-principles.md
sappelhoff: ping @yarikoptic

src/02-common-principles.md
sappelhoff: sounds good to me.

ping @satra

satra: i like @effigies paragraph - but i think he still had the question out for the community.

src/02-common-principles.md
yarikoptic: TL;DR: because it should be less of fields we allow to overload than of the fields we do not, may be we should keep original specification, but we would need to 1. come up with the annotation to when "conflict" is allowed (e.g. for some "description" fields etc) in side cars. 2. when allowed in metadata from filename entities (e.g. subject or session).

<details>
<summary>FWIW My longer thinking/line of argument</summary> 

It seems then again to require per sidecar field "equality" specification that some could be allowed to disagree (thus it would be only a WARNING if sidecar has different value) whenever others MUST agree (and thus ERROR from validator). Most likely the strict requirements would then be added a explicit `check`s in the schema, like ATM checks being added for lengths consistency e.g. in https://github.com/bids-standard/bids-specification/blob/HEAD/src/schema/rules/checks/asl.yaml#L26 . That would indeed make it explicit but standard itself IMHO somewhat "inconsistent" since it would allow for some fields to be overloaded but not the others, and "by default" allow for them being different. Note that adding a restriction (from a default of "could different") in principle constitutes backward incompatible change, thus should be avoided.

In general though it is a tough aspect to allow such differences since it mandates that tools must now be able to read that "overloaded" metadata from sidecar files. That is why, ideally, we should **really** strive for consistency. And that is why e.g. we are working with NWB file format folks to allow for ".overwrite.json" file to accompany ".nwb" file so that metadata could be overloaded at the level of the file format, and thus allowing for "cheap" adjustment of metadata if so desired.  For NIfTI, the primary workhorse of BIDS, since most of the tools would need correct base metadata such as pixdim, we better indeed enforce their consistency with side car files.

So, overall, I think we MUST enforce consistency and if so desired, relax only for some sidecar metadata fields AND strive to allow for metadata overload at the original file format whenever possible/necessary.
</details>

src/02-common-principles.md
```diff
@@ -125,6 +125,25 @@ The solutions will change from case to case and publicly available datasets will
...
+If BIDS metadata is defined,
+format-specific metadata MUST NOT conflict to the extent permitted by the format.
+
+If an exact representation of the BIDS metadata is possible in the format,
+then the metadata SHOULD be duplicated inside the data file.
+If an exact representation of the BIDS metadata is not possible in the format,
+then the closest possible approximation of the BIDS metadata SHOULD be used.
+Special null or undefined values MAY be used when available.
+If the format-specific metadata field is defined,
+the BIDS metadata SHOULD also be defined.
```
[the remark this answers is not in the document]
effigies: If NIfTI `slice_code != 0`, then `SliceTiming` SHOULD be defined?

robertoostenveld (approved):

effigies:
I'm not sure what approving reviews are at this point. There is an open proposal to completely reverse the priority to have file metadata take precedence over BIDS metadata.

src/02-common-principles.md
effigies: @yarikoptic I intended to write it to mean (and still read it as saying) "floating point fields in a binary format take precedence over JSON, which is inherently decimal".

src/02-common-principles.md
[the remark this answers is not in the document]
yarikoptic: `slice_code != 0` is not just that `slice_code` is defined but rather meets some criteria (`!=0`) and looking at https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/slice_code.html describing it as " slice_code = If this is nonzero, AND if slice_dim is nonzero, AND if slice_duration is positive,..." so it is even more complex than just *defined*.  Moreover from above sentence it somehow suggests that it is just a matter of "mirroring" some value between file and BIDS metadata. But in this case it is not the case since `SliceTiming` is not even in NIfTI, which only specifies the order and timing is not contained in the file.  So I will suggest ...

src/02-common-principles.md
```diff
@@ -184,6 +184,25 @@ The solutions will change from case to case and publicly available datasets will
...
```
yarikoptic, line 204:
```suggestion
Some BIDS metadata can be REQUIRED to be defined depending on the values of the format-specific metadata fields.
```

although even then the `slice_code` target use case is quite hard to guess ;)  but it also has nothing to do with the overloading really.  May be should just be removed?

yarikoptic:
> There is an open proposal to completely reverse the priority to have file metadata take precedence over BIDS metadata.

yeah, and it seems that we are arriving that overall there might be no *effective* precedence and metadata must be in agreement between file metadata and BIDS metadata.

sappelhoff:
migrating to 1.9.0 milestone because apparently this needs more discussion.

src/02-common-principles.md
yarikoptic: to freshen up: re my comment on floating point representation in text. data representation would also be truncated and it would just be a matter of expressing enough precision in json form. 

<details>
<summary>E.g. specifically for my own example messing around in ipython</summary> 

```
In [1]: repr(1/3)
Out[1]: '0.3333333333333333'

In [2]: 0.3333333333333333 * 3
Out[2]: 1.0

In [3]: 3.333333333333333e-1 * 3
Out[3]: 1.0

In [4]: 3.33333333333333e-1 * 3
Out[4]: 0.9999999999999989

In [5]: 1/3 == 3.333333333333333e-1
Out[5]: True

In [6]: 1/3 == 3.33333333333333e-1
Out[6]: False

```
</details>

so may be we should just add somewhere wording that "serialization of floating point numbers in json should be to maximal precision of the underlying data type used in the data file" or something like that?

effigies:
I've updated this PR to master because it's listed in the 1.9.0 milestone. I don't think there's a clear consensus on what to do. I've personally kind of given up hope on it, and would appreciate if a motivated party took over leading the discussion in a new issue/PR.

effigies:
Abandoning this. Others are welcome to pick it up.

PR #740, opened 2021-02-21
[DOC] move schema documentation into the schema folder

Remi-Gau:
fixes #716 

- [x] add a README in the schema folder
- [x] move some of the schema related content to that README and only keep in contributing.md info about how to modify the schema

Issue #699, opened 2020-12-31
Add Required/Recommended Metadata info to Schema


dlevitas:
This is in regards to the BIDS [schema](https://github.com/bids-standard/bids-specification/tree/master/src/schema), where yaml files specify the required/recommended suffixes, entities, and extensions for BIDS file names. However, the [BIDS spec](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html) specifies select fields in the JSON metadata that are either required, recommended, or optional. For example, functional MRI acquisitions must have the `RepetitionTime` field in the corresponding JSON file. Is this something that can be added to the schema? I'd be happy to open a PR if this seems worthwhile.

tsalo:
This is definitely a long-term goal for the schema. I started working with the NIDM-Terms folks on this (see #423), although I've been pulled into other things recently and I haven't made much progress on it (see #609, which is probably woefully out of date at this point). The relevant issue is probably #604.

Two elements that we'll want to have working before we move the metadata into the schema are:
1. **Supporting logic within the schema (#620).** There are relationships between metadata fields that we'll want to represent in the schema.
2. **Rendering schema elements in the specification automatically (#610).** If we don't have the schema represented in the specification directly, we're just asking for drift between the two information sources. We've already noticed the difficulty in keeping what the schema up-to-date w.r.t. the specification, so I'd hate to add all of the metadata fields to the schema and then have it sit, growing more and more out-of-date, as has happened with #609.

I'd be happy to have any help you're willing to provide!

dlevitas:
Sure, I'd be happy to help. Regarding your points:
**1).** That's a good example, one that I wasn't aware of. I suppose that would need to be fleshed out at some point.

**2).** My thought was to use a web scrapping library (e.g. `Beautiful Soup`) to select the schema elements; unsure though if that would address the issue. If so, grab the schema elements and place them into yaml files based on DataType (and ModalityLabel)

tsalo:
**1.** :+1:

[...]

Issue #620, opened 2020-09-21
Mutual exclusion and conditional relationships in the schema


tsalo:
There are a number of cases of mutually exclusive items or conditional relationships between items in the specification, and we'll need to support them in the schema.

One prime example is [timing metadata](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#timing-parameters_1) for fMRI. In that situation, there are five metadata fields (`RepetitionTime`, `SliceTiming`, `AcquisitionDuration`, `DelayTime`, and `VolumeTiming`) with specific relationships. There are five supported patterns for these five fields.

Another, less idiosyncratic case is entities work for the dual-purpose `beh` suffixes (`physio`, `stim`, `beh`, and `events`). In all cases, they have two options for entities: (1) place the files under `beh`, in which case there are a set of applicable entities, or (2) place the files under the associated imaging data type, in which case they much "match" entities, plus a reduced set of applicable entities. 

To handle the latter case, I was thinking that the schema could include a `match` "entity" (basically a special stand-in for a set of entities that may vary based on the associated imaging data). We would need to protect this term from the main specification, even though it would be almost exclusively internal.

To handle mutual exclusion, I think we could start using nested lists or something, so if we originally had:

```YAML
- suffixes:
    - stim
    - physio
  extensions:
    - .tsv.gz
    - .json
  entities:
    sub: required
    ses: optional
    task: required
    acq: optional
    run: optional
    recording: optional
```

we could instead have:

```YAML
- suffixes:
    - stim
    - physio
  extensions:
    - .tsv.gz
    - .json
  entities:
    - matches: required
      recording: optional
    - sub: required
      ses: optional
      task: required
      acq: optional
      run: optional
      recording: optional
```

In that case there would be two options for entities for `stim`/`physio` data, represented as a list of dictionaries.

An alternative that might be more flexible would be to use keywords like OR, AND, and NOT as keys to dictionaries, like so:

```YAML
  entities:
    - OR:
        - matches: required
        - sub: required
          ses: optional
          task: required
          acq: optional
          run: optional
    - recording: optional
```

When applied to the timing information for fMRI, this option would look like:

```YAML
  metadata:
    - OR:
        - AND:
            RepetitionTime: required
            NOT:
                - AcquisitionDuration
                - VolumeTiming
        - AND:
            SliceTiming: required
            VolumeTiming: required
            NOT:
                - RepetitionTime
[...]

PR #762, opened 2021-03-17
[SCHEMA] Add metadata term files

tsalo:
References #699.

This is a draft that might be useful for determining how to structure metadata YAML files for the schema. 

This approach does not incorporate any interactions between metadata terms, or requirement levels, which are both dependent on the suffix or data type to which the metadata term is being applied. This fits well with #603.

Changes proposed:

- Take the metadata JSON files from [the bids-validator](https://github.com/bids-standard/bids-validator/tree/master/bids-validator/validators/json/schemas) and convert them to YAML format.
- Add "description" and "key_name" fields to the metadata YAML files.
- Add functions to render the metadata information in tables within the schema. These functions require the specification text to include the requirement levels.
- Create files for a number of metadata terms that aren't present in the validator, but _are_ present in specification tables.

TODO:

- [x] Different units dependent on different data types?
    - Yes
- [x] Include "unit" for string/integer data types?
    - No
- [x] Limits, like minimum and maximum values
    - Taken from validator
- [x] Should we retain references to other info (e.g., lists of valid coordinate spaces for iEEG, EEG, and MEG)?
    - Currently, I have the references still in the files, with the referenced info in "private" yaml files (i.e., files with leading underscore).
- [x] Links that have to be broken up aren't looking great.
    - This is because the flattening step adds a space between separate lines.
- [x] Would be great to replace relative link paths with absolute ones.
- [X] Type determination for objects.
    - `additionalProperties` defines type.
    - `properties` should create additional rows. How should individual sub-object-level requirement levels be annotated then?? Take `Genetics.Dataset`, `Genetics.Database`, and `Genetics.Descriptors` as an example.
        - SOLUTION: Instead of trying to expand nested dictionaries into new rows, I've replaced the internal definitions with `$ref`s and added new files for the properties that can be referenced on their own. So now Genetics' `properties` fields is a dict with three `$ref`s and there are separate files for `Genetics.Dataset.yaml` etc.
- [x] Make type determination function recursive to deal with nested objects
    - See `SoftwareFilters` as an example.
- [x] Document new structure once #740 is merged.
- [x] Add "relative" and "absolute"(?) path formats for strings?
    - I decided to go with `dataset_relative` and `participant_relative` for now.
- [x] Add "unit" format for strings?
    - With PET now merged in, we have a bunch of new metadata fields that just define units of other fields instead of committing to a standard like we have for most other fields with units. Maybe we could define a "unit" format that will let the validator check for valid units?
- [ ] Identify and incorporate section-specific definition modifications.


src/schema/metadata/PostLabelingDelay.yaml
```diff
@@ -0,0 +1,18 @@
...
+  Can be a number (for a single-PLD time series) or an array of numbers
+  (for multi-PLD and Look-Locker).
+  In the latter case, the array of numbers contains the PLD of each volume,
+  namely each `control` and `label`, in the acquisition order.
+  Any image within the time-series without a PLD, for example an `m0scan`,
+  is indicated by a zero.
+  Based on DICOM Tags 0018,9079 `Inversion Times` and 0018,0082
+  `InversionTime`.
+data_type:
+  - number
```
tsalo, line 16:
```suggestion
  - number:
    - minimum: 0
```

What about this kind of notation for limits?

src/schema/metadata/PostLabelingDelay.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - array of numbers
```
tsalo, line 17:
```suggestion
  - array:
      number:
        minimum: 0
```

Maybe this for nesting rules for the data types within the array?

effigies:
Is there an advantage to a single file per term? Grouping terms feels more natural to me.

tsalo:
I don't want to group terms into multiple files, since relationships between terms (including their categories) probably vary across the specification, along with any logic that needs to be applied. I figured one file per term would be easier to search and edit than one file for all terms.

tsalo:
Based on the current code, the following macro call in the specification produces the table at the bottom. Some are missing descriptions right now and others are just missing completely, but it's still pretty cool!

```Text
{{ MACROS___make_metadata_table(
   {
      "Manufacturer": "RECOMMENDED",
      "ManufacturersModelName": "RECOMMENDED",
      "DeviceSerialNumber": "RECOMMENDED",
      "StationName": "RECOMMENDED",
      "SoftwareVersions": "RECOMMENDED",
      "HardcopyDeviceSoftwareVersion": "DEPRECATED",
      "MagneticFieldStrength": "RECOMMENDED, but REQUIRED for Arterial Spin Labeling",
      "ReceiveCoilName": "RECOMMENDED",
      "ReceiveCoilActiveElements": "RECOMMENDED",
      "GradientSetType": "RECOMMENDED",
      "MRTransmitCoilSequence": "RECOMMENDED",
      "MatrixCoilMode": "RECOMMENDED",
      "CoilCombinationMethod": "RECOMMENDED",
   }
) }}
```

image

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -10,6 +10,24 @@ by Ben Inglis:
...
+      "HardcopyDeviceSoftwareVersion": "DEPRECATED",
+      "MagneticFieldStrength": "RECOMMENDED, but REQUIRED for Arterial Spin Labeling",
+      "ReceiveCoilName": "RECOMMENDED",
+      "ReceiveCoilActiveElements": "RECOMMENDED",
+      "GradientSetType": "RECOMMENDED",
+      "MRTransmitCoilSequence": "RECOMMENDED",
+      "MatrixCoilMode": "RECOMMENDED",
+      "CoilCombinationMethod": "RECOMMENDED",
+   }
+) }}
```
tsalo, line 29: Until we add conditional logic to the schema, this would be how we define requirement levels in the specification.

src/schema/metadata/AnatomicalLandmarkCoordinateSystem.yaml
```diff
@@ -0,0 +1,13 @@
...
+  Defines the coordinate system for the anatomical landmarks.
+  Preferably the same as the `EEGCoordinateSystem`.
+  See Appendix VIII for a list of restricted keywords for coordinate systems.
+  If `Other`, provide definition of the coordinate system in
+  `AnatomicalLandmarkCoordinateSystemDescription`.
+anyOf:
+- $ref: common_definitions.json#/definitions/MEGCoordSys
+- $ref: common_definitions.json#/definitions/EEGCoordSys
+- $ref: common_definitions.json#/definitions/StandardTemplateCoordSys
+- $ref: common_definitions.json#/definitions/StandardTemplateDeprecatedCoordSys
```
tsalo, line 13: I'm not so sure about this one. Do we want to include references in the schema? If so, is this the best way to do it?

tsalo:
I can keep trucking along, adding terms, but I'd like to get input from interested folks.

@satra @dbkeator would this format be useable for NIDM-Terms?

@yarikoptic I know this isn't exactly how you wanted it (e.g., files are grouped in subfolders by type), but WDYT? The actual interactions between terms will still need to be defined elsewhere in the schema.

@rwblair could the validator use files like this? They're basically a translation of the validator JSON files to individual YAML files, with added descriptions and units, so I assume it wouldn't require too much effort to support them.

src/schema/metadata/AnatomicalLandmarkCoordinateSystem.yaml
tsalo: Alternatively, we could make sure that all referenced definitions are in "private" files (e.g., `_ StandardTemplateCoordSys.yaml`), since those definitions shouldn't be useful on their own.

src/02-common-principles.md
```diff
@@ -498,6 +498,16 @@ Note that if a field name included in the data dictionary matches a column name
...
+{{ MACROS___make_metadata_table(
+   {
+      "LongName": "OPTIONAL",
+      "Description": "RECOMMENDED",
+      "Levels": "RECOMMENDED",
+      "Units": "RECOMMENDED",
+      "TermURL": "RECOMMENDED",
+   }
+) }}
+
```
yarikoptic, line 510: table seems to be (rendered) empty, shortcut link https://bids-specification--762.org.readthedocs.build/en/762/02-common-principles.html#tabular-files

src/03-modality-agnostic-files.md
```diff
@@ -14,6 +14,23 @@ Templates:
 The file `dataset_description.json` is a JSON file describing the dataset.
 Every dataset MUST include this file with the following fields:
 
+{{ MACROS___make_metadata_table(
+   {
+      "Name": "REQUIRED",
```
yarikoptic, line 19: I don't see name in rendered version: https://bids-specification--762.org.readthedocs.build/en/762/03-modality-agnostic-files.html#dataset_descriptionjson

src/03-modality-agnostic-files.md
```diff
@@ -69,6 +86,13 @@ In addition to the keys for raw BIDS datasets,
 derived BIDS datasets include the following REQUIRED and RECOMMENDED
 `dataset_description.json` keys:
 
+{{ MACROS___make_metadata_table(
+   {
+      "GeneratedBy": "REQUIRED",
+      "SourceDatasets": "RECOMMENDED",
```
yarikoptic, line 92: SourceDatasets is missing: https://bids-specification--762.org.readthedocs.build/en/762/03-modality-agnostic-files.html#derived-dataset-and-pipeline-description

I guess script should be adjusted to error out if some field is not "handled"?

src/03-modality-agnostic-files.md
```diff
@@ -14,6 +14,23 @@ Templates:
...
+      "DatasetType": "RECOMMENDED",
+      "License": "RECOMMENDED",
+      "Authors": "OPTIONAL",
+      "Acknowledgements": "OPTIONAL",
+      "HowToAcknowledge": "OPTIONAL",
+      "Funding": "OPTIONAL",
+      "EthicsApprovals": "OPTIONAL",
+      "ReferencesAndLinks": "OPTIONAL",
+      "DatasetDOI": "OPTIONAL",
+   }
```
yarikoptic, line 31: having such specifications in the document text would be a big step back from toward the target of having a "machine readable" schema.

Why not to add `metadata` key to the corresponding records, e.g. in this case of https://github.com/bids-standard/bids-specification/blob/master/src/schema/top_level_files.yaml#L14 where similar to `entities` of https://github.com/bids-standard/bids-specification/tree/master/src/schema/datatypes to list the 'requirements' for each one of those? and then in the macros just point to what metadata of to render here (e.g. of `top_level_files.dataset_description` here)?  or you can immediately see that this would not work?

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -362,12 +423,28 @@ also be specified.
...
 | IntendedFor  | RECOMMENDED           | [string][]    | This can be an MRI/CT or a file containing the operative photo, x-ray or drawing with path relative to the project folder. If only a surface reconstruction is available, this should point to the surface reconstruction file. Note that this file should have the same coordinate system specified in `iEEGCoordinateSystem`. For example, **T1**: `"sub-<label>/ses-<label>/anat/sub-01_T1w.nii.gz"`  **Surface**: `"/derivatives/surfaces/sub-<label>/ses-<label>/anat/sub-01_T1w_pial.R.surf.gii"` **Operative photo**: `"/sub-<label>/ses-<label>/ieeg/sub-0001_ses-01_acq-photo1_photo.jpg"` **Talairach**: `"/derivatives/surfaces/sub-Talairach/ses-01/anat/sub-Talairach_T1w_pial.R.surf.gii"` |
 
 Fields relating to the iEEG electrode positions:
 
+{{ MACROS___make_metadata_table(
+   {
+      "iEEGCoordinateSystem": "REQUIRED",
+      "iEEGCoordinateUnits": "REQUIRED",
+      "iEEGCoordinateSystemDescription": 'RECOMMENDED, but REQUIRED if `iEEGCoordinateSystem` is `"Other"`',
```
yarikoptic, line 442: oh, that is an interesting case...  in light of my eager desire for concentrating this information in machine readable form, I can suggest for such cases to get to temp middle point  where requirement could be "spelled out" in human form for now with the last one listed as "catch all", e.g. 

```yaml
metadata:
 iEEGCoordinateUnits: required
 iEEGCoordinateUnits: required
 iEEGCoordinateSystemDescription:
  required: if `iEEGCoordinateSystem` is `"Other"`
  recommended:
 iEEGCoordinateProcessingDescription: recommended
```

yarikoptic (changes requested):
> @yarikoptic I know this isn't exactly how you wanted it (e.g., files are grouped in subfolders by type), but WDYT? The actual interactions between terms will still need to be defined elsewhere in the schema.

GREAT STUFF HERE, thank you @tsalo for all the work on this!  I blissfully forgotten (and decided to not encumber my review with prior opinion by looking it up) what I wanted.  But I immediately saw what I would have really preferred to avoid here -- "leaking" schema into the document.   I have dropped some question/idea(s) in other comments on how possibly may be those metadata fields "requirements" could be centralized within schema.  WDYT?

src/03-modality-agnostic-files.md
tsalo: Ultimately yes, but since I'm still in the process of converting all of the metadata terms, I don't want everything to error out just yet. Right now it just filters out any missing terms and renders the available ones.

src/03-modality-agnostic-files.md
tsalo: > having such specifications in the document text would be a big step back from toward the target of having a "machine readable" schema.

I see it as an incremental move forward. It definitely doesn't solve a lot of the problems (e.g., requirement levels, conditional relationships between metadata fields), but to be fair those are the _hard_ problems so I don't feel too bad about it.

> Why not to add metadata key to the corresponding records, e.g. in this case of https://github.com/bids-standard/bids-specification/blob/master/src/schema/top_level_files.yaml#L14 where similar to entities of https://github.com/bids-standard/bids-specification/tree/master/src/schema/datatypes to list the 'requirements' for each one of those? and then in the macros just point to what metadata of to render here (e.g. of top_level_files.dataset_description here)? or you can immediately see that this would not work?

I'll have to try it out, but that sounds like a good plan for the PR _after_ this one. I figure that, since we want to separate the core definitions from the rules of the schema, we can tackle those two pieces in separate stages.

tsalo: I think the best next step for the schema (after this) is to try to identify the types of rules we can expect from the schema, along with any really weird edge cases. That's what I've tried to do in #620, although I worry that my current list isn't exhaustive. Without a list of necessary rules, I'm not sure if we can define them in YAML files in a way that will satisfy everyone.

yarikoptic: Sure could and should be a multistage process. I am ok with your proposal. I just thought that  it might be nice to just put rules identification for later pr, thus keep description free form, while already allowing for machine readable specification of possible requirement levels and keeping everything under src/schema. But such a move could indeed be just yet another follow up PR to not block this one

rwblair:
> @rwblair could the validator use files like this? They're basically a translation of the validator JSON files to individual YAML files, with added descriptions and units, so I assume it wouldn't require too much effort to support them.

The validator certainly could build its json schema files from this. Is there any where encoded in yaml which terms are acceptable in which json files?

tsalo:
>Is there any where encoded in yaml which terms are acceptable in which json files?

Not yet. I think I'll need to translate over the context files from the validator, with some minor additions.

sappelhoff:
I think the URL the linkchecker flags has moved to `https://www.hedtags.org/docs/HEDTaggingStrategyGuide.pdf`, we should probably fix that in a PR that will be merged soon, or make a dedicated PR

tsalo:
>The validator certainly could build its json schema files from this. Is there any where encoded in yaml which terms are acceptable in which json files?

@rwblair, I'm sorry, it looks like I misunderstood your question. Right now the _rules_ applied to the metadata terms (e.g., which terms apply to each suffix or datatype) are not implemented, and I wasn't planning to put that in this PR. I figured that the validator could retain those rules for now, but without the information that is in the YAML files I'm proposing.

I'm hopeful that someone will be able to tackle the rules (and thus #699) in a future PR.

tsalo:
Okay, I think I'm happy with this, so I'm marking it as ready for review.

src/schema/metadata/PowerLineFrequency.yaml
```diff
@@ -0,0 +1,11 @@
...
+key_name: PowerLineFrequency
+description: |
+  Frequency (in Hz) of the power grid at the geographical location of the MEG
+  instrument (for example, 50 or 60).
+anyOf:
+  - type: number
+    exclusiveMinimum: 0
+    unit: Hz
+  - type: string
+    pattern: ^n/a$
```
tsalo, line 11: Could we replace the `^n/a$` pattern with a one-item `enum`?

src/schema/metadata/BackgroundSuppressionNumberPulses.yaml
```diff
@@ -0,0 +1,8 @@
+---
+key_name: BackgroundSuppressionNumberPulses
+description: |
+  The number of background suppression pulses used.
+  Note that this excludes any effect of background suppression pulses applied
+  before the labeling.
+type: integer
+minimum: 0
```
tsalo, line 8: Is `minimum` equivalent to `inclusiveMinimum`?

src/schema/metadata/BolusCutOffDelayTime.yaml
```diff
@@ -0,0 +1,20 @@
...
+  saturation pulse(s), in seconds.
+  This can be a number or array of numbers, of which the values must be
+  non-negative and monotonically increasing, depending on the number of bolus
+  cut-off saturation pulses.
+  For Q2TIPS, only the values for the first and last bolus cut-off saturation
+  pulses are provided.
+  Based on DICOM Tag 0018,925F `ASL Bolus Cut-off Delay Time`.
+anyOf:
+  - type: number
+    Minimum: 0
```
tsalo, line 14: I've noticed there's both `minimum` and `Minimum`.

src/schema/metadata/IntendedFor.yaml
```diff
@@ -0,0 +1,14 @@
+---
+key_name: IntendedFor
+description: |
+  Contains one or more filenames with paths relative to the participant subfolder.
+  The path needs to use forward slashes instead of backward slashes.
+  This field is OPTIONAL, and in case the fieldmaps do not correspond to any
+  particular scans, it does not have to be filled.
+anyOf:
+  - type: string
+    minLength: 1
```
tsalo, line 10: What if we added a `format: relative_path` option?

src/schema/metadata/MEGCoordinateSystem.yaml
```diff
@@ -0,0 +1,13 @@
...
+  Defines the coordinate system for the MEG sensors.
+  See [Appendix VIII](/99-appendices/08-coordinate-systems.html)
+  for a list of restricted keywords for coordinate systems.
+  If `"Other"`, provide definition of the coordinate system in
+  `MEGCoordinateSystemDescription`.
+anyOf:
+  - $ref: _MEGCoordSys.yaml
+  - $ref: _EEGCoordSys.yaml
+  - $ref: _StandardTemplateCoordSys.yaml
+  - $ref: _StandardTemplateDeprecatedCoordSys.yaml
```
tsalo, line 13: This currently renders as "string or string or string or string". Not sure how best to clean it up.

src/schema/metadata/Genetics.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: |
+  An object containing information about the genetics descriptor.
+type: object
+properties:
+  Database:
+    $ref: Genetics.Database.yaml
+  Dataset:
+    $ref: Genetics.Dataset.yaml
+  Descriptors:
+    $ref: Genetics.Descriptors.yaml
```
tsalo, line 12: What does everyone think of this instead of defining those fields in here directly?

src/schema/metadata/MEGCoordinateSystem.yaml
rwblair: I tried rendering it with the enum typed fields joined with " or ", unfortunately the resulting list seems unwieldy to put directly into the table. This also wouldn't make sense for other fields like GeneticLevel where the enumerations are listed directly in the description.

Is there any current situation where removing duplicates of  what `_resolve_metadata_type` returns when generating the string for anyOf would cause a problem?

tsalo: Currently it's just a single single that's appended to, so it's difficult to identify and remove duplicates. I could reorganize it to a list of lists or something that would be easier to identify and remove duplicates though.

sappelhoff (commented):
How can we most effectively review these 234 files? Do you have a "guide" @tsalo ? Where is another set of eyes most useful?

src/schema/metadata/PowerLineFrequency.yaml
sappelhoff: yes, technically that would be equivalent except for the message that a user gets upon validating that schema. (I.e., "foo must be pattern X" versus "foo must be one of X" or the like). But that small difference would be no problem IMHO.

tsalo:
I think the best place to start is the inline comments/questions I added. I also added a lot of information to `src/schema/README.md` that I _hope_ is accurate, and I'd love input on it. Other than that, I guess the best things to do would be to check the generated tables, which are rendered alongside the original tables for easy comparison.

src/schema/metadata/PowerLineFrequency.yaml
tsalo: The current issue is that the tables render these kinds as `object or string` instead of `object or "n/a"`. I thought maybe checking against an enum list and rendering those values directly if the list is short would be easier than figuring something out for pattern strings, but maybe it'll be easier if I figure out what the characters in the pattern represent, and what patterns are possible?

src/schema/metadata/IntendedFor.yaml
tsalo: I've added `dataset_relative`, `participant_relative`, and `unit` formats. They may not be used yet, but they could ultimately be useful, I think.

src/schema/metadata/Columns.yaml
```diff
@@ -0,0 +1,8 @@
+---
+key_name: Columns
+description: |
+  Names of columns in file.
+type: array
+items:
+  type: string
+  minLength: 1
```
tsalo, line 8: Could we just assume that strings have a minimum length of 1? If the field is provided with a string, then I think it should have at least one character to be valid, from a BIDS standpoint. Missing data should be provided with `"n/a"`, so there's not really a case for `""` appearing in jsons.

src/schema/metadata/DatasetDOI.yaml
```diff
@@ -0,0 +1,10 @@
+---
+key_name: DatasetDOI
+description: |
+  The Digital Object Identifier of the dataset (not the corresponding paper).
+  DOIs SHOULD be expressed as a valid
+  [URI](/02-common-principles.html#uniform-resource-indicator);
+  bare DOIs such as `10.0.2.3/dfjj.10` are
+  [DEPRECATED](/02-common-principles.html#definitions).
```
tsalo, line 8: I started replacing internal relative links (e.g., `../02-common-principles.html#definitions`) with absolute versions (`/02-common-principles.html#definitions`). How does everyone feel about this change? I think it would be good to apply throughout the whole specification.

src/04-modality-specific-files/09-positron-emission-tomography.md
```diff
@@ -130,18 +130,60 @@ which we divide into several categories:
...
 | InstitutionName           | RECOMMENDED           | [string][]    | The name of the institution in charge of the equipment that produced the composite instances. Corresponds to DICOM Tag 0008, 0080 InstitutionName.                                     |
 | InstitutionAddress        | RECOMMENDED           | [string][]    | The address of the institution in charge of the equipment that produced the composite instances. Corresponds to DICOM Tag 0008, 0081 InstitutionAddress.                               |
-| InstitutionDepartmentName | RECOMMENDED           | [string][]    | The department in the institution in charge of the equipment that produced the composite instances. Corresponds to DICOM Tag 0008, 1040 Institutional Department Name.                 |
+| InstitutionalDepartmentName | RECOMMENDED           | [string][]    | The department in the institution in charge of the equipment that produced the composite instances. Corresponds to DICOM Tag 0008, 1040 Institutional Department Name.                 |
```
tsalo, line 152: I _think_ this was a typo in the PET BEP?

src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -354,12 +405,30 @@ head.
 
 General fields:
 
+{{ MACROS___make_metadata_table(
+   {
+      "IntendedFor": (
+         "OPTIONAL",
+         "This identifies the MRI or CT scan associated with the electrodes, "
+         "landmarks, and fiducials.",
+      )
```
tsalo, line 414: What does everyone think of this approach for extending descriptions for modality- or table-specific purposes?

src/schema/metadata/PowerLineFrequency.yaml
sappelhoff: > maybe it'll be easier if I figure out what the characters in the pattern represent, and what patterns are possible?

There are some libraries that allow you to generate strings from a given regular expression, see: https://github.com/asciimoo/exrex

So perhaps it wouldn't even be too difficult. And coincidentally, it may provide us with a nice double check of our regexps :-)

src/schema/metadata/BackgroundSuppressionNumberPulses.yaml
sappelhoff: > Is minimum equivalent to inclusiveMinimum?

conceptually yes, but according to the JSON schema spec, there is no dedicated keyword for `inclusiveMinimum`: https://json-schema.org/understanding-json-schema/reference/numeric.html?highlight=minimum#range



        x ≥ minimum
        x > exclusiveMinimum
        x ≤ maximum
        x < exclusiveMaximum

sappelhoff: related: https://github.com/bids-standard/bids-validator/pull/1261

src/schema/metadata/BolusCutOffDelayTime.yaml
sappelhoff: That's weird, I am pretty sure that in JSON schema, the keys are case sensitive, and that `minimum` is a key and `Minimum` is not. I am just surprised that we did not run into errors so far.

will fix in https://github.com/bids-standard/bids-validator/pull/1261

src/schema/metadata/Genetics.yaml
sappelhoff: looks clean to me

src/schema/metadata/Columns.yaml
sappelhoff: agreed! And then users can still abuse it by providing `" "`, but it might be sufficient to make them think twice :-)

src/schema/metadata/DatasetDOI.yaml
sappelhoff: Why are you doing that? I thought it's an advantage to point to the relative files and have Mkdocs generate the suitable links for us? Pointing to html directly feels error prone but I can't put my finger on it right now :thinking:

src/04-modality-specific-files/09-positron-emission-tomography.md
sappelhoff: yes, most likely. fix: https://github.com/bids-standard/bids-specification/pull/771

src/schema/metadata/Columns.yaml
tsalo: I think the same can apply to `minItems` when the type is `array`, so I'll remove all cases of `minLength: 1` and `minItems: 1` from the YAML files.

src/schema/metadata/BackgroundSuppressionNumberPulses.yaml
tsalo: Awesome! I'll replace `Minimum` with `minimum` and `inclusiveMaximum` with `maximum`. Thanks!

src/schema/metadata/DatasetDOI.yaml
tsalo: One reason is that relative links depend on where the metadata field is rendered, so what might work for one file might not work for another. It should also make it easier to resolve links in external applications (e.g., NIDM-Terms).

sappelhoff: I see :thinking: I couldn't find info on whether that's good or bad: https://www.mkdocs.org/user-guide/writing-your-docs/#internal-links

it's not really mentioned in the docs over there. So maybe we should carefully try it, and if it doesn't break anything it's fine.

PR #773, opened 2021-04-12
FIX: Rename "Unit" metadata to "Units" for consistency with existing fields

effigies:
As noted in https://github.com/bids-standard/bids-specification/pull/622#issuecomment-817734317, we previously had in [anatomy imaging data](https://bids-specification.readthedocs.io/en/bep-009/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data):

https://github.com/bids-standard/bids-specification/blob/a7faded33a96404b07edbe57e4c1f1630ee0400e/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md#L222-L233

Given that this metadata field has the same intent, to describe the interpretation of voxel values in an imaging file, we should be consistent.

@mnoergaard @melanieganz I do not know if this means we should also change, e.g., `InjectedMassUnit`.

@tsalo I notice that we don't seem to have this field in a metadata table anywhere. Do you have a suggestion of where to add it?


tsalo:
> @tsalo I notice that we don't seem to have this field in a metadata table anywhere. Do you have a suggestion of where to add it?

In many cases, suffixes are restricted to one specific unit (usually `"arbitrary"`). I'll need to dig around to see which suffixes (or entity-value combinations) allow "Units".

sappelhoff:
> @mnoergaard @melanieganz I do not know if this means we should also change, e.g., InjectedMassUnit.

that'd sound better (more consistent) to me.

mnoergaard:
I am also fine with this (for all names including units). I will go ahead and update the example data sets and the validator.

effigies (commented):
Full list of changed fields:

* Units
* InjectedRadioactivityUnits
* InjectedMassUnits
* SpecificRadioactivityUnits
* TracerMolecularWeightUnits
* InjectedMassPerWeightUnits
* MolarActivityUnits
* InfusionSpeedUnits
* PharmaceuticalDoseUnits
* ReconMethodParameterUnits

sappelhoff (approved):

effigies:
Can we get a second review?

mnoergaard (approved):

PR #781, opened 2021-04-20
[MISC] Add mini entity tables after filename templates

tsalo:
This implements a proposal of @Remi-Gau's to make it easier for readers to link to relevant entity definitions.

Changes proposed:
- Call `MACROS___make_entity_table` after each call of `MACROS___make_filename_template` to generate a limited entity table for the suffixes/data types in the filename template.

### DIFF
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -138,6 +138,8 @@ whenever possible. See also
 
 {{ MACROS___make_filename_template(datatypes=["anat"]) }}
 
+{{ MACROS___make_entity_table(datatypes=["anat"]) }}
+
 Anatomical (structural) data acquired for that participant. Currently supported
 non-parametric structural MR images include:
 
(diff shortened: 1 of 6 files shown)


tsalo:
One small bug I'm noticing- there are weird rows in some tables (see below for an example) where the datatype is set, but the suffix list (contents of the parens) is empty:

image

tsalo:
I've resolved the issue by just skipping parts of the filtered schema where no suffixes are found. This is a hack, but until we have a better schema structure and/or filtering function, I think it's a solid solution.

sappelhoff (approved):
I think it looks nice and helps ... good overview and accessible links :+1:

Remi-Gau (approved):

tsalo:
Pending any negative feedback or requests for substantive changes, I think we can merge this on Tuesday (May 4).

effigies:
TBH I find it a bit cluttering, but I don't want to stand in the way if others disagree.

Remi-Gau:
> TBH I find it a bit cluttering, but I don't want to stand in the way if others disagree.

If you feel that way I suspect some others will too.

As this is more a user experience than pure content should I ping the google group and/or neurostars just to get a couple more opinions on this?

yarikoptic:
In response to the ping:  looking at HTML version (didn't check PDF) - even though I love the structure, it does look a bit cluttered indeed... It is useful but it seems that `template` above shows the same information, isn't it?

Also, I wondered, if in HTML version we could simply wrap it into 
```
<details>
<summary>Entities table</summary>

... table HTML goes here

</details>
```
so it would become expandable:

<details>
<summary>Entities table</summary>

... table HTML goes here

</details>

Remi-Gau:
I am mostly "nays" on that one, so might be better to close it, no?

PR #784, opened 2021-04-21
FIX: Typos discovered by codespell

yarikoptic:
I did not go wild and automate it but codespell is nice and someone might.  There would need to be some exclusions given all abbreviations etc, but here is what it would say now after these fixes and excluding embedded JS

```shell
$> codespell src | grep -v '\.js'
src/01-introduction.md:119: TE ==> THE, BE, WE
src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md:97: TE ==> THE, BE, WE
src/04-modality-specific-files/images/PET_scan_overview.svg:1: bre ==> be, brie
src/04-modality-specific-files/images/PET_scan_overview.svg:1: ba ==> by, be
src/04-modality-specific-files/images/PET_scan_overview.svg:1: ba ==> by, be
src/99-appendices/08-coordinate-systems.md:92: ALS ==> ALSO
src/99-appendices/08-coordinate-systems.md:94: ALS ==> ALSO
src/99-appendices/08-coordinate-systems.md:95: ALS ==> ALSO
```


effigies (approved):
All typos look good. I'm a little surprised that the linter didn't complain about misaligned tables. Did you fix those and it just doesn't look like it?

yarikoptic:
hm, no - I didn't do anything extra beyond that `sed` command

sappelhoff (commented):
Typos look good thanks.

I'd be in favor of (before merging) pushing a commit that obviously violates a table to see if CIs are triggered by that ...

we can then revert the commit and know more about whether or not we need to fix CIs.

effigies:
@sappelhoff I manually fixed the table alignment. I would rather get this in, release 1.6.0, and check on the CI after. I don't want a potential rabbit hole blocking obviously good fixes, and I don't want to release with typos we know are there.

WDYT?

sappelhoff:
yes, makes sense!

sappelhoff (approved):

PR #786, opened 2021-04-23
[FIX] consistently refer to DICOM Tags throughout the specification

Hboni:
Only some changes for Dicom tag representation to be consistent across the page and as used in MRI.
Also, remove what I think is duplicate in a figure caption


sappelhoff (approved):
Thanks for seeing this and fixing a bunch of these inconsistencies @Hboni.

I took your start and made a pass through the whole spec, so that we always use `DICOM Tag \d+, \d+`

Also thanks for removing that duplication in the caption!

PS: You made this PR from your `master` branch. Just for next time, I recommend that you make a separate branch off your (up to date) master branch to submit changes/proposals --> that can avoid some git difficulties that we luckily didn't run into here (so far)

src/schema/entities.yaml
```diff
@@ -164,7 +164,7 @@ part:
   description: |
     This entity is used to indicate which component of the complex
     representation of the MRI signal is represented in voxel data.
-    The `part-<label>` key/value pair is associated with the DICOM tag
+    The `part-<label>` key/value pair is associated with the DICOM Tag
     `0008,9208`.
```
effigies, line 168: I think you've been adding spaces.

```suggestion
    `0008, 9208`.
```

effigies (approved):

sappelhoff:
Thanks both. To be clear: I don't know how we would *want* DICOM tags to be formatted in the spec. I am simply completing/fixing up @Hboni's start to make the formatting consistent based on which convention was initially used in the spec. Prior to PET, ASL, qMRI adding some more tags.

PR #787, opened 2021-04-26
[SCHEMA] Use macro for filename templates in file collections appendix

tsalo:
Closes None.

Changes proposed:
- Auto-generate the filename templates for anat and fmap file collections in **Appendix X: File collections**.


tsalo:
@agahkarakuzu would you be willing to take a look at the rendered filename templates? [Here](https://bids-specification--787.org.readthedocs.build/en/787/99-appendices/10-file-collections.html#anatomy-imaging-data) is the rendered pattern for anat data.

agahkarakuzu:
Hey @tsalo sure, I'll look into it, thank you!

agahkarakuzu:
Most of the issues I raise below are probably because of the way we defined requirements in the first place, but now it has become much easier to see what is reasonable and what is less. Thanks for the effort! 

### MP2RAGE

`[_flip-<index>]_inv-<index>[_part-<label>]_MP2RAGE.json` 

I think it is safe to drop `OPTIONAL` from the flip entity and make `part` required if possible. Optional `echo` is the only alternative use case.  

### MPM 

`[_echo-<index>]_flip-<index>_mt-<label>[_part-<label>]_MPM.json`

Echo should be required for MPM (at least that's what I infer from the article), @ChristophePhillips @lazaral can confirm if that's the case. 

### B1maps 

It all looks good to me, I think people will check the table below and not rely solely on the template, which should make it clear how to use `acq` for some fmap file collections.

sappelhoff:
> I think it is safe to drop OPTIONAL from the flip entity and make part required if possible. Optional echo is the only alternative (file-collection entity relevant) use case.

> Echo should be required for MPM (at least that's what I infer from the article), @ChristophePhillips @lazaral can confirm if that's the case.

@agahkarakuzu et al. those sound like quite important changes: Changing requirement levels can have a large impact on downstream software. We should (1) be sure that this is needed, (2) if it is needed, act quickly on it, before more and more people start to adopt the current (then to be outdated) way of specifying.

effigies:
We're a couple releases in. I think we need to make them RECOMMENDED and make sure there's a warning in the validator.

tsalo:
@agahkarakuzu can you open an issue about the requirement changes in the [bids 2.0 repository](https://github.com/bids-standard/bids-2-devel)? That way, at least it'll be on the to-do list for when we're able to make backwards-incompatible changes.

Regarding this PR, is there anything I need to change or is it good as-is?

tsalo:
>@tsalo the PR LGTM.

Thanks!

>If possible, as @effigies suggested, switching to RECOMMENDED is a good solution short term.

Definitely worth discussing, but we haven't used "RECOMMENDED" with entities in the spec. Just REQUIRED and OPTIONAL.

sappelhoff:
@tsalo for the sake of the original intent of this PR (rendering text in the spec from the schema), can that be done without changes to what's in the text? 

Then we could merge this, and discuss potential changes to qMRI requirement levels in a separate, dedicated issue.

tsalo:
>@tsalo for the sake of the original intent of this PR (rendering text in the spec from the schema), can that be done without changes to what's in the text?

Sorry, could you clarify what you mean by "can that be done"? I think the PR, as-is, could be merged, with the caveat that two issues should be opened:
1. Adding "RECOMMENDED" as a valid entity requirement level.
2. In `bids-2-devel`, changing entity requirements for some qMRI suffixes, according to Agah's initial comment.

sappelhoff (approved):
Ah, I see now - thanks. So this PR just made obvious what was already there: Namely that it wouldn't hurt to refine a few requirement level settings for qMRI.

But as that's not fully possible in a backward compatible way, let's 1) push it back to BIDS 2.0, and 2) discuss in a new issue whether/how we can RECOMMEND in file path templates (backward compatible).

+1 for merging then, as the issues can be raised by the respective stakeholders later.

tsalo:
Awesome! We have two approving reviews, so I'll wait until Thursday and merge then, unless someone raises a concern.

PR #774, opened 2021-04-13
[SCHEMA] Add metadata term files

tsalo:
References #699. Continues from #762.

Changes proposed:

- Take the metadata JSON files from the bids-validator and convert them to YAML format.
- Add "description" and "key_name" fields to the metadata YAML files.
- Add functions to render the metadata information in tables within the schema. These functions require the specification text to include the requirement levels.
- Create files for a number of metadata terms that aren't present in the validator, but are present in specification tables.

To do:

- [x] Identify and incorporate section-specific definition modifications.
- [x] Fix rendering of multiple equivalent data types (e.g., `MEGCoordinateSystem`, which renders as `string or string or string or string`).
- [x] update w.r.t. https://github.com/bids-standard/bids-specification/pull/806
- [x] Remove hardcoded tables (after review)
- [x] deprecate ScanDate field after #798 is merged (could be anytime now)
- [x] potentially update with respect to #825 (if merged)

## Reviewing

I think the best approach to reviewing this PR is to (1) review the proposed metadata schema structure defined in the schema README and a handful or randomly chosen example terms, and (2) compare the rendered metadata tables against the hardcoded ones in the generated specification. These tasks can be split across multiple reviewers.

- [x] 02-common-principles
- [x] 03-modality-agnostic-files
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Common Metadata Fields
    - [x] Scanner Hardware
    - [x] Sequence Specifics
    - [x] remaining sections
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Anatomy imaging data
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Task (including resting state) imaging data
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Diffusion imaging data
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Arterial Spin Labeling perfusion data
- [x] 04-modality-specific-files/01-magnetic-resonance-imaging-data Fieldmap data
- [x] 04-modality-specific-files/02-magnetoencephalography
- [x] 04-modality-specific-files/03-electroencephalography
- [x] 04-modality-specific-files/04-intracranial-electroencephalography
- [x] 04-modality-specific-files/08-genetic-descriptor
- [x] 09-positron-emission-tomography
- [x] 05-derivatives/02-common-data-types
- [x] 05-derivatives/03-imaging
- [x] 99-appendices/11-qmri
[...]

PR #792, opened 2021-05-03
[SCHEMA] Apply schema rules to entity values

### LINKED ISSUE #809: Specify allowed values for entities that requires specific values in bids schema
Ref > 
https://github.com/bids-standard/bids-specification/blob/c0434124f4b07fd53af925aecae124c2a9102a0b/src/schema/entities.yaml#L169

Would it make sense to enumerate all allowed values for certain entities like this?

```
    options:
      - phase
      - mag
      - real
      - imag 
```

Downstream apps (ezBIDS, bids-validator, etc) can then use this information to let user pick / validate from one of the allowed values?

tsalo:
Closes #809.

The goal of this PR is to move toward applying the rules that govern metadata values (in #774) to entity values as well. Those rules support a wide range of possible values and restrictions, but the most important one right now is `type: string` combined with `enum`, which lets us list specific acceptable values for the `part` and `mt` entities.

Changes proposed:
- Add `type: string` to all entity definitions in the schema.
- Replace `format: label` with `enum: <list>` for the part and mt entities.
- Add an "Allowed values" subsection to entity sections in the Entities appendix for those entities with restricted values.
- Default to `format = label` when format is not defined in entity schema definitions (just to work around the new enums).

Pending questions:
1. Should I use `type: integer` for any of the `index`-format entities? I kept `type: string` because of the leading zeros issue, although I'm not sure if that applies to all entities or just run.
    - Answered. They should be strings but we'll ultimately have an index-specific format to validate against.


tsalo:
I'm thinking it might also be a good idea to create a new `formats.yaml` that defines the different formats used in the schema. Something like:

```yaml
label:
    description: |
        Freeform labels without special characters.
    pattern: ^[0-9a-zA-Z]+$
index:
    description: |
        Non-negative, non-zero integers, optionally prefixed with leading zeros for sortability.
    pattern: ^[0-9]+$
unit:
    description: |
        A unit.
    pattern: ^[0-9a-zA-Z]+$
```

sappelhoff:
> Should I use type: integer for any of the index-format entities? I kept type: string because of the leading zeros issue, although I'm not sure if that applies to all entities or just run.

I think the leading zeros is recommended for each index-format entity ... to make sorting more reasonable: That's nothing `run`-specific. Perhaps we can say it must be a string of digits? So that a conversion to "int" of the string does not result in a ValueError, as in this Python example:

```python
int("0001")  # 1
float("0001")  # 1.0
float("0001a")  # ValueError, could not convert string to float: '0001a'

```

tsalo:
Sounds good. I'll leave it as string then.

sappelhoff (approved):
LGTM

For future reviewers, you can look at:
- https://bids-specification--792.org.readthedocs.build/en/792/99-appendices/09-entities.html#mt
- https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#mt

tsalo:
Thanks! The second approval triggers the five-day waiting period and I'll merge next Thursday if no one raises any concerns.

Regarding `formats.yaml`... I think I'll open a new issue about it rather than try to shoehorn it in here.

sappelhoff:
I think you can merge starting from today, as it's been 5 days :-)

image

PR #798, opened 2021-05-20
[FIX] Deprecate ScanDate (PET) in favor of AcquisitionTime in scans.tsv files

mnoergaard:
This PR updates the description of the `ScanDate`field in the PET specification to be in accordance with the correct units for dates. The updates have been discussed here: https://github.com/bids-standard/bids-validator/pull/1283#issuecomment-844892636

and the corresponding PR for the validator can be found here: https://github.com/bids-standard/bids-validator/pull/1290


sappelhoff (dismissed):
LGTM, but would be interested in what others think.

Remi-Gau:
In summary, it mentions the "official" way to write units for dates with the extra that time is optional in this case.

I did not remember we had kept that piece of metadata for PET and thought we had fallen back on using scans.tsv to keep track of that. Doesn't that open the door for this piece of metadata to be stored in 2 different places and also have potentially conflicted info if that piece of information is stored in both places? (Sorry to have missed that one before the final merge and sorry to not remember how this discussion was resolved... Feel free to ignore).

mnoergaard:
Yeah - that's the alternative. I guess it would be most clean to have everything in the scans.tsv/json file.

effigies:
+1 for deprecating ScanDate and recommending using acq_time in scans.tsv.

VisLab:
Putting in scans.tsv is a better alternative -- it allows tool builders to consistently look for the time irrespective of modality

mnoergaard:
Thank you all - I believe we seem to have a consensus. I will go ahead and make the necessary changes to the specification and the validator, and then make a PR.

effigies:
We need to keep the field around for backwards compatibility and mark it deprecated. I made a proposal in edf0237.

rwblair (approved):

sappelhoff:
>  I made a proposal in edf0237.

@effigies it's a bit unclear to me where that commit is. Is it on your fork? Maybe you could make a PR to Martin's branch (or push directly since he's agreed with your suggestion)

image

effigies (approved):

sappelhoff (approved):

sappelhoff:
Thanks @mnoergaard for starting this and everybody else for reviewing.

PR #806, opened 2021-05-22
Add device and acquisition metadata for physio files

Remi-Gau:
fixes #803 

Add the following keys for physio files (NOT for stim files)

- [x] Manufacturer
- [x] ManufacturersModelName
- [x] DeviceSerialNumber
- [x] SoftwareVersion

PR #811, opened 2021-05-30
Update physio metadata

Remi-Gau:
Adds new physio metadata fields into macro to generate tables.

Implements #806 into #774


tsalo (approved):
Thanks!

sappelhoff (approved):
Thanks!

CI failures are fixed in master and show up here because the branches are slightly outdated. Merging now.

Issue #779, opened 2021-04-16
Add sample entity and samples.tsv file


mariehbourget:
## Context and motivation

Hi BIDS community!

As part of the development of the Microscopy BEP ([BEP031](https://bids.neuroimaging.io/bep031)), we want to add a new `sample` entity to BIDS. This `sample` entity was introduced in order to distinguish different tissue samples from the same subject.

The `sample` entity may also be used by the Animal Ephys BEP ([BEP032 ](https://bids.neuroimaging.io/bep032) @SylvainTakerkart) and could benefit other modalities as well.

This issue aims to start a discussion about the details of the sample entity between the 2 BEP groups and with the BIDS community. It will also facilitate the breaking down of BEPs in smaller modules by adding the sample entity as a separate PR.

## Definition of the sample entity

To ensure compatibility with BIDS other modalities, the `subject` entity should correspond to the participant (e.g. a human, a mouse, etc). To identify multiple tissue samples from the same subject, we define the sample entity in BEP031 as:
~~~
A tissue sample, volume or slice pertaining to a subject.
~~~

It is positioned after the optional `session` entity in the filename:
~~~
sub-<label>[_ses-<label>]_sample-<label>_<modality_suffix>.<ext>
~~~

## samples.tsv file

In BEP031, a **samples.tsv** file was added at the root of the dataset along with **participants.tsv**.

The **samples.tsv** file would have 2 required columns:

- `sample_id`: corresponding to `sample-<label>` of the filename
- `participant_id`: corresponding to `sub-<label>` of the filename

Another column `sample_type` was also suggested as required:

- `sample_type`: kind of sample from [ENCODE BiosampleType](https://www.encodeproject.org/profiles/biosample_type) 

We should also discuss if (and how) we want to encode an additional identifier when a sample is derived from another sample (e.g., a slice is derived from a block of tissue).

## participants.tsv file

As part of the subject vs. sample definitions, we would also like to add 2 columns to the **participants.tsv** file:

1. `species`: string corresponding to the Binomial species name from [NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi), required when different from “Homo sapiens”
We think species should be in **participants.tsv** and not samples.tsv as it is an attribute of the subject and not the sample.

2. `pathology`: required when different from “Healthy”
[...]

Issue #751, opened 2021-03-09
RFC: stimuli BEP


yarikoptic:
As @Remi-Gau hinted by #695 , we still lack total clarity on original stimuli storage and annotation.

We do have
- `stimuli/` folder which, like `sourcedata/` is nohow "prescribed" for a specific structure.
- `stim_file` column in `_events.tsv` as to point to a (unregulated) location under `stimuli/` and then populating that `stim_file` description/HED tags within `_events.json` (bless the inheritance principle).
- ["human wording"](https://github.com/bids-standard/bids-specification/blob/1d7b363b77c9ecd78443ac76633ef890c1f4c3bd/src/04-modality-specific-files/05-task-events.md#stimuli-databases) to point to the origin of a stimuli within `_events.json` as possibly coming from some DB
- `_stim.tsv.gz` files for "signals related to the stimulus" (but not necessarily stimulus)

In respect to the first 3 items, and in conjunction with
- stimuli collections/datasets should be self-sufficient/described
- incoming requests to store stimuli datasets on DANDI archive

I wondered if there either an ongoing effort to standardize "stimuli datasets" so 
- they could be readily reusable across studies by simply placing them under `stimuli/<name>`  and avoiding necessity to describe stimuli in `_events.json` since information could be picked from their standardized layout 
- derivatives of them could be created and possibly shared along (e.g. all the feature extractions done by [pliers](https://github.com/psychoinformaticslab/pliers) (thanks to @tyarkoni , @qmac, et al)

With that in mind I am even thinking such datasets could follow BIDS mantra and just get "participant/subject" and `sub-` renamed to "stimulus"/`stim-`, and preserve `README.md`, `dataset_description.json`, `stimuli.tsv` etc

Worth a BEP/effort or may be it is already a "solved problem"? ;) WDYT?
 
Related:
- fresh #750 for storing `_stim.{mp3.mp4,...}` along side with neural data.  sidecar fields in that file could come handy why instrumenting specification of stimuli as presented from the shared across subjects `sourcedata/` (e.g. `StartTime`, possibly `TimeDrift` between scanner and stimuli delivery for lengthy (an hour) presentation, etc)

effigies:
At one point I talked to @Gilles86 about how he was storing stimuli, but don't clearly recall how deep we went. He might have some thoughts here.

[...]

PR #812, opened 2021-05-31
[ENH] BEP031 - New entity: sample and samples.tsv file

mariehbourget:
Dear BIDS community,

### Context

As part of the development of the Microscopy BEP ([BEP031](https://bids.neuroimaging.io/bep031) @mariehbourget, @jcohenadad) and Animal Ephys BEP ([BEP032](https://bids.neuroimaging.io/bep032) @SylvainTakerkart, @JuliaSprenger), the “sample” entity was introduced in order to distinguish different tissue samples from the same subject.

In brief, the scope of the BEP031 proposal is to extend the Brain Imaging Data Structure (BIDS) specification to 2D and 3D microscopy images of histological samples, both ex vivo and in vivo. In this context, we were advised to split the BEP into smaller separate PRs when possible.

### Contribution

The purpose of this PR is to add a new “sample” entity and `samples.tsv/json` files with required column (`sample_type`) and recommended columns (`pathology`, `derived_from`).

The "sample" entity is described as:
```
A sample pertaining to a subject such as tissue, primary cell or cell-free sample.
The `sample-<label>` key/value pair is used to distinguish between different samples from the same subject.
The label MUST be unique per subject and is RECOMMENDED to be unique throughout the dataset.
```

See issue #779 for related discussions on this topic.
Related PR #816 for new columns in `participants.tsv` file.


satra:
@mariehbourget - shouldn't this PR also describe a samples.tsv file providing additional details about a sample?

mariehbourget:
> @mariehbourget - shouldn't this PR also describe a samples.tsv file providing additional details about a sample?

Hi @satra,
From previous discussions in the BEP031 meeting and with the maintainers, and from the comments in #779 ([here](https://github.com/bids-standard/bids-specification/issues/779#issuecomment-840186739), [here](https://github.com/bids-standard/bids-specification/issues/779#issuecomment-840321709) and [here](https://github.com/bids-standard/bids-specification/issues/779#issuecomment-840564178)), my understanding was that dealing with the sample entity and the samples.tsv file in different PRs was preferred. So the plan is to open 2 separate PRs: 1) sample entity 2) participants.tsv and samples.tsv.

satra:
they would likely have to be merged at the same time, so perhaps doing it together instead of linking would work better for review.  @bids-standard/maintainers thoughts?

mariehbourget:
> they would likely have to be merged at the same time, so perhaps doing it together instead of linking would work better for review. @bids-standard/maintainers thoughts?

I suggest to add the `samples.tsv` file description in this PR, and open another PR for the additional recommended columns in the `participants.tsv` file.

effigies:
Sounds good to me.

mariehbourget:
Hi everyone!
As discussed, this PR now includes the description of the new `samples.tsv` file.
A new PR #816 is also open for the addition of recommended/optional columns in `participants.tsv`.

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+## Samples file
+
+Template:
+
+```Text
+samples.tsv
+samples.json
+```
+
+The purpose of this RECOMMENDED file is to describe properties of samples
```
satra, line 263: should this be recommended or required if you have a sample entity. i don't think i would want people to use the sample entity to run disambiguation routes, and we would want to know the type of sample.

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+you do use them, we RECOMMEND to use the following values for them:
+
+-   `pathology`: string value describing the pathology of the sample or type of control.
+    When different from `healthy`, pathology SHOULD be specified in `samples.tsv`.
+    The pathology MAY instead be specified in [Sessions files](06-longitudinal-and-multi-site-studies.md#sessions-file)
+    in case it changes over time.
+
+-   `derived_from`: `sample-<label>` key/value pair from which a sample is derived from,
+    for example a slice of tissue (`sample-02`) derived from a block of tissue (`sample-01`),
+    as illustrated in the example below.
```
satra, line 289: this makes me think that a samples.tsv could refer to entities that are not in the filenames. which would be fine. but we may want to illustrate or clarify.

satra:
@mariehbourget - this looks good. just a few clarification comments.

Remi-Gau:
Couple of open questions that relate more to where those changes should go rather than about their content:

Currently those changes read to me like the sample entity will be made "available" for ALL modalities.

--- 

Describing it in the common principles puts it at the same level as the other entities that are described there and that are pretty much available to all modalities (`subject`, `session`, `run`). Though this is not the case for `event` so there is a precedent although it is a suffix not an entity.

Should we possibly only have the entity defined only the in the entity table? Or do we feel that sample is a concept important enough that it deserves to be defined in the common principle?

I think a good case can be made for the latter but I just want to make sure we are on the same page.

---

Similarly putting the description of the `samples.tsv` in the modality agnostic section feels strange. Sample will be used for microscopy  and ephys. Therefore `samples.tsv` is not like the `scans.tsv` or `sessions.tsv` (the latter is being transferred into the modality agnostic section in another PR).

So maybe it should go somewhere else?

What has been done so far is to duplicate, into their respective page, a description of the file: for example with the `electrodes.tsv` and `channels.tsv` files in EEG, MEG, iEEG (also because they might have slightly different requirements). See also my regular mentions of the details about the task entity being repeated across several pages. `#IAmABrokenRecord`

As a user or reader of the spec this is usually fine because this concentrates all the info about a modality in a single page. As maintainer I am not a big fan of duplication.

The other problem is that we don't even have a microscopy or ephys page so far to put this in.

I would not be against the creation of a new page where we can put information about things that are shared by some but not all modalities: this could definitely help with "refactoring".

That latter point would probably deserves its own issue and PR, so feel free to disregard this for this PR.

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+
+Template:
+
+```Text
+samples.tsv
+samples.json
+```
+
+The purpose of this RECOMMENDED file is to describe properties of samples
+when the `sample-<label>` key/value pair is used in filenames.
```
effigies, line 264: I agree with @satra that this should be REQUIRED if `sample` is present. Something like the following?

```suggestion
The purpose of this file is to describe properties of samples, indicated by the `sample` entity.
This file is REQUIRED if `sample-<label>` is present in any file name within the dataset.
```

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+It is RECOMMENDED to accompany each `samples.tsv` file with a sidecar
+`samples.json` file to describe the TSV column names and properties of their values
+(see also the [section on tabular files](02-common-principles.md#tabular-files)).
+
+`samples.json` example:
+
+```JSON
+{
+    "sample_type": {
+        "Description": "type of sample from ENCODE Biosample Type (https://www.encodeproject.org/profiles/biosample_type)",
```
effigies, line 311:
```suggestion
        "Description": "type of sample from ENCODE Biosample Type",
        "TermURL": "https://www.encodeproject.org/profiles/biosample_type"
```

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+    },
+    "derived_from": {
+        "Description": "sample_id from which the sample is derived from",
```
effigies, line 314:
```suggestion
        "Description": "sample_id from which the sample is derived"
```

src/03-modality-agnostic-files.md
```diff
@@ -251,6 +251,71 @@ to date of birth.
...
+when the `sample-<label>` key/value pair is used in filenames.
+If this file exists, it MUST contain the three following columns:
+
+-   `sample_id`: MUST consist of `sample-<label>` values identifying one row
+    for each sample
+
+-   `participant_id`: MUST consist of `sub-<label>`
+
+-   `sample_type`: MUST consist of sample type values (for example: tissue,
+    primary cell or cell-free sample) from [ENCODE Biosample Type](https://www.encodeproject.org/profiles/biosample_type)
```
effigies, line 273: This may be overly restrictive. We do have the concept of TermURL, that allows researchers to associate a column with any ontology.

No strong opinion, here, though. If we want to require this TermURL for now, it can be loosened in the future to permit other ontologies.

effigies (commented):
Some quick thoughts. I'll try to have a more detailed read this week.

src/03-modality-agnostic-files.md
mariehbourget: I went with “recommended” because I think all the other `tsv` files in the specs are either recommended or optional, even `participants.tsv`. But I agree with @satra that the `sample_type` column is reason enough to make the file REQUIRED, I was just wondering if that makes sense with the rest of the spec? If so, I agree with the suggested changes.

src/03-modality-agnostic-files.md
mariehbourget: I addressed this in another similar question [here](https://github.com/bids-standard/bids-specification/pull/812#discussion_r646894795).

src/03-modality-agnostic-files.md
mariehbourget: I’m sorry, I don’t understand what you mean, could you give me an example. Thanks!

src/03-modality-agnostic-files.md
mariehbourget: For Apps down the line, it seems easier to restrict the possible values.

I have question about validation though:
1. If the value MUST be from a specific ontology, does it mean that we have to plan a validation for it?
2. And if the value is REQUIRED but the ontology is not, what would we validate?

mariehbourget:
@Remi-Gau, thank you for the feedback about where those changes should go, those are important questions and I’m glad to discuss them here.
> Should we possibly only have the entity defined only the in the entity table? Or do we feel that sample is a concept important enough that it deserves to be defined in the common principle?

The rationale for this was twofold:
1)	It was discussed in (https://github.com/bids-standard/bids-specification/issues/779#issuecomment-822660725) and in meetings that `sample` could eventually be used by modalities other than microscopy/ephys. For example, sample could be scans from different body part of the same subject in anat MRI.
2)	Moreover, `sample` is closely related to `subject` because it serves to identify “what” is scanned rather than “how” (e.g.: acq, task, ce). To preserve the BIDS subject definition, we added the sample entity. When sample is used, it is the combination of `sub-id` + `sample-id` that serves as the “unique identifier” for the specimen scanned.

On the other hand, since no modality is using the `sample` entity yet in the spec, we could add the entity in the appendix list, and postpone its integration to the common principles for when microscopy is merged.

> Similarly putting the description of the `samples.tsv` in the modality agnostic section feels strange. Sample will be used for microscopy and ephys. Therefore `samples.tsv` is not like the `scans.tsv` or `sessions.tsv`

For the same reason as above, I put `samples.tsv` in the modality agnostic files section. Another particularity is that the `samples.tsv` file is global and sits at the root of the dataset (unlike `channels.tsv` or `electrodes.tsv`).

Finally, as you mentioned, without an official page for microscopy yet, that information has nowhere to go! I am not against creating a page for microscopy with only the `sample` description and `samples.tsv` file for now. That page could eventually become the official microscopy page and the samples info could be moved to the page “about things that are shared by some but not all modalities” you suggested.

So I’m open to make changes to facilitate the integration and maintenance of the specs but I'm not sure what is best, let me know your thoughts. Thanks!

SylvainTakerkart:
> For the same reason as above, I put `samples.tsv` in the modality agnostic files section. Another particularity is that the `samples.tsv` file is global and sits at the root of the dataset (unlike `channels.tsv` or `electrodes.tsv`).

+1 on listing the samples.tsv file in the modality-agnostic files section ; on top of @mariehbourget 's arguments, I think any future modality that can be recorded in animal models will probably use this file...

src/03-modality-agnostic-files.md
satra: imagine you submit data from a slab (slab-01), which came from a block (sample-01). both are samples, and one may want to provide properties for both, even though the data itself is about the slab. so:

```
sample-01, tissue, sub-01
slab-01, slice, sample-01
```

src/03-modality-agnostic-files.md
mariehbourget: The current set of Encode Biosample is fairly small:
`"cell line", "in vitro differentiated cells", "primary cell", "cell-free sample", "cloning host", "tissue", "whole organisms", "organoid", "technical sample"`. I could list them all in the description as accepted values if we go with a fixed ontology.

As for TermURL, I understand it the same way as @satra, as a link to the definition of what `sample_type` means rather than the link to the ontology itself. If we want to point to one of the Ontobee definition, I would go with `Biospecimen` (http://www.ontobee.org/ontology/NCIT?iri=http://purl.obolibrary.org/obo/NCIT_C70699) rather than `TissueSample` (http://purl.obolibrary.org/obo/NCIT_C19697). However, it feels strange to have a `TermURL` linking to a different ontology than the one we wish to use for the values.

src/03-modality-agnostic-files.md
mariehbourget: In the current proposal, if `slab-01` and `sample-01` are both samples, they are required to be identified with a `sample-<label>`. This was the idea behind `sample` to avoid adding several entities to the specs like `block`, `slice`, `slab`, etc. So any of these should be identified as a `sample`. From your example, that would give something like this, where the `id` is any alphanumeric string:

|sample_id|participant_id   |sample_type|derived_from|
|---------|-----------------|-----------|------------|
|sample-01|sub-01         |tissue     |n/a         |
|sample-slab01|sub-01         |tissue     |sample-01|

In the first line: because every sample is always associated with a subject (`ex: sub-01_sample-01_SEM.ome.tif`), `sub-01` goes in `participant_id` and not in `derived_from`.
In the second line: `slice` is not one of the `sample_type` accepted in ENCODE biosample, they would both be `tissue` in this case.

Let me know if I interpreted correctly your comment. Thanks!

satra: thank you @mariehbourget - this looks great.

mariehbourget:
Thank you @effigies and @Remi-Gau for your feedback, I left a few suggestions and questions regarding the validation and BIDS-compliance, let me know what you think and I’ll make the appropriate changes to the PR. Thanks!

mariehbourget:
Hi @bids-standard/maintainers,
The doc build failed after my last change 2 weeks ago. I’m not sure why this error occurs, I can build the doc locally with mkdocs serve. I was wondering if there is a way that I can re-trigger the build of the doc or correct the error in the PR? Thanks!

sappelhoff:
yes, you can merge `master` into your branch, these issues have been fixed on there already. Or you can rebase your changes on top of `master` and force push.

Look at this screenshot: 

image


you are 37 commits behind `master`, that should be removed so that you are even with `master`

src/03-modality-agnostic-files.md
effigies: participants.tsv is not REQUIRED because there is no required metadata in it. if `sample_type` must exist for a dataset to be interpretable, then this should be REQUIRED.

src/03-modality-agnostic-files.md
effigies: You're right about TermURL. Apologies for the confusion. I would probably go with a fixed set, and we can say that these are taken from Encode Biosample.

effigies (approved):
LGTM.

satra (approved):

yarikoptic (approved):
Thanks everyone who worked on this BEP -- looks great.  My only concern is a bit of "special casing" `samples` as to be promoted to the top level, although it is "per-subject(/session)" level description.

FWIW: Following my "generalization blurb" at https://github.com/bids-standard/bids-specification/issues/751#issuecomment-820815484 , I still think that it might have been more consistent if as long as we do not have `sample-<sample_id>/` top level directories, to have `sub-<label>/sub-<label>_samples.{tsv,json}` (and with inheritance principle do allow for `samples.json` on top) analogous to `_scans.{tsv,json}` we have.

But if, as proposed in this PR, `subject_id` column is retained in such `*samples.tsv` at any level -- we "kinda have it", just that this PR promotes it to the top level via inheritance principle, so no conflict for the future if we do decide to support `sub-<label>/sub-<label>_samples.{tsv,json}` etc.  Thus -- approve! ;)

effigies:
@yarikoptic I initially (https://github.com/bids-standard/bids-specification/issues/779#issuecomment-823303226) preferred using a per-subject table or the for a global table that applied to all via the inheritance principle, but the BEP working group argued that a global table would need to be recomputed by every tool, so dropped the objection (https://github.com/bids-standard/bids-specification/issues/779#issuecomment-825777533).

mariehbourget:
Thank you everyone for the reviews!
I’ll tag @SylvainTakerkart and @lazaral here if they want to take a look at the latest changes.

Also, this is our first PR for BEP031, are there other things we have to do before the merge? Thanks!

lazaral:
This looks great to me, and very much in line with the discussions of the BEP031 working group. Thank you @mariehbourget and all the BIDS maintainers who helped with it. Exciting to see the first PR from BEP031 is about to come to life!

Remi-Gau (approved):

CPernet:
@mariehbourget @jcohenadad @SylvainTakerkart @JuliaSprenger @satra @Remi-Gau @sappelhoff is there any overlap with https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/08-genetic-descriptor.html   

SampleOrigin: Describes from which tissue the genetic information was extracted
TissueOrigin: Describes the type of tissue analyzed for SampleOrigin value 'brain'
BrainLocation: Refers to the location in space of the TissueOrigin
CellType: Describes the type of cell analyzed

mariehbourget:
> is there any overlap with https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/08-genetic-descriptor.html

Hi @CPernet,
I double-checked the genetic descriptor part of the spec and I don’t see any overlaps.
If I understand correctly, the genetic information in `genetic_info.json` applies to all participants, whereas the `sample-<label>` entity from this PR is meant to distinguish between different samples from the same participant.

For example, a participant `sub-01` could have 2 tissue samples extracted (`sample-01` and `sample-02`). Each sample would be imaged by a microscope and the `sample-<label>` is used in the filename of those images, ex: `sub-01_sample-01_TEM.png`, `sub-01_sample-02_TEM.png`.

The `samples.tsv` file is meant to list all the different samples present in the dataset with their attributes. In particular, the column `sample_type` which describe the type of sample (ex: tissue, primary cell, cell-free sample). This is similar to the `CellType` field in `genetic_info.json` but does not directly overlap as it is an attribute of the samples that are imaged and not the cells used for genetic analysis.

There are no suggested columns in `samples.tsv` similar to `SampleOrigin`, `TissueOrigin` or `BrainLocation` from `genetic_info.json`. In the Microscopy BEP, we use the metadata fields `BodyPart` and `BodyPartDetails` to describe the tissue location (PET also uses BodyPart). However, those fields will be included in the JSON sidecar files associated to the image files (for each sample of each participant).

Please let me know if that answer your question or if your have any other concerns.

CPernet:
ok cool - I just wanted to make sure the names and possibly ontologies to pick information from are in agreement (and by that I mean we can amend genetic if that's easier, re cell type for instance) - thx

effigies:
Definitely a good thing to keep genetics concepts in mind as we add microscopy and electrophysiology. Thanks for the nudge, @CPernet!

satra:
i think the sample stuff can be pulled over into the genetic side too, or the columns of samples.tsv augmented with some of the keys. spatial proteomics and transcriptomics across tissues/samples from humans and others are going to share a bunch of things.

PR #802, opened 2021-05-22
[ENH] BEP030: Functional Near-Infrared Spectroscopy

rob-luke:
BEP030: Functional Near-Infrared Spectroscopy
This PR adds functional near-infrared spectroscopy to BIDS.
Leads: @rob-luke & @lpollonini 

- validator: https://github.com/bids-standard/bids-validator/pull/952
- examples: https://github.com/bids-standard/bids-examples/pull/333
- rendered specification draft: https://bids-specification.readthedocs.io/en/bep030/04-modality-specific-files/11-functional-near-infrared-spectroscopy.html

Progress:
- [x] Community reviewed proposal ([google docs](https://docs.google.com/document/d/1FzPt7TVpIAtlhrF9_9b2VBCZOrRXJASEMyHndhDQKl8/edit?ts=5f589efa#heading=h.nxhqi3lai5gs))
- [x] Transition google doc to markdown (this PR). [View current render here](https://bids-specification.readthedocs.io/en/bep030/04-modality-specific-files/11-functional-near-infrared-spectroscopy.html).
- [x] Fix language in document. It currently reads as if it was written by many people (as it was), and needs to be reworked to fix sentence structure and grammar. 
- [x] Extend validator to support this BEP (WIP: [GitHub PR link](https://github.com/bids-standard/bids-validator/pull/952)) 
- [x] Generate open datasets from several labs and ensure they are compliant with specification (e.g. [dataset from rob-luke](https://github.com/rob-luke/BIDS-NIRS-Tapping), [dataset from robertoostenveld](http://www.fieldtriptoolbox.org/example/bids_nirs/), these are listed in the manuscript)
- [x] Add example dataset to [bids-examples](https://github.com/bids-standard/bids-examples) ([dataset 1](https://github.com/bids-standard/bids-examples/pull/305), [dataset 2](https://github.com/bids-standard/bids-examples/pull/323))
- [ ] Resolve remaining issues in this PR as described [here](https://github.com/bids-standard/bids-specification/pull/802#pullrequestreview-1121994111)
- [ ] Preprint of the new extension (manuscript hs been reviewed by half the authors, about to be circulated to remaining authors).
- [ ] Press release 

NOTE: At all stages in this process we will actively advertise and invite feedback on the proposal. We invite all feedback no matter how minor or major. And we will advertise the draft website to the community for review once it is rendering correctly.

closes #438

---------------

Scope of proposal:

[...]

PR #813, opened 2021-06-01
[FIX] What is a composite instance? Change to measurement for non MRI modalities?

rob-luke:
What is a composite instance? I had never heard this term before and don't believe it is used in EEG or fNIRS. I googled it and it appears to be a DICOM specific term and thus fMRI specific? But I found the terminology in EEG MEG iEEG and Physiological pages.

I suggest either changing it to measurement (as in this PR), or adding a definition somewhere in the specification.

This is an issue as a PR. Hope that's ok.


src/04-modality-specific-files/06-physiological-and-other-continuous-recordings.md
```diff
@@ -133,12 +133,12 @@ sub-01_task-cuedSGT_run-1_echo-3_bold.nii.gz
...
+| Manufacturer           | RECOMMENDED           | [string][]    | Manufacturer of the system used to record the physiological data                                                                                                                                                |
+| ManufacturersModelName | RECOMMENDED           | [string][]    | Manufacturer's designation of the system used to record the physiological data                                                                                                                                  |
+| SoftwareVersions       | RECOMMENDED           | [string][]    | Manufacturer's designation of the acquisition software.                                                                                                                                                         |
+| DeviceSerialNumber     | RECOMMENDED           | [string][]    | The serial number of the equipment that produced the physiological data. A pseudonym can also be used to prevent the equipment from being identifiable, as long as each pseudonym is unique within the dataset. |
```
sappelhoff, line 141: I think I would keep the text for the different fields (here `DeviceSerialNumber`) the same everywhere. "measurements" would work here as well.

sappelhoff (commented):
thanks @rob-luke, this looks sensible.

Note that https://github.com/bids-standard/bids-specification/pull/774 aims to render all these tables from the schema, hence my comment to try and make the texts consistent. cc @tsalo is there something from the schema side to do here?

sappelhoff:
PS: linkchecker fails due to biosemi.com failing to renew their certificates (?) ... https fails, but http works. You could just change the link from https to http for now. I am not sure when they'll manage to get their https back up.

src/04-modality-specific-files/06-physiological-and-other-continuous-recordings.md
rob-luke: Here, I just tried to make the minimal change, and the lines above already had `physiological data`. Do you want me to make it measurements in all modality files? Including fMRI?

rob-luke:
FYI @mshader (https://github.com/bids-standard/bids-specification/pull/802#discussion_r642800061)

src/04-modality-specific-files/06-physiological-and-other-continuous-recordings.md
sappelhoff: That's something to discuss together with @tsalo: If we want to render all these metadata fields from the schema, the schema entry should/must be consistent.

tsalo: Yes, thank you. The "canonical" metadata definitions in the schema (see #774) are going to just use the generic definition (i.e., "measurement" instead of "physiological data"), so it would be preferable to just stick with "measurement" in the current tables.

rob-luke: Thanks @tsalo

rob-luke:
@sappelhoff I think I addressed all issues. How/who do I ping for review? There seems to be ways to ping a _team_, is that what I should do? And which team?

tsalo (approved):
The changes LGTM.

sappelhoff:
@rob-luke these are the teams you can ping: https://github.com/orgs/bids-standard/teams/everyone --> all nested under "everyone", so if you ping [at]bids-standard/everyone, EVERYBODY will get a notification ... if you go down the nesting, the amount of pings reduces accordingly.

Usually it's fine to ping one of the maintainers that you last were in touch with about an issue/pr.

sappelhoff (approved):

rob-luke:
Thanks @sappelhoff for explaining. Looks like this is good to go then :)

sappelhoff:
Yes, I think so too, will merge! Thanks a lot :-)

Just a note for you: Usually we wait about 5 days after the most recent commit before merging a PR, to give the community enough time to see it, see: https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules

But in this case it's a relatively uncontroversial fix that has been approved and open for many days now without many changes.

rob-luke:
Will do @sappelhoff, and Ill be sure to read through that document too.

PR #816, opened 2021-06-04
[ENH] BEP031 - New columns to participants.tsv file

mariehbourget:
Dear BIDS community,

### Context

As part of the development of the Microscopy BEP ([BEP031](https://bids.neuroimaging.io/bep031) @mariehbourget, @jcohenadad) and Animal Ephys BEP ([BEP032](https://bids.neuroimaging.io/bep032) @SylvainTakerkart, @JuliaSprenger), the “sample” entity was introduced along with a `samples.tsv` file describing properties of the samples in PR #812.

To better describe animal participants, new properties of the subjects also need to be included in `participants.tsv`.

### Contribution

The purpose of this PR is to add new recommended column (`species`) and optional columns (`strain`, `strain_rrid` and `diagnosis`) to the participants.tsv file.

See issue #779 for related discussions on this topic.

### DIFF
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -162,16 +162,20 @@ participants.json
 ```
 
 The purpose of this RECOMMENDED file is to describe properties of participants
-such as age, sex, handedness.
+such as age, sex, handedness, species, strain, strain_rrid, diagnosis.
 If this file exists, it MUST contain the column `participant_id`,
 which MUST consist of `sub-<label>` values identifying one row for each participant,
 followed by a list of optional columns describing participants.
 Each participant MUST be described by one and only one row.
 
-Commonly used *optional* columns in `participant.tsv` files are `age`, `sex`,
-and `handedness`. We RECOMMEND to make use of these columns, and
-in case that you do use them, we RECOMMEND to use the following values
-for them:
+When different from `homo sapiens`, `participants.tsv` SHOULD include a `species`
+column, and the value MUST be the string of the binomial species name from
+[NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi).
+
+Commonly used *optional* columns in `participants.tsv` files are `age`, `sex`,
+`handedness`, `strain`, `strain_rrid` and `diagnosis`. We RECOMMEND to make use
+of these columns, and in case that you do use them, we RECOMMEND to use the
+following values for them:
 
 -   `age`: numeric value in years (float or integer value)
 
@@ -197,6 +201,15 @@ for them:
     -   for "ambidextrous", use one of these values: `ambidextrous`, `a`, `A`,
         `AMBIDEXTROUS`, `Ambidextrous`
 
+-   `strain`: string value indicating the strain of the species
+
+-   `strain_rrid`: research resource identifier ([RRID](https://scicrunch.org/resources/Organisms/search))
+    of the strain of the species
+
+-   `diagnosis`: string value describing the diagnosis of the participant.
+    The diagnosis MAY instead be specified in [Sessions files](06-longitudinal-and-multi-site-studies.md#sessions-file)
+    in case it changes over time.
+
 Throughout BIDS you can indicate missing values with `n/a` (for "not
 available").
 
@@ -213,9 +226,9 @@ It is RECOMMENDED to accompany each `participants.tsv` file with a sidecar
 `participants.json` file to describe the TSV column names and properties of their values (see also
 the [section on tabular files](02-common-principles.md#tabular-files)).
 Such sidecar files are needed to interpret the data, especially so when
-optional columns are defined beyond `age`, `sex`, and `handedness`, such as
-`group` in this example, or when a different age unit is needed
-(for example, gestational weeks).
+optional columns are defined beyond `age`, `sex`, `handedness`, `species`, `strain`,
+`strain_rrid` and `diagnosis`, such as `group` in this example, or when a different
+age unit is needed (for example, gestational weeks).
 If no `units` is provided for age, it will be assumed to be in years relative
 to date of birth.
 


src/03-modality-agnostic-files.md
```diff
@@ -162,16 +162,20 @@ participants.json
...
 If this file exists, it MUST contain the column `participant_id`,
 which MUST consist of `sub-<label>` values identifying one row for each participant,
 followed by a list of optional columns describing participants.
 Each participant MUST be described by one and only one row.
 
-Commonly used *optional* columns in `participant.tsv` files are `age`, `sex`,
-and `handedness`. We RECOMMEND to make use of these columns, and
-in case that you do use them, we RECOMMEND to use the following values
-for them:
+When different from `homo sapiens`, `participants.tsv` SHOULD include a `species`
```
satra, line 171: shouldn't this be MUST, given that all of BIDS assumes humans at this point.

```suggestion
When different from `homo sapiens`, `participants.tsv` MUST include a `species`
```

src/03-modality-agnostic-files.md
```diff
@@ -197,6 +201,15 @@ for them:
     -   for "ambidextrous", use one of these values: `ambidextrous`, `a`, `A`,
         `AMBIDEXTROUS`, `Ambidextrous`
 
+-   `strain`: string value indicating the strain of the species
+
+-   `strain_rrid`: research resource identifier ([RRID](https://scicrunch.org/resources/Organisms/search))
+    of the strain of the species
+
+-   `diagnosis`: string value describing the diagnosis of the participant.
```
satra, line 209: i don't know if this has to be a string value. in many datasets on openneuro diagnosis/dx is present and can be an enumerated type. also, this is one place, where one can have multiple designations depending on the study. we should allow for some notion of that, or simply remove diagnosis from this file.

src/03-modality-agnostic-files.md
effigies: I don't know how we would validate a `MUST`, here. Also, there are rodent datasets that may not have this column in this form at this point, so we would be breaking backwards compatibility if we could validate. What about:

```Markdown
The RECOMMENDED `species` column MUST be a binomial species name from the
[NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi).
For backwards compatibility, if `species` is absent, the participant is assumed to be
`homo sapiens`.
```

Also, REQUIRE-ing a species name from `NCBI Taxonomy` feels like it's going to be difficult to validate, as we will need to either query the database or maintain a list of accepted names, updating the validator as new use cases arise... Is there a validation plan?

src/03-modality-agnostic-files.md
```diff
@@ -197,6 +201,15 @@ for them:
     -   for "ambidextrous", use one of these values: `ambidextrous`, `a`, `A`,
         `AMBIDEXTROUS`, `Ambidextrous`
 
+-   `strain`: string value indicating the strain of the species
```
effigies, line 204: Examples for each of these would be useful.

src/03-modality-agnostic-files.md
mariehbourget: Thank you @effigies  for the suggestion, I think assuming `homo sapiens` if the column is omitted is a strong incentive without breaking backward compatibility. I would be in favor of that.

However, I had not thought about the validation, querying the database seems like the best option to not have to maintain an up-to-date list in the validator but it may be difficult to implement. Are there similar requirements elsewhere in the spec? Would the alternative of “SHOULD” or “strongly RECOMMENDED” be advisable?

Also, thinking about it, I think I should add examples other than `homo sapiens` like `mus musculus` and `rattus norvegicus` in the description.

src/03-modality-agnostic-files.md
mariehbourget: Just to clarify, do you mean example directly in the description, like above for handedness with `ambidextrous`, `a`, `A`, etc, or an example of `particpants.tsv` for an animal.

I did not change the example of `participants.tsv` below which is an example for human. I thought having complete examples for both human and animal would maybe be too much.

src/03-modality-agnostic-files.md
mariehbourget: The aim of this PR is to add new columns to describe animal properties. In that context, I agree that diagnosis may be out of scope.

For context, I added the columns following this discussion https://github.com/bids-standard/bids-specification/issues/779#issuecomment-822660725, https://github.com/bids-standard/bids-specification/issues/779#issuecomment-822662620 and https://github.com/bids-standard/bids-specification/issues/779#issuecomment-823303226 because we also introduced pathology in `samples.tsv`.

src/03-modality-agnostic-files.md
satra: @effigies - regarding validation, i will raise the same issue here as the other post. i'm not sure we actually validate values for example for sex or anything that could have levels or enumerations. 

> I don't know how we would validate a MUST

while one could at least detect presence, i agree that keeping with the current perspective of the participants.tsv being a recommended file, we can keep things recommended instead of required.

species does get a little complicated, especially for animals, as you start going into species + genotype notions. here is our generic participant at a timepoint model in dandi: https://github.com/dandi/dandischema/blob/master/dandischema/models.py#L642 (technically all of those properties could come into play, with some being more important for animal studies).

src/03-modality-agnostic-files.md
satra: perhaps a few examples from: https://www.jax.org/jax-mice-and-services/find-and-order-jax-mice/most-popular-jax-mice-strains

sappelhoff (commented):
I have a more general and perhaps controversial comment, that I'd like to hear opinions on.

Up until now BIDS has been mostly about humans and their brain+behavior data. I assume that >80% of users will continue to come to BIDS from this angle.

Adding coverage for animal ephys, and microscopy into BIDS is great, and I fully support it. However is it a good idea to introduce very specific concepts (such as "strain") that are not used in 80% of BIDS use-cases`*` into a very general section that discusses things like age, sex, and handedness (which are ubiquitous)? I feel like this would be better discussed in a dedicated, separate section of the spec (animal ephys / microscopy respectively), as it avoids clutter and potential confusion / annoyance with the spec.

`*`(this is my assumption/prediction, feel free to challenge).

src/03-modality-agnostic-files.md
```diff
@@ -162,16 +162,20 @@ participants.json
...
-Commonly used *optional* columns in `participant.tsv` files are `age`, `sex`,
-and `handedness`. We RECOMMEND to make use of these columns, and
-in case that you do use them, we RECOMMEND to use the following values
-for them:
+When different from `homo sapiens`, `participants.tsv` SHOULD include a `species`
+column, and the value MUST be the string of the binomial species name from
+[NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi).
+
+Commonly used *optional* columns in `participants.tsv` files are `age`, `sex`,
+`handedness`, `strain`, `strain_rrid` and `diagnosis`. We RECOMMEND to make use
```
rob-luke, line 176: Could group be added to this list as its used in the example below?

src/03-modality-agnostic-files.md
mariehbourget: @effigies, I modified the species description as you suggested and added examples in 8b41137.
For validation purposes, I kept both the column and the taxonomy as RECOMMENDED and not REQUIRED.
Let me know what you think.

src/03-modality-agnostic-files.md
mariehbourget: Following your suggestion, I removed the diagnosis column in 7146144 as being out of scope for this PR (animal properties).

src/03-modality-agnostic-files.md
mariehbourget: Examples added in b451671.

mariehbourget:
> Adding coverage for animal ephys, and microscopy into BIDS is great, and I fully support it. However is it a good idea to introduce very specific concepts (such as "strain") that are not used in 80% of BIDS use-cases into a very general section that discusses things like age, sex, and handedness (which are ubiquitous)? I feel like this would be better discussed in a dedicated, separate section of the spec (animal ephys / microscopy respectively), as it avoids clutter and potential confusion / annoyance with the spec.

Thanks @sappelhoff for your feedback!
I changed the wording to make it clearer that it applies to species other than human and added an example for `strain` and `strain_rrid`, so it may create less confusion in that way.

The main goal of this PR is to promote `species` as a recommended field in `participants.tsv` to introduce animal data in the spec, as discussed previously with BEP032 and with the steering group.

We foresee that the BIDS-specification for microscopy will be adopted by many users, some have already started converting their datasets or working on compatible acquisition workflows, but I would not be able to quantify such adoption at this point. 
So for microscopy and animal ephys, it is likely that `strain` and `strain_rrid` will apply for many users, but I agree that it would not be the case for 80% of BIDS at large.

I'm not against moving those field into the dedicated microscopy and animal ephys sections if we want to keep the general section more general. The main drawback being duplication in the specification.

src/03-modality-agnostic-files.md
mariehbourget: Good question, I'm not sure we should because `group` is also used below to illustrate how to describe an additional column that is not part of the optional ones in the `participants.json` example. From my understanding, `group` may have different meanings depending on the study and it would be hard to define it with a general example

mariehbourget:
Hi @effigies and @sappelhoff,
I’m currently working on the microscopy .md file for the main BEP031 PR. I would like to know your thoughts about the optional columns `strain` and `strain_rrid`, following the latest changes and comment above.

I can move them to the microscopy part if it makes more sense to have them per modalities rather than in the “general” participants file.

In that case, this PR would only add the species columns to participants.
Thank you!

sappelhoff:
tagging @Remi-Gau to chime in instead of me :-)

Remi-Gau:
Tempted to keep those `strain` and `strain_rrid` in the modality agnostic part, because at least 2 BEP will be using them and if we only start allowing in this part of the spec things that can REALLY ONLY apply to any modality then, we might have to move `handedness` out as I am not sure how it applies to rodents, invertebrates, cell cultures...

SylvainTakerkart:
> Tempted to keep those `strain` and `strain_rrid` in the modality agnostic part, because at least 2 BEP will be using them and if we only start allowing in this part of the spec things that can REALLY ONLY apply to any modality then, we might have to move `handedness` out as I am not sure how it applies to rodents, invertebrates, cell cultures...

agreed! if we only put them in the modality-specific part, we will have to repeat it for several future modalities and it's going to get more complex to handle the specs...

mariehbourget:
Thank you everyone for your feedback!
All the comments have been addressed and the PR is ready for re-review.

effigies (approved):

mariehbourget:
Great! thank you @effigies and @tsalo, should I do the merge or are the PRs usually merged by maintainers?

PR #824, opened 2021-06-25
[FIX] correct file location of scans.tsv file in example

ghisvail:


sappelhoff (approved):
Thanks for spotting this @ghisvail

PR #827, opened 2021-07-06
[SCHEMA] Add TSV column files

tsalo:
Closes None.

Tagging @kayrobbins so she can take a look.

Changes proposed:

- Add a YAML file describing columns for TSV files in the specification.
- Add a macro and Python function to render column tables.
- Replace hardcoded tables in the specification with macro-generated ones.
    - Currently, the generated tables are provided _in addition to_ the hardcoded tables, for easy comparison during review.

To do:
- [x] Add columns from #816 once it's merged.
- [x] Remove hardcoded tables, after the necessary approvals.


CPernet:
can I ask why using YAML? for many people BIDS is scary just because we have tsv and json ... this adds yet another layer -- I think we need to be careful

tsalo:
The main reason we chose yaml over json is that yaml allows comments. Also, the schema files describing the specification aren't things that BIDS users will interact with, any more than they spend time with the `pybids` or `bids-validator` code.

CPernet:
ah I see - ok cool thx, just wanted to know more

Remi-Gau:
Made a mental note yesterday though that I will need to double check how this splitting of content in several files and format might affect "on-boarding" of people who want to contribute to specifications (especially for BEP content but also in general).

Will try to remember to raise the issue at the next maintainers meeting I attend.



Sent from my Galaxy

erdalkaraca:
> Made a mental note yesterday though that I will need to double check how this splitting of content in several files and format might affect "on-boarding" of people who want to contribute to specifications (especially for BEP content but also in general). Will try to remember to raise the issue at the next maintainers meeting I attend.
> […](#)
> Sent from my Galaxy

Normally (speaking from experience), a technical schema is somehow self-contained. Indeed, the many yaml files may scare off contributors, but, as Taylor mentioned, they are rather intended for automatic processing via tooling

tsalo:
I will consolidate the column files into a single file if #883 is merged. I'm planning to hold off on changing the structure in this PR until that PR is handled.

effigies:
@tsalo Apologies for taking so long. Which PR should be considered higher priority?

tsalo:
This PR contains content that needs to be reviewed, while the other really just needs a quick once-over from anyone who _uses_ the schema. The changes in #883 include updates to the internal schema-processing code, so nothing in the specification should really change there.

In summary, I think that #883 is higher priority, but this PR will need a more thorough review.

tsalo:
@effigies just a small note to follow up from our call earlier this week. It looks like the JSON fields used to describe TSV columns (e.g., "LongName", "Description", "Levels") are included in the metadata YAML files, so they won't need to be added here.

src/schema/columns/type.yaml
```diff
@@ -0,0 +1,35 @@
+---
+name: type
+description: |
+  Type of channel; MUST use the channel types listed below.
+  Note that the type MUST be in upper-case.
+type: string
+enum:
```
Remi-Gau, line 7: Am I correct that you are pulling all the available types for EEG, MEG and iEEG ?

If I remember MEG has some that might not exist for EEG for example.

From the macros in the specs I see that you are not using this to generate the channel type tables. Am I getting this right?

Remi-Gau (approved):
Not sure if you are planning to add more to this but so far this looks good to me.

src/schema/columns/type.yaml
tsalo: That's true. Unfortunately, we don't have any way to account for that just yet. I guess I could do an anyOf enum, though. That would separate the different options out a little bit.

Yeah, the channel tables aren't being generated from these files yet.

tsalo:
>Not sure if you are planning to add more to this but so far this looks good to me.

It turns out I completely missed the contents of `participants.tsv`, `samples.tsv`, `scans.tsv`, and `sessions.tsv`, but I think it should be complete now.

src/schema/columns/type.yaml
Remi-Gau: Okidoki. Thanks for the clarification. :smiley:

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -227,6 +227,14 @@ MUST be present **in this specific order**:
 | type            | REQUIRED              | Type of channel; MUST use the channel types listed below. Note that the type MUST be in upper-case.                                                                           |
 | units           | REQUIRED              | Physical unit of the value represented in this channel, for example, `V` for Volt, or `fT/cm` for femto Tesla per centimeter (see [Units](../02-common-principles.md#units)). |
 
+{{ MACROS___make_columns_table(
+   {
+      "name_channels": "REQUIRED",
```
sappelhoff, line 232: this now contains iEEG specific info, which seems out of place in the MEG spec :thinking: 

> name and all ieeg type channels are required to have a position. The reference channel name MAY be provided in the reference column

Can we have this info only in the iEEG channel name column description?

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+name_segmentations:
+  name: name
+  description: |
+    The unique label name.
+  type: string
+notch:
+  name: notch
+  description: |
+    Frequencies used for the notch filter applied to the channel, in Hz.
+    If no notch filter applied, use n/a.
```
sappelhoff, line 223:
```suggestion
    If no notch filter applied, use `n/a`.
```

src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -222,6 +222,14 @@ MUST be present **in this specific order**:
 | type            | REQUIRED              | Type of channel; MUST use the channel types listed below. Note that the type MUST be in upper-case.                                                                           |
 | units           | REQUIRED              | Physical unit of the value represented in this channel, for example, `V` for Volt, or `fT/cm` for femto Tesla per centimeter (see [Units](../02-common-principles.md#units)). |
 
+{{ MACROS___make_columns_table(
+   {
+      "name_channels": "REQUIRED",
```
sappelhoff, line 227: same comment as for MEG above --> there is now ieeg specific info here, which seems out of place.

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+    Polar metabolite fraction of the radiotracer (0-1).
+  type: number
+  minimum: 0
+  maximum: 1
+name_channels:
+  name: name
+  description: |
+    Label of the channel.
+    The label must correspond to `_electrodes.tsv` name and all ieeg type channels are
+    required to have a position.
```
sappelhoff, line 206: This doesn't make sense technically. An electrode is an electrical contact, a channel is a potential measurement between two electrodes. When applying an average reference, or a dedicated (common to all) reference electrode, it is kind of sensible to name a channel like an electrode, and this is arguably the most common practice, and where a lot of confusion between electrode vs channel comes from.

The concept breaks apart for bipolar recordings: Imaging an EOG recording with an electrode named `VEOG+` above the left eye, and its counterpart named `VEOG-` below the left eye. Both these electrodes together form the channel named `VEOG` ... however, in electrodes.tsv, this bipolar recording would need two rows, not a single one. Then again in channels.tsv, there would only be `VEOG`, a single row.

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+    Indicate `n/a` in the absence of software filters applied.
+  anyOf:
+  - type: string
+  - type: string
+    enum:
+    - n/a
+status:
+  name: status
+  description: |
+    Data quality observed on the channel `(good/bad)`.
```
sappelhoff, line 380: I think we should also clarify that `n/a` is a valid entry next to good and bad.

src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -303,6 +333,14 @@ SHOULD be present:
 | material        | RECOMMENDED           | Material of the electrode  (for example, Tin, Ag/AgCl, Gold).          |
 | impedance       | RECOMMENDED           | Impedance of the electrode, units MUST be in `kOhm`.                   |
 
+{{ MACROS___make_columns_table(
+   {
+      "type": "RECOMMENDED",
```
sappelhoff, line 338: the electrode type is different from channel type --> for channel types, we have a list of restricted keywords, for electrode type, we don't (the latter is completely freeform text right now)

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+  type: number
+  minimum: 0
+  maximum: 1
+name_channels:
+  name: name
+  description: |
+    Label of the channel.
+    The label must correspond to `_electrodes.tsv` name and all ieeg type channels are
+    required to have a position.
+    The reference channel name MAY be provided in the reference column.
```
sappelhoff, line 207: make obvious that this is a standardized column name

```suggestion
    The reference channel name MAY be provided in the `reference` column.
```

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -246,6 +256,18 @@ SHOULD be present:
 | status             | OPTIONAL              | Data quality observed on the channel (good/bad). A channel is considered bad if its data quality is compromised by excessive noise. Description of noise type SHOULD be provided in `[status_description]`.                                                                |
 | status_description | OPTIONAL              | Freeform text description of noise or artifact affecting data quality on the channel. It is meant to explain why the channel was declared bad in `[status]`.                                                                                                               |
 
+{{ MACROS___make_columns_table(
+   {
+      "reference": "OPTIONAL",
```
sappelhoff, line 261: this makes references to the EEG section here ... but we are in the iEEG section of the spec - so these references don't make sense here.

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -343,6 +365,16 @@ MUST be present **in this specific order**:
 | z               | REQUIRED              | Z position. If electrodes are in 2D space this should be a column of `n/a` values.                                               |
 | size            | REQUIRED              | Surface area of the electrode, units MUST be in `mm^2`.                                                                          |
 
+{{ MACROS___make_columns_table(
+   {
+      "name_electrodes": "REQUIRED",
+      "x": "REQUIRED",
+      "y": "REQUIRED",
+      "z": "REQUIRED",
```
sappelhoff, line 373: iEEG can contain 2D elec locations, ...originally the spec had a comment about that here, that `z` may be fully `n/a`. I think it'd be good to keep that info.

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -343,6 +365,16 @@ MUST be present **in this specific order**:
 | z               | REQUIRED              | Z position. If electrodes are in 2D space this should be a column of `n/a` values.                                               |
 | size            | REQUIRED              | Surface area of the electrode, units MUST be in `mm^2`.                                                                          |
 
+{{ MACROS___make_columns_table(
+   {
+      "name_electrodes": "REQUIRED",
+      "x": "REQUIRED",
```
sappelhoff, line 371: This previously contained the following information:

> The positions of the center of each electrode in xyz space. Units are specified in space-<label>_coordsystem.json.

I think that should be preserved. Possibly in the spec text above the table.

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -352,6 +384,15 @@ SHOULD be present:
 | group           | RECOMMENDED           | The group that the electrode is a part of. Note that any group specified here should match a group specified in `_channels.tsv`. |
 | hemisphere      | RECOMMENDED           | The hemisphere in which the electrode is placed, one of `['L' or 'R']` (use capital).                                            |
 
+{{ MACROS___make_columns_table(
+   {
+      "material": "RECOMMENDED",
+      "manufacturer": "RECOMMENDED",
+      "group": "RECOMMENDED",
```
sappelhoff, line 391: This is the electrodes.tsv table, but it says:

> Note that any groups specified in _electrodes.tsv must match those present here.

The reference should be to channels.tsv instead

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+  - ambidextrous
+  - a
+  - A
+  - AMBIDEXTROUS
+  - Ambidextrous
+  - n/a
+hemisphere:
+  name: hemisphere
+  description: |
+    The hemisphere in which the electrode is placed, one of `['L' or 'R']` (use capital).
```
sappelhoff, line 124: consistency

```suggestion
    The hemisphere in which the electrode is placed, one of `['L' or 'R']` (MUST be in upper-case).
```

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+    Physical unit of the value represented in this channel,
+    for example, `V` for Volt, or `fT/cm` for femto Tesla per centimeter
+    (see [Units](/02-common-principles.html#units)).
+  type: string
+value:
+  name: value
+  description: |
+    Marker value associated with the event (for example, the value of a TTL
+    trigger that was recorded at the onset of the event).
+  type: number
```
sappelhoff, line 474: string or number, currently in the spec.

src/schema/objects/columns.yaml
```diff
@@ -0,0 +1,507 @@
...
+    Size of the group (grid/strip/probe) that this electrode belongs to.
+    Must be of form `[AxB]` with the smallest dimension first (for example, `[1x8]`).
+  type: string
+duration:
+  name: duration
+  description: |
+    Duration of the event (measured from onset) in seconds.
+    Must always be either zero or positive.
+    A "duration" value of zero implies that the delta function or event is so
+    short as to be effectively modeled as an impulse.
```
sappelhoff, line 72: why did you cut out:

>  (or "n/a" if unavailable)

? :)

sappelhoff (commented):
Thanks for the work here @tsalo. I have several comments, some of which are probably beyond the scope of this PR (you decide).

I didn't review:

- src/05-derivatives/03-imaging.md
- src/04-modality-specific-files/09-positron-emission-tomography.md

src/03-modality-agnostic-files.md
```diff
@@ -266,6 +250,18 @@ samples.json
 
 The purpose of this file is to describe properties of samples, indicated by the `sample` entity.
 This file is REQUIRED if `sample-<label>` is present in any file name within the dataset.
+Each sample MUST be described by one and only one row.
+
+{{ MACROS___make_columns_table(
+   {
+      "sample_id": "REQUIRED",
+      "participant_id": "REQUIRED",
```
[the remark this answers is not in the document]
mariehbourget: Hi! I may be able to help here (my apologies if I'm missing some context for this PR).

`sample_id` and `samples.tsv` were introduced in PR #812.

In `samples.tsv`, the "unique" identifier per row is `sample_id`.

So the descriptions were:
`sample_id: MUST consist of sample-<label> values identifying one row for each sample`
and
`participant_id: MUST consist of sub-<label>` (without mention of one row per participant).

The resulting table for participants with more than 1 sample each looks like this:
```
sample_id participant_id
sample-01 sub-01
sample-02 sub-01
sample-03 sub-01
sample-04 sub-02
sample-05 sub-02
```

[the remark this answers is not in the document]
effigies: It seems we're getting `participants.tsv`-specific text in `samples.tsv` table, and more generally we're combining the term description and its usage in a specific file.

Here I think the correct fix is probably to make `participant_id` say something like "A participant identifier of the form `sub-<label>`." and then the `participants.tsv` macro can add text that says "There MUST be exactly one row for each participant."

src/schema/objects/columns.yaml
tsalo: I honestly have no idea 😕. I'll add it back now though!

src/04-modality-specific-files/02-magnetoencephalography.md
tsalo: Just to be sure, does the additional info not apply to MEG/EEG columns at all, or is the problem just that the term `ieeg` is used in the definition? If it's the former then I can definitely just append this info to the description within the table like we do with metadata fields. If it's the latter then I'd prefer to generalize the description.

src/04-modality-specific-files/03-electroencephalography.md
tsalo: I'll split it into `type_channels` and `type_electrodes` then. Thanks!

src/04-modality-specific-files/04-intracranial-electroencephalography.md
tsalo: The definitions for EEG and iEEG are _really_ different, so I decided to split it into two schema objects. I don't _love_ splitting definitions for the same column in the same file across modalities, so if there's any way to write a single definition that works for both that would be awesome. Do you think it's possible?

sappelhoff (approved):
I am happy with this, thanks Taylor.

effigies (approved):

PR #834, opened 2021-07-21
[SCHEMA] Separate schema object definitions from rules

### LINKED ISSUE #603: Separate schema item definitions from layout rules

tsalo:
Closes #603. This is still a WIP.

The original scope of this PR was to reorganize the entities to follow a similar convention to the suffixes and metadata fields within the schema.

Changes proposed:

- Move entity definitions from the `entities.yaml` file to entity-specific files in the `entities/` folder.
- Keep `entities.yaml` to define the order in which entities should appear.
- Move all definition files into appropriate subfolders of a new `objects/` folder.
- Move all rule files into a new `rules/` folder.

Reorganized structure:

- objects/
    - datatypes/
         - anat.yaml
    - entities/
         - subject.yaml
    - metadata/
         - EchoTime.yaml  
- rules/
    - entities.yaml
    - datatypes/
        - anat.yaml


src/schema/entities.yaml
```diff
@@ -1,307 +1,26 @@
+This file simply defines the order in which entities should appear within filenames.
```
tsalo, line 1: Might need to be renamed so it doesn't conflict with the `entities` folder.

src/schema/rules/metadata/mri.yaml
```diff
@@ -0,0 +1,83 @@
+---
+# Metadata rules that apply to all/most MRI data types
```
tsalo, line 2: Would love input on this general structure.

src/schema/rules/metadata/mri.yaml
```diff
@@ -0,0 +1,83 @@
...
+MatrixCoilMode: recommended
+CoilCombinationMethod: recommended
+PulseSequenceType: recommended
+ScanningSequence: recommended
+SequenceVariant: recommended
+ScanOptions: recommended
+SequenceName: recommended
+PulseSequenceDetails: recommended
+
+# Required if PET data are present
```
tsalo, line 23: How?

src/schema/rules/metadata/mri.yaml
```diff
@@ -0,0 +1,83 @@
...
+NonlinearGradientCorrection: recommended
+
+MRAcquisitionType: recommended
+MTState: recommended
+
+# Required if MTState is True
```
tsalo, line 29: How can this be done?

src/schema/rules/metadata/entities.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
tsalo, line 1: This isn't anywhere close to something usable, but I'm not sure what we can do here.

src/schema/rules/datatypes/func.yaml
```diff
@@ -19,6 +19,45 @@
...
+    # VolumeTiming: required
+    TaskName: required
+    NumberOfVolumesDiscardedByScanner: recommended
+    NumberOfVolumesDiscardedByUser: recommended
+    # DelayTime: recommended
+    # AcquisitionDuration: recommended
+    DelayAfterTrigger: recommended
+    # Above commented-out requirements overriden by the logic below,
+    # along with SliceTiming from the generic MRI metadata rules.
+    oneOf:
```
tsalo, line 35: Does this make sense?

erdalkaraca:
@tsalo I am currently experimenting with the schema and think it would be better to have the entities as an array at the top level of the entities.yaml files.
Example:

subject:
  name: Subject
  entity: sub
  description: |
    A person or animal participating in the study.
  type: string
  format: label

would become

- name: subject:
  **label**: Subject
  **key**: sub
  description: |
    A person or animal participating in the study.
  type: string
  format: label

It is also a bit confusing to have the whole object being called an "entity", then have one of its properties also named "entitiy".
Instead use name/label/key to differentiate between those aspects.

tsalo:
>I am currently experimenting with the schema and think it would be better to have the entities as an array at the top level of the entities.yaml files.

I really would prefer to separate the object definitions into separate files. Hypothetically, we could have one YAML file for each type of object (metadata fields, data types, suffixes, etc.), but I would like to avoid massive files for now. While `entities.yaml` isn't currently very large (and probably won't ever be), I'm leaning toward splitting it up to match our convention for other, similar objects, like the metadata fields.

>It is also a bit confusing to have the whole object being called an "entity", then have one of its properties also named "entitiy".
>Instead use name/label/key to differentiate between those aspects.

That's reasonable. I know that others have raised concerns with the field names in the entity definitions. I'd prefer to have that in a separate PR though, since hopefully this one will be more about reorganization and the addition of rules than redefining the objects.

tsalo:
I'm closing this in favor of a more limited version that doesn't include pseudo-code rules.

PR #845, opened 2021-08-06
[ENH] Add links to example datasets for each modality

### LINKED ISSUE #842: Add links to specific bids-example in the specification
I just realized that the EEG page links to some specific examples from the BIDS example repo:
https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html

I wonder if it would not be a good thing to generalize to other modalities too.

Any thoughts on this?

Remi-Gau:
Fixes #842 

Add links to example from the BIDS example repo in the following section:

- [x] MEG
- [x] iEEG
- [x] genetics
- [x] PET
- [ ] MRI
  - [ ] anat
  - [ ] func
  - [x] dwi
  - [x] fmap
  - [x] asl
  - [x] qmri
- [ ] behavior ?
- [x] physio

### DIFF
--- a/src/04-modality-specific-files/03-electroencephalography.md
+++ b/src/04-modality-specific-files/03-electroencephalography.md
@@ -10,7 +10,7 @@ The following example EEG datasets have been formatted using this specification
 and can be used for practical guidance when curating a new dataset.
 
 -   Single session per subject: [`eeg_matchingpennies`](https://doi.org/10.17605/OSF.IO/CJ2DR)
--   Multiple sessions per subject: [`eeg_rishikesh`](https://doi.org/10.5281/zenodo.1490922)
+-   Multiple sessions per subject: [`eeg_rishikesh`](https://github.com/bids-standard/bids-examples/tree/master/eeg_rishikesh)
 -   Combined with fMRI: [`eeg_rest_fmri`](https://osf.io/94c5t/files/)
 
 Further datasets are available from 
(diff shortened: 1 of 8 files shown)


src/04-modality-specific-files/03-electroencephalography.md
```diff
@@ -13,7 +13,8 @@ and can be used for practical guidance when curating a new dataset.
 -   Multiple sessions per subject: [`eeg_rishikesh`](https://doi.org/10.5281/zenodo.1490922)
```
Remi-Gau, line 13: The example in EEG link to "actual" data sets and not those in the bids-example repository.

Should we only point all links to the bids-example repo, given that the link to the real data is listed there anyway?

sappelhoff: yes, I think that'd be good! :-)

Remi-Gau:
While working on this I am tempted to split the table in the README of BIDS-example to match better with the BEP:
https://github.com/bids-standard/bids-examples#dataset-index

- get the PET example out of the multomodal table
- get the qMRI and the ASL examples out of the MRI dataset and in their own respective tables

I would be easier to then have a single link to those sub-tables in the spec, than a whole list of links in the spec. Also this way we can add more examples in BIDS example without having to do the same in the spec.

sappelhoff (approved):
LGTM! Thanks Remi.

tsalo (approved):
Looks good to me. Thanks!

PR #853, opened 2021-08-18
[INFRA] LGTM recommendation: Unused local variable

DimitriPapadopoulos:
The value assigned to local variable 'data' is never used.

### DIFF
--- a/pdf_build_src/process_markdowns.py
+++ b/pdf_build_src/process_markdowns.py
@@ -410,7 +410,7 @@ def edit_titlepage():
                 "\\end{titlepage}")
 
     with open('cover.tex', 'w') as file:
-        data = file.writelines(data)
+        file.writelines(data)
 
 
 def process_macros(duplicated_src_dir_path):
--- a/pdf_build_src/process_markdowns.py
+++ b/pdf_build_src/process_markdowns.py
@@ -16,7 +16,7 @@ from datetime import datetime
 import numpy as np
 
 sys.path.append("../tools/")
-from schemacode import macros
+from schemacode import macros  # noqa   (used in "eval" call later on)
 
 
 def run_shell_cmd(command):


yarikoptic (approved):
LGTM

sappelhoff (commented):
LGTM, you have a conflict to solve - but other than that, I think it's good to be merged.

Thanks for your recent series of PRs

pdf_build_src/process_markdowns.py
```diff
@@ -16,7 +16,7 @@
 import numpy as np
 
 sys.path.append("../tools/")
-from schemacode import macros
+from schemacode import macros  # lgtm [py/unused-import]
```
sappelhoff, line 19: is importing `macros` somehow required? or could we also remove this line? cc @tsalo

DimitriPapadopoulos: I believe its is required because we later call `eval("macros.some_function(...)")`:
https://github.com/bids-standard/bids-specification/blob/dd377ae/pdf_build_src/process_markdowns.py#L457

Thus, the `lgtm` suppression comment is quite useful, it reminds maintainers of this atypical call of `macros.` functions. I don't know if there's an equivalent `noqa` error code, but note that `noqa` support in LGTM.com is buggy  (github/codeql#6517).

sappelhoff: thanks! Yes, I have used `...  # noqa: F401` before. It that works, it'd be preferable to the LGTM thing, I think.
Perhaps even a comment like: `...  # noqa: F401 --> it's used in "eval" below`

DimitriPapadopoulos: A simple `noqa` would be taken into account by LGTM.com, but not `noqa: F401`. What about these options?
```
# noqa   (used in "eval" call  below)
```
```
# noqa: F401   (noqa because used in "eval" call  below)
```
```
# noqa: F401   lgtm [py/unused-import]   (used in "eval" call  below)
```

sappelhoff: +1 for option 2, if that works, else option 1

DimitriPapadopoulos: :+1:  OK for option 2, I believe the LGTM regex will catch the second lone `noqa` :-)

DimitriPapadopoulos: Option 2 is too long, and I have just noticed the second `noqa` requires spaces around it.

I've pushed option 1.

sappelhoff:
Thanks @DimitriPapadopoulos

Issue #596, opened 2020-09-03
Support internal links in the PDF version


tsalo:
Internal linking to entity definitions introduced in #568 does not work in the PDF version of the specification. While we're comfortable with leaving these links as a web-only feature, if anyone does have the time and interest to figure out how to support the links in the PDF version, that would be very helpful.

This stems from https://github.com/bids-standard/bids-specification/pull/568#issuecomment-686564674.

sappelhoff:
+1, would be awesome to get the internal links in the PDF working.

Currently we are explicitly *removing* all internal links from the MD source as a build step on the way from markdown to PDF, see: 

https://github.com/bids-standard/bids-specification/blob/1d2c9722c2ccf0eb493619a754173db18ffd497c/pdf_build_src/process_markdowns.py#L102

If I remember correctly that is because our internal links often look like this:

`[inheritance-principle](../some-file.md#some-section-header)`

That is, the internal links are often across different files.

--> This works well for the html generated from the MD source with mkdocs, because the html specification still consists out of several documents (html documents) that stay interlinked that way.

*However* when building a PDF, we append all the distinct documents into a single document, and the links that were previously across different files stop working, because we have a single file now.

A solution would thus probably be to replace the above linked `remove_internal_links` function with a function that replaces the cross-file links with links how they are predicted to be once all distinct files have been appended into a single file.

PR #855, opened 2021-08-18
[INFRA] fixing "remove_internal_links" for PDF build

### LINKED ISSUE #852: "LGTM" tool warns about regexp for "remove_internal_links" function (for PDF build)
Here are the two LGTM warnings:
* https://lgtm.com/projects/g/bids-standard/bids-specification/snapshot/225bf2292f8dc5754a4829c837074b91ad9d9980/files/pdf_build_src/process_markdowns.py?sort=name&dir=ASC&mode=heatmap#x48aac9687c9e224f:1
* https://lgtm.com/projects/g/bids-standard/bids-specification/snapshot/225bf2292f8dc5754a4829c837074b91ad9d9980/files/pdf_build_src/process_markdowns.py?sort=name&dir=ASC&mode=heatmap#x6ea466000e63cf12:1

The relevant lines of code:
* [line 114 of `pdf_build_src/process_markdowns.py`](https://github.com/bids-standard/bids-specification/blob/28ad463/pdf_build_src/process_markdowns.py#L114)
```python
        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-\w]+))\)')  # noqa: E501
```
* [line 117 of `pdf_build_src/process_markdowns.py`](https://github.com/bids-standard/bids-specification/blob/28ad463/pdf_build_src/process_markdowns.py#L117)
```python
        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._\w]+)\)')
```

Indeed, there are two **`\w`** within **classes** of characters delimited by [] and not groups of characters delimited by ():
* **`[\w\-\w]`**
* **`[#\w\-._\w]`**

See [Duplication in regular expression character class](https://lgtm.com/rules/7880099/):
> Character classes in regular expressions represent sets of characters, so there is no need to specify the same character twice in one character class. Duplicate characters in character classes are at best useless, and may even indicate a latent bug.
> 
> ### Recommendation
> 
> Determine whether a character is simply duplicated or whether the character class was in fact meant as a group. If it is just a duplicate, then remove the duplicate character. If was supposed to be a group, then replace the square brackets with parentheses.

DimitriPapadopoulos:
This regular expression includes duplicate character `\w` in a set of characters.

Fixes #852.

Or at least attempts to fix that issue. This patch should not modify the current behaviour of the program. Yet, please review it carefully as the duplication might hide a conceptual error - but then I suppose such an error would have been detected by now!

### DIFF
--- a/pdf_build_src/process_markdowns.py
+++ b/pdf_build_src/process_markdowns.py
@@ -112,10 +112,10 @@ def remove_internal_links(root_path, link_type):
     if link_type == 'cross':
         # regex that matches cross markdown links within a file
         # TODO: add more documentation explaining regex
-        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-\w]+))\)')  # noqa: E501
+        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-]+))\)')  # noqa: E501
     elif link_type == 'same':
         # regex that matches references sections within the same markdown
-        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._\w]+)\)')
+        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._]+)\)')
 
     for root, dirs, files in sorted(os.walk(root_path)):
         for file in files:
--- a/pdf_build_src/process_markdowns.py
+++ b/pdf_build_src/process_markdowns.py
@@ -112,10 +112,10 @@ def remove_internal_links(root_path, link_type):
     if link_type == 'cross':
         # regex that matches cross markdown links within a file
         # TODO: add more documentation explaining regex
-        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-]+))\)')  # noqa: E501
+        primary_pattern = re.compile(r'\[((?!http).[\w\s\.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-]+))\)')  # noqa: E501
     elif link_type == 'same':
         # regex that matches references sections within the same markdown
-        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._]+)\)')
+        primary_pattern = re.compile(r'\[([\w\s\.\(\)`*/–]+)\]\(([#\w\-\._]+)\)')
 
     for root, dirs, files in sorted(os.walk(root_path)):
         for file in files:


yarikoptic (approved):
IMHO looks LGTM

pdf_build_src/process_markdowns.py
```diff
@@ -111,10 +111,10 @@ def remove_internal_links(root_path, link_type):
     if link_type == 'cross':
         # regex that matches cross markdown links within a file
         # TODO: add more documentation explaining regex
-        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-\w]+))\)')  # noqa: E501
+        primary_pattern = re.compile(r'\[((?!http).[\w\s.\(\)`*/–]+)\]\(((?!http).+(\.md|\.yml|\.md#[\w\-]+))\)')  # noqa: E501
     elif link_type == 'same':
         # regex that matches references sections within the same markdown
-        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._\w]+)\)')
+        primary_pattern = re.compile(r'\[([\w\s.\(\)`*/–]+)\]\(([#\w\-._]+)\)')
```
effigies, line 117:
```suggestion
        primary_pattern = re.compile(r'\[([\w\s\.\(\)`*/–]+)\]\(([#\w\-\._]+)\)')
```

These `.`s should be escaped, too.

DimitriPapadopoulos: Indeed, I have fixed it. I have fixed the previous regex too. Can you have a look?

sappelhoff (commented):
It seems to me that there is something more basic that is broken with `remove_internal_links`.

The purpose of this function is a bit better described in #596, but the TLDR is: for the PDF build only, we need to remove Markdown links that refer to headings within the specification text.

However, looking at the "build_docs_pdf" build artifact (see screenshot): 

image

... you can see that there are several "internal" links that are NOT removed and that are broken (try clicking one).

Without having looked deeper into it, I suspect that either (i) the regexp doesn't properly work for the majority of our cases, or (ii) it doesn't work for markdown links that are declared at the bottom of the page and referred to via this syntax: `[I am a link text][I-am-a-ref-to-be-declared-somewhere-else]`, or (iii) both of the above.

Could you look into this @DimitriPapadopoulos?


EDIT: For example, look at page 167 in the pdf build: `dataset-level metadata is included in Derived dataset and pipeline description.` The link on that text should have been removed --- but it's there and broken.

DimitriPapadopoulos:
Yes, that's why I had asked for a careful review. I suspect that the ***`[]`*** classes might need to be changed into actual **`()`** groups or something like that. I will have a look at #596.

sappelhoff:
> Yes, that's why I had asked for a careful review. I suspect that the [] classes might need to be changed into actual () groups or something like that. I will have a look at #596.

I know you are currently busy with two fresh PRs @DimitriPapadopoulos - but do you have news on this one? Or an expected timeline?

DimitriPapadopoulos:
Not really. Perhaps a few weeks? I believe it's not an easy one, but maybe I haven't taken the time to find the right angle yet.

sappelhoff:
I also think it might be hard. So no rush, any work here is highly appreciated!

PR #856, opened 2021-08-18
[INFRA] eval() → getattr()

### LINKED ISSUE #854: eval() → getattr()
I suggest changing the call to [eval()](https://docs.python.org/3/library/functions.html#eval) into a call to [gettatr()](https://docs.python.org/3/library/functions.html#getattr) in [line 457 of `pdf_build_src/process_markdowns.py`](https://github.com/bids-standard/bids-specification/blob/28ad463/pdf_build_src/process_markdowns.py#L456-L457):
```python
                # Run the function to get the output
                new = eval(function_string)
```

See https://github.com/github/codeql/issues/6504#issuecomment-901099742:
> Normally I would recommend rewriting `eval("foo." + attrname)` into `getattr(foo, attrname)`, but I don't understand the code well enough to see whether that is possible here.

It seems possible to me 😀

DimitriPapadopoulos:
Fixes #854.


pdf_build_src/process_markdowns.py
```diff
@@ -447,14 +447,11 @@ def process_macros(duplicated_src_dir_path):
...
-                # Replace prefix with module name
-                function_string = function_string.replace(
-                    "MACROS___",
-                    "macros."
-                )
-                # Run the function to get the output
-                new = eval(function_string)
+                function_name = m.strip("{} ")
+                # Remove prefix
+                function_name = function_name[len("MACROS___"):]
```
yarikoptic: I would prefer to avoid changing the logic of the program here, and do 

```suggestion
                function_name = re.sub("^MACROS___", "", function_name)
```
so whatever assumptions brought us to this point, would not truncate function_name "blindly"

DimitriPapadopoulos: Done, actually re-used `replace("MACROS__", "")` to be as close as possible to the initial logic.

pdf_build_src/process_markdowns.py
```diff
@@ -447,14 +447,11 @@ def process_macros(duplicated_src_dir_path):
...
-                    "MACROS___",
-                    "macros."
-                )
-                # Run the function to get the output
-                new = eval(function_string)
+                function_name = m.strip("{} ")
+                # Remove prefix
+                function_name = function_name.replace("MACROS___", "")
+                # Find function_name in imported module macros
+                new = getattr(macros, function_name)
```
yarikoptic: Didn't analyze but CI seems to report

```
Traceback (most recent call last):
  File "/root/project/pdf_build_src/process_markdowns.py", line 470, in <module>
    process_macros(duplicated_src_dir_path)
  File "/root/project/pdf_build_src/process_markdowns.py", line 454, in process_macros
    new = getattr(macros, function_name)
AttributeError: module 'schemacode.macros' has no attribute 'make_metadata_table(
   {
        "LongName": "OPTIONAL",
        "Description": (
            "RECOMMENDED",
            "The description of the column.",
        ),
        "Levels": "RECOMMENDED",
        "Units": "RECOMMENDED",
        "TermURL": "RECOMMENDED",
        "HED": "OPTIONAL",
   }
)'

Exited with code exit status 1
```

DimitriPapadopoulos: Was the error there in my first attempt? Will look into it again.

DimitriPapadopoulos: I have totally misunderstood what `eval`() is doing here. Will close this PR.

PR #861, opened 2021-08-23
[INFRA] Enforce consistent line endings via .gitattributes

### LINKED ISSUE #860: [INFRA] Add a .gitattributes file to enforce line endings
As seen in #859.

Suggested solution documented in this comment: https://github.com/bids-standard/bids-specification/pull/859#issuecomment-903549146

DimitriPapadopoulos:
As documented in:
* [gitattributes](https://git-scm.com/docs/gitattributes)
* [Configuring Git to handle line endings](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings).

Fixes #860.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto
+*.png -text
+*.jpg -text


Remi-Gau (approved):

.gitattributes
```diff
@@ -0,0 +1,3 @@
+*     text=auto
```
sappelhoff, line 1:
```suggestion
* text=auto
```

DimitriPapadopoulos: I agree. That's what the [gitattributes](https://git-scm.com/docs/gitattributes) documentation suggests, but it looks ugly. Fixed now.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto
+*.png -text
+*.jpg -text
```
sappelhoff, line 3: The GitHub guide has the following for these lines:

```

# Denote all files that are truly binary and should not be modified.
*.png binary
*.jpg binary
```

Is `-text` equivalent? I think I find the GitHub way (together with the comment) more straight forward.

DimitriPapadopoulos: I guess "not text" is the same as "binary". I just copied over from the source - the gitattributes documentation.

DimitriPapadopoulos: Actually `binary` is equivalent to `-text -diff`. Since this patch is about line endings, I'd rather stick with the documentation.

sappelhoff (approved):

sappelhoff:
Thanks @DimitriPapadopoulos!

PR #883, opened 2021-09-18
[SCHEMA] Consolidate schema files by term type

### LINKED ISSUE #603: Separate schema item definitions from layout rules

### LINKED ISSUE #877: Consolidate schema files into a single file for each term type
Multiple people have raised concerns with having the schema definitions separated into one file per term. I found that structure easier to work with as I started adding terms to the schema, but that doesn't mean we have to stick with it in the long run. One proposal would be to consolidate all terms for each term type (`metadata`, `suffix`, `datatype`) into their own file.

I'd like to know how people feel about this.

Here are some potential blockers that came to mind:

- We currently use the filenames of metadata YAML files to support different definitions of the same fields (see [`EchoTime.yaml`](https://github.com/bids-standard/bids-specification/blob/811e17e2fa173b78456d4b3c6edd126344265633/src/schema/metadata/EchoTime.yaml) and [`EchoTime_fmap.yaml`](https://github.com/bids-standard/bids-specification/blob/811e17e2fa173b78456d4b3c6edd126344265633/src/schema/metadata/EchoTime_fmap.yaml) for an example). If all fields are in the same file, then we'll need some other way to support this behavior.
- Inherited fields are stored in files starting with an underscore (for example, [`_EEGCoordSys.yaml`](https://github.com/bids-standard/bids-specification/blob/811e17e2fa173b78456d4b3c6edd126344265633/src/schema/metadata/_EEGCoordSys.yaml)). It's probably not a problem when placing them in a single file, but I want to make sure everything still works that way.
- `datatypes` files have a _very_ different structure than the `metadata`, `suffixes`, or `columns` (see #827) files.
    - The closest I have to a solution for this is in #834, where I move most of the `datatypes` information to a new "rules" folder, and create new `datatypes` _object_ files with just the names and definitions. We need to separate object definitions from rules anyway.

Some potential benefits:

- Fewer files to manage. Though it's the same amount of information of course.
- The schema loading code could be simplified if we don't need to walk through folders to load all files.

Tagging some folks who have expressed opinions on this or who I think might have an opinion: @erdalkaraca, @effigies, @yarikoptic, @CPernet, @Remi-Gau, and @Tokazama. Apologies to anyone I might have missed.

tsalo:
Closes #877 and closes #603.

Changes proposed:

- Combine the schema files in each term type (for example, datatypes, metadata fields, and suffixes) into single files.
- Separate term definitions from term rules.
- Update the schema code to work with the new structure.
- Document the new structure in the schema folder's README.


src/schema/rules/datatypes/beh.yaml
```diff
@@ -1,5 +1,5 @@
 ---
```
tsalo, line 1: These `datatypes` files are essentially unchanged. I didn't come up with a clean way to reorganize them.

src/schema/rules/entities.yaml
```diff
@@ -0,0 +1,27 @@
+---
```
tsalo, line 1: This is a new "rules" file for entities.

tsalo:
Given the ~400 files that are removed or just moved, I've added comments to the most relevant ones.

src/schema/README.md
```diff
@@ -9,57 +9,122 @@ Any changes to the specification should be mirrored in the schema.
...
-
-    -   `array`: If `type` is `array`, then there MUST be an `items` field at
-        the same level as `type`. `items` describes the data type and rules that
-        apply to the individual items in the array. The same rules that apply to
-        describing data types for the field itself apply to `items`.
-        Additionally, there may be any of the following fields at the same level
-        as `type`: `minItems`, `maxItems`.
-        Here is an example of a field that MUST have three `integer` items:
-        ```yaml
+The schema is divided into two parts: the object definitions and the rules.
```
tsalo, line 12: Is this documentation clear enough?

src/schema/rules/datatypes/beh.yaml
tsalo: I did end up changing the indents for lists because apparently `yamllint`'s default behavior doesn't match `libyaml`'s defaults. I don't think there's even an easy way to write out lists with that extra indent with `pyyaml`, so this should make it easier to write out schema files programmatically in the future.

tsalo:
Tagging @bids-standard/schema-users. Is this going to cause any major problems for any tools that use the schema?

nellh:
> Tagging @bids-standard/schema-users. Is this going to cause any major problems for any tools that use the schema?

Looks like just path updates needed for bids-validator but otherwise it looks compatible with the use case so far.

Tokazama:
I haven't finalized any schema related code yet at `https://github.com/JuliaNeuroscience/NeuroStudies.jl`, but this certainly makes it easier to manage and update with changes here.

src/schema/objects/datatypes.yaml
```diff
@@ -0,0 +1,37 @@
...
+  name: Behavioral Data
+  description: |
+    Behavioral data.
+dwi:
+  name: Diffusion-Weighted Imaging
+  description: |
+    Diffusion-weighted imaging (DWI).
+eeg:
+  name:
+  description:
```
effigies, line 18:
```suggestion
  name: Electroencephalography
  description: Electroencephalography
```

src/schema/objects/datatypes.yaml
```diff
@@ -0,0 +1,37 @@
...
+fmap:
+  name: Field-Maps
+  description:
```
effigies, line 21:
```suggestion
  name: Field maps
  description: |
    MRI scans for estimating B0 inhomogeneity-induced distortions.
```

src/schema/objects/datatypes.yaml
```diff
@@ -0,0 +1,37 @@
...
+func:
+  name: Task-Based Magnetic Resonance Imaging
+  description: |
+    Task (including resting state) imaging data
+ieeg:
+  name:
+  description:
```
effigies, line 28:
```suggestion
  name: Intracranial electroencephalography
  description: |
    Intracranial electroencephalography (iEEG) or electrocorticography (ECoG) data
```

src/schema/objects/datatypes.yaml
```diff
@@ -0,0 +1,37 @@
...
+meg:
+  name:
+  description:
```
effigies, line 31:
```suggestion
  name: Magnetoencephalography
  description: Magnetoencephalography
```

src/schema/objects/datatypes.yaml
```diff
@@ -0,0 +1,37 @@
...
+perf:
+  name:
+  description:
```
effigies, line 34:
```suggestion
  name: Perfusion imaging
  description: |
    Blood perfusion imaging data, including arterial spin labeling (ASL)
```

tools/schemacode/utils.py
```diff
@@ -132,3 +133,55 @@ def flatten_multiindexed_columns(df):
     df.index.name = "Entity"
     df = df.drop(columns=["DataType"])
     return df
+
+
+def _get_link(string):
```
effigies, line 138: In `utils.py`, these make sense as public functions:

```suggestion
def get_link(string):
```

tools/schemacode/utils.py
```diff
@@ -132,3 +133,55 @@ def flatten_multiindexed_columns(df):
...
+
+        elif "type" in definition.get("additionalProperties", {}):
+            # Values within objects
+            string += " of " + _get_link(
+                definition["additionalProperties"]["type"] + "s"
+            )
+
+    elif "anyOf" in definition:
+        # Use dictionary to get unique substrings while preserving insertion order
+        substrings = {_resolve_metadata_type(subdict): None
```
effigies, line 176:
```suggestion
def resolve_metadata_type(definition):
    """Generate string of metadata type from dictionary."""
    if "type" in definition.keys():
        string = get_link(definition["type"])

        if definition.get("enum") == ["n/a"]:
            # Special string case of n/a
            string = '`"n/a"`'

        elif "type" in definition.get("items", {}):
            # Items within arrays
            string += " of " + get_link(definition["items"]["type"] + "s")

        elif "type" in definition.get("additionalProperties", {}):
            # Values within objects
            string += " of " + get_link(
                definition["additionalProperties"]["type"] + "s"
            )

    elif "anyOf" in definition:
        # Use dictionary to get unique substrings while preserving insertion order
        substrings = {resolve_metadata_type(subdict): None
```

tools/schemacode/schema.py
```diff
@@ -527,7 +503,7 @@ def make_metadata_table(schema, field_info, tablefmt="github"):
             "[DEPRECATED](/02-common-principles.html#definitions)",
         )
 
-        type_string = _resolve_metadata_type(metadata_schema[field])
+        type_string = utils._resolve_metadata_type(metadata_schema[field])
```
effigies, line 506:
```suggestion
        type_string = utils.resolve_metadata_type(metadata_schema[field])
```

effigies (approved):
LGTM. Small suggestions.

src/schema/objects/datatypes.yaml
tsalo: I totally forgot I left these empty. Thanks!

src/schema/objects/entities.yaml
```diff
@@ -81,23 +36,25 @@ ceagent:
...
   description: |
-    The `trc-<label>` key/value can be used to distinguish
-    sequences using different tracers.
-    The key `TracerName` MUST also be included in the associated JSON file,
-    although the label may be different.
+    Density of non-parametric surfaces.
+    MUST have a corresponding `Density` metadata field to provide
+    interpretation.
+
+    This entity is only applicable to derivative data.
```
Remi-Gau, line 47: Side note: I suspect you have thought about this, but we might need at some point to create a rule to keep track of derivatives only entities.

src/schema/README.md
Remi-Gau: I think so. Definitely helped me at least.

Remi-Gau (approved):
It all makes sense to me. This should definitely help a lot. Thanks @tsalo

src/schema/README.md
```diff
@@ -9,57 +9,122 @@ Any changes to the specification should be mirrored in the schema.
...
+The `entity` field is the entity as it appears in filenames. For example, the `entity` for `ceagent` is `ce`.
+
+The `description` field is a freeform description of the entity.
+The description should apply to _all_ uses of the entity,
+so if there is some additional information about the entity that applies specifically to a subset of suffixes or datatypes,
+then that information should be provided in the specification, rather than the schema.
+
+The `type` field defines the representation type for the value associated with the entity.
+All entities are represented in filenames as key-value pairs, and in all cases the value should be a string,
+so the `type` field should always be `string`.
```
tsalo, line 103:
```suggestion
The `type` field defines the representation type for the value associated with the entity.
Given that all entities appear in filenames, they should all be strings and the `type` field should always be `string`.
This is true of both alphanumeric entities (such as `acq-desc`) and numeric ones (such as `run-1`).
```

src/schema/objects/entities.yaml
tsalo: You're right. We can always add whatever fields we want, but it's probably something that can be formalized with the changes proposed by @erdalkaraca in #884.

src/schema/README.md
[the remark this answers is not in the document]
tsalo: I rewrote this portion in 70e2c4e. I hope it's clearer now.

Remi-Gau:
> Now that we have two approvals and I think I've addressed the review comments, I believe this is ready to merge. Do we want to wait five days before merging?

Technically this does not affect the content itself so I am not sure if the 5 days rule apply.

https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules

I know that there have been precedents for infra related PR where we merged sooner than that. @sappelhoff @effigies

sappelhoff:
I haven't had a chance to take a look yet, but yeah - if no content is impacted and most major schema-stakeholders agree, then I think merging is fine

tsalo:
Well, I pinged the schema users team 11 days ago and no one raised any concerns, so I guess I'll merge once CI finishes. Thanks!

Issue #876, opened 2021-09-13
Clarify use of n/a on required fields


nicholst:
In a BIDS validator issue https://github.com/bids-standard/bids-validator/issues/1326, the value of `n/a` threw a "should be a number" error for a REQUIRED field, which made me think carefully about about the handling of `n/a`s for mandatory fields.  

The entirety of general guidance on `n/a` in the spec appears to be in `02-common-principles.md` 
```
Missing and non-applicable values MUST be coded as `n/a`.
```
and in `03-modality-agnostic-files.md`:
```
Throughout BIDS you can indicate missing values with `n/a` (for "not  available"). 
```
(Should "can" here should be replaced with "MAY"?)

I would like input on whether a small addition would be useful in these generic settings to clarify use of `n/a` in required fields.  It seems like we should clarify on or more of the following:

* Entering a value of `n/a` for a required field does not create a valid BIDS document.
* It is RECOMMENDED to _omit_ a required field than to specify it as `n/a`. (The validator can make clearer recommendations if a field is entirely missing, rather than having a out-of-scope value).
* If data for a required field is unknown, it is RECOMMENDED to _specify_ it as `n/a`. (This produces more consistent documents that can be easier to parse???)

*Or*, is it the case that some required fields may anticipate and expect a value of `n/a` as being meaningful, and we can't make blanket assertions like these?  (In which case, no changes is needed).

Also, I'd be interested if those working on scheme (e.g. @tsalo) have considered n/a as a special value... e.g. in the validator problem I encountered, `n/a` was specified for a required numeric field, and so was caught that way.  If `n/a` was given for a free text required field, would the schemea complained?

tsalo:
I personally like your first two proposals. I agree that a validator can more easily make sense of missing fields than invalid values, and using "n/a" in lieu of required information won't make that document interpretable.

>Or, is it the case that some required fields may anticipate and expect a value of n/a as being meaningful, and we can't make blanket assertions like these? (In which case, no changes is needed).

After browsing through the metadata fields in the schema where "n/a" is a valid value, the only one that seems to treat it as being meaningful, as far as I can tell, is `SubjectArtefactDescription`, where "n/a" is interpreted as meaning that no artifact was found. Essentially, in that one field it means more "not applicable" than "not available". Fortunately, that field is recommended, rather than required.

[...]

PR #885, opened 2021-09-24
[SCHEMA] Add patterns for format validation

### LINKED ISSUE #832: Define valid value formats in schema
I'm thinking it might also be a good idea to create a new `formats.yaml` that defines the different formats used in the schema. Something like:

```yaml
label:
    description: |
        Freeform labels without special characters.
    pattern: ^[0-9a-zA-Z]+$
index:
    description: |
        Non-negative, non-zero integers, optionally prefixed with leading zeros for sortability.
    pattern: ^[0-9]+$
unit:
    description: |
        A unit.
    pattern: ^[0-9a-zA-Z]+$
```

Unfortunately, I don't actually know how most of those patterns would look. Units, for example, can be freeform text if necessary, but optimally they should be valid combinations of the values in the Units appendix. Plus we have relative paths of multiple flavors, along with URIs. I think `uri`, at least, is offloaded to JSON Schema in the validator.

_Originally posted by @tsalo in https://github.com/bids-standard/bids-specification/issues/792#issuecomment-831462364_

EDIT: Labeling the different types of paths, and thus defining patterns for them, will have to wait until after #820 is resolved, at least.

tsalo:
Closes #832.

This currently doesn't include any code that would _use_ the patterns, except for tests.

Changes proposed:
- Add `formats.yaml` file to schema, with regular expressions with which to validate different entity values, metadata types, and metadata string formats. Individual metadata fields should still be allowed to supply a specific "pattern" keyword.

To do:
- [x] write tests (post-#892)
- [x] add rrid pattern (for #827 and the existing `SoftwareRRID`) 
- [ ] review the URI and path patterns


src/schema/formats.yaml
```diff
@@ -0,0 +1,66 @@
...
+index:
+  description: |
+    Non-negative, non-zero integers, optionally prefixed with leading zeros for sortability.
+  pattern: ^[0-9]+$
+# Metadata types
+string:
+  description: |
+    A basic string type (not a specific format).
+    This should allow any free-form string *except* "n/a".
+  pattern: ^(?!(n/a)$).*$
```
tsalo, line 17: The "n/a" exception is included for #876.

src/schema/formats.yaml
```diff
@@ -0,0 +1,66 @@
...
+    The validation for this format is minimal.
+  pattern: ^[0-9a-zA-Z/_-\.]+$
+uri:
+  description: |
+    A uniform resource indicator.
+  pattern: ^[a-zA-Z]+:[0-9a-zA-Z/_-\.]+$
+bids_uri:
+  description: |
+    A BIDS uniform resource indicator.
+  pattern: ^bids:[0-9a-zA-Z/_-\.]+$
```
tsalo, line 66: The patterns for all four of the path/URI formats should be revised. I don't know enough about the rules to define solid patterns just yet.

src/schema/formats.yaml
```diff
@@ -0,0 +1,66 @@
...
+number:
+  description: |
+    A number.
+  pattern: ^[+-]?([0-9]+([.][0-9]*)?|[.][0-9]+)$
+boolean:
+  description: |
+    A boolean.
+  pattern: ^(true|false)$
+# String patterns
+date:
```
tsalo, line 31: `date` is included in the schema as a "pattern" rather than a "format". That was just an oversight on my part. I would fix it here, but that would cause a conflict with #883, so I will address it in that PR.

src/schema/formats.yaml
tsalo: The same goes for `stimuli_relative` (added after the initial comment).

src/schema/formats.yaml
tsalo: Fixed in e9c4f5f (part of #883).

Remi-Gau (commented):
This will for sure be superuseful. I don't speak regex fluently so not sure I am the best to review this though.

rwblair (approved):
Structure makes sense to me. Eyeballing the regex it seems ok, but the best way to verify it is for it to be used by the validator in tests.

tsalo:
Thanks for taking a look. I'm thinking that I should try writing some unit tests for the schema code, which will necessitate reorganizing it into a proper package. I will open a PR to do that after #883 is merged, so that there aren't too many merge conflicts.

src/schema/formats.yaml
```diff
@@ -0,0 +1,88 @@
...
+  description: |
+    Non-negative, non-zero integers, optionally prefixed with leading zeros for sortability.
+    An index may not be all zeros.
+  pattern: ^[0-9]*[1-9]+[0-9]*$
+# Metadata types
+string:
+  description: |
+    The basic string type (not a specific format).
+    This should allow any free-form string *except* "n/a".
+  pattern: ^(?!(n/a)$).*$
```
sappelhoff, line 18: This is related to #876 

> This should allow any free-form string *except* "n/a".

I am not sure whether disallowing `n/a` as a string for free-form text fields would break some datasets. For example for the required `EEGReference` field, users could currently pass `n/a` according to bids-validator:

https://github.com/bids-standard/bids-validator/blob/75c54e4f7a17bf58c7cfc47c9698eb4e0841315b/bids-validator/validators/json/schemas/eeg.json#L25

Whether or not that's a good idea should be discussed in #876, this is just a note that what you introduce here has not been discussed and has not been formally part of the spec :thinking:

yarikoptic:
I guess the next step for this pr is adding some tests as @tsalo [was planning todo](https://github.com/bids-standard/bids-specification/pull/885#issuecomment-934535138) (and #833 is merged) or that would be postponed till later?

tsalo:
I'd prefer to add the tests before merging. When they have time, Anthony and Chris are planning to take a look at #892. Once that's merged, I can add tests for the formats to this PR.

src/schema/objects/formats.yaml
```diff
@@ -0,0 +1,99 @@
+---
+# This file defines valid patterns for different formats
+# Entity patterns
+label:
+  description: |
+    Freeform labels without special characters.
+  pattern: ^[0-9a-zA-Z]+$
```
yarikoptic, line 7: what is the motivation for explicit anchoring?
We with @TheChymera we are thinking to establish regex based validation of BIDS filenames for DANDI project... for that, we likely just to compose a list of regexes for a valid BIDS filename.  Then, `<label>` would be replaced with regex from here.  If it is anchored -- it would require us to strip those first.  I think anchors should be explicitly added in the code whenever processing already "extracted" label.

tsalo: I believe I kept the anchors from the validator expressions. I can definitely remove them here, and then the validator would just add them back I think.

tsalo: Are quotes a problem? Some of the regular expressions seem to cause problems with YAML if I don't wrap them in quotes.

src/schema/objects/formats.yaml
```diff
@@ -0,0 +1,115 @@
...
+integer:
+  name: Integer
+  description: |
+    An integer which may be positive or negative.
+  pattern: '[+-]?\d+'
+number:
+  name: Number
+  description: |
+    A number which may be an integer or float, positive or negative.
+  pattern: '[+-]?([0-9]+([.][0-9]*)?|[.][0-9]+)'
```
effigies, line 31: I would expect scientific notation (e.g., `1e3`, `-2.1E+5`) to be permissible.

```suggestion
  pattern: '[+-]?([0-9]+([.][0-9]*)?|[.][0-9]+)([eE][+-]?[0-9]+)?'
```

src/schema/objects/formats.yaml
```diff
@@ -0,0 +1,115 @@
...
+  description: |
+    A datetime in the form `"YYYY-MM-DDThh:mm:ss[.000000][Z]"`,
+    where [.000000] is an optional subsecond resolution between 1 and 6 decimal points,
+    and [Z] is an optional, valid timezone code.
+  pattern: '[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}?(\.[0-9]{1,6})?([A-Z]{3,4})'
+time:
+  name: Time
+  description: |
+    A time in the form `"hh:mm:ss"`.
+  pattern: '(?:2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9]'
```
effigies, line 56: Should we permit `1:00:00`? Or just `01:00:00`?

```suggestion
  pattern: '(?:2[0-3]|[01]?[0-9]):[0-5][0-9]:[0-5][0-9]'
```

effigies (commented):
Couple quick thoughts.

src/schema/objects/formats.yaml
tsalo: ohhhh good catch!

src/schema/objects/formats.yaml
tsalo: I think that's reasonable. Thanks!

src/schema/objects/formats.yaml
```diff
@@ -0,0 +1,115 @@
...
+  name: Research resource identifier
+  description: |
+    A [research resource identifier](https://scicrunch.org/resources).
+  pattern: 'RRID:.+_.+'
+uri:
+  name: Uniform resource indicator
+  description: |
+    A uniform resource indicator.
+  pattern: '(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?'
+bids_uri:
```
yarikoptic, line 107: might be worth to sort the entries to ease locating any particular one

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+            "05/01/2022",  # MM/DD/YYYY or DD/MM/YYYY
+        ],
+        "datetime": [
+            "2022-01-05T13:16:30.1000005",  # too many decimal points
+            "2022-01-05T13:16:30U",  # time zone too short
+            "2022-01-05T13:16:30UTCUTC",  # time zone too long
+            "2022-01-05T34:10:10",  # invalid time
+        ],
+        "time": [
+            "34:10:10",  # invalid time
```
yarikoptic, line 104: I would have added some extra boundary conditions
```suggestion
            "34:10:10",  # invalid time
            "24:00:00",  # should be 00:00:00
            "00:60:00",  # should be 01:00:00
            "00:00:60",  # should be 00:01:00
            "01:23",     # lacks either hours or seconds
```

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+            "stimuli/path/file.txt",
+        ],
+        "dataset_relative": [
+            "/path/with/starting/slash/file.txt",
+        ],
+        "participant_relative": [
+            "/path/with/starting/slash/file.txt",
+            "sub-01/path/file.txt",
+        ],
+        "rrid": [],
```
yarikoptic, line 118:
```suggestion
        "rrid": [
            "RRID:",  # empty one
        ],
```

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+        "uri": [],
```
yarikoptic, line 119:
```suggestion
        "uri": [
            "ftp://"  # lacks anything but protocol
        ],
```
didn't check if that would be caught by regex, but clearly such URI would be quite useless

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+        "bids_uri": [],
+    }
+    for pattern, test_list in BAD_PATTERNS.items():
+        pattern_format = schema_obj["objects"]["formats"][pattern]["pattern"]
+        search_pattern = "^" + pattern_format + "$"
```
yarikoptic, line 124: I guess I just love f-strings too much?

```suggestion
        search_pattern = f"^{pattern_format}$"
```

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+        search = re.compile(search_pattern)
+        for test_string in test_list:
+            assert bool(
+                search.fullmatch(test_string)
+            ), f"'{test_string}' is not a valid match for the pattern '{search.pattern}'"
+
+    # Check that invalid strings do not match the search pattern.
+    BAD_PATTERNS = {
+        "label": ["test_01", "!", "010101-"],
```
yarikoptic, line 87:
```suggestion
        "label": ["test_01", "!", "010101-", "01-01", "-01"],
```
just to have one with `-` in the middle or at the beginning

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,101 @@ def test_object_definitions(schema_obj):
...
+        "index": ["test", "0.1", "0-1", "0_1"],
+        "string": [],
+        "integer": ["3.14", "-3.14", "1.", "-1.", "string", "s1", "1%"],
```
yarikoptic, line 90:
```suggestion
        "integer": ["3.14", "-3.14", "1.", "-1.", "string", "s1", "1%", "one"],
```
AI is not supported!

yarikoptic (commented):
I just love testing boundary conditions I guess ;)

src/schema/objects/formats.yaml
tsalo: Good idea. I'll sort them after merging in your recommendations.

tools/schemacode/schemacode/tests/test_schema.py
```diff
@@ -27,3 +27,109 @@ def test_object_definitions(schema_obj):
...
+            "2022-01-05T13:16:30.000005",  # up to 6 decimal points
+            "2022-01-05T13:16:30UTC",  # timezones are allowed
+            "2022-01-05T13:16:30.05UTC",
+        ],
+        "time": [
+            "13:16:30",
+            "09:00:00",
+            "9:00:00",  # leading zeros are not required for hours
+        ],
+        "unit": ["any string is valid."],
```
yarikoptic, line 64: oh, even with spaces and unicodes?  I thought we had discussion and there was a decision on some unicodes to be ok (but not any), then with #411 CMIXF-12 ascii recommended. https://github.com/bids-standard/bids-specification/blob/HEAD/src/99-appendices/05-units.md mentions `µ` (could be only as prefix) `Ω` (only as suffix) and `°` (all by itself?).  So may be regex and test here could be made more stringent?  well, aforementioned page gives example just as "NOT RECOMMENDED" instead of "WRONG": `volts per microsecond` so may be my whining is all in vein?  But then I would restrict to composition of alphanumerics, space, `-` and those 3 unicodes, and thus disallow everything else.

tsalo:
It looks like `ftp://` passes validation, unfortunately. Not sure how to update the regex appropriately.

tools/schemacode/schemacode/tests/test_schema.py
tsalo: TBH I allowed any string because I don't know much about the unit validation. Do you mind if we allow any string in this PR and make sure to follow up with a more accurate regex before the next release? Assuming we merge this after 1.7.0 is released, so there'll be a bit of time between 1.7.0 and 1.8.0.

yarikoptic: I would not mind. Please just create an issue for that so we do not forget?

yarikoptic:
> It looks like `ftp://` passes validation for the `uri` format, unfortunately. Not sure how to update the regex appropriately.

I can try later to create one following the "Syntax" of https://en.wikipedia.org/wiki/Uniform_Resource_Identifier which would demand to have a non-degenerate `authority` after //

tools/schemacode/schemacode/tests/test_schema.py
tsalo: Done. #999.

sappelhoff (approved):

tsalo:
Now that I have two approvals, I'm going to merge. I think we can keep improving these patterns (i.e., #999) before the next release.

PR #892, opened 2021-10-07
[SCHEMA] Reorganize schema code into a package

tsalo:
This stems from https://github.com/bids-standard/bids-specification/pull/885#issuecomment-934535138. The idea is to make the schema code operate as a proper Python package, so we can also start writing tests for it.

We still may want to move schema-related code into a different repository at some point (e.g., `bids-schema`, `pybids`), but for now I'd just like to have the code organized nicely within this repo.

Changes proposed:

- Reorganize `schemacode` into a Python package, so that it can be installed and tests can be written for it.
    - This involves symlinking the schema files in `src/schema` to `tools/schemacode/schemacode/data/schema`. Thanks to @effigies for that.
- Write a few initial tests for `schemacode`. These can be improved in future PRs.
- Add a GitHub Action to run continuous integration for the new package and upload a coverage report to CodeCov.
- Add new schema-related folders to the CODEOWNERS file with myself listed.


effigies:
@tsalo Made several updates. RTD is fixed. The PDF build looks like it needs updating to new modules. I assume you'll be able to figure that out faster than me, but let me know if you want me to tackle it.

tsalo:
Thanks @effigies! How does it work? It looks like the schema files are duplicated in `tools/schemacode/schemacode/data/schema/`. Are those files updated automatically when the files in `src/schema/` are updated?

effigies:
It's just a symlink, so they will be copied into the package. If you're going to develop, it probably makes sense to `pip install -e` so you don't freeze them into a package and keep using the symlink. But from a CI build perspective, it's all based on the current commit, so fine to freeze.

tsalo:
Awesome! I fixed the module paths, so I think everything is good to go.

tsalo:
I'd like to have a CI job that runs the package's tests (at least with Python 3.7, if not other supported versions). Is there any preference for CircleCI vs. GitHub Action?

effigies:
GHA.

tsalo:
Okay, I think it's looking good. I've added a few tests and a GitHub Action to run said tests. The tests are passing and the coverage report is uploading to CodeCov.

In a perfect world, I think we'd only run the tests when files in `src/schema/` or `tools/schemacode/` have been edited, but I don't want to hold this up on that basis. We could always try it in a future PR.

tools/schemacode/versioneer.py
```diff
@@ -0,0 +1,1886 @@
+# Version: 0.18
```
sappelhoff, line 1: why not use version 0.21?

https://github.com/python-versioneer/python-versioneer/tags

requirements.txt
```diff
@@ -3,7 +3,4 @@ mkdocs-material>=5.4
 pymdown-extensions>=7.0.0
 mkdocs-branchcustomization-plugin~=0.1.3
 mkdocs-macros-plugin
-numpy
-pandas
-PYYaml
-tabulate
```
sappelhoff, line 9: [process_markdowns.py](https://github.com/bids-standard/bids-specification/blob/master/pdf_build_src/process_markdowns.py) depends on numpy, and depends on it being in requirements.txt --> https://github.com/bids-standard/bids-specification/blob/89115cd9ef0d4676ae78cfa1e9e78fa1048869a2/.circleci/config.yml#L66

It's also mentioned in the docs: https://github.com/bids-standard/bids-specification/blob/master/pdf_build_src/README.md

I don't know why the job still seems to pass :shrug: 

Could you either:
1. remove the numpy dependency in process_markdowns.py (preferred, seems possible at first glance, maybe even easy)
2. update the documentation (pdf build README) and ensure that numpy is installed whenever someone wants to build the pdf?
3. any other solution you might have,

sappelhoff (commented):
the linkchecker issue seems to have to do with the glossary PR we recently merged. Is that an easy one to fix here as well @tsalo? Or separate PR?

tools/schemacode/versioneer.py
tsalo: Good idea. I updated it in 31bcd196b601481b73c5c6f1101d9f60cdb37fc9.

tsalo:
>the linkchecker issue seems to have to do with the glossary PR we recently merged. Is that an easy one to fix here as well @tsalo? Or separate PR?

@sappelhoff I think it's just because there's a "canonical" link on all pages that goes to the stable version based on the MkDocs config file (which says that the `site_url` is the stable version). Until we have a new release with the glossary page included, I think that linkchecker is going to fail.

requirements.txt
tsalo: It looks like the `requirements.txt` file includes `schemacode`, which, in turn, includes `numpy` as a dependency. That explains why it's working. However, I agree that `numpy` probably shouldn't have been removed from the main requirements file, since it's used for specification code outside of `schemacode`. I'll add it back in.

The other requirements that were removed from that file all seem to be `schemacode`-specific, so I'll keep them out.

tsalo: I added it back in in f766804.

sappelhoff: > It looks like the requirements.txt file includes schemacode, which, in turn, includes numpy as a dependency. That explains why it's working.

thanks!

> I'll add it back in.

:+1: 

> The other requirements that were removed from that file all seem to be schemacode-specific, so I'll keep them out.

excellent, thanks

sappelhoff:
> @sappelhoff I think it's just because there's a "canonical" link on all pages that goes to the stable version based on the MkDocs config file (which says that the site_url is the stable version). Until we have a new release with the glossary page included, I think that linkchecker is going to fail.

Mh, perhaps this could be solved by changing the `site_url` value as part of the release protocol. During release (and the version bumps) we change `site_url` to stable ... then after release (when changing the version to the dev version), we change it to "dev"

would that work?
https://github.com/bids-standard/bids-specification/blob/89115cd9ef0d4676ae78cfa1e9e78fa1048869a2/mkdocs.yml#L2

tsalo:
I think so, but I don't know if it's worth the effort. I don't think the links are a real problem outside of linkchecker. When you click links to other pages in the specification, they go to the same version you started from, rather than defaulting to `stable`. I believe that the messed up links only appear in HTML comments in the built site.

tsalo:
@bendhouseart sorry for not requesting your review earlier. I know you wanted to take a look, though we didn't end up trying out `poetry`.

sappelhoff (approved):
> I believe that the messed up links only appear in HTML comments in the built site.

I see :| but it's still annoying to have linkchecker now fail until we release ...

But that's not a blocker for this PR, which I approve of!

tools/schemacode/setup.cfg
```diff
@@ -0,0 +1,68 @@
...
+license = MIT
+classifiers =
+    Development Status :: 2 - Pre-Alpha
+    Intended Audience :: Science/Research
+    Topic :: Scientific/Engineering :: Information Analysis
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
yarikoptic, line 18: what about
```suggestion
    Programming Language :: Python :: 3.9
    Programming Language :: Python :: 3.10
```
?
versioneer 0.21 is compatible according to changes done since 0.21 (just addition of 3.10 marker)

.github/workflows/schemacode_ci.yml
```diff
@@ -0,0 +1,64 @@
...
+
+  latest:
+    needs: check_skip
+    if: ${{ needs.check_skip.outputs.skip == 'false' }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+        fail-fast: false
+        matrix:
+            os: ["ubuntu-latest"]
+            python-version: ["3.8"]
```
yarikoptic, line 34: what about
```suggestion
            python-version: ["3.8", "3.10"]
```
to test against "target" and the "most recent"?  or even better
```suggestion
            python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
```
against all supported? (if adding 3.10)

.github/workflows/schemacode_ci.yml
```diff
@@ -0,0 +1,64 @@
+name: "schemacode_ci"
+
+on:
+  push:
+    branches:
```
yarikoptic, line 5:
```suggestion
    paths:
      - 'tools/schemacode/**'
      - 'src/schema/**'
    branches:
```
and similar one for `pull_request` to not bother testing the package (if takes long, didn't check) if no package related changes?

yarikoptic (commented):
just some minor comments/suggestions - ok to ignore

.github/workflows/schemacode_ci.yml
tsalo: Ohhh... I didn't know that was an option! The tests are currently pretty fast, but I was worried about spamming PRs with a bunch of unnecessary checks. This should deal with that nicely.

sappelhoff:
>>    I believe that the messed up links only appear in HTML comments in the built site.

> I see :| but it's still annoying to have linkchecker now fail until we release ...

We could add a small Python script that checks whether the current commit is tagged (=stable), or not (=dev), and then in CI **before** building with mkdocs, that Python scripts changes `stable` to `latest` in the following line:

https://github.com/bids-standard/bids-specification/blob/98ca3fa8795a2ad29910cc54327c3cedb8f84906/mkdocs.yml#L2

:thinking: or would we screw up search engine results with that? I am not very familiar with "canonical URL" metadata in HTML headers.

tsalo:
>We could add a small Python script that checks whether the current commit is tagged (=stable), or not (=dev), and then in CI before building with mkdocs, that Python scripts changes stable to latest

I have no clue if that's possible, but having the ability to update content files with Python scripts _within_ CI would be really useful for adding tooltips (#954). The problem I have been bumping into for that is that the macros-generated content is not available for the auto-linked tooltips to grab onto. Generating the glossary within CI would circumvent that issue, I believe.

That said, I don't even know where to start to do it.

bendhouseart (approved):
Looked over the changed files and managed to build and install locally w/ setup.py 👍. I'd call it good to go. 

```python
Python 3.9.9 (main, Nov 21 2021, 03:22:47)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import schemacode
>>>
```

bendhouseart:
CircleCI disagrees, but eh.

tools/schemacode/schemacode/render.py
```diff
@@ -539,7 +379,7 @@ def make_entity_table(schema, tablefmt="github", **kwargs):
     def _remove_numeric_suffixes(string):
         import re
 
-        suffix_str = re.findall("\((.+)\)", string)
+        suffix_str = re.findall(r"\\((.+)\\)", string)
```
effigies, line 382: I think you want `r"\"` or `"\\"`, but not both.

```suggestion
        suffix_str = re.findall(r"\((.+)\)", string)
```

Unless you're specifically trying to match two `\` characters?

effigies (approved):
Looks good. A couple bits of code caught my eye as likely able to be simplified. I haven't tested though...

tsalo:
The built site looks good to me, as does the PDF. Now that we have two approvals, is this alright to merge or do we want to wait 5 days?

effigies:
The 5 days is more for semantic changes than infrastructure.

Issue #86, opened 2018-11-10
SimultaneousRecording and SimultaneousRecordingWith


CPernet:
# 2 fields

add to dataset_description.json the field _SimultaneousRecording_ to indicates the different imaging modalities acquired at the time e.g. SimultaneousRecording: 'func','EEG','eye tracker', 'physio', 'behav'.  

add _SimultaneousRecordingWith_ field within the sidecar files to indicates the different imaging modalities acquired at the time and on different hardware.  

e.g. ds000117 SimultaneousRecording: {'MEG','EEG'} but no SimultaneousRecordingWith because meg and eeg data are in the same file and recorded togehter on the same hardware.  

# SimultaneousRecordingWith
Tibor solution is to point to all files related to each others: for instance in the run-02_bold.json we would have SimultaneousRecordingWith = { 'eeg/sub-01_ses-01_task-something_run-02_eeg.vhdr', 'eyetracker/sub-01_ses-01_task-something_run-02_eyetracker.asc'}.  

Chris G proposed to add the relative root path (e.g. {'/sub-01/func/sub-01_task-something_run-02_bold.nii.gz']) to accommodate hyperscanning (simultaneous recordings across participants).  

Robert pointed out that 'It might also be relevant to know whether behaviour (behav) was recorded during the functional brain recordings (e.g. func+behav), or prior to (or after) the functional scan. It is often assumed (e.g. in the main bids spec) that these are recorded simultaneously, but e.g. section 8.7 and 8.8 already have to deal with the simultaneous versus sequential (or separate) measurement of the two. When considering SimultaneousRecording as a general field, it might have the side effect of “behav" becoming a mature data type on its own, rather than an addition to functional brain data. This would make it more symmetric'.    

Mainak has an open issue: should we split concurent recordings: eg split the meg and eeg data of ds000117 ; most felt it is not necessary nor recommended since it comes from the same hardware - the isue to solve is ensure metadata cover both MEG and EEG witjout conflict (seems largely possible to me)  

# synchronization issue
Since we have multiple modalities, each one should have their own timing information recorded using events.tsv files. This is important since different frequency sampling exists for each modality and often clock on different hardware run a different speed. If there is no event (rest) we need at least one marker to ensure this is synchronized (a typical case in point is starting the eeg recording before the MRI starts).  

Chris G pointed out that to compare events.tsv they should probably have the same number of rows since there is no unique identifier for events.

CPernet:
[...]

PR #894, opened 2021-10-12
[SCHEMA] add physio to schema or EEG, MEG, iEEG, PET, ASL, DWI

### LINKED ISSUE #890: [SCHEMA] physio and stim not included for EEG, MEG, iEEG and PET
`physio` and `stim` suffixes are currently included for [func](https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/func.yaml) and [beh](https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/beh.yaml).

But they are not included for the following data types:

- [x] EEG: https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/eeg.yaml
- [x] iEEG: https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/ieeg.yaml
- [x] MEG: https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/meg.yaml
- [x] PET: https://github.com/bids-standard/bids-specification/blob/master/src/schema/datatypes/pet.yaml
- [x] ASL
- [x] DWI

Implement validation for validator:
- [ ] EEG
- [ ] iEEG
- [ ] MEG
- [ ] PET
- [ ] ASL
- [ ] DWI
- [ ] mri anat

I can't think of a reason for why that's the case, should we add those to the schema ?

Remi-Gau:
Fixes #890 

This updates the schema so that all modalities support the physio and stim suffixes.

This might need an update counterpart on the validator side.


Remi-Gau:
pinging our schema master @tsalo & BEP lead @mnoergaard @melanieganz @patsycle @HenkMutsaerts

Remi-Gau:
I think we can definitely think of different ways the schema could be "refactored" (and I usually am not a big fan of duplications) but I am not sure we are there yet in terms . Though we could easily get there if we apply to this `anat`.

But whatever is decided should be done in a separate PR IMHO

Remi-Gau:
> @Remi-Gau It looks like `proc` and `split` come after `recording` in those rules, and that seems to be reflected in the filename templates, so it must be an issue with how the entities themselves are organized. On the bright side, it doesn't look like `proc`, `split`, and `space` can appear with `recording` in any valid filenames at the moment, so we could reorder those entities if we want. Since `space` is a mixed raw/derivative entities (raw for MEG/EEG/iEEG and derivative for other datatypes), I think we still want that at the end, but I believe we could reorder the entities to be `proc` --> `split`--> `recording` --> `space` without being backwards-incompatible. WDYT?

yes that was exactly my thought too.

> It does seem strange that `proc` is allowed for physio data. Shouldn't it denote multiple versions of the same data processed using different devices or algorithms? Other entities that indicate multiple versions from the same run are dropped for physio data (`echo`, for example). Then again, fMRI physio allows `rec`, which seems strange in the same way.
 
> Also, should `split` be included? Since the MEG data are separated based on file size, are physio files going to be split up in the same way?

Tempted to say that if we have `rec` for `func` data then we should have it too for `proc` for `meg`.

For `split`: I am not as sure. 

Pinging several of the people discussing these types simultaneous recording in issues #86 

@CPernet @dorahermes @robertoostenveld

CPernet:
yes physio is assumed to be .. but does not have to ..

tsalo:
I think it really comes down to whether we can consider the entity "within-acquisition" or "between-acquisition". Entities denoting data from the same recording, like `echo` and `part`, should not be included in the physio filenames. The same rule applies to events files. I think there must have been a decision point in the past where the maintainers had to decide if events and physio files would be duplicated across echoes or not, but instead they chose to leverage the inheritance principle. I lean toward that, but the big issues for me are (1) `rec`'s behavior doesn't seem consistent with that idea, and (2) I don't know how `split` should be treated. 

Based on the following line from the **Scans file** section of the spec, I'd assume that `split` should be treated like `echo`:

>Each neural recording _file_ SHOULD be described by exactly one row. Some recordings consist of multiple parts, that span several files, for example through `echo-`, `part-`, or `split-` entities. Such recordings MUST be documented with one row per file.

sappelhoff (approved):
Conceptually this PR LGTM, didn't dive too deep into the discussion on simultaneous recordings. Once the merge conflict is resolved, I think this PR is mergeable.

Remi-Gau:
I am still unsure about what how to deal with this `split` case for physio file with MEG.

People having to split their acquired physio data in 2 to match the content of the MEG feels like an extra curation step and potential source of error.

Also as mentioned by @tsalo the `events.tsv` are supposed to follow a `<match>` pattern but don't have the `proc` and `split` entity.

This is template for MEG that would come out of merging this PR as is.

```
sub-<label>/
    [ses-<label>/]
        meg/
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>]_meg.<extension>
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>]_meg.json
            sub-<label>[_ses-<label>]_acq-<calibration>_meg.dat
            sub-<label>[_ses-<label>]_acq-<crosstalk>_meg.fif
            sub-<label>[_ses-<label>][_task-<label>][_acq-<label>][_space-<label>]_markers.mrk
            sub-<label>[_ses-<label>][_task-<label>][_acq-<label>][_space-<label>]_markers.sqd
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>]_events.json
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>]_events.tsv
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>][_recording-<label>]_physio.json
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>][_recording-<label>]_physio.tsv.gz
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>][_recording-<label>]_stim.json
            sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_run-<index>][_proc-<label>][_split-<index>][_recording-<label>]_stim.tsv.gz
```

sappelhoff:
Right - I agree with @tsalo opinion from https://github.com/bids-standard/bids-specification/pull/894#issuecomment-947701229

> I think it really comes down to whether we can consider the entity "within-acquisition" or "between-acquisition". Entities denoting data from the same recording, like echo and part, should not be included in the physio filenames. The same rule applies to events files. I think there must have been a decision point in the past where the maintainers had to decide if events and physio files would be duplicated across echoes or not, but instead they chose to leverage the inheritance principle.

He points out two issues:

> (1) rec's behavior doesn't seem consistent with that idea, and (2) I don't know how split should be treated.

re: 1) --> no idea

re: 2) --> Just looked into this ... are we being inconsistent in the spec? :grimacing: 

# split --> rows in scans.tsv :zap: inconsistent? 

The (MEG-centric) documentation of the [split entity](https://bids-specification.readthedocs.io/en/latest/99-appendices/09-entities.html#split) and [MEG split files](https://bids-specification.readthedocs.io/en/latest/99-appendices/06-meg-file-formats.html#split-files) both say:

> If there are multiple parts of a recording and the optional scans.tsv is provided, remember to list all files separately in scans.tsv and that the entries for the acq_time column in scans.tsv MUST all be identical, as described in Scans file.

Yet the [scans section](https://bids-specification.readthedocs.io/en/latest/03-modality-agnostic-files.html#scans-file) says:

> Each neural recording file SHOULD be described by exactly one row. Some recordings consist of multiple parts, that span several files, for example through echo-, part-, or split- entities. Such recordings MUST be documented with one row per file.

:face_with_head_bandage: 

MNE-BIDS writes one row per "split" (instead of one row per neural recording, even if there are multiple "splits"), see: https://github.com/mne-tools/mne-bids/blob/c85dafaaac0ff79127b9513e130c566bc2313e0e/mne_bids/write.py#L458-L474 or try it yourself (click on details):

<details>

```python
# %%
# do first:
# pip install mne
# pip install mne-bids
import os

import mne
import mne_bids
import numpy as np

# %%
# Create raw
nchannels = 100
ntimes = int(1e7)
data = np.random.randn(nchannels, ntimes)

info = mne.create_info(ch_names=[f"{i}" for i in range(nchannels)],
                       sfreq=1000, ch_types="mag")

raw = mne.io.RawArray(data, info)

# %%
# Save raw to FIFF
raw_fname = os.path.join(os.path.expanduser("~"), "Desktop", "test_raw.fif")
raw.save(raw_fname)
raw = mne.io.read_raw_fif(raw_fname)


# %%
# Save raw to BIDS
root = os.path.join(os.path.expanduser("~"), "Desktop", "split_test_bids")
os.makedirs(root, exist_ok=True)
bids_path = mne_bids.BIDSPath(subject="1", session="1", task="rest", root=root)
mne_bids.write_raw_bids(raw, bids_path)

# %%

```

</details>

sappelhoff:
or am I misunderstanding the above :point_up: This part from scans.tsv

> Such recordings MUST be documented with one row per file

I am confused now whether "a recording" is the combination of all splits ... OR each split file.

tsalo:
>I am confused now whether "a recording" is the combination of all splits ... OR each split file

@sappelhoff the recording is the combination of all splits, but you must log each _file_ separately within the scans.tsv file.

Remi-Gau:
suggestion for this PR: 

we get all the non MEG related things in with version 1.7 and we have an open PR left to only deal with the MEG physio aspect of this.

sappelhoff:
Coming back to this ...

> we get all the non MEG related things in with version 1.7 and we have an open PR left to only deal with the MEG physio aspect of this.

generally I agree.

BUT: is the only "MEG related thing" that is open the question of what to do with the `split` entity for physio and stim files?

If yes, the I think @tsalo's comment from https://github.com/bids-standard/bids-specification/pull/894#issuecomment-947701229 is valid (now that also my misunderstanding above was cleared up, thanks btw):

> Based on the following line from the Scans file section of the spec, I'd assume that split should be treated like echo:

Remi-Gau:
> is the only "MEG related thing" that is open the question of what to do with the `split` entity for physio and stim files?
> 
> If yes, the I think @tsalo's comment from [#894 (comment)](https://github.com/bids-standard/bids-specification/pull/894#issuecomment-947701229) is valid (now that also my misunderstanding above was cleared up, thanks btw):
> 
> > Based on the following line from the Scans file section of the spec, I'd assume that split should be treated like echo:

I think this is the only MEG specific issue.

Happy to make `split` for MEG behave the way `split` does for MRI.

sappelhoff (approved):

tsalo (approved):
LGTM!

src/schema/rules/datatypes/dwi.yaml
Remi-Gau, line 1: task not supported

PR #899, opened 2021-10-14
[SCHEMA] Add MeasurementToolMetadata and Derivative metadata fields

tsalo:
Closes None, but adds fields that we previously missed.

Changes proposed:

- Add two new metadata fields that are described in the phenotypic data section of the spec.
- Add a metadata table for the MeasurementToolMetadata field.
- Add Derivative to the existing tabular data metadata table.


effigies (approved):

tsalo:
Given that this is a fix, rather than an enhancement, I'm going to merge now instead of waiting five days.

Issue #905, opened 2021-10-20
Adding sourcedata filename to a column in the scans.tsv file


adam2392:
### Problem
I've been working with conversion of `sourcedata` files over to BIDS for the sake of i) speeding up my analysis work streams and ii) speeding up sharing of datasets. However, many times you'll have new datasets coming in, or maybe you want to determine if the file you uploaded with some filename (e.g. `subject001_eeg_001.edf`) was converted or not.

Moreover, many of my collaborators (i.e. clinicians) only remember their original file naming scheme, not the organized BIDS files. Unfortunately, then there's a lot of back and forth about which file is which unless there is a backwards trace of which BIDS file corresponds to which source file. There is no easy way to check this right now. 

### Suggestion
My proposal is to add a SHOULD requirement in the `scans.tsv` that suggests that users add a column to the file for `original_filename`, which adds the filename of the source file. This way, one can backtrack what was converted easily. To be honest, I think it should be a MUST, unless there is some sort of PHI embedded in the source filename?

adam2392:
You mean https://bids-specification.readthedocs.io/en/stable/05-derivatives/02-common-data-types.html here?

I suppose that serves the same purpose as adding `original_filename` in scans.tsv. However, where would these go? Would they go in the sidecar JSON? 

I'm okay w/ either option as long as it's specified in BIDS, then we can support it in `mne-bids`.

effigies:
> You mean https://bids-specification.readthedocs.io/en/stable/05-derivatives/02-common-data-types.html here?

Yes.

> I suppose that serves the same purpose as adding `original_filename` in scans.tsv. However, where would these go? Would they go in the sidecar JSON?

Yes.

---

Another approach to this could just be a table in `sourcedata/` or `code/` with `source`/`destination` columns. It could serve as a log or an input to a tool that performs the conversions.

adam2392:
Should this just go in the sidecar json part for each MEG, EEG and iEEG?

sappelhoff:
closed, because we discussed in #906 that implementing this on the tooling side of things would suffice ... see https://github.com/mne-tools/mne-bids/pull/890

PR #906, opened 2021-10-21
[ENH] Add Sources link to all raw datatypes sidecar JSON

adam2392:
Fixes: #905 

Add `Sources` to the sidecar json that point to the original source file.

### DIFF
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -449,6 +449,8 @@ For example, the [EDF](https://www.edfplus.info/)
 data format can only contain recording dates after 1985.
 Shifting dates is RECOMMENDED, but not required.
 
+The source file path of each recording is also OPTIONAL and can be stored under the ``sources`` column. Similar to [derivatives](./05-derivatives/02-common-data-types.md), it is OPTIONAL to include ``sources``, specifying the filename(s) of the source file used to generate this dataset. If the filename(s) contains patient identifiable information, then it should not be stored.
+
 Additional fields can include external behavioral measures relevant to the
 scan.
 For example vigilance questionnaire score administered after a resting
@@ -460,11 +462,11 @@ All such included additional fields SHOULD be documented in an accompanying
 Example `_scans.tsv`:
 
 ```Text
-filename	acq_time
-func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
-func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
-meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27
-meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27
+filename	acq_time  sources
+func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30 sourcedata/sub01/dicoms
+func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33 sourcedata/sub01/motordicoms
+meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27 sourcedata/sub01/restdicoms
+meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27 sourcedata/sub01/restdicoms2
 ```
 
 ## Sessions file
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -492,6 +492,10 @@ combined image rather than an image from each coil.
 
 ### Other RECOMMENDED metadata
 
+#### Source Filename
+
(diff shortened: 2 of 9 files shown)


effigies:
* I would just use `Sources`. My inclination, with #820, is to deprecate `RawSources`, and have a single `Sources` list that can contain any URI.
* I think this has to be an OPTIONAL, as there's no expectation that there's a linkable reference for raw data. SHOULD is supposed to come with validator warnings, and this would be very noisy.
* And agree that this should apply to all data files. I would be inclined to put it into common principles as a generally applicable piece of metadata, though this is something we don't currently have, so that may be a different conversation...

sappelhoff:
> Thoughts?

technically, nothing is preventing you to already add a `sources` column to `scans.tsv` and to document it in `scans.json`. If you want to make this a standardized "optional" column, I think that'd be fine. It'd be similar to how we recommend a "handedness" column in participants.tsv, and suggest some values to put there ([source](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#participants-file)), only that a `sources` column would be optional.

Not sure whether that solution would be preferable to extending the derivative `Sources` to raw BIDS data. We could also do both. To me, both changes feel like they would not destroy anything and/or make anything harder downstream, so personally I am open to whatever you think solves your problem best.

adam2392:
I think adding to `scans.tsv` is a good addition too. I'll modify it to include both possibilities. I think downstream in MNE-BIDS for example, we would just write it in `scans.tsv` and check there for existing files when writing.

adam2392:
@effigies, @sappelhoff just pinging to see if there's any other issues here for me to address, so I don't forget it :p. Thanks!

adam2392:
Not sure why link checker is breaking.

sappelhoff:
> Not sure why link checker is breaking.

see #910 - nothing you need to fix here :-)

> just pinging to see if there's any other issues here for me to address, so I don't forget it :p. Thanks!

will take a look soon!

src/03-modality-agnostic-files.md
```diff
@@ -452,6 +452,8 @@ For example, the [EDF](https://www.edfplus.info/)
 data format can only contain recording dates after 1985.
 Shifting dates is RECOMMENDED, but not required.
 
+The source file path of each recording is also OPTIONAL and can be stored under the ``sources`` column. Similar to [derivatives](./05-derivatives/02-common-data-types.md), it is OPTIONAL to include ``sources``, specifying the filename(s) of the source file used to generate this dataset. If the filename(s) contains patient identifiable information, then it should not be stored.
```
sappelhoff, line 455: double backticks render code in RST, but for MD, it's single backticks :wink: 

Also: the files under the `filename` column are relative to the `scans.tsv` file (this is implicit unfortunately, but relatively clear from the context that this is meant by the text on line 435). --> What should the files under the `sources` column be relative to? ... should we wait for #918 and make this BIDS-URIs? ... or should it be relative to the dataset root - as you seem to do in your current example?

sappelhoff (commented):
Having read about the `Sources` field in JSON files again (see [Sources](https://bids-specification.readthedocs.io/en/latest/05-derivatives/02-common-data-types.html)), I think that a simple `sources` column in `scans.tsv` may be a better solution, see my review comment.

I think the `Sources` field would be a bit unwieldy for the goal you have ... it'd mean having to create a JSON sidecar for each file you want to describe a source of :thinking: That'd quickly become messy when you have lots of task and run entities.

thus I suggest to remove all "Sources" JSON references. But @adam2392 please hold off any more changes, I plan to discuss this with the other maintainers tomorrow evening, so my opinion might change :-) (don't want you to do more work than necessary)

adam2392:
> Having read about the `Sources` field in JSON files again (see [Sources](https://bids-specification.readthedocs.io/en/latest/05-derivatives/02-common-data-types.html)), I think that a simple `sources` column in `scans.tsv` may be a better solution, see my review comment.

Sounds good. I agree here.

> thus I suggest to remove all "Sources" JSON references. But @adam2392 please hold off any more changes, I plan to discuss this with the other maintainers tomorrow evening, so my opinion might change :-) (don't want you to do more work than necessary)

Okay, any update here? Thanks!

sappelhoff:
> Okay, any update here? Thanks!

After some discussion, the consensus was that this problem at this point is too little a user-issue to warrant making some `sources` column a "first-class-citizen" in the BIDS spec, when instead, users can just add a `sources` column themselves and document it in their JSON file.

In the original issue, you make these two arguments:

> ... many times you'll have new datasets coming in, or maybe you want to determine if the file you uploaded with some filename (e.g. subject001_eeg_001.edf) was converted or not.

This is something you can easily fix and track in your conversion scripts, I don't think this needs to be done in BIDS.

> Moreover, many of my collaborators (i.e. clinicians) only remember their original file naming scheme, not the organized BIDS files. Unfortunately, then there's a lot of back and forth about which file is which unless there is a backwards trace of which BIDS file corresponds to which source file. There is no easy way to check this right now.

You are the first to bring up this issue (thanks for that) - but it's arguably a relatively small issue (and apparently not widespread), and I pointed out the solution in my first few sentences above (using arbitrary columns).

---

@adam2392 I think we could add support for this in mne-bids without having to have this explicitly in the spec.

Overall, I suggest closing the issue and this PR and to solve your problems from the tooling side.

adam2392:
Sounds good to me if we can PR mne-bids.

PR #918, opened 2021-10-29
[ENH] Add BIDS URIs and deprecate relative paths, RawSources and (possibly unused) BasedOn

### LINKED ISSUE #471: Specification of relative or absolute path in IntendedFor and SpatialReference  is not consistent

### LINKED ISSUE #718: `IntendedFor` fields needs to be harmonized
Thanks a lot for your feedback @effigies. I investigated the `IntendedFor` field a bit, see below:

I get the feeling that the different `IntendedFor` fields should be harmonized.

I think this should be its own issue - for the sake of this PR, I will see how much "easy improvement" I can salvage without touching/impacting `IntendedFor`.

in particular, we should clarify that IntendedFor contains a filepath (str) or list of filepaths ... and how these filepaths are to be made up ... if absolute or relative ... to what reference if relative, etc. see #471 

Also I think all of those should accept a str *or* array of str (currently some only accept str).

Lastly, the "descriptions" should be homogenized ... and then the special aspect of using "IntendedFor" with a particular usecase can be added at the end of the description ... or something similar. Just to get rid of this currently confusing diversity in descriptions.

# linking fieldmap data (Recommended, str or array of str)

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#using-intendedfor-metadata

> Contains one or more filenames with paths relative to the participant subfolder. The path needs to use forward slashes instead of backward slashes. This field is OPTIONAL, and in case the fieldmaps do not correspond to any particular scans, it does not have to be filled.

# link ASL timeseries and m0 scan (Required, str or array or str)

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#m0scan-metadata-fields

> One or more filenames with paths relative to the subject subfolder, with forward slashes, referring to ASL time series for which the `*_m0scan.nii[.gz]` is intended.

# MEG: link MRI(s) (Optional, str or array of str)

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/02-magnetoencephalography.html#coordinate-system-json-_coordsystemjson

> Path or list of path relative to the subject subfolder pointing to the structural MRI, possibly of different types if a list is specified, to be used with the MEG recording. The path(s) need(s) to use forward slashes instead of backward slashes (for example, ses-/anat/sub-01_T1w.nii.gz).

# EEG: Link MRI/CT (Optional, str)

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/03-electroencephalography.html#coordinate-system-json-_coordsystemjson

> Relative path to associate the electrodes, landmarks and fiducials to an MRI/CT.

# iEEG: Link MRI7CT/Photograph (Recommended, str)

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/04-intracranial-electroencephalography.html#coordinate-system-json-_coordsystemjson

> This can be an MRI/CT or a file containing the operative photo, x-ray or drawing with path relative to the project folder. If only a surface reconstruction is available, this should point to the surface reconstruction file. Note that this file should have the same coordinate system specified in iEEGCoordinateSystem. For example, T1: sub-<label>/ses-<label>/anat/sub-01_T1w.nii.gz Surface: /derivatives/surfaces/sub-<label>/ses-<label>/anat/sub-01_T1w_pial.R.surf.gii Operative photo: /sub-<label>/ses-<label>/ieeg/sub-0001_ses-01_acq-photo1_photo.jpg Talairach: /derivatives/surfaces/sub-Talairach/ses-01/anat/sub-Talairach_T1w_pial.R.surf.gii

_Originally posted by @sappelhoff in https://github.com/bids-standard/bids-specification/issues/700#issuecomment-768960234_

### LINKED ISSUE #790: Specification of AssociatedEmptyRoom
Hello,

MEG experimental data often comes with an associated "empty-room" recording, which is supposed to capture the signal recorded with no participant present.

BIDS allows to indicate **which** empty-room recording belongs to an experimental recording via the `AssociatedEmptyRoom` key in [`*_meg.json`](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#sidecar-json-_megjson). Specifically, the specs state the following:

> Relative path in BIDS folder structure to empty-room file associated with the subject’s MEG recording. The path needs to use forward slashes instead of backward slashes (for example, "sub-emptyroom/ses-/meg/sub-emptyroom_ses-_task-noise_run-_meg.ds").

Two questions came to my mind:

- The path is supposed to be "relative", but it's unclear to me what the reference point is supposed to be. Intuitively, I assumed it's either the location of `*_meg.json`, or of the experimental recording. However, the path in the example seems to be relative to the **BIDS root**. What is the correct way to specify `AssociatedEmptyRoom`?
- The path in the example has empty `ses` and `run` entities. I didn't know this was even valid. Is that a mistake?

For reference, this is the schema currently used for validation of `AssociatedEmptyRoom`. It's … under-complex and therefore didn't provide me with any additional clues:
https://github.com/bids-standard/bids-validator/blob/2f0ca2458cff47e1fd1f898c3a81aef370558085/bids-validator/validators/json/schemas/meg.json#L48-L59

Thanks for you help,

Richard

effigies:
This PR replaces #820, which I'm leaving open for comparison purposes.

I have moved BIDS URI as a sub-section of URI, and drastically reduced the amount of text, including justification and examples. I have also reduced the scope to purely include files in BIDS datasets, as standard URIs are sufficient for other resources.

I am happy to consider re-adding the justification and some examples to an appendix, but I think keeping Common Principles lean is a good idea.

One significant semantic change: Reading up on `file:///` URLs, there is no scope for these to be relative. Relative paths are simply that, and are resolved based on context. I have thus permitted `DatasetLinks` to include `derivatives/xyz` as a valid location.

closes #790
closes #471
closes #718
#757 (or at least, pave the way for it to be closed)


src/02-common-principles.md
```diff
@@ -674,7 +674,85 @@ specificity and persistence.
...
+
+```plain
+bids:<dataset-name>:<absolute-path>
+```
+
+The scheme component `bids` identifies a BIDS URI,
+which defines a `path` component of the form `<dataset-name>:<absolute-path>`.
+The `dataset-name` component is an identifier for a dataset,
+and the `absolute-path` component is a path, starting with a slash (`/`),
+followed by the location of a resource within the dataset.
```
satra, line 691: it seems absolute path simply means path relative to the root of the dataset. also to make it clear relative to unix systems, you could consider dropping the leading `/` so that there is no confusion with absolute locations on a file system.

effigies: I kept this language, since that had been in @sappelhoff's version, but I'm okay moving away from that. Also okay with dropping the `/`.

Will wait a day or two to see if there are objections before moving forward with these changes.

sappelhoff: No objections against dropping the leading `/` from my side.

src/02-common-principles.md
```diff
@@ -674,7 +674,85 @@ specificity and persistence.
...
+
+Examples:
+
+```plain
+bids::/sub-01/fmap/sub-01_dir-AP_epi.nii.gz
```
sappelhoff, line 696:
```suggestion
bids:/sub-01/fmap/sub-01_dir-AP_epi.nii.gz
```

sappelhoff (commented):
thanks for narrowing the scope of this and writing it up so clearly. I am gonna do a double check of all fields that would need to be adjusted - but perhaps you already caught them all.

Some comments meanwhile.

src/02-common-principles.md
```diff
@@ -674,7 +674,85 @@ specificity and persistence.
...
+root.
+
+BIDS URIs cannot be interpreted outside a BIDS dataset,
+as they require a `dataset_description.json` file to resolve.
+
+#### Examples
+
+Consider this example `dataset_description.json`:
+
+```YAML
```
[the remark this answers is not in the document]
effigies: Yes. JSON will highlight syntax errors, YAML is a more forgiving superset, so I tend to use it in Markdown when I don't want to render an entire JSON document.

src/02-common-principles.md
```diff
@@ -674,7 +674,85 @@ specificity and persistence.
...
+
+To reference files in BIDS datasets, the following URI scheme may be used:
+
+```plain
+bids:<dataset-name>:<absolute-path>
+```
+
+The scheme component `bids` identifies a BIDS URI,
+which defines a `path` component of the form `<dataset-name>:<absolute-path>`.
+The `dataset-name` component is an identifier for a dataset,
```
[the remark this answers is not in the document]
effigies: I believe I have done this in all cases where there could be any ambiguity.

sappelhoff (commented):
I think the `BasedOn` field is the only one that remains to be adjusted (right now).

Taken from my list in #820 ... Some of the items I had on there were invalid, one of them is now deprecated, and the rest should be in this PR - and only BasedOn is missing as far as I can see.

- `IntendedFor`
- `SpatialReference` (invalid)
- `AssociatedEmptyRoom`
- `B0FieldIdentifier` (invalid)
- `B0FieldSource` (invalid)
- `BasedOn`
- `RawSources` (deprecated)
- `Sources`

effigies:
Should `BasedOn` be `Sources`? It's only specified in an appendix as a qMRI-specific derivative metadata field, and seems nearly identical to `RawSources` (which will be merged into `Sources`).

cc @agahkarakuzu @Gilles86 for input.

agahkarakuzu:
Thanks @effigies ! `BasedOn` may be a field introduced early in BEP001, then not adapted to following conventions. As you noted, `RawSources` (or `Sources` after the merge)  is identical. I am OK for the modifications this requires, i.e. changing `BasedOn` --> `Sources` in the qMRI appendix. Example datasets will need to be updated as well.

effigies:
Thanks, Agah! @sappelhoff I guess technically we should probably deprecate `BasedOn` like we will `RawSources`?

sappelhoff:
> Thanks, Agah! @sappelhoff I guess technically we should probably deprecate BasedOn like we will RawSources? 

Sounds good! Too bad we didn't catch this earlier - but better late than never.

nicholst:
I've reviewed this at a high level, and have no comments except to say that issues that I caught in #820 have been addressed here.

effigies:
@sappelhoff I guess for something this big, we should probably patch the validator and update some examples to ensure that somebody following these rules won't be unable to validate.

sappelhoff:
> @sappelhoff I guess for something this big, we should probably patch the validator and update some examples to ensure that somebody following these rules won't be unable to validate.

agreed :+1: we should have a PR ready for the validator to merge once this is merged/released ... and updating the examples is a good idea as well.

VisLab:
I have gone over the changes and support this pull request.

src/02-common-principles.md
```diff
@@ -674,7 +674,89 @@ specificity and persistence.
...
+### BIDS URI
+
+To reference files in BIDS datasets, the following URI scheme may be used:
+
+```plain
+bids:[<dataset-name>]:<relative-path>
+```
+
+The scheme component `bids` identifies a BIDS URI,
+which defines a `path` component of the form `<dataset-name>:<absolute-path>`.
```
sappelhoff, line 688:
```suggestion
which defines a `path` component of the form `<dataset-name>:<relative-path>`.
```

src/02-common-principles.md
```diff
@@ -674,7 +674,89 @@ specificity and persistence.
...
+Consider this example `dataset_description.json`:
+
+```YAML
+{
+    ...
+    "DatasetLinks": {
+        "deriv1": "derivatives/derivative1",
+        "phantoms": "file:///data/phantoms",
+        "ds000001": "doi:10.18112/openneuro.ds000001.v1.0.0"
+    }
```
sappelhoff, line 734: What would an example look like for referencing a raw file from derivatives dataset nested in a raw BIDS dataset?
(will often be needed instead of the now deprecated `RawSources`)

Like this?

```yaml
{
    ...
    "DatasetLinks": {
        "raw": "../../"
    }
}
```

see e.g., https://github.com/bids-standard/bids-examples/issues/303 the `qmri_mp2rage` dataset, and turning `RawSources` there into `Sources` with BIDS-URIs

sappelhoff:
We have decided to take our time with this one and not squeeze it into 1.7, so that it does not block microscopy. A release is long overdue, and we can do another release soon after 1.7, where this PR can then get in.

src/02-common-principles.md
sappelhoff: @effigies any opinion?

src/05-derivatives/02-common-data-types.md
```diff
@@ -51,33 +51,36 @@ A guide for using macros can be found at
...
 {
-    "RawSources": ["sub-01/func/sub-01_task-rest_bold.nii.gz"]
+    "Sources": ["bids:raw:sub-01/func/sub-01_task-rest_bold.nii.gz"]
 }
 ```
 
-If this file was generated with prior knowledge from additional sources, such as
-the same subject's `T1w`, then both files MAY be included in `RawSources`.
+Note that `"raw"` must appear in the `DatasetLinks` metadata in
+`dataset_description.json`.
```
sappelhoff, line 59:
```suggestion
`dataset_description.json`.
For example in the case that the given derivatives dataset is nested within the 
"derivatives" directory of a raw dataset, the entry in `DatasetLinks` may say:
`"raw": "../../"`.
```

src/02-common-principles.md
sappelhoff: see suggestion https://github.com/bids-standard/bids-specification/pull/918/files#r927333853

src/05-derivatives/02-common-data-types.md
```diff
@@ -51,33 +51,39 @@ A guide for using macros can be found at
...
+For example, in the case that the given derivatives dataset is nested within the
```
effigies, line 60:
```suggestion
For example, in the case that the given derivatives dataset is nested within the
```

src/02-common-principles.md
```diff
@@ -808,7 +808,89 @@ specificity and persistence.
...
+bids:[<dataset-name>]:<relative-path>
+```
+
+The scheme component `bids` identifies a BIDS URI,
+which defines a `path` component of the form `<dataset-name>:<relative-path>`.
+The `dataset-name` component is an identifier for a BIDS dataset,
+and the `relative-path` component is the location of a resource within the
+BIDS dataset, relative to the root of that dataset.
+
+Examples:
```
sappelhoff, line 827:
```suggestion
and the `relative-path` component is the location of a resource within that
BIDS dataset, relative to the root of that dataset.
The `relative-path` MUST NOT start with a forward-slash character (`/`).

Examples:
```

sappelhoff (approved):
apart from my code suggestion on leading slashes (based on https://github.com/bids-standard/bids-examples/pull/304#discussion_r927711558), I think this can finally go in

Issue #912, opened 2021-10-27
Render valid value restrictions in tables based on object definitions in schema


tsalo:
It would be great if tables in the specification automatically included restrictions on valid values, without having to hardcode those restrictions in the object "description" fields.

This stems from https://github.com/bids-standard/bids-specification/pull/827#discussion_r723356782.

As a simple example, we can look at the `CASLType` metadata field:

https://github.com/bids-standard/bids-specification/blob/251f4400ee44ea5555d418dc12c852b5c3ef7aa0/src/schema/objects/metadata.yaml#L299-L307

Note that there are currently two valid values for this field: "single-coil" and "double-coil". These values are written out in two places: (1) the "enum" that actually restricts valid values for the field and (2) the "description", which redundantly includes the values. Optimally, we wouldn't need to include a list of valid values in the description, since that information is already available, in a more useful form, elsewhere in the object definition. One step in this direction is to update our table-rendering functions to attempt to convert restrictions on valid values to a human-readable form and append that new string to the "description" cell in the table automatically.

NOTE: Preferably, any enhancements to the code to implement this would also work with #904.

Issue #658, opened 2020-11-01
Define supported coordinate systems in the BIDS specification instead of FieldTrip wiki


sappelhoff:
The FieldTrip wiki is a great resource full of information that's hard to get by otherwise.

We are currently linking to this wiki for the coordinate systems that are "accepted" in M(i)EEG.

See: https://bids-specification.readthedocs.io/en/latest/99-appendices/08-coordinate-systems.html

Screenshot for example:
![image](https://user-images.githubusercontent.com/9084751/97799306-bcb4be80-1c2d-11eb-97cd-a664bac2fa7b.png)

That's linking to this [page in the wiki](https://www.fieldtriptoolbox.org/faq/how_are_the_different_head_and_mri_coordinate_systems_defined/) where coordinate systems are defined.

I want to gather some opinions: Should we rather define the BIDS accepted coordinate systems within the BIDS specification?

Benefits:
- have everything in one place
- the information are released, and thus *stable* instead of captured in a wiki

Drawbacks:
- information is maintained in two places
- takes some initial work to carry over the information from wiki to spec

cc @bids-standard/raw-electrophys

hoechenberger:
> I want to gather some opinions: Should we rather define the BIDS accepted coordinate systems within the BIDS specification?
> 
> Benefits:
> 
> * have everything in one place (the spec)
> * the information are released with the spec, and thus _stable_ instead of captured in a wiki

+1 for this. I'm hoping that maybe that in the long the situating will invert: BIDS becomes the authority, and other places like the FT Wiki will link to the BIDS spec. :)

VisLab:
+1 Agree that it should be in BIDS.

I suspect that there will need to be a great deal more work to be done in
this area on coordinate systems and frames of reference as BEP's such as
motion capture and eye tracking are incorporated. A good starting point is
to have the Head coordinate systems nailed down in the spec and the
terminology clarified.

adam2392:
> > I want to gather some opinions: Should we rather define the BIDS accepted coordinate systems within the BIDS specification?
> > Benefits:
> > 
> > * have everything in one place (the spec)
> > * the information are released with the spec, and thus _stable_ instead of captured in a wiki
> 
> +1 for this. I'm hoping that maybe that in the long run, the situation will invert: BIDS becomes the authority, and other places like the FT Wiki will link to the BIDS spec. :)

Would also greatly prefer it to all be in the spec and explicitly clarified.

teonbrooks:
[...]

PR #919, opened 2021-11-01
[SCHEMA] Add full object definitions for valid values in schema

### LINKED ISSUE #904: Support full object definitions for valid values in schema
Currently, valid values (i.e., `enum`s) for objects in the schema are written as lists. However, in some cases, we also want to have each value defined, and perhaps even summarized in a separate table. As an example, take the [channels file for MEG data](https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/02-magnetoencephalography.html#channels-description-_channelstsv). Each of the valid values for the `type` column has a description, which is currently rendered  with a hardcoded table, but which could be incorporated into the schema directly.

tsalo:
- Closes #904

This PR adds new information to the schema, but should not affect how it is rendered in the specification.

will simplify finding a solution for:

- #1436 
- https://github.com/bids-standard/bids-specification/issues/1126

I'm not going to write a rendering function based on tags in this PR. I think it can go in a future one.

Changes proposed:
- Convert certain valid values (enums) in the schema to full objects. Specifically, these object definitions contain "name" and "description" fields. In most cases, they're still nested within the parent object definitions, rather than stored as distinct objects and referenced in the objects that use them.

To do:

- [x] Ensure that the schema rendering and searching functions can handle the fact that many lists of strings are now lists of objects.
- [x] Fill in the TODOs. I may need to ask for help in describing some (many?) of the values.
- [x] ~~Think about tackling #912 in this PR as well.~~ Dropped in favor of #921.


src/schema/objects/metadata.yaml
```diff
@@ -1497,6 +1524,7 @@ MTPulseShape:
     `"SINC"`, `"SINCHANN"` (sinc pulse with Hanning window),
     `"SINCGAUSS"` (sinc pulse with Gaussian window), `"FERMI"`.
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 1527: What do the different `MTPulseShape` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -1705,6 +1733,7 @@ PCASLType:
     Type the gradient pulses used in the `"control"` condition:
     `"balanced"` or `"unbalanced"`.
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 1736: What do the different `PCASLType` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2082,6 +2145,7 @@ SampleOrigin:
     Values MUST be one of `"blood"`, `"saliva"`, `"brain"`, `"csf"`,
     `"breast milk"`, `"bile"`, `"amniotic fluid"`, `"other biospecimen"`.
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2148: Should we describe the different `SampleOrigin` values?

src/schema/objects/metadata.yaml
```diff
@@ -2387,6 +2453,7 @@ SpoilingType:
     Specifies which spoiling method(s) are used by a spoiled sequence.
     Accepted values: `"RF"`, `"GRADIENT"` or `"COMBINED"`.
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2456: What do the different `SpoilingType` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2468,6 +2535,7 @@ TissueOrigin:
     Values MUST be one of `"gray matter"`, `"white matter"`, `"csf"`,
     `"meninges"`, `"macrovascular"` or `microvascular`.
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2538: Should we describe the different `TissueOrigin` values?

src/schema/objects/metadata.yaml
```diff
@@ -2550,6 +2618,7 @@ Type:
     Reserved values: `Brain` - brain mask, `Lesion` - lesion mask,
     `Face` - face mask, `ROI` - ROI mask
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2621: Should we describe the different `Type` values?

src/schema/objects/metadata.yaml
```diff
@@ -2623,13 +2692,15 @@ _CoordUnits:
   - n/a
 _EEGCoordSys:
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2695: What do the different `_EEGCoordSys` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2623,13 +2692,15 @@ _CoordUnits:
...
   enum:
   - CapTrak
   - EEGLAB
   - EEGLAB-HJ
   - Other
 _GeneticLevelEnum:
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2703: What do the different `_GeneticLevelEnum` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2647,6 +2718,7 @@ _LandmarkCoordinates:
     maxItems: 3
 _MEGCoordSys:
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2721: What do the different `_MEGCoordSys` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2656,6 +2728,7 @@ _MEGCoordSys:
   - Other
 _StandardTemplateCoordSys:
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2731: What do the different `_StandardTemplateCoordSys` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -2698,6 +2772,7 @@ _StandardTemplateDeprecatedCoordSys:
   - UNCInfant2V23
 _iEEGCoordSys:
   type: string
+  # TODO: Add definitions for these values.
```
tsalo, line 2775: What do the different `_iEEGCoordSys` values mean?

src/schema/objects/metadata.yaml
```diff
@@ -1983,9 +2036,19 @@ RecordingType:
     interest (for example, stimulus presentations or subject responses).
   type: string
   enum:
-  - continuous
-  - epoched
-  - discontinuous
+  - continuous:
```
tsalo, line 2039: I would prefer to have meaningful descriptions for the continuous and discontinuous values.

src/schema/objects/metadata.yaml
sappelhoff: These are string identifiers that point to a free-form definition of

1. the origin of a coordinate system, defined in some meaningful way (e.g., by making cardinal axes go through well defined anatomical landmarks on a subject's head)
2. where the cardinal axes point to (e.g., y axis from origin (0) through nasion (positive))
3. the units for x,y,z values in this coordinate system

src/schema/objects/metadata.yaml
sappelhoff: see https://github.com/bids-standard/bids-specification/pull/919/files#r785334647

src/schema/objects/metadata.yaml
sappelhoff: see https://github.com/bids-standard/bids-specification/pull/919/files#r785334647

src/schema/objects/metadata.yaml
sappelhoff: I guess this is related to: https://github.com/bids-standard/bids-specification/issues/658

src/schema/objects/metadata.yaml
```diff
@@ -2868,60 +3036,322 @@ _LandmarkCoordinates:
...
+  - scanner:
+      name: scanner
+      description: |
+        The intrinsic coordinate system of the original image (the first entry of `RawSources`)
+        after reconstruction and conversion to NIfTI or equivalent for the case of surfaces and dual volume/surface files.
+
+        The `scanner` coordinate system is implicit and assumed by default if the derivative filename does not
+        define **any** `space-<label>`.
+        Please note that `space-scanner` SHOULD NOT be used,
+        it is mentioned in this specification to make its existence explicit.
```
tsalo, line 3093: I'm not sure if these values should be enumerated here. I took them from https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#nonstandard-coordinate-system-identifiers and https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#non-template-coordinate-system-identifiers.

src/schema/objects/metadata.yaml
tsalo: I think I figured it out. The descriptions aren't exactly extensive, but they're something.

effigies:
Noting that this would be a backwards incompatible change (cf #1013), but easily handled by detecting a list of single-keyed objects to a list of keys.

effigies:
@tsalo Sorry, this is a nasty one to try to resolve on my own. Conflicts from both #1230 and #1235.

tsalo:
@effigies I think it's all good now.

tsalo:
Per discussion today, I will (1) resolve conflicts, (2) move the value definitions to a new YAML file, and (3) ensure that there are links in the "allowed values" sections in glossary entries.

src/schema/objects/enums.yaml
```diff
@@ -0,0 +1,1113 @@
+---
+# This file defines valid values in BIDS key-value pairs.
+_CoordUnits:
```
effigies, line 3: Are these used for anything? Kept around for future use?

tsalo: I thought they were used elsewhere, but it looks like I was wrong. I'll just remove this group and any others that aren't actually used.

src/schema/objects/enums.yaml
```diff
@@ -0,0 +1,1113 @@
...
+  value: FERMI
+  display_name: Fermi pulse
+  description: |
+    A Fermi-shaped pulse.
+i:
+  value: i
+  display_name: i
+  description: |
+    Phase/Slice encoding direction is along the first axis of the data in the NIFTI file,
+    with phase encoding going from the zero index to the maximum index.
```
effigies, line 172:
```suggestion
    The encoding direction is along the first axis of the data in the NIFTI file,
    and the encoding value increases from the zero index to the maximum index.
```

src/schema/objects/enums.yaml
```diff
@@ -0,0 +1,1113 @@
...
+  display_name: k
+  description: |
+    Phase/Slice encoding direction is along the third axis of the data in the NIFTI file,
+    with phase encoding going from the zero index to the maximum index.
+iMinus:
+  value: i-
+  display_name: i-
+  description: |
+    Phase/Slice encoding direction is along the first axis of the data in the NIFTI file,
+    with phase encoding going from the maximum index to the zero index.
```
effigies, line 190:
```suggestion
    The encoding direction is along the first axis of the data in the NIFTI file,
    and the encoding value decreases from the zero index to the maximum index.
```

effigies (commented):
The Phase/SliceEncodingDirection values are slightly weirdly worded. Would something like this work?

Alternately, we at least should use `phase/slice` in both lines.

src/schema/objects/enums.yaml
tsalo: I'm happy to just remove mention of phase and slice, as in your suggestion.

sappelhoff (commented):
I have reviewed this PR and I approve of it, thanks a lot for this work.

Just one question:

For some items you have added the comment

> # TODO: Add definitions for these values.

While other items *don't* have such a comment, yet are not converted to the "new form" either, e.g.:

```
  enum:
    - IODINE
    - GADOLINIUM
    - CARBON DIOXIDE
    - BARIUM
    - XENON
```

Shouldn't we just add information for **all** enums (except those that only have "n/a")?

Or is that kind of superfluous? (e.g., for m, cm, mm enums ... meter, centimeter ...)

tsalo:
Thanks for reviewing @sappelhoff!

>While other items don't have such a comment, yet they are not converted to the "new form" either

Those could be enums that were added after I started this PR, or they might even be ones I just missed in my original pass-through.

>Shouldn't we just add information for all enums (except those that only have "n/a")?

I think some are self-explanatory (e.g., I don't know if I want to write descriptions for all of the allowed values for the `handedness` column, since they include l, L, left, Left, LEFT, etc.), but I do think most valid values could use a description. I think we can tackle that in future PRs though, since this one adds in the infrastructure we need to make that work.

src/schema/objects/entities.yaml
```diff
@@ -132,8 +132,8 @@ hemisphere:
   type: string
   format: label
   enum:
-    - 'L'
-    - 'R'
+    - $ref: objects.enums.left_hemisphere.value
+    - $ref: objects.enums.right_hemisphere.value
```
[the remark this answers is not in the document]
tsalo: I think they end up being treated the same.

src/schema/objects/columns.yaml
```diff
@@ -603,159 +603,62 @@ trigger:
...
+    - $ref: objects.enums.PPG.value
+    - $ref: objects.enums.PUPIL.value
+    - $ref: objects.enums.REF.value
+    - $ref: objects.enums.RESP.value
+    - $ref: objects.enums.SEEG.value
+    - $ref: objects.enums.SYSCLOCK.value
+    - $ref: objects.enums.TEMP.value
+    - $ref: objects.enums.TRIG.value
+    - $ref: objects.enums.VEL.value
+    - $ref: objects.enums.VEOG.value
```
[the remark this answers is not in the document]
tsalo: Awesome!

sappelhoff (approved):
ok, great. I am going to merge this today and then do follow up PRs:

1. rendering the channel type tables in the modality sections from the schema
1. creating a new appendix page rendering ALL channel types (and putting links to this appendix below the modality ch type tables)
1. adding remaining enums (or adding comments saying that we won't add them as of now ... for cases where such comments are not currently there already)

tsalo:
Thanks!

sappelhoff:
thanks a lot again @tsalo!

PR #921, opened 2021-11-01
[ENH] Render valid value restrictions in tables based on object definitions in schema

### LINKED ISSUE #912: Render valid value restrictions in tables based on object definitions in schema

tsalo:
Closes #912.

Changes proposed:

- Add new function to `schemacode.utils`, called `describe_valid_values`, that creates a string describing valid values for an object based on the schema.
- In the case of referenced lists of valid values (i.e., enums) being used for an object, combined with an anyOf, try to flatten that set of lists into a single enum list. This makes rendering valid values for things like `iEEGCoordinateSystem` much easier.
- Incorporate valid value strings into metadata and column tables.

To do:

- [x] Remove hardcoded lists of values.
- [X] ~~Figure out how objects should be described.~~ Dropped.
- [X] ~~Figure out how anyOf situations should be described.~~ Dropped.
- [X] ~~Some cases will just be too complicated to describe automatically. How should we handle those?~~ Dropped.


tsalo:
Some highlights:

### Restrictions on numbers are shown

image

### Restrictions on arrays (and their items!)

image

### Objects with many possible restricted values can end up being a bit much

image

### No restrictions beyond type means no added text

image

### I haven't tackled `anyOf`s yet.

image

### Also haven't tackled objects

image

tsalo:
@effigies is this in line with what you were hoping for?

effigies:
Much of this is covered by the type, already, or more clearly written in the description.

The enumeration of a finite set of possible values seems like the one that's really useful.

tsalo:
Okay, I'll scale it back to only include enums then.

sappelhoff (commented):
I like this feature (the downscaled, current version). However, if we want to merge this, we should go over the description texts and make sure the description is not duplicating the allowed values in an unnecessary way. As for example here:

grafik

tsalo:
@sappelhoff done!

sappelhoff (approved):
Thanks! LGTM now

yarikoptic (approved):

PR #932, opened 2021-11-17
[INFRA] Run linkchecker in stock docker image

effigies:
Updated recipe, rebuilt, and pushed to my own repository for now. Let's see if this works.


effigies:
The linkchecker is now failing because https://thewinnower.com is down.

sappelhoff:
> The linkchecker is now failing because https://thewinnower.com is down.

sounds good because then it works :-) thanks!

tools/linkchecker-docker/neurodocker-build.sh
```diff
@@ -9,19 +9,19 @@
 #     for instructions
 
 # tag for patched linkchecker -- will also serve our image version
-version="9.4.0.anchorfix1"
+version="v10.0.1"
```
yarikoptic, line 12: Just FTR: properly (?) fixed in https://github.com/linkchecker/linkchecker/pull/462 (issue https://github.com/linkchecker/linkchecker/issues/460) which was `v10.0.0~14^2` (thus this version above includes it) instead of my adhoc https://github.com/linkchecker/linkchecker/pull/194 (which I believe was what that version pointed to).

.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
   linkchecker:
     docker:
-      - image: yarikoptic/linkchecker:9.4.0.anchorfix1-1
+      - image: markiewicz/linkchecker:v10.0.1-1
```
yarikoptic, line 24: should we may be just establish `bids` or `bids-specification` docker hub org? (otherwise we would just keep footballing it across private orgs)

yarikoptic:
Thank you @effigies !!!

.circleci/config.yml
effigies: There is a `bids`, but it's maxed out in users. Guess we could create a `bids-maintenance` one for things like this, as we did with Zenodo.

effigies:
Checked the Winnower Google cache and found PDFs that we can use that seem more stable than the website.

.circleci/config.yml
sappelhoff: > but it's maxed out in users

looks like we can have 3 users for a free-plan dockerhub org. As long as these three are also BIDS maintainers, I think that sufficiently minimizes the bus factor, so I'd rather we check and redistribute the "members" of current `bids` dockerhub than creating a new `bids-maintenance` dockerhub org (which would also be fine, but not preferred from my side)

https://www.docker.com/blog/how-to-manage-docker-hub-organizations-and-teams/

effigies: `bids` already has 9 users. Should we kick 8 people off to add 2?

sappelhoff: ah, that was the "grandfathered" status that @rwblair was talking about.

Luckily, out of those 9 people, 3 regularly attend the maintainers meetings (Ross, Franklin, myself). Then there is a certain `bidsappdev` user ... would be nice to get the credentials for that account, because then we could "repurpose" those to be shared among bids maintainers, wdyt?

effigies:
@sappelhoff @yarikoptic Having finally looked into what was involved in the docker container, it occurs to me that we can just use a standard container and install linkchecker in it. No need to push to any user's Docker repo. Let me know if there was a good reason I missed for using a custom image.

.circleci/config.yml
yarikoptic: I think creating a shared account between trusted parties is the easiest way forward, and IIRC might avoid some issues etc whenever working with individual accounts. May be just `bidsdev` or just `bids`? (IIRC there is difference between organizations and users on docker hub; so may be should be ok)

yarikoptic:
everything ingenious is simple ;)  I don't think there was a good reason really besides having a dedicated container/easier to test locally etc.

sappelhoff:
> it occurs to me that we can just use a standard container and install linkchecker in it. No need to push to any user's Docker repo

that would be the best solution then :)

sappelhoff (approved):

PR #934, opened 2021-11-18
[FIX] Remove repeated words

DimitriPapadopoulos:


Remi-Gau (approved):
wow. 

you would think that after being seen by so many people we would have spotted those.

Thanks!

sappelhoff (approved):

PR #945, opened 2021-12-03
Address BEP031 macro requests

tsalo:
Changes proposed:

- Add an optional parameter to `schema.make_filename_template`, `n_dupes_to_combine`, that determines the number of suffixes or extensions that should be combined into `<suffix>` or `<extension>`, respectively, when the filename pattern otherwise matches.
- Refactor `utils.combine_extensions` to only combine extensions of compressed versions (for example, `nii.gz` and `nii` will combine to form `nii[.gz]`, but `ome.tif` and `ome` would not combine).

### DIFF
--- a/src/04-modality-specific-files/10-microscopy.md
+++ b/src/04-modality-specific-files/10-microscopy.md
@@ -35,7 +35,7 @@ The current MACRO list ".ome.btf", ".png" and ".tif[.ome]", is it possible to ha
 If not, ".tif" and ".ome.tif" should be 2 separate lines but are concatenated as ".tif[.ome]"
 -->
 {{ MACROS___make_filename_template(datatypes=["micr"], suffixes=["TEM", "SEM", "uCT", "BF", "DF",
-"PC", "DIC", "FLUO", "CONF", "PLI", "CARS", "2PE", "MPE", "SR", "NLO", "OCT", "SPIM"]) }}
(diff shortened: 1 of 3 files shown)


tsalo:
The **Microscopy imaging data** template:

```
sub-<label>/
    [ses-<label>/]
        micr/
            sub-<label>[_ses-<label>]_sample-<label>[_acq-<label>][_stain-<label>][_run-<index>][_chunk-<index>]_<suffix>.<extension>
            sub-<label>[_ses-<label>]_sample-<label>[_acq-<label>][_stain-<label>][_run-<index>][_chunk-<index>]_<suffix>.json
```

The photo one:

```
sub-<label>/
    [ses-<label>/]
        micr/
            sub-<label>[_ses-<label>]_sample-<label>[_acq-<label>]_photo.<extension>
```

mariehbourget (commented):
Thank you @tsalo!
For BEP031, the new `n_dupes_to_combine` works perfectly for extensions. I will also be able to drop the long suffix list in the macro.
I tested without the `n_dupes_to_combine` as well, and `.ome.tif` is rendered correctly.

I don't know if there are other cases where the number of dupes to combine would be different for suffixes and extensions? For BEP031, this is not an issue.
Thanks!

tsalo:
>I don't know if there are other cases where the number of dupes to combine would be different for suffixes and extensions? 

I can see that coming up in the future, but I don't think we need it yet.

Thanks for your comments. I'll merge then.

Issue #482, opened 2020-05-26
Ambiguity in example used for inheritance principle


Lestropie:
From [Common principles](https://github.com/bids-standard/bids-specification/blob/v1.3.0/src/02-common-principles.md):

> Example 2: Multiple runs and recs with same acquisition (acq) parameters
> ```
> sub-01/
>     anat/
>     func/
>         sub-01_task-xyz_acq-test1_run-1_bold.nii.gz
>         sub-01_task-xyz_acq-test1_run-2_bold.nii.gz
>         sub-01_task-xyz_acq-test1_rec-recon1_bold.nii.gz
>         sub-01_task-xyz_acq-test1_rec-recon2_bold.nii.gz
>         sub-01_task-xyz_acq-test1_bold.json
> ```
> For the above example, all NIfTI files are acquired with same scanning parameters (acq-test1). Hence a JSON file describing the acq parameters will apply to different runs and rec files. Also if the JSON file (task-xyz_acq-test1_bold.json) is defined at dataset top level directory, it will be applicable to all task runs with test1 acquisition parameter.

The NIfTI files here indicate that there are:
- Two different "`run`s", i.e. the protocol acquisition was executed twice;
- Two different "`rec`s", i.e. reconstruction algorithms.

However what is not clear from this example is: *what data were provided as input to the two distinct reconstructions*? Given they don't include "`_run-#`", should it be assumed that both reconstructions used as input the concatenation of all data across the two runs?

Moreover, the specific details of "_run-`" and "`_rec-`" aren't provided until a later section ([Modality specific - MRI](https://github.com/bids-standard/bids-specification/blob/v1.3.0/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md)).

I would propose that an alternative, more simple example may be preferable here: one that does not introduce such potential ambiguity or dependence on other sections.

sappelhoff:
I agree that this example is more complicated than necessary. +1 from my side to work on a simpler one.

yarikoptic:
As to me inheritance principle lacks necessary formal algorithmic definition and thus is ambiguous for any (like here) but trivial scenario, thus examples indeed need to be dumbed down more.  More of formalized whining in https://github.com/bids-standard/bids-specification/issues/102 .

sappelhoff:
@Lestropie do you want to open a PR to improve the example?

Issue #259, opened 2019-06-29
Relaxation of inheritance principle?


Lestropie:
In working on a fairly wide-sweeping change proposal for BEP 016 (Diffusion derivatives), I've encountered a bit of a snag with respect to the existing spec, and want to see if there are any opinions on the matter in the wider community, since it's near the top of the specification hierarchy.

Tagging BEP016 owners: @oesteban @francopestilli 

Relevant section of current spec:
https://github.com/bids-standard/bids-specification/blob/v1.2.0/src/02-common-principles.md#the-inheritance-principle

## The problem

A key advantage of diffusion MRI is the sensitisation of images to, and subsequent modelling of, anisotropic information in each voxel. In the context of diffusion models, there are a wide range of formats in which data encapsulating such anisotropic information may be mapped to a NIfTI image (I'm currently up to 9). Many of these could easily be interpreted erroneously if adequate sidecar information were to not be provided alongside the raw image intensities.

I'll demonstrate why this presents a problem with an example. Here's the current state of my working proposal for representing the outcome of the Neurite Orientation Dispersion and Density Imaging (NODDI) model:

```Text
my_diffusion_pipeline/
    sub-01/
        dwi/
            sub-01_parameter-icvf_noddi.nii.gz
            sub-01_parameter-isovf_noddi.nii.gz
            sub-01_parameter-od_noddi.nii.gz
            sub-01_parameter-direction_noddi.nii.gz
            sub-01_parameter-direction_noddi.json
            sub-01_noddi.json
```

"`icvf`", "`isovf`" and "`od`" are 3D images containing quantitative scalar parameters from the model; "`direction`" is a 4D image with 3 volumes, which encode the estimated fibre orientation in each voxel.

Note that there are *two* JSON files in one directory here:

1. File "`sub-01_noddi.json`" contains information regarding the NODDI model, and any parameters related to the fitting of that model, which are therefore applicable to *all* output parameters from the model.

2. File "`sub-01_parameter-direction_noddi.json`" contains *only* data relevant for the correct interpretation of the fibre orientation data contained within image "`sub-01_parameter-direction_noddi.nii.gz`", which are *not* applicable to the other parameter images; this includes the data representation used to encode such orientation information within the NIfTI image, and the reference axes with respect to which the directions are defined.

This design however **clashes with the existing spec**:

>Any metadata file ... may be defined at any directory level, but no more than one applicable file may be defined at a given level ...

## Non-ideal solutions

[...]

Issue #102, opened 2018-12-05
Inheritance principle: clarify the procedure of which files would be considered


yarikoptic:
It might be just me, but from current wording it is not 100% clear if it is only `_run` and `_rec` are of special treatment to be "generalized over" or any other possible key/value pair (`_acq` etc).  I.e. for a file

```
sub-<label>/[ses-<label>/]
    func/
        sub-<label>[_ses-<label>]_task-<label>[_acq-<label>][_ce-<label>][_dir-<label>][_rec-<label>][_run-<index>][_echo-<index>]_<contrast_label>.json
```

would `task-<label>_<contrast_label>.json` at the top level be considered, disregarding any possibly present OPTIONAL key (like `_acq`, `_ce`, etc)?

What about anatomy, e.g. for a file
```
sub-<label>/[ses-<label>/]
    anat/
        sub-<label>[_ses-<label>][_acq-<label>][_ce-<label>][_rec-<label>][_run-<index>]_<suffix>.json
```
would `<suffix>.json` be considered regardless any of the possible `acq`, `ce` etc present in the target subject/session specific file?

To make it totally clear, it might be worth to formulate explicitly a generic rule to generate "higher level" filename considered for inheritance for any given "leaf" file (might be already as implemented in pybids -- didn't check yet):

- strip `sub-<label>_` prefix while providing generalization across subject(s)
- remove `_ses-<label>` while providing generalization across session(s)
- remove any `_key-<value>` if generalizing across various types of acquisition
- strip possibly present leading `_` if left only with the suffix, such as `_<suffix>.json` in  anat example

what do you think?  Or it is just me, and current wording is sufficient (I would be ok with that)

<details>
<summary>
FWIW, here is a list (click to expand) of all interesting "corner cases" across openneuro datasets
</summary>

```shell
$> ls -l */*json | grep -v -e dataset_de -e participa -e task-  
-rw------- 1 yoh yoh   240 Dec  4 15:36 ds000101/T1w.json
-rw------- 1 yoh yoh   247 Dec  4 15:37 ds000102/T1w.json
-rw------- 1 yoh yoh  1372 Dec  4 16:08 ds000117/acq-mprage_T1w.json
-rw------- 1 yoh yoh    82 Dec  4 16:08 ds000117/run-1_echo-1_FLASH.json
-rw------- 1 yoh yoh    82 Dec  4 16:08 ds000117/run-1_echo-2_FLASH.json
[...]

PR #946, opened 2021-12-05
[FIX] Rewrite inheritance principle

Lestropie:
Draft PR in search of feedback.

This change relates to #102, #259, #482; am escalating due to re-invigoration of BEP016, in particular bids-standard/bids-bep016#32.

IMO, there are multiple issues with the current description of the inheritance principle:

1. Some details are reiterated in different ways, which shouldn't be necessary if the principle is stated unambiguously;
2. Despite the availability & use of filesystem tree examples, many aspects of the principle are described through *hypothetical* modifications rather than presenting such;
3. Some instances of point 2 result in reference to either files that don't exist or would need to be renamed in order to satisfy the change suggested;
4. There is a persistent blurring between official description of the principle that should be interpreted literally, and recommendation / guidance.
5. Inheritance is described with respect to the lower-level JSON file, rather than with respect to a data file to which multiple JSONs may be applicable.
6. The logic by which software tools should classify which JSONs are or are not applicable to a given data file is not adequately clear.

Given the breadth of these issues I think a re-write is warranted. I am completely open to proposals for modifications or proposal of outright alternatives, but the ball needs to start rolling somewhere.

Full disclosure: as per #259, I would like to pursue the prospect of modifying the inheritance principle itself, not just the description of such; specifically removal of the preclusion of having multiple applicable JSONs at one level of the hierarchy. However here I am focusing solely on modifying the text describing the principle as it currently stands, which would be applicable to a patch update. Changing the principle itself should IMO be deferred to a minor or potentially major update; in addition, the isolation of that specific aspect and modification thereof should be somewhat facilitated by the clarification of the current state of the principle following these changes.


Remi-Gau:
pinging @yarikoptic and @VisLab as they both have expressed interest in this issue in the past

src/02-common-principles.md
```diff
@@ -400,108 +400,129 @@ datasets and non-compliant derivatives.
...
+Any metadata file (such as `.json`, `.bvec` or `.tsv`) MAY be defined at any
+directory level. For any given data file, any metadata file at that directory
+level or higher that does not include any entities absent from the name of the
+data file and possesses the same suffix are applicable to that data file. Such
+files are loaded from the top of the directory hierarchy downwards, such that
+values from the top level are inherited by all data files at lower levels to
+which it is applicable unless overridden by a value for the same key present
+in another metadata file at a lower level (though it is RECOMMENDED to minimise
+the extent of such overrides). There is no notion of "unsetting" a
+key/value pair.
```
Remi-Gau, line 412:
```suggestion
Any metadata file (such as `.json`, `.bvec` or `.tsv`) MAY be defined at any directory level. 
For a given data file, any metadata file at that directory level or higher 
is applicable to that data file if:
- the metadata file and the data file possess the same suffix,
- the metadata file does not include any entities absent from the name of the data file. 
Such files are loaded from the top of the directory hierarchy downwards, 
such that values from the top level are inherited by all data files 
at lower levels to which it is applicable unless overridden 
by a value for the same key present in another metadata file at a lower level 
(though it is RECOMMENDED to minimise the extent of such overrides). 
There is no notion of "unsetting" a key/value pair.
```

Lestropie: Yes, I'd contemplated bullet-pointing this myself; or potentially even enumerating so that datasets violating specific requirements / software code can be cross-referenced more precisely?

@VisLab: Good catch. There's going to need to be a hard split between JSON and other metadata files. For `.bvec` / `.bval` / `.tsv`, it I think needs to be simply selection of whichever applicable file is lowest in the directory tree; anything above it gets ignored, there's no reason to describe the contents of such as being "overwritten" by the content of files lower down. Only edge case that comes to mind would be `.tsv` files that happen to contain exactly the same columns (including their titles), but even then it's hard to justify choosing between a concatenation and a replacement. Are there any other metadata types that are worth being familiar with for the sake of getting the language / logic of the inheritance principle right?

Remi-Gau: Good idea about using an ordered list and referring to those points in the example.

VisLab: So am I understanding that for JSON the resolution is top-down, with top (closest to the root) having precedence.
For the others it is bottom up, with the bottom (farthest from the root) having precedence?  

The notions of "lowest" and "highest" level in the tree isn't necessarily clear.  Does lowest mean farthest from the root? 

I honestly thought only JSON files got inherited in the sense that there could be multiple applicable ones for the same suffix.

Remi-Gau: I think the only case of "official" of TSV inheritance we have around is having a single `task-foo_events.tsv` in the root of the directory in case all subjects / runs had the same design:

see from the BIDS examples

```bash
╰─⠠⠵ ls -l */*task*tsv
-rw-rw-r-- 1 remi remi  143 Nov 10 11:33 ds114/task-covertverbgeneration_events.tsv
-rw-rw-r-- 1 remi remi  280 Nov 10 11:33 ds114/task-fingerfootlips_events.tsv
-rw-rw-r-- 1 remi remi  143 Nov 10 11:33 ds114/task-overtverbgeneration_events.tsv
-rw-rw-r-- 1 remi remi  127 Nov 10 11:33 ds114/task-overtwordrepetition_events.tsv
-rw-rw-r-- 1 remi remi 1054 Nov 10 11:33 eeg_ds000117/task-facerecognition_channels.tsv
-rw-rw-r-- 1 remi remi 1807 Nov 10 08:42 synthetic/task-nback_events.tsv
```

VisLab: We probably need to expand the discussion of events inheritance.  Inherited `events.tsv files` are needed at several levels.  

For example, because BIDS forces multi-modality data to be separated into separate directories, simultaneously recorded MEG/EEG or fMRI/EEG will have the same event files for each run.  

In the following example from [hed-examples](https://github.com/hed-standard/hed-examples/tree/main/datasets/eeg_ds003654s_inheritance), EEG and MEG were recorded simultaneously, but separated into separate modality directories.  

```
ds_003654s:
       task-FacePerception_events.json
       sub-02:
               sub-002_task-FacePerception_events.json
               sub-002_task-FacePerception_run-1_events.tsv
               eeg:
                      sub-002_task-FacePerception_run-1_eeg.set 
               meg:
                      sub-002_task-FacePerception_run-1_meg.fif
   . . .
```

Right now we are assuming that multiple `events.json` files can be applicable to a given `events.tsv` files and that the keys are resolved from directory root down in the JSON files.  I think this is the right behavior for JSON files, but it needs to be clarified in the specification.

We have only made a single `events.tsv`  applicable to any given recording, but that `events.tsv` file could appear anywhere in the directory hierarchy.   

Events are a special case, and I think it might be feasible/useful to allow multiple events files at different levels to apply to a given run. The events are the union of the rows/join of columns.  This might be too complicated.  

The issue also comes up with derivatives.  Here the derivatives might have their own set of events. If the events from the raw data also apply do they need to always be copied into the events files? There was some discussion of this in the past, but there didn't seem to be a clear resolution or consensus on this when the discussion occurred.

VisLab: I also have a general question about inheritance.  Is the matching of entity components and their names case sensitive?

Lestropie: > The notions of "lowest" and "highest" level in the tree isn't necessarily clear. Does lowest mean farthest from the root?

I always have to cross-reference elsewhere in the spec to remind myself which is which. Does anyone know if there's a robust reference-able definition somewhere? Or does this need to be defined upfront in the spec?

Lestropie:
Lots of changes (not expecting this to be a quick PR). I think that it is by necessity transitioning from the original text being along the lines of "here's how you should do this thing" to what is really required here for the sake of stringent enforcement and validation being "here's the rules", with the downstream *consequences* of such expressed in a more user-friendly way separate to the rules themselves (what's currently called "corollaries" here could conceivably be expanded quite a lot).

Note that while @Remi-Gau's change proposal above was merged, I've explicitly unresolved the conversation as there is a lot of content there unrelated to that specific proposal that have not been resolved. Would suggest branching separate chains of thought into separate comment threads tied to the relevant lines; this one is invariably going to be messy, best to mitigate it where we can.

VisLab:
I have reviewed the commits and the bullet points make things much clearer.  Thank you!

src/02-common-principles.md
```diff
@@ -644,6 +540,147 @@ for more information.
...
+## The Inheritance Principle
+
+1.  Any metadata file (such as `.json`, `.bvec` or `.tsv`) MAY be defined at any directory level.
+
+1.  For a given data file, any metadata file is applicable to that data file if:
+    1.  It is stored at the same directory level or higher;
+    1.  The metadata and the data filenames possess the same suffix;
+    1.  The metadata filename does not include any entity absent from the data filename.
+
+1.  A metadata file MUST NOT have a file name that would be otherwise applicable
```
yarikoptic, line 552:
```suggestion
1.  A metadata file MUST NOT have a filename that would be otherwise applicable
```
if there would be agreement on https://github.com/bids-standard/bids-specification/pull/962 prompted by this line ;-)

src/02-common-principles.md
```diff
@@ -644,6 +540,147 @@ for more information.
...
+    to some data file based on rules 2.2 and 2.3 but is made inapplicable based on its
+    location in the directory structure as per rule 2.1.
+
+1.  There MUST NOT be multiple metadata files applicable to a data file at one level
+    of the directory hierarchy.
```
yarikoptic, line 557: FWIW: it is such a nice concise rule which helps to avoid ambiguity and the workaround in Example 3 is quite cute ;)

src/02-common-principles.md
```diff
@@ -644,6 +540,147 @@ for more information.
 }
 ```
 
+## The Inheritance Principle
+
+1.  Any metadata file (such as `.json`, `.bvec` or `.tsv`) MAY be defined at any directory level.
+
+1.  For a given data file, any metadata file is applicable to that data file if:
```
yarikoptic, line 547: although markdown does numbering for us, since there are direct references to them in the text (e.g. `(rule 5.2.2)`) I think it would be better just to number them here as well, so it is easier to refer back while reading markdown and not rendered version

yarikoptic (approved):
I really like it, well done.  IMHO should be taken out of the draft! ;)

Left some minor comments

src/02-common-principles.md
yarikoptic: OH -- I remembered what bothered me in my original #102 and outlined in Edit 1 there. Citing an example from there:

> I placed myself into a corner with an example of having e.g.
> 
> ```
> sub-1_task-task1_run-1_bold.json and
> sub-1_task-task1_acq-X_run-1_bold.json
> ```
>
> per subject (should be ok), and then trying to aggregate over them while retaining also _acq- if defined. 

and there are many other entities/use-cases which would fall into similar situation.  What I am thinking is to add a clarification as subitem here or a separate rule:

- A metadata file at the data files level of hierarchy MUST not be considered for inheritance if there is a matching in entities data file.

This would make `sub-1_task-task1_run-1_bold.json` not a to be considered for providing metadata to enrich `sub-1_task-task1_acq-X_run-1_bold.json`, and examples below would stay valid (we might want to add an example for this rule though)

Lestropie: > it is such a nice concise rule which helps to avoid ambiguity and the workaround in Example 3 is quite cute ;)

Well, actually (from first post):

> Full disclosure: as per #259, I would like to pursue the prospect of modifying the inheritance principle itself, not just the description of such; specifically removal of the preclusion of having multiple applicable JSONs at one level of the hierarchy.

😬 

The proposed changes here do at least I think demonstrate what would need to change in order to facilitate such. Any text I propose for such will need to be very clear for both users and software creators exactly what is permitted vs. not permitted and in what order JSONs should be loaded. But as I said in the original post, this is not a change in descriptive language but a change in prescriptive permissible structure, and so IMO necessitates either a minor or major version change rather than just a patch.

yarikoptic: Gotcha. Given the currently loose formulation of the principle, depending on the changes to it, I think it might be feasible to get it done within minor revision. So it would be great to see this PR be finalized/merged soon.

src/02-common-principles.md
Lestropie: Better would be enumeration cross-references; but as far as I can tell no standardized solution for such exists?

Lestropie: > it would be better just to number them here as well

Not permitted: Lestropie/bids-specification#2

yarikoptic: eh, thanks for trying it out! 
Indeed from https://www.npmjs.com/package/remark-lint-ordered-list-marker-value I see that
- the default is `"one"`
- there seems to be no I do not see some way to override locally in a file/for this specific list.
   - https://stackoverflow.com/a/63753061/1265472 suggests that a local to a directory `.remarkrc` could be used to overload some settings, but that would not help here/more confusing than helpful in the long run
- the other places in BIDS specification follow the "one" convention and there it seems to make sense

so I guess I will "retract" my invaluable comment ;-) Thanks again for trying

yarikoptic: actually according to https://github.com/remarkjs/remark-lint#ignore-warnings there is an inline way to disable/re-enable specific warnings in a specific block, but I do not see there a way to change configuration setting (i.e. make it "ordered") . I will suggest this disable/reenable in the PR just to see if works -- not sure if worth it if we can't lint the order

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+    1.  It is stored at the same directory level or higher;
+    1.  The metadata and the data filenames possess the same suffix;
+    1.  The metadata filename does not include any entity absent from the data filename.
+
+1.  A metadata file MUST NOT have a filename that would be otherwise applicable
+    to some data file based on rules 2.2 and 2.3 but is made inapplicable based on its
+    location in the directory structure as per rule 2.1.
```
sappelhoff, line 554: I like these back-references to 2.2, 2.3 etc. -> however in the rendered version, the nested enumerated list has letters, not numbers (see screenshot below)

image

So I would suggest to refer to 2a, 2b, 2c here.

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+
+1.  There MUST NOT be multiple metadata files applicable to a data file at one level
+    of the directory hierarchy.
+
+1.  If multiple metadata files are applicable to a given data file:
+
+    1.  For [tabular files](#tabular-files) and other simple metadata files
+        (for instance, [`bvec` / `bval` files for diffusion MRI](#04-modality-specific-files/01-magnetic-resonance-imaging#required-gradient-orientation-information),
```
sappelhoff, line 562:
```suggestion
        (for instance, [`bvec` / `bval` files for diffusion MRI](#04-modality-specific-files/01-magnetic-resonance-imaging#required-gradient-orientation-information)),
```

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+        only the file lowest in the filesystem hierarchy SHALL be treated as being
```
sappelhoff, line 563: could we use MUST instead of SHALL here? Simply because so far there is zero mention of SHALL in the spec, but lots of MUST - and I thought these were equivalent?

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+            "sub-01_task-rest_acq-default_bold.nii.gz": "",
+            "sub-01_task-rest_acq-longtr_bold.nii.gz": "",
+            "sub-01_task-rest_acq-longtr_bold.json": "",
+            }
+        },
+    "task-rest_bold.json": "",
+    }
+) }}
+
+Contents of file "task-rest_bold.json":
```
sappelhoff, line 603:
```suggestion
Contents of file `task-rest_bold.json`:
```

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+
+```JSON
+{
+    "EchoTime": 40.0,
+    "RepetitionTime": 1000.0
+}
+```
+
+Contents of file "sub-01/func/sub-01_task-rest_acq-longtr_bold.json":
```
sappelhoff, line 612:
```suggestion
Contents of file `sub-01/func/sub-01_task-rest_acq-longtr_bold.json`:
```

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+```JSON
+{
+    "RepetitionTime": 3000.0
+}
+```
+
+When reading image `sub-01/func/sub-01_task-rest_acq-default_bold.nii.gz`, only
+metadata file `task-rest_bold.json` is read; file
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.json` is inapplicable as it contains
+entity "`acq-longtr`" that is absent from the image path (rule 2.3). When reading image
```
sappelhoff, line 623: Rule 2.c, see my comment above.

I see that below this would be "5.b.ii" instead of 5.2.2, so maybe we need to discuss my suggestion - as 5.b.ii is kind of ugly.

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.nii.gz`, metadata file
+`task-rest_bold.json` at the top level is read first, followed by file
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.json` at the bottom level (rule 5.2.1);
+the value for field "`RepetitionTime`" is therefore overridden to the value 3000.0.
```
sappelhoff, line 627:
```suggestion
the value for field "`RepetitionTime`" is therefore overridden to the value `3000.0`.
```

sappelhoff (commented):
This is wonderful work, thanks a lot! I left a few comments

src/02-common-principles.md
Lestropie: I had thought intuitively that there was a distinction between MUST and SHALL, but that appears to be incorrect based on RFC2119. My thinking here was that it's entirely plausible that an app may have no interest in the content of a sidecar file, and just load an e.g. image in isolation, and therefore stating that such sidecar information MUST be loaded is to me unnecessarily dictatorial. It's more like: if you are loading a data file, *and* you are interested in the sidecar data associated with that file, *then* only the lowest sidecar file MUST be utilised.

src/02-common-principles.md
Lestropie: Closed by 8e808a12.

src/02-common-principles.md
Lestropie: Closed by 8e808a12.

src/02-common-principles.md
Lestropie: See if you're happy with 8e808a12.

src/02-common-principles.md
Lestropie: Closed by 8e808a12.

src/02-common-principles.md
Lestropie: Closed by 8e808a12.

src/02-common-principles.md
Lestropie: Revised by 8e808a12, but not yet resolving in case the enumeration formatting needs to be discussed further.

src/02-common-principles.md
Lestropie: Closed by 8e808a12.

Lestropie:
Invested parties please also see Lestropie/bids-specification#3, which proposes an alteration to the contents of this PR.

src/02-common-principles.md
sappelhoff: yes, thank you - resolving.

src/02-common-principles.md
sappelhoff: Looking at it again, I don't find it ugly anymore :-) Thanks for making the change.

src/02-common-principles.md
```diff
@@ -644,6 +540,150 @@ for more information.
...
+        accessing metadata associated with a data file MUST consider only the
+        applicable file that is lowest in the filesystem hierarchy.
+
+    1.  For [JSON files](#key-value-files-dictionaries):
+
+        1.  Files are loaded from the top of the directory hierarchy downwards,
+            such that values from the top level are inherited by all data files
+            at lower levels to which it is applicable unless overridden
+            by a value for the same key present in another metadata file at a lower level
+            (though it is RECOMMENDED to minimise the extent of such overrides).
```
sappelhoff, line 572: nitpick: Are we using American or British spelling throughout the spec? I had assumed we use American spelling, and indeed there are two occurrences of `minimize` and none of  `minimise` in the spec currently. Perhaps we should add a CI step (in a different issue/PR) that checks for American/British spelling. Just to be consistent.

Remi-Gau: Not a bad idea, but I am not quite ready to write that PR. :-)

src/02-common-principles.md
Remi-Gau: > Is the matching of entity components and their names case sensitive?

I think the case sensitivity issues was solved in https://github.com/bids-standard/bids-specification/pull/858

Remi-Gau:
Generally in favor of merging this after https://github.com/Lestropie/bids-specification/pull/3 is in.

sappelhoff (approved):

Lestropie:
@sappelhoff I'm done tinkering, content as it currently stands. Will have a go at #259 in a separate PR as discussed.

"`linkchecker`" CI test failed, naively it looks like it might have been a one-off failure that just requires a re-run since it doesn't seem to me to relate to the proposed changes?

sappelhoff:
> "linkchecker" CI test failed, naively it looks like it might have been a one-off failure that just requires a re-run since it doesn't seem to me to relate to the proposed changes?

That issue was solved in `master`, so I merged those changes back here - should all be fine now.

>  I'm done tinkering, content as it currently stands. 

awesome, let's try to get a final approval by a bunch of people and merge this!

Remi-Gau (approved):

VisLab:
There is a problem with the bullet point numbering when I do the view file on GitHub.  The first set of bullet points uses 2 (with subcategories i, ii, and iii), but the text refers to 2 (a, b, c).  

Similarly in the corollaries, rule three refers to "per rule 5.2"

effigies (commented):
Overall looks good. Two minor issues.

src/02-common-principles.md
```diff
@@ -644,6 +540,152 @@ for more information.
...
+Corollaries:
+
+1.  As per rule 3, metadata files applicable only to a specific participant / session
+    MUST be defined in or below the directory corresponding to that participant / session;
+    similarly, a metadata file that is applicable to multiple participants / sessions
+    MUST NOT be placed within a directory corresponding to only one such participant / session.
+
+1.  It is permissible for a single metadata file to be applicable to multiple data
+    files at that level of the hierarchy or below. Where such metadata content is consistent
+    across multiple data files, it is RECOMMENDED to store metadata in this
```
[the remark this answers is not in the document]
VisLab: I'm okay with taking out the recommended here.

VisLab (approved):

src/02-common-principles.md
```diff
@@ -644,6 +540,152 @@ for more information.
...
+    "EchoTime": 40.0,
+    "RepetitionTime": 1000.0
+}
+```
+
+Contents of file `sub-01/func/sub-01_task-rest_acq-longtr_bold.json`:
+
+```JSON
+{
+    "RepetitionTime": 3000.0
```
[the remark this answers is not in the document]
Lestropie: Closed by 33cc25a.

src/02-common-principles.md
```diff
@@ -644,6 +540,152 @@ for more information.
...
+```
+
+When reading image `sub-01/func/sub-01_task-rest_acq-default_bold.nii.gz`, only
+metadata file `task-rest_bold.json` is read; file
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.json` is inapplicable as it contains
+entity "`acq-longtr`" that is absent from the image path (rule 2.c). When reading image
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.nii.gz`, metadata file
+`task-rest_bold.json` at the top level is read first, followed by file
+`sub-01/func/sub-01_task-rest_acq-longtr_bold.json` at the bottom level (rule 5.b);
+the value for field "`RepetitionTime`" is therefore overridden to the value `3000.0`.
```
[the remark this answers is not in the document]
Lestropie: Closed by 33cc25a.

src/02-common-principles.md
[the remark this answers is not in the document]
Lestropie: >  This is a best practice written in the language of a validatable rule.

1.  This seems to extend beyond the RFC2119 definition. Is this an established BIDS-specific interpretation that is documented somewhere?

2.  Would the same criticism not also apply to 5.2?

    > For [JSON files](#key-value-files-dictionaries), key-values are loaded from files from the top of the directory hierarchy downwards, such that key-values from the top level are inherited by all data files at lower levels to which it is applicable unless overridden by a value for the same key present in another metadata file at a lower level (though it is RECOMMENDED to minimise the extent of such overrides).

    Issuing a validator warning regarding the presence of individual key-value overrides may be unexpected.

Lestropie:
> There is a problem with the bullet point numbering when I do the view file on GitHub.

I think the most important correspondence here is going to be not with a GitHub rendering, but with the website / PDF renderings. I couldn't find an existing example; perhaps if someone could duplicate this branch on the main repo, then set up ReadTheDocs to render that branch, we could check what the current behaviour is there?

sappelhoff:
> perhaps if someone could duplicate this branch on the main repo, then set up ReadTheDocs to render that branch, we could check what the current behaviour is there?

We should have all the needed checks available from the CI:

- [pdf](https://10077-150465237-gh.circle-artifacts.com/0/bids-spec.pdf)
- [html](https://bids-specification--946.org.readthedocs.build/en/946/)

src/02-common-principles.md
```diff
@@ -645,6 +541,152 @@ for more information.
...
+    MUST be defined in or below the directory corresponding to that participant / session;
+    similarly, a metadata file that is applicable to multiple participants / sessions
+    MUST NOT be placed within a directory corresponding to only one such participant / session.
+
+1.  It is permissible for a single metadata file to be applicable to multiple data
+    files at that level of the hierarchy or below. Where such metadata content is consistent
+    across multiple data files, it is RECOMMENDED to store metadata in this
+    way, rather than duplicating that metadata content across multiple metadata files.
+
+1.  Where multiple applicable JSON files are loaded as per rule 5.2, key-values can
```
sappelhoff, line 586:
```suggestion
1.  Where multiple applicable JSON files are loaded as per rule 5.b, key-values can
```

src/02-common-principles.md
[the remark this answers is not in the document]
sappelhoff: >  Is this an established BIDS-specific interpretation that is documented somewhere?

AFAIK this is not documented anywhere, but among maintainers we have often discussed that RECOMMENDED corresponds to a "warning" level in the validator, and MUST corresponds to an "error" level in the validator. The main point of this consideration is to not design a specification that we cannot reasonably validate.

Having that said, there are many cases in the spec where we RECOMMEND or require (MUST), where validation is currently not happening and might be difficult to implement --- as you correctly point out in your second point.

Personally I am fine with having some recommendations that we cannot "warn" about (see especially 5.b here, but maybe also the point raised by Chris.).

src/02-common-principles.md
```diff
@@ -645,6 +541,152 @@ for more information.
...
+        (for instance, [`bvec` / `bval` files for diffusion MRI](#04-modality-specific-files/01-magnetic-resonance-imaging#required-gradient-orientation-information)),
+        accessing metadata associated with a data file MUST consider only the
+        applicable file that is lowest in the filesystem hierarchy.
+
+    1.  For [JSON files](#key-value-files-dictionaries), key-values are loaded
+        from files from the top of the directory hierarchy downwards, such that
+        key-values from the top level are inherited by all data files at lower
+        levels to which it is applicable unless overridden by a value for the
+        same key present in another metadata file at a lower level
+        (though it is RECOMMENDED to minimise the extent of such overrides).
```
sappelhoff, line 572:
```suggestion
        (though it is RECOMMENDED to minimize the extent of such overrides).
```

sappelhoff:
Thanks for leading this effort @Lestropie! I think it's a huge improvement in terms of clarity.

PR #947, opened 2021-12-06
[FIX] Common principles: Add "entity" to list of definitions

Lestropie:
Just realised why the "entity" terminology when referring to key-values in file names hadn't clicked for me: it's missing from the list of definitions!

While this is a relatively trivial addition, I am nevertheless listing as a draft PR. I have noticed that in subsequent text, both "entity / entities" and "key-value(s) / key-value pairs" are used interchangeably. If the definition of "entity" is presented unambiguously in the specification, then it is worth considering the prospect of enforcing this nomenclature throughout the specification, and I think this is an appropriate location for that discussion.

### DIFF
--- a/src/02-common-principles.md
+++ b/src/02-common-principles.md
@@ -105,6 +105,9 @@ misunderstanding we clarify them here.
     The modality may overlap with, but should not be confused with
     the **data type**.
 
+1.  **Entity** - a portion of a file name, consisting of a **key** and corresponding
+    **value** separated by a hyphen.
+
 1.  **`<index>`** - a nonnegative integer, possibly prefixed with arbitrary number of
     0s for consistent indentation, for example, it is `01` in `run-01` following
     `run-<index>` specification.


Remi-Gau (changes requested):
I am generally in favor of this addition.

But I think it would be worth to link to the appendix:
- entities list: https://bids-specification.readthedocs.io/en/latest/99-appendices/09-entities.html
- https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html

sappelhoff (commented):
Thanks for this proposal, I am also in favor, and I agree with Remi that links to the entity pages should be supplied.

> If the definition of "entity" is presented unambiguously in the specification, then it is worth considering the prospect of enforcing this nomenclature throughout the specification

I think replacing mentions of "key value pairs" with entity (and back referencing to the definition) is a good idea.

tsalo:
Just a thought, but I think it's probably worth it to have a schema file for the common principles, which could then be rendered into a list with a macro.

Lestropie:
> But I think it would be worth to link to the appendix:

Done (00f9cea); but feel free to propose alternative phrasing.

> I think replacing mentions of "key value pairs" with entity (and back referencing to the definition) is a good idea.

I can give that a go, but I might try it in a separate branch based off of this one and generate a dedicated PR for that proposal.

> it's probably worth it to have a schema file for the common principles, which could then be rendered into a list with a macro.

Were you thinking of such for specifically the *definitions* within the common principles? Regardless, might be better documented as a standalone Issue so it doesn't get lost in here.

Lestropie:
For anyone interested in the prospect of adopting the "entity" term more broadly across the specification, please see Lestropie/bids-specification#1.

Remi-Gau:
Actually read this too quickly and now I have an existential question.

Is an `entity`, just the `key` or is is the `key-value` pair?

I had pretty much understood `entity` as being the `key` and would use `label` when referring to the `value`.

Seems that defining `entity` will be good thing if even a maintainer is not clear on its meaning.

sappelhoff:
> Is an entity, just the key or is is the key-value pair?

I understood entity as the key value pair :-D

sappelhoff (approved):
LGTM, even better together with https://github.com/Lestropie/bids-specification/pull/1

although as you mention in #950, the 

> allowed, OPTIONAL or REQUIRED

should maybe be amended

Lestropie:
> Is an entity, just the key or is is the key-value pair?

Yeah, crossed my mind too.

Given it is used strictly in reference to file names, where keys cannot appear in the absence of a value, I think that from a utilitarian perspective having the term refer to the pair is preferable. It does make things slightly muddy when e.g. referring to appendices that relate to entities; but that documentation doesn't only specify a list of keys, they also specify for each the requirements for the corresponding value in order to form a valid entry in the file name. So one can kinda still point to a "list of entities", even if the list is indexed by "entity keys", as the expectations for the entity values in order to form a valid entity is still there in the list content.

Is a GitHub poll needed? :-P

sappelhoff:
> Is a GitHub poll needed? :-P

maybe we first ask @bids-standard/maintainers who have not seen this, and then decide whether we need to discuss this more widely.

Remi-Gau:
I think the only thing that feels weird by defining the `key-value` pair as an entity is that it makes weird to read the schema that seems to imply that it is the key. Or it is just my reading of it.

Either way I am sure I can adjust my mental model but some clarification would be good.

VisLab:
Sorry, I have always been confused about entities and I am more confused after the discussion.  

The section called [Entity table](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html) might be helped by some clarification.   

Are `T1w` and  `eeg` an entities? They don't appear as key value pairs in the file names.  They are parsed as "suffixes" for files.  Yet, the first line of the appendix says entities (key-value pairs).   As far as I can tell `T1w` never is part of a key-value pair.  Another issue is that `events` are also suffixes but they are not treated in the same way as modalities such as `eeg`.

tsalo:
FWIW, I've always considered the entity to be the key only (e.g., "run entity"), with the combination of entity and label/index being the "key-value pair" we refer to in the specification, but I'm not heavily invested in that distinction. I will note that I think folks say "the run entity" more often than they say "the run entity key" or "the run-index entity", though.

>I think the only thing that feels weird by defining the key-value pair as an entity is that it makes weird to read the schema that seems to imply that it is the key.

@Remi-Gau I mostly took descriptions from the specification, but I can't guarantee that my own interpretation didn't bleed into any changes I made in the schema and in the appendices.

>Are T1w and eeg entities? They don't appear as key value pairs in the file names. They are parsed as "suffixes" for files. Yet, the first line of the appendix says entities (key-value pairs).

@VisLab, `T1w` and `eeg` are both modalities (and thus suffixes), rather than entities. Also, `eeg` (when it's the folder, rather than part of the filename) is a datatype as well.

>Another issue is that events are also suffixes but they are not treated in the same way as modalities such as eeg.

Could you expand on this? Some modalities/suffixes are sort of "auxiliary" in that they are associated with some other file containing imaging data, such as events files or physio files associated with a BOLD scan, but they can also exist on their own. In the case of events files, if you wanted to have those files in the absence of imaging files, you would have to have them in the `beh` folder. When they're in that folder, I think they operate just like other "primary" suffixes.

VisLab:
The [Entity table](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html) lists the things that I thought of as "modalities"  (or "suffixes") in the **Entity** column.   I think that re-titling and re-captioning this table as well as the appendix itself could improve clarity.  

@tsalo:  This was a useful observation:

> In the case of events files, if you wanted to have those files in the absence of imaging files, you would have to have them in the beh folder. When they're in that folder, I think they operate just like other "primary" suffixes.

Lestropie:
There's actually *three* options here, not two.

Which of the following is an example of an entity?

1. "Subject"
2. `sub`
3. `sub-01`

From subsequent text in "Common principles" - "File name structure":

> Note that `sub-<label>` corresponds to the `subject` entity because it has the `sub-` "key" and`<label>` "value"

This implies option 1, which has been thus far omitted from the discussion...

VisLab:
Note:  Option 1 seems to be what the title of the Entity Table in the appendix is using.  (Though I think that most people reading the table would think that the Entity title refers to items in column 1 and not to the meaning of the other titles in the header.)

satra:
at a generic level, one could borrow the description of entity from the prov ontology: https://www.w3.org/TR/prov-dm/#term-entity . as we are working on a prov extension for bids, aligning with such a description, and potentially augmenting it with bids/neuroimaging examples, may also lay some semantic groundwork for the future.

src/02-common-principles.md
```diff
@@ -105,6 +105,12 @@ misunderstanding we clarify them here.
     The modality may overlap with, but should not be confused with
     the **data type**.
 
+1.  **Entity** - a portion of a file name, consisting of a **key** and corresponding
+    **value** separated by a hyphen. Supported entities are defined in
+    [Appendix IX](99-appendices/09-entities.md); further, whether they are allowed,
```
effigies, line 110: I can't think of a case that is allowed but neither OPTIONAL or REQUIRED, so we should avoid introducing a term that implies a distinction.

```suggestion
    [Appendix IX](99-appendices/09-entities.md); further, whether they are
```

effigies (approved):
Overall looks good. Minor suggestion.

RE @satra's suggestion that we derive from "An entity is a physical, digital, conceptual, or other kind of thing with some fixed aspects; entities may be real or imaginary." I don't see how that fits with our usage, which is closer to the concept of "morpheme", IMO.

Lestropie:
And the existential descent continues... 😬 

"Morpheme" is rather obscure and doesn't convey the 2-tuple nature. "[Attribute](https://en.wikipedia.org/wiki/Attribute_(computing))" may be closer? Would however require fairly extensive changes to substitute such a core terminology entirely.

VisLab:
To me key-value pairs are entity descriptors, entity designators, or entity
specifiers not entities.

Lestropie:
Following from my [prior comment](https://github.com/bids-standard/bids-specification/pull/947#issuecomment-993099091):

- [Appendix IV](https://bids-specification.readthedocs.io/en/v1.6.0/99-appendices/04-entity-table.html) implies that "Subject" is an entity, with "`sub-<label>`" being the "format" for that entity.

- [Appendix IX](https://bids-specification.readthedocs.io/en/v1.6.0/99-appendices/09-entities.html) implies that "`sub`" is an entity, with "Subject" being the "full name" of that entity.

I'm happy to have a go at redressing this myself, but it's going to be relatively laborious to adopt a common definition *across the entire specification*, so I don't really want to be generating hypothetical PRs for all of the different alternatives.

-----

Let's see if this works...

[](https://api.gh-polls.com/poll/01FQD9YKAYFMWZH7577VBW6PSW/%22Subject%22%20%3D%20Entity%3B%20%22sub%22%20%3D%20entity%20key%3B%20%22%3Clabel%3E%22%20%3D%20entity%20value/vote)
[](https://api.gh-polls.com/poll/01FQD9YKAYFMWZH7577VBW6PSW/%22sub%22%20%3D%20Entity%3B%20%22Subject%22%20%3D%20full%20name%3B%20%22%3Clabel%3E%22%20%3D%20corresponding%20value/vote)
[](https://api.gh-polls.com/poll/01FQD9YKAYFMWZH7577VBW6PSW/%22sub-01%22%20%3D%20Entity%3B%20%22sub%22%20%3D%20entity%20key%3B%20%2201%22%20%3D%20entity%20value/vote)
[](https://api.gh-polls.com/poll/01FQD9YKAYFMWZH7577VBW6PSW/%22Entity%22%20%3D%20wrong%20word/vote)

sappelhoff:
regarding the discussion on entity/morpheme/attribute, I think we should stick with entity - both for "backwards compatibility" (at least language wise) and because (at least until now) none of the proposed alternatives have a clear advantage over "entity" IMHO.

I see a BIDS "entity" as a distinct, independent part of a BIDS filename that conveys some meaning. So I'll vote for "Subject" as the entity, which in the filename consists out of the `sub-<label>` key-value pair (`sub`=entity key, `<label>`=entity value).


@Lestropie thanks for setting up this poll, first time I saw one like this in GitHub, how did you do it / what did you use?


@VisLab your comment from https://github.com/bids-standard/bids-specification/pull/947#issuecomment-993515525 has a point: 

> Though I think that most people reading the table would think that the Entity title refers to items in column 1 and not to the meaning of the other titles in the header

do you have a suggestion how we could clarify the entity table in that regard?

Lestropie:
> regarding the discussion on entity/morpheme/attribute, ...

Agreed.

> ... thanks for setting up this poll ...

Well, it's certainly demonstrated the fact that there was a hidden lack of consensus on how this term has been used up to this point. It's almost a tie across the board, which makes it difficult to motivate investing in proposed modifications based on just one.

I almost wonder whether it needs a formulation that forces a disambiguation at *every usage*. So "entity" does not refer to one specific attribute, but a higher-order structure containing multiple members. So e.g. in clumsy Python it'd look something like:

```Python
class Entity:
    def __init__(self, fullname, key, valuetype):
        assert valuetype in ('index', 'value')
        self.fullname = fullname
        self.key = key
        self.valuetype = valuetype
        self.format = key + '-<' + valuetype + '>'

all_entities = ( Entity('Subject', 'sub', 'label'),
                 Entity('Session', 'ses', 'label'),
                 Entity('Run', 'run', 'index'),
                 ...
               )

Class EntityInstance:
    def __init__(self, key, value):
        entity = [item for item in all_entities if item.key == key]
        assert len(entity) == 1
        entity = entity[0]
        if entity.valuetype == 'value':
            assert isinstance(value, int)
        self.key = key
        self.value = value
    def __str__(self):
        return self.key + '-' + self.value
```

In this way, the term "entity" does not refer to the full name (poll item 1), or the key (poll item 2), or the string generated according to the format by a specific instance of such (poll item 3), but the superstructure that encapsulates all of these things. One would therefore have to *always* refer to the "entity full name", or the "entity key", or the "entity format", or the "entity (instance) string".

> ... how did you do it / what did you use?

https://app.gh-polls.com/

Unfortunately doesn't look very good in dark mode. Seems to be plenty of alternatives also.

> do you have a suggestion how we could clarify the entity table in that regard?

I think that my proposal above would address this: row 1 first column would be "Entity full name", and row 2 first column would be "Entity format".

VisLab:
Regarding my comment from #947:

>    Though I think that most people reading the table would think that the Entity title refers to items in column 1 and not to the meaning of the other titles in the header

I would suggest having a separating the information about format and the information about modality into separate tables.

I would have one table that is common to all modalities at the top.

| Entity  |   Format  |   Example |
| ------- | ---------- | ---------- |
| Subject | `sub-<label>` |  `sub-big1` |
 

Then for each modality I would entity the sections like this:

### Magnetic Resonance Imaging entity requirements


| Data-type  |   Subject  |   Session | .... |
| ------------ | ---------- | --------- | ---|
| anat (etc.) | REQUIRED |  OPTIONAL | etc.....|

Lestropie:
> I would suggest having a separating the information about format and the information about modality into separate tables.

Given the PR topic and complexity of the ensuing discussion, it might be preferable to have this suggestion as its own standalone Issue so as not to be skipped over. While clarification of the definition of "entity" in this PR may help to disambiguate the contents of the entity table, that's a secondary effect of such, whereas for your suggestion that's the primary focus.

Lestropie:
Bumping seeking maintainer feedback on alternative proposal in [comment](https://github.com/bids-standard/bids-specification/pull/947#issuecomment-1005387248) above; I'm unable to progress on this proposal based on poll results.

sappelhoff:
I like the proposal from https://github.com/bids-standard/bids-specification/pull/947#issuecomment-1005387248. Let's collect some more opinions 

cc @bids-standard/maintainers @hoechenberger @adam2392

Lestropie:
Invested / interested parties please see Lestropie/bids-specification#4, which is intended as an augmentation of this PR based on prior discussion here.

Lestropie:
https://github.com/Lestropie/bids-specification/pull/4 has been merged, and 699b74e brings this up-to-date with master (see commit message for conflicts). PR is no longer flagged as draft.

src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
```diff
@@ -513,24 +513,24 @@ other label than `singleband` and `multiband` as long as they are consistent
 across subjects and sessions and consist only of the legal label characters.
 
 Similarly the OPTIONAL [`ce-<label>`](../99-appendices/09-entities.md#ce)
-key/value can be used to distinguish
+entity can be used to distinguish
 sequences using different contrast enhanced images. The label is the name of the
-contrast agent. The key ContrastBolusIngredient MAY be also be added in the JSON
+contrast agent. The key "`ContrastBolusIngredient`" MAY be also be added in the JSON
```
Remi-Gau, line 518:
```suggestion
contrast agent. The key `"ContrastBolusIngredient"` MAY be also be added in the JSON
```

Remi-Gau (approved):

ericearl (approved):
Nice work!

PR #955, opened 2021-12-12
[FIX] Entity table: Clarify meaning of empty cells

### LINKED ISSUE #950: Entity table: Neither "REQUIRED" nor "OPTIONAL" = ?
From [Appendix IX: Entities](https://bids-specification.readthedocs.io/en/latest/99-appendices/09-entities.html):

> The ordering of entities and whether it is allowed, OPTIONAL, or REQUIRED for each is given in the Entity Table.

-  It seems to me that "allowed" and "OPTIONAL" are functionally equivalent.
   Potentially the usage of "allowed" was intended to be understood with regards to the *antonym* of such? Let's use "forbidden" here for now.

-  It's also unusual to me to be mixing stringently-defined terms from RFC2119 with a term that is not.

------

From [Appendix IV: Entity table](https://bids-specification.readthedocs.io/en/latest/99-appendices/04-entity-table.html):

> REQUIRED and OPTIONAL entities for a given file type are denoted.

The inference here is that for any field that is *blank*, i.e. neither REQUIRED nor OPTIONAL, then it is *forbidden* to use that entity for that format. It is however unusual to me that this not be stated somewhere *explicitly* in the specification, given that it relates directly to potential violation of the specification. 

The solution to such would not necessarily have to fill all cells in the table with some appropriate term (especially since RFC2119 doesn't really have a good candidate), but it does IMO necessitate some statement before the table regarding the interpretation of empty cells. 

Thoughts welcome.

Lestropie:
Closes #950.


sappelhoff (approved):

tsalo (approved):
LGTM. Thanks for this.

sappelhoff:
Thanks @Lestropie

Issue #960, opened 2021-12-16
Add HTML comments linking to documentation before macro calls in specification


tsalo:
Recently, a number of users have mentioned that the schema-related macros in the specification are difficult to work with for folks who are not already familiar with them. The maintainers are working on improving documentation of the schema-related code, and one idea was to add an HTML comment before each macro call in the specification linking to documentation and clarifying what the code snippet means.

Something like this (proposed by @effigies):

```
<!-- This block generates a metadata table. The definitions of these fields can be found in src/schema/objects/metadata.yaml and a guide for editing at <link>. -->
```

The comment would vary based on the type of macro call, so suffix tables would link to the suffixes YAML file, etc.

VisLab:
I think this would be very helpful.

PR #963, opened 2021-12-20
[DOC] add documentation for macros

Remi-Gau:
Several of the items in here could / should probably be tackled in separate PRs

- [x] transfer content from hackmd
- [ ] remove current doc from CONTRIBUTING (or integrate it in the new doc)
- [ ] decide where this doc on macros should go (root folder, in tools, somewhere else?)
- [ ] finalize content of doc
  - [ ] possibly remove direct reference to the schema to avoid having too many "new" concepts in that doc
  - [ ] improve doc string of each macro
- [ ] preface each macro call with link to the macro doc

---

- [ ]  add a link checker for our doc (as the issue was raised that we had undetected dead links in the doc)

### DIFF
--- a/macros_doc.md
+++ b/macros_doc.md
@@ -1,0 +1,310 @@
+# Using MkDocs macros in the BIDS specification
+
+## What are macros and why use them?
+
+A macro is a rule or pattern that specifies how an input should be mapped to
+output. Macros are very useful for standardizing the output format for items
+such as tables. You might already be familiar with using macros from other tools
+such as Excel.
+
+MkDocs (the tool we use to turn the markdown version of the specification into
+HTML pages) supports macros. In the BIDS specification document, we use these
+macros to standarize the format of items such as tables and examples.
+
+Increasingly, parts of the BIDS specification are being formalized into a
+"schema" so that requirements in the specification can be automatically checked
+by validators. Several of the macros incorporate information from the schema to
+assure consistency.
+
+Table of macros here: | Name | Purpose | Uses schema | Link to example of use |
+| ---- | ------- | ----------- | ---------------------- |
+
+## When should I use a macro?
+
+Most typo corrections and minor changes to the BIDS specification do not require
+you to use macros. Even adding a table may not require you to use macros unless
+the table falls into one of the categories listed in the macros table.
+
+If you want to add content with a macro and need help, do not hesitate to
+contact a member of the bids-maintainers for help.
+
+Nope. No need to learn how to program to use those macros. But some macros
+require some specific "input", so you will just need to know what input to give
+it. In many cases, it will be just a metadata term whose definition or
+description should be displayed somewhere in the specification.
+
+Once you have given the macro its input, it usually knows where to look for the
+rest of the info it should use to do the rest of its work.
+
+The only exception to this is when you want to add new content. Say you want to
+add a completely new piece of metadata to the specification. In that case, you
+will need to create the input for the macro to be able to do its job. For that,
+you will need to update the schema itself.
+
+## How to
+
+Say you want to edit the content of table of the `Reconstruction` section for
+the PET page.
+
+The HTML version of the this section is here:
+
+https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/09-positron-emission-tomography.html#reconstruction
+
+In the markdown document, it is here:
+
+https://github.com/bids-standard/bids-specification/blob/master/src/04-modality-specific-files/09-positron-emission-tomography.md#reconstruction
+
+GitHub's way of directly rendering markdown documents makes it a bit harder to
+read, so if you opened the markdown document in your code editor it would look
+like this.
+
+```python
+#### Reconstruction
+
+{{ MACROS___make_metadata_table(
+   {
+      "AcquisitionMode": "REQUIRED",
+      "ImageDecayCorrected": "REQUIRED",
+      "ImageDecayCorrectionTime": "REQUIRED",
+      "ReconMethodName": "REQUIRED",
+      "ReconMethodParameterLabels": "REQUIRED",
+      "ReconMethodParameterUnits": "REQUIRED",
+      "ReconMethodParameterValues": "REQUIRED",
+      "ReconFilterType": "REQUIRED",
+      "ReconFilterSize": "REQUIRED",
+      "AttenuationCorrection": "REQUIRED",
+      "ReconMethodImplementationVersion": "RECOMMENDED",
+      "AttenuationCorrectionMethodReference": "RECOMMENDED",
+      "ScaleFactor": "RECOMMENDED",
+      "ScatterFraction": "RECOMMENDED",
+      "DecayCorrectionFactor": "RECOMMENDED",
+      "DoseCalibrationFactor": "RECOMMENDED",
+      "PromptRate": "RECOMMENDED",
+      "RandomRate": "RECOMMENDED",
+      "SinglesRate": "RECOMMENDED",
+   }
+) }}
+```
+
+---
+
+**HINT:** if you want to see the "raw" content of a file on Github you can
+always press the `raw` button that is on the top right of the document you are
+browsing on Github.
+
+---
+
+This section calls the macro `make_metadata_table` to create the table when
+building the HTML version of that page.
+
+The macro will create the different columns of the table:
+
+- Key name
+- Requirement level
+- Data type
+- Description
+
+A general description of that Macro call would look like this:
+
+```python
+{{ MACROS___make_metadata_table(
+   {
+      "TermToRender": "REQUIREMENT_LEVEL plus anything else after", "Extra content you want to append after the description of that term."
+   }
+) }}
+```
+
+To know what to put in the different columns, the macro will go and look into
+the `metadata.yaml` file in the BIDS schema and find the entry that correspond
+to the term you want to add.
+
+https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/metadata.yaml
+
+And in the above example, all the information about `AcquisitionMode` would be
+read from that section:
+
+https://github.com/bids-standard/bids-specification/blob/05f7ce1ae9cc72b087b57c39183e36afcf37cb00/src/schema/objects/metadata.yaml#L20
+
+If you had to write the markdown equivalent of the general example for the Macro
+call above it would give a table that would look like this:
+
+| Key name     | Requirement level                          | Data type | Description                                  |
+| ------------ | ------------------------------------------ | --------- | -------------------------------------------- |
+| TermToRender | REQUIREMENT_LEVEL plus anything else after | string    | whatever description was in the metadata.yml |
+
+### Modifying a term in the table
+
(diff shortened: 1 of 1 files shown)


VisLab (commented):
I only edited the first 31 lines of this while it was in the other format.  After the table of macros gets filled in, I think the rest can be edited.  I would suggest that the first example be a simple example that doesn't rely on any information from the schema.  

After that the meta data table is by far the most important and requires more explanation. 

If someone can fill in the table, I think the rest of us can more easily make contributions on the wordsmithing.

tsalo (commented):
Just some initial thoughts. It would also be nice to link to the MkDocs, mkdocs-material, and mkdocs-macros-plugin documentation.

One other thought- can we use MkDocs to build RTD-like package documentation for the schemacode package, and even maybe the macros module? Then we could link to a canonical list of the macros instead of managing multiple lists across files.

sappelhoff:
related: #960

Remi-Gau:
@VisLab created a table of available macros (very rough but better than nothing), also made sure that the first example is "schema less". Feel free to make wordsmithing suggestions.

@tsalo added links at the end

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
 
 MkDocs (the tool we use to turn the markdown version of the specification into
```
VisLab, line 35: Add BIDS before specification here and remove BIDS from next sentence:

MkDocs (the tool we use to turn the markdown version of the BIDS specification into

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
...
 HTML pages) supports macros. In the BIDS specification document, we use these
-macros to standarize the format of items such as tables and examples.
+macros to standardize the format of items such as tables and examples.
+
+## What kind of input information are required by macros?
+
+It will depend for each macro.
```
VisLab, line 41: Replace lines 41-45 with:

Some macros only use the arguments you directly supply in the macro call.

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
...
+
+For some of them, all the inputs will be directly available in the markdown
+document where you have put the call to the macro.
+
+Other macros will need to be pointed to some external input (like a metadata
```
VisLab, line 46: Replace the first sentence in lines 46-47 with:

Other macros use information (such as metadata terms) from external sources, and you will need to provide links to this information as part of the call.

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
...
-Increasingly, parts of the BIDS specification are being formalized into a
-"schema" so that requirements in the specification can be automatically checked
-by validators. Several of the macros incorporate information from the schema to
-assure consistency.
+## What macros are available? What can we use macros for?
+
+All the macros we use are in listed in this
+[python file](https://github.com/bids-standard/bids-specification/blob/master/tools/mkdocs_macros_bids/macros.py).
+
+| Name                    | Purpose                                                                                | Uses schema | Link to example of use |
```
VisLab, line 57: Change last item in title of the table to:

Link to example   or just to Example

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
 
 MkDocs (the tool we use to turn the markdown version of the specification into
 HTML pages) supports macros. In the BIDS specification document, we use these
-macros to standarize the format of items such as tables and examples.
+macros to standardize the format of items such as tables and examples.
```
VisLab, line 37: I would move the example of writing the file tree to here.

The following is an example of a macro used to create consistent "file tree" layouts in the documentation.
The macro takes a single parameter, the directory tree to be displayed in JSON format.
If you insert the following in the BIDS markdown document:

```
{{ MACROS___make_filetree_example(

   {
   "sub-01": {
      "func": {
         "sub-control01_task-nback_bold.json": "",
         },
      }
   }

) }}
```
The result would be rendered in the specification document as:

```Text
└─ sub-01/
   └─ func/
      └─ sub-control01_task-nback_bold.json
```

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
...
-"schema" so that requirements in the specification can be automatically checked
-by validators. Several of the macros incorporate information from the schema to
-assure consistency.
+## What macros are available? What can we use macros for?
+
+All the macros we use are in listed in this
+[python file](https://github.com/bids-standard/bids-specification/blob/master/tools/mkdocs_macros_bids/macros.py).
+
+| Name                    | Purpose                                                                                | Uses schema | Link to example of use |
+| ----------------------- | -------------------------------------------------------------------------------------- | ----------- | ---------------------- |
```
VisLab, line 58: `make_filetree_example` doesn't use the schema does it?  Shouldn't it say No in the third column

The third column should have consistent capitalization.

Also, should the macros be in alphabetical order in the table?

macros_doc.md
```diff
@@ -26,7 +76,11 @@ you to use macros. Even adding a table may not require you to use macros unless
 the table falls into one of the categories listed in the macros table.
 
 If you want to add content with a macro and need help, do not hesitate to
-contact a member of the bids-maintainers for help.
+contact a member of the bids-maintainers for help. To do this, you can either
+mention an individual maintainer by their GitHub username or mention the whole
+team (`@bids-standard/maintainers`).
+
+## Do I need learn how to code in Python to use those macros?
```
VisLab, line 83: This is a rewrite of lines 83-96

## Do I need to know how to program to use macros?

Macros don't require programming knowledge to use.  You do need to know what arguments the macro expects. The examples linked in the above table provide useful guidance in this respect.

Macros that extract information from the schema also require you to use the correct terms in the schema. This process is illustrated in the next section.

Note that under the hood the macros themselves call python code that can be found in the [`tools` folder](https://github.com/bids-standard/bids-specification/tree/master/tools). If you are interested in creating a new macro for users, this would be useful.

macros_doc.md
```diff
@@ -38,10 +92,105 @@ rest of the info it should use to do the rest of its work.
 
 The only exception to this is when you want to add new content. Say you want to
 add a completely new piece of metadata to the specification. In that case, you
-will need to create the input for the macro to be able to do its job. For that,
-you will need to update the schema itself.
+will need to create the input for the macro to be able to do its job.
+
+## Why use macros at all?
```
VisLab, line 97: I think lines 97 to 141 should be moved to the end of the document and probably ultimately scrapped... but there might be some good nuggets to extract.

macros_doc.md
```diff
@@ -9,15 +34,40 @@ such as Excel.
...
+| make_entity_definitions | Generate definitions and other relevant information for entities in the specification. | Yes         |                        |
+| make_glossary           |                                                                                        | yes         |                        |
+| make_suffix_table       | Generate a markdown table of suffix information.                                       | yes         |                        |
+| make_metadata_table     | Generate a markdown table of metadata field information.                               | Yes         |                        |
+| make_columns_table      | Generate a markdown table of TSV column information.                                   | Yes         |                        |
+| make_filetree_example   | Generate a filetree snippet from example content.                                      | Yes         |                        |
 
-Table of macros here: | Name | Purpose | Uses schema | Link to example of use |
-| ---- | ------- | ----------- | ---------------------- |
+Note that under the hood the macros themselves call some more python code that
```
VisLab, line 68: Delete lines 68-69 here --- proposed moving them later.

VisLab (commented):
I stopped at this point, since I wanted to make sure there was agreement before going on.  I think the table with example links will be great and is the key.

macros_doc.md
Remi-Gau: For the moment, I am not moving it but copy pasting it. I feel that because this is an FAQ, it is very possible people will not read from top to bottom.

macros_doc.md
Remi-Gau: Correct. Fixing this.

> `make_filetree_example` doesn't use the schema does it? Shouldn't it say No in the third column

Correct. Fixing this.

Remi-Gau: > The third column should have consistent capitalization.
> 
> Also, should the macros be in alphabetical order in the table?

Done and done.

macros_doc.md
Remi-Gau: Soft push back on this. I think that giving some explanation as to the "why" of some infrastructure / design choice is important.

VisLab:
Thank you for making the changes.... Here are a few more:

In the table of contents, I would suggest changing the title to:

Below you will find answers to frequently asked questions regarding using [macros in the BIDS specification](#using-mkdocs-macros-in-the-bids-specification):


and then eliminating the first level of the table of contents.

VisLab:
The section that titles the table, I would suggest changing to:

> ## What macros are available?

VisLab:
In the **Adding a new term to the table section**.  There is a typo:

> If this is the case because the term already exists

VisLab:
In the **Should I create a macro if I need a new kind of table?** section, I would change the second paragraph to start with:

> ... unless it is clear that this kind of table will reappear many times in the future in the specification.

Otherwise it isn't a sentence...

VisLab:
Thank you again for producing this document.  I am looking forward to the links to the examples... but I think that it has cleared up many of my own confusions, and hopefully others as well.

Remi-Gau:
@VisLab made the suggested changes in this commit: https://github.com/bids-standard/bids-specification/pull/963/commits/be1a2ed51c194321c2dbb9e16996aa933e304895

Remi-Gau:
> Thank you again for producing this document. I am looking forward to the links to the examples... but I think that it has cleared up many of my own confusions, and hopefully others as well.

Regarding examples, I am uncertain what the best way to proceed. 

2 suggestions that are not mutually exclusive (a mix and match is possible depending on the examples)

- link to some (simple) "live examples" that exist in the specification.
  - pros: 
    - this is more "ecological" and stays close to what users might have to do
    - can allow users to "run" the examples but using `mkdocs serve` to see how they render for real
  - cons: 
    - harder to maintain because the link might break (though quite a few macros are used only once and in location that don't change often)
 
- add examples that in the same documentation
  - pro: some examples already exist in that documentation
  - con: 
    - examples cannot be run to be rendered
      - there is [notebook](https://github.com/bids-standard/bids-specification/blob/master/tools/filetree_example.ipynb) for the file tree one but it is not an example of the macro per se but one for its underlying code.
    - those examples will have to be updated if the macro change
    
Feedback welcome.

VisLab:
Either way or mix is okay with me.   One way is to have links in the table to examples in the document.  Some of the examples could have links to live examples at the end of the example.

tsalo:
@Remi-Gau, I'm hoping we'll be able to transition to proper auto-executed examples as we move toward a separate repo and/or documentation, but for right now I think static examples would be easier to draft.

CPernet (approved):
thx Remi, now users know what to do :-)

Remi-Gau:
Added some links to examples as used in the spec currently. Those might "break" (by not pointing the the correct line) but this is better than nothing for now.

So this should be ready for merging IMHO: the unticked boxes in the top messages should probably be tackled in different issues.

PR #968, opened 2022-01-07
ENH: Update B0Field metadata to accommodate single-blip fieldmaps

effigies:
Follow-up to #622. We realized that as currently worded, there are unresolvable ambiguities for pepolar schemes. Consider the case where you have two BOLD series and one reverse-PE EPI fieldmap. The inputs to estimate the fieldmap could be one or both BOLD files (along with the reverse-PE EPI), and the estimated fieldmap could then apply to one or both.

This allows files to be listed as part of multiple identifiers, and for a BOLD file to be part of an `B0FieldIdentifier` AND have a `B0FieldSource`.

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -174,30 +174,33 @@ B0FieldIdentifier:
   description: |
     The presence of this key states that this particular 3D or 4D image MAY be
     used for fieldmap estimation purposes.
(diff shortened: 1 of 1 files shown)


sappelhoff (commented):
I'm not an expert: why do you remove the `participant_relative` format?

effigies:
Because these are strings, not paths.

oesteban (approved):

oesteban:
Thanks for this.

PR #969, opened 2022-01-07
BIDS Validator XS

TheChymera:
Draft PR to track the diff for a smaller BIDS validator.


TheChymera:
NOTES from meeting:

* Example data for testing: https://github.com/bids-standard/bids-examples/
* Repository of path-accessible schema versions, only https://github.com/bids-standard/bids-schema , older versions will need to be added, which might break due to logic updates — potentially old schemas can be updated wrt. logic but not content. This might shape up to be a lot of work when updating schema logic.

TheChymera:
@tsalo @yarikoptic path validation should be done pretty reliably now, and a slight variation of the code is already used “in production” as part of DANDI. I think this is ready for a review.

Only thing missing are integration tests (commented out right now), since I'm unsure how you would like to handle the reference data download. DANDI uses a fixture, though maybe a static archive could also do, or even an included archive. The issue with pulling from git is that tests might break due to asynchronous test data and validator updates. Though maybe that's desired to make enough noise to ensure the code always stays up to date....

TheChymera:
@effigies @tsalo updated it to work after the latest update following the hackathon, also has good test coverage. Not fully featured as of right now (only checking filenames, still, and not using the error codes), but I think merging the current version already offers good value, and I can add further features incrementally, possibly in sync with what we consider most definitive-ish.

tools/schemacode/schemacode/validator.py
```diff
@@ -0,0 +1,733 @@
...
+        regex_extensions = fixed_variant_extensions[0]
+    regex_string = f"{regex_string}{regex_extensions}"
+
+    return regex_string
+
+
+def _add_subdirs(regex_string, variant, datatype, entity_definitions, modality_datatypes):
+    """Add appropriate subdirectories as required by entities present."""
+
+    label = "([a-z,A-Z,0-9]*?)"
```
effigies, line 160: I don't think you're using special options with the `re` module, so I'm pretty sure these commas are out of place.

```suggestion
    label = "([a-zA-Z0-9]*?)"
```

```Python
>>> import re
>>> re.match(r"[a-z,A-Z]*", "abc,def")
<re.Match object; span=(0, 7), match='abc,def'>
>>> re.match(r"[a-zA-Z]*", "abc,def")
<re.Match object; span=(0, 3), match='abc'>
```

tools/schemacode/schemacode/validator.py
```diff
@@ -0,0 +1,733 @@
...
+    * More issues in comments.
+    * Using pre 3.8 string formatting for legibility.
+
+    Returns
+    -------
+    regex_schema : list of dict
+        A list of dictionaries, with keys including 'regex' and 'mandatory'.
+    """
+
+    label = "([a-z,A-Z,0-9]*?)"
```
effigies, line 262:
```suggestion
    label = "([a-zA-Z0-9]*?)"
```

tools/schemacode/schemacode/validator.py
TheChymera: thank you, solved!

rwblair:
Does not seem to validate sidecar files at the root of the dataset:
`2022-05-19 14:25:52,937 [ WARNING] The `/home/rwblair/projects/datasets/ds000003/task-rhymejudgment_bold.json` file was not matched by any regex schema entry.`

tools/schemacode/schemacode/validator.py
```diff
@@ -0,0 +1,733 @@
...
+        regex_extensions = fixed_variant_extensions[0]
+    regex_string = f"{regex_string}{regex_extensions}"
+
+    return regex_string
+
+
+def _add_subdirs(regex_string, variant, datatype, entity_definitions, modality_datatypes):
+    """Add appropriate subdirectories as required by entities present."""
+
+    label = "([a-zA-Z0-9]*?)"
```
effigies, line 160: I think we want at least one character, and greedy. These should always end with `_` or `.` or `/`, right? And shouldn't we have indices for things like `run` and `echo`?

```suggestion
    label = "([a-zA-Z0-9]+)"
    index = "(\d+)" 
```

TheChymera: loaded it from the schema, should be the most.... schema-tic ;) way to do it.

TheChymera:
@effigies loading variable field strings from the schema now. You said we could merge before the inheritance thing? It's in the notes, but might be best for a separate PR.

effigies (approved):
Agreed. Let's merge and have a base to work off of.

@yarikoptic Would you mind doing a review?

PR #984, opened 2022-01-22
[FIX] add microscopy to modalities in schema

### LINKED ISSUE #983: add microscopy to yml for modalities
discovered when trying to index microscopy examples with bids matlab by using the schema

https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/modalities.yaml

Remi-Gau:
fixes #983


tsalo (approved):
LGTM. Thanks for catching this.

It looks like `micr` is already covered in `rules/datatypes/micr.yaml` and `rules/modalities.yaml`, so this seems to be the only missing element.

sappelhoff (approved):

mariehbourget (approved):
LGTM! Thanks @Remi-Gau for catching and correcting this.
I checked the rest of the schema files and it looks like it was the only thing missing.

Issue #948, opened 2021-12-07
`acq_time` for `sessions.tsv` ?


Remi-Gau:
It seems that `acq_time` and `pathology` are now mentioned as columns for sessions TSV files.

I am not sure why this happened, but I am not sure it makes sense to have acquisition time in `sessions.tsv` AND `scans.tsv`.

And if we do have an `acq_time` then the current definition is taken from the scan level and read bizzarrely for sessions.

Am I missing something here?

For `pathology`, was that introduced through the `sample entity` PR?

 @tsalo is it possible we did not cross check things properly when we started using the schema for this?

See the latest version:
https://bids-specification.readthedocs.io/en/latest/03-modality-agnostic-files.html#sessions-file

See the stable version
https://bids-specification.readthedocs.io/en/stable/06-longitudinal-and-multi-site-studies.html#sessions-file

tsalo:
I believe I added `acq_time` based on the example TSV (see below for screenshot from stable version):

![image](https://user-images.githubusercontent.com/8228902/145048876-4083a12b-5294-4a9e-838b-1bcdbb547a08.png)

I didn't realize the definition read weirdly (I probably didn't really look at it in context), so we should probably adjust that. I assume we want to allow/support `acq_time` for sessions files since it's been in the example snippet for quite some time.

I added `pathology` in the TSV schema PR as well because the definition added in #812 says "the pathology MAY instead be specified in Sessions files in case it changes over time."

Remi-Gau:
> I didn't realize the definition read weirdly (I probably didn't really look at it in context), so we should probably adjust that. I assume we want to allow/support `acq_time` for sessions files since it's been in the example snippet for quite some time.

Well it is just a bit incompatible with the fact that there will be one row per session.

Current version for sessions.tsv:
```
Acquisition time refers to when the first data point in each run was acquired. Furthermore, if this header is provided, the acquisition times of all files that belong to a recording MUST be identical.
```

Maybe something like:
```
Acquisition time refers to when the first data point of the first run was acquired.
```

> I added `pathology` in the TSV schema PR as well because the definition added in #812 says "the pathology MAY instead be specified in Sessions files in case it changes over time."

ah OK I forgot about this one.

sappelhoff:
Damn ... I think both of you are right (Taylor for including it; Remi for pointing out it seems weird).

[...]

PR #985, opened 2022-01-22
[SCHEMA] Add objects and rules for sessions and scans files

### LINKED ISSUE #978: `sessions.tsv` representation in YAML
Is [the `*_sessions.tsv`](https://bids-specification.readthedocs.io/en/stable/06-longitudinal-and-multi-site-studies.html#sessions-file) represented anywhere in the YAML schema data? I couldn't find it mentioned other than in a comment, and it would be pretty cool to have it there for automatic specification loading, as seen [here](https://github.com/bids-standard/bids-specification/pull/969).

tsalo:
Closes #978. This continues from Thursday's schema conversation.

Changes proposed:

- Add object definitions for the scans and sessions files to `objects/suffixes.yaml`.
- Create a new rules file for "tabular metadata files", such as participants.tsv, scans.tsv, and sessions.tsv files. The file is currently called `tabular_metadata.yaml`.
    - The proposed structure formalizes the assumption that presence or absence of allowed entities _also_ determines where they should be located in the dataset (e.g., if a file requires the `sub` entity, then it MUST go in the subject subfolder).
- Remove cases of `required: false` from `rules/top_level_files.yaml`, given that the default should be `false` moving forward.
- Move object definitions in `objects/top_level_files.yaml` to `objects/suffixes.yaml`. These filenames are essentially suffixes without entities, even though some of them have underscores in the names (e.g., `dataset_description`), which are normally prohibited for suffixes. 
    - I just think this organization is a little cleaner, overall.

To do:

- [ ] Discuss the proposed changes.
- [ ] Update the rendering code to work with the new rules.
- [ ] Add a test or two for the new rules.


src/schema/objects/suffixes.yaml
```diff
@@ -554,6 +554,45 @@ sbref:
...
+    recording MUST be identical.
+
+    Datetime should be expressed as described in Units.
+
+    For anonymization purposes all dates within one subject should be shifted by a randomly chosen
+    (but consistent across all recordings) number of days.
+    This way relative timing would be preserved, but chances of identifying a person based on the
+    date and time of their scan would be decreased.
+    Dates that are shifted for anonymization purposes SHOULD be set to the year 1925 or earlier to
+    clearly distinguish them from unmodified data. Shifting dates is RECOMMENDED, but not required.
```
sappelhoff, line 581: I think the details about anonymization should only be in the "Units" section of the spec, not in the schema entries, see also: #948 #986

TheChymera:
I was also working on this and trying to make it work with our validator.
I created a draft PR (including the suffix description from here): https://github.com/bids-standard/bids-specification/pull/987

I thought the suggestion was to treat these files similarly to datatypes and not create a separate YAML file for them, but rather have the hierarchy level defined by the presence of the subject and session entities.
If the subject or session entities are present, so are the respective directories — that seemed very elegant.

TheChymera:
In as far as anybody finds this relevant at this point, this is how we could ~seamlessly integrate directory addition with all the other datatypes:

https://github.com/TheChymera/bids-specification/blob/e92d094a20135992cc4f9d8c7ee27e0c94bc1316/tools/schemacode/schemacode/validator.py#L81-L103

I've tried to base this code closely off of `render.py` and maybe we could use similar functions for both (or the same, if there's an easy way to output either regex or human-readable styling)... in any case this approach is usable with the logic seen here: https://github.com/bids-standard/bids-specification/blob/f8882915a55613b26d60d1789a75da037c8dd63a/tools/schemacode/schemacode/render.py#L152

tsalo:
>If the suggestion was to treat these files similarly to datatypes and not create a separate YAML file type for them, perhaps the YAML documenting them can live inside the datatype directory? Might also be good as an intermediary step, i.e. those are files which behave the same in as far as for the time being others still do not.

I think I'd prefer to avoid having them in the datatypes folder and to instead pursue adding a `datatype` field, as proposed by (I believe) @effigies. Then, these files would either have `datatype: None` or just nothing, with the default being that datatype is None.

>In as far as anybody finds this relevant at this point, this is how we could ~seamlessly integrate directory addition with all the other datatypes:
>
>https://github.com/TheChymera/bids-specification/blob/e92d094a20135992cc4f9d8c7ee27e0c94bc1316/tools/schemacode/schemacode/validator.py#L81-L103

That's awesome! I haven't had a chance to take a close look at either of your code snippets, but I will try to soon.

TheChymera:
> I'd prefer to avoid having them in the datatypes folder and to instead pursue adding a `datatype` field

@tsalo so you'd like to flatten the datatype file hierarchy as well and have everything live under `rules/` ?

TheChymera:
> I've tried to base this code closely off of `render.py`and maybe we could use similar functions for both

Also, do you have any input on this? I think it would be cool if we could have a minimal validator which draws upon the BIDS documentation functions and is thereby in sync with it by design.

tsalo:
>@tsalo so you'd like to flatten the datatype file hierarchy as well and have everything live under rules/ ?

@TheChymera I meant more that the subfolders would still exist, but for convenience, rather than as an actual source of information.

>Also, do you have any input on this? I think it would be cool if we could have a minimal validator which draws upon the BIDS documentation functions and is thereby in sync with it by design.

That sounds like a good idea, and the code itself looks solid. I think the actual discussion of the minimal validator should probably be a part of the schema meeting, but using similar patterns for rendering and regex construction seems like a solid plan.

tsalo:
Closing this in favor of #1006.

PR #986, opened 2022-01-23
[FIX] update definition acq_time for sessions.tsv

### LINKED ISSUE #948: `acq_time` for `sessions.tsv` ?

Remi-Gau:
fixes #948 

- [x] simplify `acq_time` definition by removing sections that are redundant with the Unit section
- [x] give different restrictions for scans.tsv and sessions.tsv

### DIFF
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -401,7 +401,7 @@ Such recordings MUST be documented with one row per file.
 {{ MACROS___make_columns_table(
    {
       "filename": ("REQUIRED", "There MUST be exactly one row for each file."),
-      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point in each run was acquired. Furthermore, if this header is provided, the acquisition times of all files that to a recording MUST be identical.")
+      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point in each run was acquired. Furthermore, if this header is provided, the acquisition times of all files that to a recording MUST be identical."),
    }
 ) }}
 
@@ -444,7 +444,7 @@ Column names in `sessions.tsv` files MUST be different from group level particip
 {{ MACROS___make_columns_table(
    {
       "session_id": ("REQUIRED", "There MUST be exactly one row for each session."),
-      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point of the first run was acquired.")
+      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point of the first run was acquired."),
       "pathology": "RECOMMENDED",
    }
 ) }}
--- a/src/03-modality-agnostic-files.md
+++ b/src/03-modality-agnostic-files.md
@@ -401,7 +401,7 @@ Such recordings MUST be documented with one row per file.
 {{ MACROS___make_columns_table(
    {
       "filename": ("REQUIRED", "There MUST be exactly one row for each file."),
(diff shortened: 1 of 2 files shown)


sappelhoff (approved):
Thanks! This is how I imagined it as well. 👍 Let's see what Taylor says about pulling schema description out of the schema and into the macro call.

src/03-modality-agnostic-files.md
```diff
@@ -401,7 +401,7 @@ Such recordings MUST be documented with one row per file.
 {{ MACROS___make_columns_table(
    {
       "filename": ("REQUIRED", "There MUST be exactly one row for each file."),
-      "acq_time": "OPTIONAL",
+      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point in each run was acquired. Furthermore, if this header is provided, the acquisition times of all files that to a recording MUST be identical."),
```
tsalo, line 404:
```suggestion
      "acq_time": ("OPTIONAL", "Acquisition time refers to when the first data point in each run was acquired. Furthermore, if this header is provided, the acquisition times of all files from the same recording MUST be identical."),
```

src/schema/objects/columns.yaml
```diff
@@ -12,23 +12,7 @@ abbreviation:
...
-    chosen (but consistent across all recordings) number of days.
-    This way relative timing would be preserved, but chances of identifying a person based
-    on the date and time of their scan would be decreased.
-    Dates that are shifted for anonymization purposes SHOULD be set to the year 1925 or earlier
-    to clearly distinguish them from unmodified data.
-    Note that some data formats do not support arbitrary recording dates.
-    For example, the [EDF](https://www.edfplus.info/)
-    data format can only contain recording dates after 1985.
-    Shifting dates is RECOMMENDED, but not required.
+  description: Datetime format and their anonymization are described in [Units](/02-common-principles.html#units).
```
tsalo, line 15:
```suggestion
  description: |
    The acquisition time for the first data point in each element of the file. 
    The datetime format and its anonymization are described in [Units](/02-common-principles.html#units).
```

tsalo (changes requested):
I'm not sure what the best way to describe this is, but I do think that the description needs to have meaningful information about the field, or else efforts like BIDS-Terms and our glossary won't be very effective. I almost think that it would be better to split the definition into multiple senses if there's no way to write a single interpretable description that only undergoes _minor_ modifications in the tables.

WDYT of just using `acc_time__sessions` and `acq_time__scans`?

src/schema/objects/columns.yaml
tsalo: Per today's maintainers call, splitting the definitions is probably the way to go, so you can ignore this suggestion.

Remi-Gau:
@tsalo I think this should  be done.

tsalo (approved):
LGTM!

sappelhoff:
Thanks @Remi-Gau

PR #987, opened 2022-01-23
[IGNORE] Sessions-level file draft

TheChymera:
Sessions file YAML as per https://github.com/bids-standard/bids-specification/issues/978 and videochat on Thursday 2022-01-20.


src/schema/rules/datatypes/tabular_metadata.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- suffixes:
+  - sessions
+  extensions:
+  - .tsv
+  - .json
+  entities:
+    subject: required
+    session: optional
```
tsalo, line 9: Just wanted to note that session actually should be prohibited for sessions files. I originally had `session: required` in #985 before I realized this.

TheChymera:
@tsalo Thank you, also, what's the best way to continue working on this, should I just drop this PR and somehow push to your branch? Or will both just be scrapped and a new one prepared after we discuss everything?

tsalo:
If you're comfortable pushing to my branch or opening a PR, I think that would be preferable since I incorporated scans files as well. Not sure if it will end up being merged, but I think it's moving in that direction if folks like the changes.

TheChymera:
@tsalo I'm certainly comfortable doing that, just that most of the content as it stands now (particularly in `src/schema/objects/suffixes.yaml`) is written by you, so maybe it's more appropriate if you authored the PR?

I think just submitting [the PR](https://github.com/bids-standard/bids-specification/pull/985) but without the `required` changes would be self-contained enough and probably easily discussed (maybe next Thursday meeting) and decided. If you don't want to do this at this point I can make a PR Wednesday evening so we have it ready on Thursday.

PR #989, opened 2022-01-25
[FIX] Optionally support echo entity for VFA suffix

### LINKED ISSUE #977: BIDS VFA anat datatype entry.
Is it an ommission that this variation of the `anat` datatype: https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/datatypes/anat.yaml#L147
does not list VFA as a possible suffix?

TheChymera:
@effigies this is pertaining to: https://github.com/bids-standard/bids-specification/issues/977#issuecomment-1020380516

@sappelhoff The PR text told me to ping you for questions. 
Should I add a `Closes:` line to the commit, will this automatically close the issue once merged?

Closes #977.

### DIFF
--- a/src/schema/rules/datatypes/anat.yaml
+++ b/src/schema/rules/datatypes/anat.yaml
@@ -105,6 +105,7 @@
     acquisition: optional
     ceagent: optional
     reconstruction: optional
+    echo: optional
     flip: required
     part: optional
 


effigies:
Added the closes line to the PR text.

effigies (approved):

TheChymera:
@effigies added

tsalo (approved):
The schema change looks good. My only question is, are there any changes/updates that should be made to the qMRI appendix page (https://bids-specification.readthedocs.io/en/latest/99-appendices/11-qmri.html)? For example, there's a table that links "file collection" suffixes with qMRI suffixes, and VFA is listed there. Should a new row be added with the appropriate applications for VFA acquisitions with multiple echoes?

TheChymera:
@tsalo I don't think this would be required at this point, and if it would be in the future, I presume the data would no longer be `VFA`, but something else.

> The tables in this appendix catalog applications where the use of a file collection is REQUIRED

`REQUIRED` doesn't really fit the bill.
The VFA specification indicates the collection is flip-angle relative, which remains the respective application. Multiple echoes can be added either denoising, or can selectively be varied for select flip angles to provide better signal — or, of course, for explorative analysis.

effigies (approved):

PR #991, opened 2022-01-27
[FIX] typo: extra sentence in anat section

Remi-Gau:
someone double check but I believe this sentence does not belong there


tsalo (approved):
LGTM

PR #996, opened 2022-02-01
[SCHEMA, FIX] Remove mentions of auxdatatypes

tsalo:
Closes None.

Changes proposed:

- Remove uses of the `auxdatatypes` keyword from macro calls, since we dropped the `auxdatatypes` "category" from the schema back in #609. For some reason the entity table functions were still using this keyword. Fortunately, it didn't have any effect on the actual entity tables, but it could be confusing for folks who look at the code.


Remi-Gau (approved):
yay!!! simpler API!!!

effigies (approved):

PR #998, opened 2022-02-03
[FIX] update highlighting of examples, JSON keys and values, and TSV headers or values in the schema

### LINKED ISSUE #746: Remaining formatting issues in tables
2 other issues I have noticed that make my inner OCD unhappy:

-  ~~inconsistency in the format of the table headers in the spec: most of the time it is in bold but not always~~
- numeric examples in the tables concerning the content of json files are not "highlighted" and might not be as obvious to the reader.

For example

```markdown
Number of EEG channels included in the recording (for example, 128).
```

Should probably be:

```markdown
Number of EEG channels included in the recording (for example, `128`).
```

- inconsistencies in the rendering of string and numeric values for tables (mostly for stings)

Sometimes examples are written:

```
like this
```

and other times

```
`like that`
```

| **Column name** | **Requirement level** | **Description**                                                                                                                                                               |
|-----------------|-----------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| name            | REQUIRED              | Channel name (for example, FC1, Cz)                                                                                                                                           |
| units           | REQUIRED              | Physical unit of the value represented in this channel, for example, `V` for Volt, or `fT/cm` for femto Tesla per centimeter (see [Units](../02-common-principles.md#units)). |

_Originally posted by @Remi-Gau in https://github.com/bids-standard/bids-specification/issues/739#issuecomment-782854311_

Remi-Gau:
closes #746 

make sure that that examples appear "as they should" be in JSON or TSV files

- [x] numeric examples appear for example like this ``1``
- [x] JSON keys or values when appearing in metadata descriptions appear with double quotes for example like `"Key"` or `"n/a"`
- [x] TSV column header or value appear without double quotes for example like `1` or like this `n/a`

I am sure I have missed some but this is a never ending battle (and I don't thing we want to automate or CI check this) so I suggest that this should close #746

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -15,13 +15,13 @@ acquisition:
     resolution.
     In such case two files could have the following names:
     `sub-01_acq-highres_T1w.nii.gz` and `sub-01_acq-lowres_T1w.nii.gz`, however
-    the user is free to choose any other label than highres and lowres as long
+    the user is free to choose any other label than `highres` and `lowres` as long
     as they are consistent across subjects and sessions.
     In case different sequences are used to record the same modality
-    (for example, RARE and FLASH for T1w)
+    (for example, `RARE` and `FLASH` for T1w)
     this field can also be used to make that distinction.
     At what level of detail to make the distinction (for example,
-    just between RARE and FLASH, or between RARE, FLASH, and FLASHsubsampled)
+    just between `RARE` and `FLASH`, or between `RARE`, `FLASH`, and `FLASHsubsampled`)
     remains at the discretion of the researcher.
   type: string
   format: label
@@ -32,7 +32,7 @@ ceagent:
     The `ce-<label>` key/value can be used to distinguish
     sequences using different contrast enhanced images.
     The label is the name of the contrast agent.
-    The key `ContrastBolusIngredient` MAY also be added in the JSON file,
+    The key `"ContrastBolusIngredient"` MAY also be added in the JSON file,
     with the same label.
   type: string
   format: label
@@ -82,9 +82,9 @@ echo:
     If files belonging to an entity-linked file collection are acquired at different
     echo times, the `_echo-<index>` key/value pair MUST be used to distinguish
     individual files.
-    This entity represents the `EchoTime` metadata field. Please note that the `<index>`
+    This entity represents the `"EchoTime"` metadata field. Please note that the `<index>`
     denotes the number/index (in the form of a nonnegative integer), not the
-    `EchoTime` value which needs to be stored in the field `EchoTime` of the separate
+    `"EchoTime"` value which needs to be stored in the field `"EchoTime"` of the separate
     JSON file.
   type: string
   format: index
@@ -95,9 +95,9 @@ flip:
     If files belonging to an entity-linked file collection are acquired at different
     flip angles, the `_flip-<index>` key/value pair MUST be used to distinguish
     individual files.
-    This entity represents the `FlipAngle` metadata field. Please note that the `<index>`
-    denotes the number/index (in the form of a nonnegative integer), not the `FlipAngle`
-    value which needs to be stored in the field `FlipAngle` of the separate JSON file.
+    This entity represents the `"FlipAngle"` metadata field. Please note that the `<index>`
+    denotes the number/index (in the form of a nonnegative integer), not the `"FlipAngle"`
+    value which needs to be stored in the field `"FlipAngle"` of the separate JSON file.
   type: string
   format: index
 hemisphere:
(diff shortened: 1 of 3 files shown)


src/schema/objects/entities.yaml
```diff
@@ -201,7 +201,7 @@ reconstruction:
   entity: rec
   description: |
     The `rec-<label>` key/value can be used to distinguish
-    different reconstruction algorithms (for example ones using motion
+    different reconstruction algorithms (for example `MoCo` for the ones using motion
```
Remi-Gau, line 204: This feels like common enough occurrence in scanners output to add it by name as an example

src/schema/objects/metadata.yaml
```diff
@@ -744,7 +744,7 @@ EffectiveEchoSpacing:
     between lines in the phase-encoding direction,
     defined based on the size of the reconstructed image in the phase direction.
     It is frequently, but incorrectly, referred to as "dwell time"
-    (see `DwellTime` parameter below for actual dwell time).
+    (see `"DwellTime"` parameter for actual dwell time).
```
Remi-Gau, line 747: removed the `below` because this might not make sense in all context if this description was used in a table where `DwellTime` does not appear

sappelhoff (approved):

tsalo (commented):
Overall, I like these changes. Just two things:

1. Including a leading underscore when referring to suffixes/specific files (e.g., `_events.tsv`) seems less readable to me.
2. I'm not sure if you want to include quotes around fields when you're referring to the actual contents of the field, rather than the key. For example, in the `RepetitionTimeExcitation` description, we have:

>Use `RepetitionTimeExcitation` (in combination with `"RepetitionTimePreparation"` if needed)

It seems like `"RepetitionTimePreparation"` should just be `RepetitionTimePreparation` in this case. That said, I don't feel strongly about this, so if you disagree then I'm happy to approve.

src/schema/objects/metadata.yaml
```diff
@@ -758,7 +758,7 @@ ElectricalStimulation:
   description: |
     Boolean field to specify if electrical stimulation was done during the
     recording (options are `true` or `false`). Parameters for event-like
-    stimulation should be specified in the events.tsv file.
+    stimulation should be specified in the `_events.tsv` file.
```
Remi-Gau, line 761:
```suggestion
    stimulation should be specified in the `events.tsv` file.
```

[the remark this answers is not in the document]
Remi-Gau: because I got carried away! :wink:

Remi-Gau:
> Overall, I like these changes. Just two things:
> 
>     1. Including a leading underscore when referring to suffixes/specific files (e.g., `_events.tsv`) seems less readable to me.
> 
>     2. I'm not sure if you want to include quotes around fields when you're referring to the actual contents of the field, rather than the key. For example, in the `RepetitionTimeExcitation` description, we have:
> 
> 
> > Use `RepetitionTimeExcitation` (in combination with `"RepetitionTimePreparation"` if needed)
> 
> It seems like `"RepetitionTimePreparation"` should just be `RepetitionTimePreparation` in this case. That said, I don't feel strongly about this, so if you disagree then I'm happy to approve.

Yeah In a way I agree but this adds another implicit rule about formatting that will be hard to keep track off and potentially confuse users, no?

tsalo (approved):
That's a fair point. Approved!

src/schema/objects/metadata.yaml
```diff
@@ -1734,11 +1734,11 @@ NumberOfVolumesDiscardedByScanner:
     before saving the imaging file.
     For example, a sequence that automatically discards the first 4 volumes
     before saving would have this field as 4.
-    A sequence that doesn't discard dummy scans would have this set to 0.
-    Please note that the onsets recorded in the `events.tsv` file should always
+    A sequence that does not discard dummy scans would have this set to 0.
+    Please note that the onsets recorded in the `_events.tsv` file should always
```
sappelhoff, line 1738:
```suggestion
    Please note that the onsets recorded in the `events.tsv` file should always
```

src/schema/objects/metadata.yaml
```diff
@@ -1747,10 +1747,10 @@ NumberOfVolumesDiscardedByUser:
     Number of volumes ("dummy scans") discarded by the user before including the
     file in the dataset.
     If possible, including all of the volumes is strongly recommended.
-    Please note that the onsets recorded in the `events.tsv` file should always
+    Please note that the onsets recorded in the `_events.tsv` file should always
```
sappelhoff, line 1750:
```suggestion
    Please note that the onsets recorded in the `events.tsv` file should always
```

sappelhoff (commented):
it seems like you removed the leading underscores again based on a comment by Taylor. I think you missed two instances (see comments).

I also like `events.tsv` better than `_events.tsv` even though I know I've been using both in the past

Remi-Gau:
> it seems like you removed the leading underscores again based on a comment by Taylor. I think you missed two instances (see comments).
> 
> I also like `events.tsv` better than `_events.tsv` even though I know I've been using both in the past

Thanks for catching those too!!

sappelhoff (approved):

PR #1000, opened 2022-02-04
[ENH] Microscopy: Add IntendedFor JSON field to photo files

### LINKED ISSUE #992: Microscopy files of suffix `_photo` do not support a `_run-` entity.
This seems like an artificial limitation, while the usefulness of doing this is indeed debatable, multiple runs can always come with multiple photos. Should I just sumbit a PR?

mariehbourget:
closes #992 

Dear BIDS community,

This PR address the issue in:

- #992

... where there was no mechanism in microscopy to associated a photo of a sample to a specific run or modality (for example: a photo could be generated at each run). Our best option seems to be a JSON file associated to the photo with the `IntendedFor` metadata field, which I added in this PR.

**Link to the rendered doc: https://bids-specification--1000.org.readthedocs.build/en/1000/04-modality-specific-files/10-microscopy.html#photos-of-the-samples-_photoextension**

Please let me know your thoughts and suggestions, thank you!

Note: If we go with this, we will need to update the bids-validator with the addition of the JSON file for photo and its associated rules.

- validator: https://github.com/bids-standard/bids-validator/pull/1415
- examples: https://github.com/bids-standard/bids-examples/pull/307


src/04-modality-specific-files/10-microscopy.md
```diff
@@ -421,14 +421,38 @@ For example:
...
+         "micr": {
+            "sub-01_ses-01_sample-01_acq-1_photo.jpg": "",
+            "sub-01_ses_01_sample-01_acq-2_photo.jpg": "",
+            },
          },
       }
    }
 ) }}
 
+Photo data MAY be accompanied by a JSON file containing the `IntendedFor` field that link
```
Remi-Gau, line 434: Should we mention that when there is no JSON or not `IntendedFor` it means that the photo applies to all images ?

mariehbourget: I could add "In case the photos do not correspond to any particular images, it does not have to be filled." in the description of `IntendedFor` in the table, similar to [fieldmap](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#using-intendedfor-metadata).
I prefer "doesn't apply to a particular image" which is less restrictive than "applies to all images".

> when there is no JSON or not IntendedFor

Just to clarify, the JSON file for photo only contains the `IntendedFor` field.

mariehbourget: Updated in [124bad7](https://github.com/bids-standard/bids-specification/pull/1000/commits/124bad7ea805ea32bc6a55fe2b2796b6f9045eba)

satra:
while we are making changes to the json document, would there be some benefit to add the possibility of some field that describes details of acquisition. 

examples:
- one can take a photo before and after clearing
- block-face image and then slice laid out image

at the simplest level an optional `description` field can be added.

cc: @Leekamentsky @gmazzamuto

LeeKamentsky:
This is kind of long-range, but it might be nice to place the photo in time with relation to the tissue protocol - somehow linking it to between two stages or after a stage. @satra is right about taking the photo before or after clearing - a huge difference in appearance and the intention of the photographer in picking the timing of the photo. I can imagine taking a picture of one of our slabs on the microscope stage would help us understand the orientation of the resulting image and other aspects.

mariehbourget:
Great idea, thanks! I added a "PhotoDescription" field with example.

mariehbourget:
I opened WIP PR for bids-examples (https://github.com/bids-standard/bids-examples/pull/307) and bids-validator (https://github.com/bids-standard/bids-validator/pull/1415) associated with the changes.

sappelhoff (approved):

gmazzamuto:
That's seems a good idea! indeed we usually take photos of the samples before and after clearing. These were also uploaded to DANDI

sappelhoff (approved):
the other modalities where "_photo" is used as a suffix file apparently don't (yet?) allow .json sidecars, but I see no reason why it shouldn't be the case.

LGTM

sappelhoff:
@TheChymera I wanted to request you as a reviewer and then noticed that you are not yet part of the org, so I've sent you some invitations for that. Please check your inbox :-)

Remi-Gau (approved):
LGTM

satra (approved):

PR #1002, opened 2022-02-04
SCHEMA: Encode metadata tables as in schema

effigies:
This is currently the first two tables of MRI encoded as schema. The idea is that a data file will have its context, and we will then walk through this document, top-to-bottom, constructing requirement levels. The first two tables apply to all files in the MRI modality. The remaining tables provides additional or updated requirement levels based on some criteria that may apply to the file (datatype == perf), dataset (PET in modalities) or values in the sidecar file (MTState == True).

Since this is not a prerequisite for constructing the sidecar information, these should all work. I have noted in the YAML that I have made two explicit assumptions about the needed logic to encode all rules; these may be invalidated by further exploration.

### DIFF
--- a/src/schema/rules/sidecars.yml
+++ b/src/schema/rules/sidecars.yml
@@ -1,0 +1,103 @@
+#
+# Groups of related metadata fields
+#
+# Assumptions: never need disjunction of selectors
+# Assumptions: top-to-bottom overrides is sufficient logic
+
+
+---
+MRIScannerHardware:
+    selectors:
+        - modality == "MRI"
+    fields:
+        Manufacturer:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0008, 0070 `Manufacturer`.
+        ManufacturersModelName:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0008, 1090 `Manufacturers Model Name`.
+        DeviceSerialNumber:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0018, 1000 `DeviceSerialNumber`.
+        StationName:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0008, 1010 `Station Name`.
+        SoftwareVersions:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0018, 1020 `Software Versions`.
+        HardcopyDeviceSoftwareVersion: DEPRECATED
+        MagneticFieldStrength:
+            level: RECOMMENDED, but REQUIRED for Arterial Spin Labeling
+        ReceiveCoilName: RECOMMENDED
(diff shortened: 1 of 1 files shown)


src/schema/rules/sidecars.yml
```diff
@@ -0,0 +1,103 @@
...
+        SpoilingState: RECOMMENDED
+        SpoilingType: OPTIONAL
+        SpoilingRFPhaseIncrement: OPTIONAL
+        SpoilingGradientMoment: OPTIONAL
+        SpoilingGradientDuration: OPTIONAL
+
+PETMRISequenceSpecifics:
+    selectors:
+        - modality == "MRI"
+        - "PET" in dataset.modalities
```
nellh, line 65: Maybe this would be more general for selectors?

```yaml
selectors:
  - field: modality
    equals: "MRI"
  - field: "dataset.modalities"
    in: "PET"
```

nellh: Here's another example from Kubernetes yaml config parser with some interesting structure for this.

```yaml
selector:
  - {key: modality, operator: Eq, values: "MRI"}
  - {key: "dataset.modalities", operator: In, values: "PET"}
```

src/schema/rules/sidecars.yml
```diff
@@ -0,0 +1,103 @@
...
+            addendum: Corresponds to DICOM Tag 0018, 1000 `DeviceSerialNumber`.
+        StationName:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0008, 1010 `Station Name`.
+        SoftwareVersions:
+            level: RECOMMENDED
+            addendum: Corresponds to DICOM Tag 0018, 1020 `Software Versions`.
+        HardcopyDeviceSoftwareVersion: DEPRECATED
+        MagneticFieldStrength:
+            level: RECOMMENDED, but REQUIRED for Arterial Spin Labeling
```
tsalo, line 30: What about a `level_addendum` field for the non-interpretable portion of this?

src/schema/rules/sidecars.yml
effigies: I'm inclined to stick with a simple language if possible; as we work through all the tables, we'll see how complex they get.

bendhouseart: I agree w/ @effigies for now that the `selectors` are written in something straight forward that that will say work w/ `eval( <selector 1 >)` in JS. It clearly (to this reader) captures the desired state and can always be updated to something more parse-able like the examples @nellh provided.

tsalo:
@effigies could you change the target to a different branch so it's easier for everyone working on #1014 to target the same branch?

effigies:
@tsalo Feel free to move this commit into a branch on bids-standard. I don't have time to attend to this right now and don't want to hold you back.

PR #1005, opened 2022-02-09
[FIX] Change recording entity to REQUIRED for pet/blood modality

ghisvail:
To make it consistent with the [rules](https://github.com/bids-standard/bids-specification/blob/e47283a268f20b98d5c05ad1bb8c63da3d2a1ceb/src/schema/rules/datatypes/pet.yaml#L30) and the [examples](https://github.com/bids-standard/bids-examples/tree/master/pet003/sub-01/ses-01/pet)

### DIFF
--- a/src/04-modality-specific-files/09-positron-emission-tomography.md
+++ b/src/04-modality-specific-files/09-positron-emission-tomography.md
@@ -315,7 +315,7 @@ If collected, blood measurements of radioactivity are be stored in
 [Tabular files](../02-common-principles.md#tabular-files) and located in
 the `pet/` directory along with the corresponding PET data.
 
-The OPTIONAL `recording` entity is used to distinguish sampling methods.
+The REQUIRED `recording` entity is used to distinguish sampling methods.
 For example, if an autosampler is used to record continuous blood samples,
 and manual measurements are also taken,
 then the files may have recording labels `autosampler` and `manual`,
--- a/src/04-modality-specific-files/09-positron-emission-tomography.md
+++ b/src/04-modality-specific-files/09-positron-emission-tomography.md
@@ -320,7 +320,7 @@ For example, if an autosampler is used to record continuous blood samples,
 and manual measurements are also taken,
 then the files may have recording labels `autosampler` and `manual`,
 respectively.
-if the sampling method is unknown, then recording is assumed `manual`.
+If the sampling method is unknown, then recording SHOULD be set as `manual`.
 All blood measurements should be reported according to a single time-scale
 in relation to time zero defined by the PET data (Figure 1).
 All definitions used below are in accordance with
--- a/src/04-modality-specific-files/09-positron-emission-tomography.md
+++ b/src/04-modality-specific-files/09-positron-emission-tomography.md
@@ -320,8 +320,7 @@ For example, if an autosampler is used to record continuous blood samples,
 and manual measurements are also taken,
 then the files may have recording labels `autosampler` and `manual`,
 respectively.
-If multiple recording methods are used on the same PET acquisition,
-the `recording` entity MUST be used to distinguish them.
+if the sampling method is unknown, then recording is assumed `manual`.
 All blood measurements should be reported according to a single time-scale
 in relation to time zero defined by the PET data (Figure 1).
 All definitions used below are in accordance with


ghisvail:
Not sure what the circleci thing is about.

mnoergaard:
Hi @ghisvail! Thanks for raising this issue. It states in the PET-BIDS spec that the recording entity is optional, unless several recordings are used (i.e. manual and autosampler).  So the example case you are referring to is not incorrect, but the recording entity could in principle be left out, and then the type of recording should be stated in the corresponding json file. I am somewhat hesitant to support your proposed change, because sometimes only the data after combining manual and autosamples is available (+ after additional modeling), and then it would be unclear what entity this data should have, as it is now a mixture between manual and autosamples. In this case, the entity could be left out (because it’s optional) and then described more thoroughly in the json. I know this is actually now derived data after combining the blood recordings, but I have seen cases where only this information is available. I would like not to block the sharing of this data. Let me know what you think of these potential concerns.

bendhouseart:
The intent is for the `recording` entity to provide clarification if there are multiple blood recording files (or to simply provide more specificity). It looks like inverse is what needs to happen such that the rules yaml needs to be changed from `required` to `optional` [here](https://github.com/bids-standard/bids-specification/blob/5e2c32a90dc5130e19c546bfee5a1abe1324137e/src/schema/rules/datatypes/pet.yaml#L30).  

Would that be satisfactory? Or do you feel there still needs to be an update or additional example added to the [examples](https://github.com/bids-standard/bids-examples/tree/master/pet003/sub-01/ses-01/pet)?

ghisvail:
The lack of consistency between the generated template in the docs, the wording of the paragraph and the example I linked to really makes the interpretation of this part of the spec quite confusing. Assuming the recording entity is indeed optional, then we should go with @bendhouseart suggestion to fix the rules generating the template and the inconsistent example I linked.

mnoergaard:
The example you are referring to won't be inconsistent independent of whether the recording entity is required or optional. But after thinking more about this I am inclined to keep the recording entity required, as it provides clarity on the recording type, which is important for the person modelling the blood data and may not be present in the corresponding json file. My previous argument was indeed for derived blood data, so this is considerations to take into account for the PET derivatives.

mnoergaard (approved):

mnoergaard:
Again, thanks for raising this issue @ghisvail.

ghisvail:
You're welcome. Thank you all for taking time reviewing it and providing the additional pieces of context I was missing.

melanieganz (commented):
Following the discussion above, I am fine with changing it to be required. The only question for me is what we do if people "only" have the merged (auto plus manual samples) left, e.g. for legacy data. Yes, this would in principle be a derivative, but I wonder if that's not in some cases the only raw data left. Do we anticipate this to become an issue?

mathesong:
This feels like the 80/20 rule ought to take prevalence here: I think it's going to be a very small minority of PET datasets for which the auto / manual nature of the blood samples is no longer known.  I would support the change to make the recording type required.

From a purely pragmatic perspective, I've dealt with data where I've not had access to which samples were auto and manual, and one just has to treat everything as if it's manual data. I think that's quite an uncontroversial take (though of course you could do something fancy with guessing which are which, but it would be need to be tailored for the specifics of the dataset). So one strategy might be to just include in the text that if the auto/manual nature of the data is unknown, that everything can just be labelled as manual?

ghisvail:
> So one strategy might be to just include in the text that if the auto/manual nature of the data is unknown, that everything can just be labelled as manual?

I'd be happy to add the proposed change to this PR. Is everyone in favour for it?

mnoergaard:
Thanks for your take on this @mathesong - I agree with you. @ghisvail please feel free to add the proposed change to this PR. Thank you.

ghisvail:
> Thanks for your take on this @mathesong - I agree with you. @ghisvail please feel free to add the proposed change to this PR. Thank you.

Done :+1:

src/04-modality-specific-files/09-positron-emission-tomography.md
```diff
@@ -315,13 +315,12 @@ If collected, blood measurements of radioactivity are be stored in
...
 
-The OPTIONAL `recording` entity is used to distinguish sampling methods.
+The REQUIRED `recording` entity is used to distinguish sampling methods.
 For example, if an autosampler is used to record continuous blood samples,
 and manual measurements are also taken,
 then the files may have recording labels `autosampler` and `manual`,
 respectively.
-If multiple recording methods are used on the same PET acquisition,
-the `recording` entity MUST be used to distinguish them.
+if the sampling method is unknown, then recording is assumed `manual`.
```
tsalo, line 323:
```suggestion
If the sampling method is unknown, then recording SHOULD be set as `manual`.
```

Hopefully I'm interpreting this correctly. Saying that recording is "assumed `manual`" implies, to me at least, that the recording entity would be absent and the sampling method would be assumed to be manual, much like how MRI files without the `part` entity are assumed to reflect magnitude values.

tsalo (changes requested):
I think the second edited sentence needs some tweaking. I do agree that this PR should be merged soon though.

tsalo (approved):
LGTM!

sappelhoff:
Thanks everyone :+1: @mnoergaard could you briefly confirm that the validator does not need to be adjusted due to the changes here?

mnoergaard:
LGTM - thanks @sappelhoff

PR #1006, opened 2022-02-10
[SCHEMA] Tabular metadata rules for both top and second-level tabular files

TheChymera:
* Representation of second-level files added.
* Homogenization of modality and non-modality file representation.
* Documenting file identifiers such as README and _sessions as
  suffixes..

Co-authored-by: Taylor Salo <tsalo006@fiu.edu>

----

As per https://github.com/bids-standard/bids-specification/pull/987#issuecomment-1027163553

@tsalo @yarikoptic

### DIFF
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -617,6 +617,33 @@ sbref:
   name: Single-band reference image
   description: |
     Single-band reference for one or more multi-band `dwi` images.
+scans:
+  name: Scans file
+  description: |
+    The purpose of this file is to describe timing and other properties of each imaging acquisition
+    sequence (each run file) within one session.
+    Each neural recording file SHOULD be described by exactly one row. Some recordings consist of
+    multiple parts, that span several files, for example through echo-, part-, or split- entities.
+    Such recordings MUST be documented with one row per file.
+    Relative paths to files should be used under a compulsory filename header.
+    If acquisition time is included it should be listed under the acq_time header.
+    Acquisition time refers to when the first data point in each run was acquired.
+    Furthermore, if this header is provided, the acquisition times of all files that belong to a
+    recording MUST be identical.
+    Datetime should be expressed as described in Units.
+    Additional fields can include external behavioral measures relevant to the scan.
+    For example vigilance questionnaire score administered after a resting state scan.
+    All such included additional fields SHOULD be documented in an accompanying _scans.json file
+    that describes these fields in detail (see Tabular files).
+sessions:
+  name: Sessions file
+  description: |
+    In case of multiple sessions there is an option of adding additional sessions.tsv files
+    describing variables changing between sessions.
+    In such case one file per participant SHOULD be added.
+    These files MUST include a session_id column and describe each session by one and only one row.
+    Column names in sessions.tsv files MUST be different from group level participant key column
+    names in the participants.tsv file.
 stim:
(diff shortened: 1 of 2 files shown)


TheChymera:
Test failure seems to be some website which isn't available and was not introduced by this PR. Might even be temporary, but apparently I don't have permission to re-run the tests.

src/schema/objects/suffixes.yaml
```diff
@@ -482,6 +505,10 @@ coordsystem:
   description: |
     A JSON document specifying the coordinate system(s) used for the MEG, EEG,
     head localization coils, and anatomical landmarks.
+dataset_description:
```
tsalo, line 508:
```suggestion
# This is a special top-level file, which is why an underscore is allowed in the "suffix".
dataset_description:
```

I'm thinking we could just add comments to this one and `genetic_info` so folks don't think they can include underscores in new suffixes.

TheChymera: Yeah, I really liked this idea, but as per the prevalent opinion in the discussion today apparently we should just strike it down as a suffix :(

Maybe we can bring the suffixification discussion back later after we get a foot in the door with this ^^

tsalo: Agreed!

TheChymera:
@tsalo @yarikoptic

I have incorporated the amendments from last week's meeting which reduced the scope of “suffixification”. This seems like a fairly simple self-contained addition of files previously not covered by the YAML which closes https://github.com/bids-standard/bids-specification/issues/978 .

So maybe we could discuss whether this can just be pulled today?

One small addition which I'd like to mention is that now we have temporarily given up on homogenizing the naming specification for all files.
Thus we have just created a representation for 2 new files which behave analogously to `rules/datatypes/*.yaml` without changing the representations of other non-modality files.
Perhaps it would be better, after all, to put the new `tabular_metadata.yaml` file in `rules/datatypes/tabular_metadata.yaml`, as [previously suggested](https://github.com/bids-standard/bids-specification/pull/985#issuecomment-1019573787)?
This would also make parsing of the schema into regex or placeholder example paths easier and compatible with current functions.

effigies:
FWIW I approve of moving participants and samples into this group, but also okay with postponing it.

effigies (approved):

tsalo (approved):
LGTM. Thanks!

src/schema/rules/tabular_metadata.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
yarikoptic, line 1: @effigies  we already have

```
❯ grep -A12 -E '^(scans|sessions)' src/schema/rules/files/common/tables.yaml
scans:
  level: optional
  suffixes:
    - scans
  extensions:
    - .tsv
    - .json
  entities:
    subject: required
    session: optional # session is required if session is present in the dataset.
sessions: # This file may only exist if session is present in the dataset.
  level: optional
  suffixes:
    - sessions
  extensions:
    - .tsv
    - .json
  entities:
    subject: required
```

seemingly with identical content (just dict not list).

How this file is used "in addition"?

yarikoptic: Sorry for the noise @effigies ! my bad -- how the hack I got to review this ? ;)    codecov woke up 2 years later and I thought that this PR got just merged, so I got all confused

PR #1007, opened 2022-02-10
BF: Use wikipedia for TIFF URL, adobe's page is 404ing now

yarikoptic:
Attn was brought up in https://github.com/bids-standard/bids-specification/pull/1006#issuecomment-1034936048
by @TheChymera

I think wikipedia would be the most reliable and informative source on TIFF format.

Attn @bids-standard/bep031


effigies (approved):

mariehbourget (approved):

yarikoptic:
IIRC 3 approvals is "enough" and don't remember if need to wait for someone to merge.  Since change is trivial and could green up other PRs, I will use my super power of pressing a green button. Thanks everyone for speedy reaction!

PR #1012, opened 2022-02-17
[SCHEMA] Add datatypes field to datatype rule files

tsalo:
Closes None.

Changes proposed:
- Add a datatypes field to the datatype rule files, so that the folder structure in the rules folder can be for readability, rather than something that implies the schema's structure.

What we won't tackle in this PR, but which is important, is that the schema code needs to be updated to use the new `datatypes` field _instead of_ the directory/filename structure that is currently used to organize the schema.


src/schema/rules/datatypes/meg.yaml
```diff
@@ -60,8 +66,10 @@
 - suffixes:
   - headshape
   extensions:
-  - .*
+  - "*"
```
effigies, line 69: I think this was a merge conflict resolution? Seems out of scope.

```suggestion
  - .*
```

effigies:
Plan on merging this with or without schemacode changes to make the field available to tooling. Note that merging as is will mean that we can't take advantage of the reorganization potential until there are schemacode changes.

src/schema/rules/datatypes/meg.yaml
tsalo: Ah, yes. I think I just merged it badly.

PR #1013, opened 2022-02-17
ENH: introduce versioning of the schema itself

yarikoptic:
Note -- it is not a version of the content of the schema, it is
a version over organization of the schema (e.g. we could add some
schema to describe how schema is organized).

The 'desire' for having a version over the BIDS schema's version
is somewhat described in README.md in added section.  Just to facilitate
tools to know what versions of bids schema they can read without explicitly
whitelisting BIDS versions themselves.  Might become more relevant as more
versions start to appear in https://github.com/bids-standard/bids-schema

I will feel totally Ok if we just drop this PR for now entirely.


tsalo:
I think this is a great idea. From what I can tell, we have four concurrent versions we need to track in this repository:
1. the specification, which is handled outside of the schema
2. the schema structure
3. the schema contents
4. the schemacode version, which is tracked with versioneer I think

Does that sound right? And this would track the schema _structure_, correct?

yarikoptic:
> Does that sound right? 

pretty much.  I feel that 1 and 3 are coming together, ie. I don't really see a need to separate out specification as what is outside or inside schema in respect of versioning.

code, as long as it is not distributed as a usable thing outside of this repo, is probably not worth its own versioning.  But we should aim indeed to move good portion of the code outside (and thus it would get its own versioning), and leave only the code needed for specification rendering etc.

> And this would track the schema _structure_, correct?

correct

yarikoptic:
attn @bids-standard/schema -- I am starting the campaign to decide on the destiny of adding versioning of schema's schema!  Please review/approve/request changes.

sappelhoff (commented):
sounds good to me, but I'd include a sentence on how the versioning changes after the `0.` series (i.e., starting with `1.0.0`)

src/schema/README.md
```diff
@@ -498,3 +498,17 @@ In cases where there is a data file and a metadata file, the `.json` extension f
...
+version (`MAJOR.MINOR.PATCH`) for the schema (how it is
+organized). Note that while in `0.` series, breaking changes are
+permitted without changing the `MAJOR` (leading) component of the
+version.  Going forward, the 2nd, `MINOR` indicator should be
+incremented whenever schema organization introduces "breaking
+changes": changes which would cause existing tools reading schema to
+adjust their code to be able to read it again.  Additions of new
+components to the schema should increment the last, `PATCH`, component
+of the version so that tools could selectively enable/disable loading
+specific components of the schema.
```
yarikoptic, line 514:
```suggestion
specific components of the schema. With the release of `1.0.0` version of the schema, we expect that
`MAJOR` component will be incremented whenever schema organization introduces "breaking
changes", `MINOR` - when adding new components to the schema, and `PATCH` - when fixing
errors in existing components.
```

effigies:
@yarikoptic @sappelhoff @tsalo Our plan this week is to continue work in separate schema branches (kept synced with changes in master). By the end of the week, we will very likely have sufficient changes to justify bumping the version in some way, so I would suggest finalizing this and merging before the end of the week so that we can have a clear break in pre/post workshop.

sappelhoff (approved):
> I'd say that the exact approach would be worked out based on our experience with versioning the 0. series, but let's indeed verbalize what we think it would be -- proposing in https://github.com/bids-standard/bids-specification/pull/1013/files#r838758576 -- looks ok'ish?

agreed, thanks for your addition. This lgtm

yarikoptic:
> @yarikoptic @sappelhoff @tsalo Our plan this week is to continue work in separate schema branches (kept synced with changes in master). By the end of the week, we will very likely have sufficient changes to justify bumping the version in some way, so I would suggest finalizing this and merging before the end of the week so that we can have a clear break in pre/post workshop.

can only wholeheartedly support the "Merge this PR" initiative!  I force pushed rebase to resolve conflicts (if I got it right due to "folder" -> "directory", so I used "directory" too)

src/schema/README.md
```diff
@@ -498,3 +498,18 @@ In cases where there is a data file and a metadata file, the `.json` extension f
...
+>>>>>>> fb273908 (ENH: introduce versioning of the schema itself)
```
tsalo, line 515: It looks like there's still a merge issue.

yarikoptic: eh, a sloppy rebase from a man on vacation near where you have been! Thank you @tsalo! force pushed the fix

src/schema/README.md
```diff
@@ -498,3 +498,17 @@ In cases where there is a data file and a metadata file, the `.json` extension f
...
+version (`MAJOR.MINOR.PATCH`) for the schema (how it is
+organized). Note that while in `0.` series, breaking changes are
+permitted without changing the `MAJOR` (leading) component of the
+version.  Going forward, the 2nd, `MINOR` indicator should be
+incremented whenever schema organization introduces "breaking
+changes": changes which would cause existing tools reading schema to
+adjust their code to be able to read it again.  Additions of new
+components to the schema should increment the last, `PATCH`, component
+of the version so that tools could selectively enable/disable loading
+specific components of the schema.
```
sappelhoff, line 514:
```suggestion
specific components of the schema.
With the release of `1.0.0` version of the schema,
we expect that the `MAJOR` component
will be incremented whenever schema organization introduces "breaking changes",
`MINOR` - when adding new components to the schema,
and `PATCH` - when fixing errors in existing components.
```

src/schema/README.md
```diff
@@ -498,3 +498,22 @@ In cases where there is a data file and a metadata file, the `.json` extension f
...
```
sappelhoff, line 514:
```suggestion
File `SCHEMA_VERSION` in the top of the directory contains a semantic
version (`MAJOR.MINOR.PATCH`) for the schema (how it is organized).
Note that while in `0.` series, breaking changes are
permitted without changing the `MAJOR` (leading) component of the version.
Going forward, the 2nd, `MINOR` indicator should be
incremented whenever schema organization introduces "breaking changes":
changes which would cause existing tools reading schema to
adjust their code to be able to read it again.
Additions of new components to the schema should increment the last,
`PATCH`, component of the version so that tools could selectively
enable/disable loading specific components of the schema.
```

Issue #1014, opened 2022-02-17
[SCHEMA] Metadata table checklist


effigies:
* Modalities
  * [ ] MRI (Started in #1002)
  * [ ] MEG
  * [ ] EEG
  * [ ] iEEG
  * [ ] Task events
  * [ ] Physiology/stimuli
  * [ ] Behavioral
  * [ ] PET
  * [ ] Microscopy
* Derivatives
  * [ ] Common
  * [ ] Imaging
* [ ] Appendices
  * [ ] MEG systems
  * [ ] Coordinate systems
  * [ ] Quantitative MRI
  * [ ] Arterial Spin Labeling

PR #1017, opened 2022-02-24
[SCHEMA] Define YAML tables for MRI common metadata fields and anatomy data

tsalo:
References #1014 and #1002.

Changes proposed:

- Add YAML definitions for the remaining tables in the MRI page's common metadata fields and anatomy imaging data sections.

### DIFF
--- a/src/schema/rules/sidecars.yml
+++ b/src/schema/rules/sidecars.yml
@@ -227,3 +227,74 @@ MRIAnatomyCommonMetadataFields:
         ContrastBolusIngredient: OPTIONAL
         RepetitionTimeExcitation: OPTIONAL
         RepetitionTimePreparation: OPTIONAL
+
+# Entities
+
+EntitiesTaskMetadata:
+    selectors:
+        - entities.task != "n/a"
+    fields:
+        TaskName: RECOMMENDED
+
+EntitiesCeMetadata:
+    selectors:
+        - entities.ce != "n/a"
+    fields:
+        ContrastBolusIngredient: OPTIONAL
+
+EntitiesTrcMetadata:
+    selectors:
+        - entities.trc != "n/a"
+    fields:
+        TracerName: REQUIRED
+
+EntitiesStainMetadata:
+    selectors:
+        - entities.stain != "n/a"
+    fields:
+        SampleStaining: RECOMMENDED
+        SamplePrimaryAntibodies: RECOMMENDED
+        SampleSecondaryAntibodies: RECOMMENDED
+
+EntitiesEchoMetadata:
+    selectors:
+        - entities.echo != "n/a"
+    fields:
+        EchoTime: REQUIRED
+
+EntitiesFlipMetadata:
+    selectors:
+        - entities.flip != "n/a"
+    fields:
+        FlipAngle: REQUIRED
+
+EntitiesInvMetadata:
+    selectors:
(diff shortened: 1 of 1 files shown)


src/schema/rules/sidecars.yml
```diff
@@ -61,21 +63,21 @@ MRISequenceSpecifics:
 
 PETMRISequenceSpecifics:
     selectors:
-        - modality == "MRI"
-        - "PET" in dataset.modalities
```
tsalo, line 65: Having quotes around PET causes `syntax error: expected <block end>, but found '<scalar>' (syntax)`

src/schema/rules/sidecars.yml
```diff
@@ -85,19 +87,215 @@ MTParameters:
...
+MRIAnatomyCommonMetadataFields:
+    selectors:
+    - modality == "MRI"
+    - datatype == "anat"
+    fields:
+        ContrastBolusIngredient: OPTIONAL
+        RepetitionTimeExcitation: OPTIONAL
+        RepetitionTimePreparation: OPTIONAL
+
+# Entities
```
tsalo, line 232: I added these because they weren't covered in #1014, but I'm not sure if they should go in this file.

src/schema/rules/sidecars.yml
```diff
@@ -85,19 +87,215 @@ MTParameters:
...
+    - sidecar.MRAcquisitionType == "2D"
+    fields:
+        SliceTiming: REQUIRED
+
+# This is technically for sparse sequences only, but I don't know how to encode that.
+SliceTimingSparse:
+    selectors:
+    - modality == "MRI"
+    # NOTE: This is my idea for "field is defined", but I don't know if it's the best approach.
+    - sidecar.DelayTime != "n/a"
```
tsalo, line 168: From today's call, we can create a keyword for "is defined".

src/schema/rules/sidecars.yml
```diff
@@ -85,19 +87,215 @@ MTParameters:
...
+    - entities.mt != "n/a"
+    fields:
+        MTState: REQUIRED
+
+EntitiesPartMetadata:
+    selectors:
+    - entities.part == "phase"
+    fields:
+        # Can we require Units to be either "rad" or "arbitrary" here?
+        Units: REQUIRED
```
tsalo, line 289: From Ross:

```suggestion
        Units:
          level: REQUIRED
          rules:
          - value in ["rad", "arbitrary"]
```

src/schema/rules/sidecars.yml
```diff
@@ -61,21 +63,21 @@ MRISequenceSpecifics:
 
 PETMRISequenceSpecifics:
     selectors:
-        - modality == "MRI"
-        - "PET" in dataset.modalities
+    - modality == "MRI"
+    - PET in dataset.modalities
```
effigies, line 67: What about the inverse?

```suggestion
    - dataset.modalities contains "PET"
```

tsalo: That works for me.

src/schema/rules/sidecars.yml
effigies: This feels like it should be its own file, but it does need to be highest priority.

Only issue I'm seeing is that these will correspond to data files, and not, say, an events.tsv.

PR #1018, opened 2022-02-28
[SCHEMA] Tests for render listings

TheChymera:
As discussed in last week's meeting.

Turns out this is a bit more tricky than I thought, because the outputs are large blocks of text which are hard to test w/o reproducing the code that created them or hard-coding output (which will automatically “break” whenever the schema is meaningfully changed — so of limited use, really, and a bit of a chore to re-write for schema contributors).

I tried to hard-code only small stable bits, and come up with some sanity/consistency checks.
Better than nothing, and it will definitely fail if the functions are *badly* broken, but really doesn't cover all the possible failures of the functions...

Another approach would be to assert line number or checksums — that would be less of a chore to update, but still of limited use since they would need to be updated very frequently.

In any case it's moderately useful as is for the addition of regex support which I wanted to do.
And testing regex consistency will in any case be easier than testing human-readable output. 

@tsalo — thoughts?

### DIFF
--- a/tools/schemacode/schemacode/tests/test_render.py
+++ b/tools/schemacode/schemacode/tests/test_render.py
@@ -5,6 +5,10 @@ from schemacode import render
 
 
 def test_make_entity_definitions(schema_obj):
+    """
+    Test whether expected format strings are present.
+    This should be stable with respect to schema format.
+    """
     schema_text = render.make_entity_definitions(schema_obj)
     expected_formats = [
         "Format: `sub-<label>`",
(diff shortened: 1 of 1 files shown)


TheChymera:
@effigies @bendhouseart thank you for looking into this whenever you have the time :)

tsalo (changes requested):
The tests look solid to me. I can definitely see what you mean about it being hard to write tests for these functions though. Validating long string outputs isn't straightforward.

The only thing I would recommend is adding docstrings to the test functions, to make it clear what is actually being tested. For example, `test_make_suffix_table` makes sure that a subset of suffixes and their "long" names appears in a rendered suffix table.

TheChymera:
@tsalo thank you for the review. I added docstrings as requested — including some commentary re robustness or fragility, which might become useful as the schema structure evolves. Should I go ahead and merge them into one?

tsalo:
It looks good to me, but we require two approvals before merging, so I'm requesting reviews from Chris and Remi.

effigies (approved):

PR #1023, opened 2022-03-04
[FIX] Add coordsystem-specific definition of DigitizedHeadPoints

### LINKED ISSUE #1022: MEG DigitizedHeadPoints definitions
For MEG, the DigitizedHeadPoints field exists in 2 different files, with 2 different definitions: 
In _meg.json it's a required boolean, while in _coordsystem.json, it's a relative path to a file.
However, in 1.7.0, both definitions now appear the same.  See [Digitized head points table in this section](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#coordinate-system-json-_coordsystemjson).

I think this is a bug due to how the tables are generated.

tsalo:
Closes #1022.

Changes proposed:

- Add definition for DigitizedHeadPoints for coordsystem files.


tsalo:
@Moo-Marc would you mind taking a look at the proposed changes?

Moo-Marc:
Thanks,
The contents looks good, though I'm not familiar with the syntax.

tsalo:
@Moo-Marc thanks for taking a look. Here is the updated definition, for posterity:

image

@sappelhoff do you know what kind of relative path `DigitizedHeadPoints` should be? I put `participant_relative`, like `IntendedFor`, but I don't know if that's correct. The definition is a little vague.

sappelhoff:
good question, I don't know :-( 

in this example, it's apparently relative to the coordsystem.json file itself (where the `DigitizedHeadPoints` metadata is specified): https://github.com/bids-standard/bids-examples/blob/df982afc719751d4b7f38d426aa38fc6832e06fa/ds000117/sub-01/ses-meg/meg/sub-01_ses-meg_coordsystem.json#L8

that looks super odd and is probably a bug. Another reason why we need #918 

Maybe @Moo-Marc knows? Or @robertoostenveld?

Moo-Marc:
I was a bit confused about paths as well and following the examples.  So for this I used the file name only (it's in the same folder).  I like the idea of https://github.com/bids-standard/bids-specification/pull/918.

robertoostenveld:
I suspect that back in the days of writing the MEG specification the idea was to have this similar as `IntendedFor` with a past specification that is relative to the JSON file that specifies it. So an example without explicit path like "sub-01_headshape.pos" would be in the same directory as the `coordinatesystem.json`

sappelhoff:
> I suspect that back in the days of writing the MEG specification the idea was to have this similar as IntendedFor with a past specification that is relative to the JSON file that specifies it.

Thanks Robert, I think you are right.

@tsalo would that mean that we need a new "format" for paths? Currently we just have `dataset_relative` and `participant_relative`:

https://github.com/bids-standard/bids-specification/blob/67c3d90dcb0bd49a81b84b1da05d82f28ed2912c/src/schema/README.md?plain=1#L219-L221

of course, all of these will be DEPRECATED anyhow once we merge #918 ... but we still need to solve this issue for backward compatibility

tsalo:
@sappelhoff I think you're right. Since it should be relative to the file in which the field is present, does a `file_relative` sound good?

sappelhoff:
> does file_relative sound good?

:+1: sounds good to me!

tsalo:
It's now set to `file_relative`.

sappelhoff (approved):
I think this is good to go then. Thanks a bunch @tsalo  (and @Moo-Marc for noticing)

PR #1029, opened 2022-03-17
[SCHEMA] Define YAML tables for DWI/ASL/(i)EEG/micr/beh metadata

tsalo:
Closes None, but is related to #1014.

Changes proposed:

- Add new `schema/rules/sidecars/dwi.yaml` file with metadata rules for DWI data.
- Add new `schema/rules/sidecars/asl.yaml` file with metadata rules for ASL data.
- Add new `schema/rules/sidecars/beh.yaml` file with metadata rules for behavioral (events/beh) data.
- Add new `schema/rules/sidecars/continuous.yaml` file with metadata rules for physiological and stimulus data.
- Add new `schema/rules/sidecars/eeg.yaml` file with metadata rules for EEG data.
- Add new `schema/rules/sidecars/ieeg.yaml` file with metadata rules for iEEG data.
- Add new `schema/rules/sidecars/micr.yaml` file with metadata rules for microscopy data.
- Split existing sidecars file into new `schema/rules/sidecars/mri.yaml`, `schema/rules/sidecars/anat.yaml`, and `schema/rules/sidecars/entities.yaml` files.


src/schema/rules/sidecars/dwi.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+---
+# Multipart (split) DWI schemes
+# NOTE: I don't think this can be schemafied, since it depends on owner intent.
+MRIDiffusionMultipart:
+    selectors:
+    - modality == "MRI"
+    - datatype == "dwi"
+    fields:
+        MultipartID: REQUIRED
```
tsalo, line 16: I'm not sure how we should handle MultipartID, since its requirement level depends on owner knowledge, rather than any validatable elements of the dataset.

Issue #1025, opened 2022-03-08
[ENH, BUG] BIDS Standard Template Coordinate Frame Definition


alexrockhill:
In trying to support the BIDS Standard Template coordinate frames in `mne-bids`, I've run into a few issues:

- Most importantly, what is the coordinate frame of standard template channel locations? The space is specified based on an atlas T1w-type image but it is not clear whether the channels are to be in Freesurfer `surface RAS`, `voxels`, `scanner RAS` or some other coordinate frame.
- Most of the links are fairly straightforward to find the atlas image but a few are not: `IXI549Space` (https://brain-development.org/ ) for instance only allows you to download all 549 T1 images, which is amazing data availability but not very helpful for defining the coordinate frame. `NIHPD`, on the other hand, has multiple allowable T1 images.

The idea with `mne-bids` is to read an electrophysiology image file in a template space and have it properly defined (some progress is here https://github.com/mne-tools/mne-bids/pull/982 but requires a lot of work for the user to find the fiducials of the atlas image when ideally that would just be done once). The issue is that we don't know in which coordinate frame the channels are stored, and, even if we did, we couldn't define the `head` coordinate frame used by MNE internally based on the nasion and left and right pre-auricular points without the T1 image defining the space since a few spaces are inaccessible or ambiguous. Ideally the coordinate frame (e.g. `voxels`) would be clearly specified and there would be a repository of T1 images and even more ideally their sidecar files with `AnatomicalLandmarkCoordiantes` in them.

Does it sound reasonable to insist that template coordinates be in `voxel` or `scanner RAS`?

And is there someone with a repository of all these atlases?

alexrockhill:
We could store the T1 atlases and fiducials in `mne-bids` but it would probably be better to store them this repository

alexrockhill:
Ok I went through all the spaces and I think I found a T1 for all of them but the only note is that the UNCInfant neonates, 1 year and 2 year olds are not affine registered so they have different scales. Not sure how much this matters for their use but it does make it difficult to define a single coordinate frame in terms of fiducial points.

alexrockhill:
Ok I did all the templates, I think they're right. They're too large for GitHub though. Maybe we'll just store the fiducials in `mne-bids` but it would be nice to have a repository of all the template T1s somewhere for people to be able to use them easier https://drive.google.com/drive/folders/1rN-Rb7D6ZTyVKOc2WZ-AM3f3dspIhZtQ?usp=sharing

[...]

PR #866, opened 2021-09-01
[ENH, MRG] Add "ScanRAS" as an accepted coordinate frame for ieeg

alexrockhill:
Fixes https://github.com/bids-standard/bids-specification/issues/747.

PR #1031, opened 2022-03-19
[FIX] Clarify that BIDS standard template data is to be in scanner coordinates (MEG, iEEG, EEG)

alexrockhill:
Fixes https://github.com/bids-standard/bids-specification/issues/1025 (most basic part).

### DIFF
--- a/src/99-appendices/08-coordinate-systems.md
+++ b/src/99-appendices/08-coordinate-systems.md
@@ -112,7 +112,7 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
 -   `4DBti`: ALS orientation and the origin between the ears
 -   `KitYokogawa`: ALS orientation and the origin between the ears
 -   `ChietiItab`: RAS orientation and the origin between the ears
--   Any keyword from the list of [Standard template identifiers](#standard-template-identifiers)
+-   Any keyword from the list of [Standard template identifiers](#standard-template-identifiers): RAS orientation and the origin at the center of the gradient coil for template NifTI images
 
 In the case that MEG was recorded simultaneously with EEG,
 the restricted keywords for
@@ -138,7 +138,7 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     For more information, see the
     [EEGLAB wiki page](https://eeglab.org/tutorials/ConceptsGuide/coordinateSystem.html#eeglab-hj-coordinate-system).
 
--   Any keyword from the list of [Standard template identifiers](#standard-template-identifiers)
+-   Any keyword from the list of [Standard template identifiers](#standard-template-identifiers): RAS orientation and the origin at the center of the gradient coil for template NifTI images
 
 In the case that EEG was recorded simultaneously with MEG,
 the restricted keywords for
@@ -182,7 +182,7 @@ Restricted keywords for the `<CoordSysType>CoordinateSystem` field in the
     is not aligned to ACPC, `ScanRAS` should be used.
 
 -   Any keyword from the list of
-    [Standard template identifiers](#standard-template-identifiers)
+    [Standard template identifiers](#standard-template-identifiers): RAS orientation and the origin at the center of the gradient coil for template NifTI images
 
 ## Image-based Coordinate Systems
 
@@ -194,11 +194,6 @@ Unless specified explicitly in the sidecar file in the
 `<CoordSysType>CoordinateUnits` field, the units are assumed to be mm.
 
 ### Standard template identifiers
-
-When using any of the below templates, coordinates should be in "world" (RAS+) space,
-and not voxel indices.
-If `<CoordSysType>CoordinateUnits` is not defined or applicable, then the units SHALL
-be assumed to be millimeters (mm).
 
 | **Coordinate System**              | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | **Used by**              | **Reference**                                                                                                                      |
 | ---------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ | ---------------------------------------------------------------------------------------------------------------------------------- |
--- a/src/99-appendices/08-coordinate-systems.md
+++ b/src/99-appendices/08-coordinate-systems.md
@@ -195,7 +195,10 @@ Unless specified explicitly in the sidecar file in the
 
 ### Standard template identifiers
 
-All BIDS standard template coordinate system data should be in ``scanner`` coordinates of the T1w template image with units corresponding to `<CoordSysType>CoordinateUnits`.
+When using any of the below templates, coordinates should be in "world" (RAS+) space,
+and not voxel indices.
+If `<CoordSysType>CoordinateUnits` is not defined or applicable, then the units SHALL
+be assumed to be millimeters (mm).
 
 | **Coordinate System**              | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | **Used by**              | **Reference**                                                                                                                      |
 | ---------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ | ---------------------------------------------------------------------------------------------------------------------------------- |
--- a/src/99-appendices/08-coordinate-systems.md
+++ b/src/99-appendices/08-coordinate-systems.md
@@ -37,7 +37,7 @@ units are expressed in mm.
 
 The coordinate systems below all relate to neuroscience and therefore to the
 head or brain coordinates.
-$Please be aware that all data acquisition starts with
+Please be aware that all data acquisition starts with
 "device coordinates" (scanner), which does not have to be identical to the
 initial "file format coordinates" (DICOM), which are again different from the
 "head" coordinates (for example, NIFTI).
@@ -194,6 +194,8 @@ Unless specified explicitly in the sidecar file in the
 `<CoordSysType>CoordinateUnits` field, the units are assumed to be mm.
 
 ### Standard template identifiers
+
+All BIDS standard template coordinate system data should be in ``scanner`` coordinates of the T1w template image with units corresponding to `<CoordSysType>CoordinateUnits`.
 
 | **Coordinate System**              | **Description**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | **Used by**              | **Reference**                                                                                                                      |
 | ---------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ | ---------------------------------------------------------------------------------------------------------------------------------- |


alexrockhill:
Not sure why the linkcheck failed, looks like it failed accessing the fieldtrip documentation...

oesteban (commented):
> Fixes #1025

After all the conversation on that issue, I'm confused about what the issue is. It seems that there is an issue, but it's not a BIDS-spec issue.

Why only T1s? That is a clear sign that the problem is very specific to your use-case, not a general problem with BIDS.

> maybe should consider storing the template T1s somewhere and linking them in the BIDS spec

I don't think BIDS should support this. As @satra mentioned, some BIDS-enthusiasts (including myself) are developing TemplateFlow. If what you need is a programmatic way of retrieving specific metadata associated with a bunch of atlases of interest, that is the IMHO best place to look. Let me know if you need to learn more about templateflow to determine whether it indeed would address your problem.

> I'm happy to use the Google Drive link I put up but it might not last forever.

Please, read every template's license first to make sure first you can actually do that, and how you can do that, if you were to do it independently of BIDS. Most of the times you will need to clearly honor attribution clauses, so having obscure links to GDrive doesn't seem a reasonable way of doing that.

alexrockhill:
> After all the conversation on that issue, I'm confused about what the issue is. It seems that there is an issue, but it's not a BIDS-spec issue.
> 
> Why only T1s? That is a clear sign that the problem is very specific to your use-case, not a general problem with BIDS.
>

In my opinion, providing links to websites that only sometimes contain a usable template definition such as a T1 is a problem. A T1 is nice because for software that supports reading BIDS data, this is a standardized way to check that the positions are reasonable.

> > maybe should consider storing the template T1s somewhere and linking them in the BIDS spec
> 
> I don't think BIDS should support this. As @satra mentioned, some BIDS-enthusiasts (including myself) are developing TemplateFlow. If what you need is a programmatic way of retrieving specific metadata associated with a bunch of atlases of interest, that is the IMHO best place to look. Let me know if you need to learn more about templateflow to determine whether it indeed would address your problem.

I think TemplateFlow is a great solution but it's not referenced in the BIDS specification so I think it's reasonable that many BIDS users would not know that it exists. It could be added to BIDS but since it doesn't have all the templates and they are not named the same, that doesn't really solve the problem, although I think that's a great step.

I think @robertoostenveld 's suggestion that we move template data to a derivative directory is the easiest and best fix which would make it less urgent that TemplateFlow have all the BIDS Standard Templates and have all the checks associated with that until more derivatives are checked by BIDS in general. I think it would be easy enough to just move template coordinate frames to derivatives as a follow-up PR to this.

> > I'm happy to use the Google Drive link I put up but it might not last forever.
> 
> Please, read every template's license first to make sure first you can actually do that, and how you can do that, if you were to do it independently of BIDS. Most of the times you will need to clearly honor attribution clauses, so having obscure links to GDrive doesn't seem a reasonable way of doing that.

I haven't checked the licenses and I agree that it isn't the ideal solution but I think having the templates all in one place for internal sharing to discuss them is helpful so that we're all talking about the same thing. I think that's a step for a future PR.

oesteban:
> I think @robertoostenveld 's suggestion that we move template data to a derivative directory

There's currently no specification for templates in BIDS. There's a specification for coordinate systems engendered by templates (yes, typically from T1w images but that's because most of the work has been done on human adults, otherwise it could well be T2w or other schemes) that were created in _standard spaces_. Please note that BIDS tries to map between _coordinates_ and _standard spaces_ intentionally skipping over templates. This is to say, there is nothing to move, IMHO.

I have heard of previous interest (including myself) in creating a specification for templates. I think it would make sense in derivatives. As the initial ideas didn't get any traction, we pushed towards TemplateFlow.

alexrockhill:
Hmm okay, I think we're almost on the same page. My issue is that in writing software that reads BIDS-formatted datasets, having channels in a template coordinate frame without the template itself to check that those channel locations make sense and to fully support data with template electrode coordinates is a problem in practice. It sounds like there is a pretty good consensus for moving the templates to derivatives which I can do as a followup PR.

This PR is much simpler, it's just that if coordinates are in terms of a standard template coordinate frame, they should be in RAS and not voxels which I think is consistent with other coordinate frames but not completely obvious because things like `AnatomicalLandmarkCoordinates` are store in voxels.

alexrockhill:
Ok, this is a good first step and happy to debate the best way to implement other, future PRs but LGTM on this one

oesteban:
> My issue is that in writing software that reads BIDS-formatted datasets, having channels in a template coordinate frame without the template itself to check that those channel locations make sense and to fully support data with template electrode coordinates is a problem in practice.

I remain unclear that this is a BIDS issue, and it's even less clear to me after #1025 (where this conversation should be had).

> It sounds like there is a pretty good consensus for moving the templates to derivatives which I can do as a followup PR.

Actually, the lack of clear consensus tempered previous initiatives to include templates into the BIDS specs. I can see more support if templates were part of derivatives.

That said, there's no point in saying _moving the templates to derivatives_ because BIDS, as it stands, does not make any prescriptions about templates. It does have COORDINATE systems engendered by templates, and to refer to those coordinate systems it actually names a bunch of templates. So, if someone wants to specify templates as part of BIDS-Derivatives, the effort would start from the ground up.


> This PR is much simpler, it's just that if coordinates are in terms of a standard template coordinate frame, they should be in RAS and not voxels which I think is consistent with other coordinate frames but not completely obvious because things like `AnatomicalLandmarkCoordinates` are stored in voxels.

Where does the spec say that `AnatomicalLandmarkCoordinates` must be stored in voxels? Because that is something we want to fix. However, I don't think this PR achieves that.

alexrockhill:
This PR is about specifying that BIDS Standard Template Coordinate Systems are in terms of scanner RAS, I think you're getting a bit off track @oesteban, although I'm happy to discuss future PRs on the referenced issue or on the PRs when they happen.

I assumed someone else that wrote the part of `mne-bids` that does `AnatomicalLandmarkCoordinates` had made sure that this was supposed to be in voxels but perhaps that's a mistake. I'll have to look into it.

alexrockhill:
> That said, there's no point in saying _moving the templates to derivatives_ because BIDS, as it stands, does not make any prescriptions about templates. It does have COORDINATE systems engendered by templates, and to refer to those coordinate systems it actually names a bunch of templates. So, if someone wants to specify templates as part of BIDS-Derivatives, the effort would start from the ground up.

To comment on this point though so that it doesn't go unaddressed, the idea would be not to allow ``iEEGCoordinateSystem``, ``EEGCoordinateSystem`` or ``MEGCoordinateSystem`` to be a BIDS Standard Template Coordinate System outside `derivatives`, but instead suggest that BIDS users put it in the same place in `derivatives` (e.g. `/users/alex/mne_data/ieeg_bids/sub-1/ieeg/sub-1_space-IXI549Space_coordsystem.json` would be suggested to be `/users/alex/mne_data/ieeg_bids/derivatives/sub-1/ieeg/sub-1_space-IXI549Space_coordsystem.json`).

alexrockhill:
Here is also where it says `AnatomicalLandmarkCoordinates` can be in voxels. Sense 1 and sense 2 seems a bit messy here https://bids-specification.readthedocs.io/en/stable/99-appendices/14-glossary.html#anatomicallandmarkcoordinates-sense-2-metadata.

alexrockhill:
Ok, just checking in on this PR once again, it's not a big change, just a super simple clarification that templates be in `scanner RAS`. Looking forward to future discussion of moving templates to derivatives etc. but not on this PR, it would be great to get this moving. cc @sappelhoff

sappelhoff:
@alexrockhill I just asked some @bids-standard/maintainers to please take a look at this. I am a little out of my depths here, especially with Oscar seeming to oppose these changes. Sorry I can't be of more help.

alexrockhill:
> @alexrockhill I just asked some @bids-standard/maintainers to please take a look at this. I am a little out of my depths here, especially with Oscar seeming to oppose these changes. Sorry I can't be of more help.

It wasn't clear to me whether he opposed the change in this PR or proposed future changes, I assume and infer from what he said more the latter although he hasn't approved this PR. I know it takes a long time and requires everyone to get on board for BIDS which requires a lot of patience and I'm not trying to rush things but this is a minuscule change that I don't really understand how it could be opposed since it's just clarifying something that was previously unspecified, which is a bit frustrating that it takes weeks to approve to be honest.

src/99-appendices/08-coordinate-systems.md
```diff
@@ -195,6 +195,8 @@ Unless specified explicitly in the sidecar file in the
 
 ### Standard template identifiers
 
+All BIDS standard template coordinate system data should be in ``scanner`` coordinates of the T1w template image with units corresponding to `<CoordSysType>CoordinateUnits`.
```
effigies, line 198:
```suggestion
When using any of the below templates, coordinates should be in "world" (RAS+) space,
and not voxel indices.
If `<CoordSysType>CoordinateUnits` is not defined or applicable, then the units SHALL
be assumed to be millimeters (mm).
```

effigies (commented):
I did not follow the entire discussion in #1025, but reading this PR in light of @oesteban's comment:

> Where does the spec say that AnatomicalLandmarkCoordinates must be stored in voxels? Because that is something we want to fix. However, I don't think this PR achieves that.

I think by "`scanner` coordinates" you mean in the RAS+ coordinates corresponding to the affine stored in some template NIfTI associated with this coordinate system, which presumably should be invariant to which specific template file is selected. I have tried to make that clearer.

I would also say that `*CoordinateUnits` doesn't apply to all use cases, so I suggest that we make the default interpretation of coordinates "mm". I don't think this would be overly restrictive if this list is ever expanded.

src/99-appendices/08-coordinate-systems.md
alexrockhill: I was copying off of `ScanRAS` above in same part of the specification:

```
ScanRAS: The origin of the coordinate system is the center of the gradient coil for the corresponding T1w image of the subject, and the x-axis increases left to right, the y-axis increases posterior to anterior and the z-axis increases inferior to superior. For more information see the [Nipy Documentation](https://nipy.org/nibabel/coordinate_systems.html). It is strongly encouraged to align the subject's T1w to ACPC so that the ACPC coordinate system can be used. If the subject's T1w in the BIDS dataset is not aligned to ACPC, ScanRAS should be used.
```

"world" might be more accurate in describing it but from what I've seen in the ieeg part of the spec, scanner has been used more often. Perhaps this should be harmonized.

alexrockhill: This is from the top of that appendix as well `Please be aware that all data acquisition starts with "device coordinates" (scanner)`

alexrockhill:
Thanks, changing `T1w` to NifTI is definitely more inclusive, I didn't realize that was the proposed fix. Also assuming `mm` seems reasonable.

effigies:
@oesteban Can you check to see if my suggestion addressed your concerns with this PR?

src/99-appendices/08-coordinate-systems.md
```diff
@@ -195,6 +195,11 @@ Unless specified explicitly in the sidecar file in the
 
 ### Standard template identifiers
 
+When using any of the below templates, coordinates should be in "world" (RAS+) space,
+and not voxel indices.
+If `<CoordSysType>CoordinateUnits` is not defined or applicable, then the units SHALL
+be assumed to be millimeters (mm).
```
oesteban, line 201: This is stated in ll. 193-194 above, I think.

effigies, line 201: Fair enough. I missed that this was a subheading of **Image-based Coordinate Systems**.

```suggestion
```

oesteban:
I don't oppose this PR - it's just I can't favor it and give it a green check if I don't understand the problem (and even less the solution). My understanding of #1025 is that there is a bottleneck implementing some utility in MNE relating to standard-space-generated coordinates systems. Not sure there is an issue with the BIDS spec.

That said, I don't think I understand this section of the BIDS specs at all. The conversations prior to the first drafts were slow, messy, and unclear, and at some point, I gave up. For instance, after reading this part of the spec, I have no idea of what `<CoordSysType>` in `<CoordSysType>CoordinateSystem` would be for an image, and whether it would be stored in the `coordsystem.json`, which seems only defined for (i)EEG. (That without entering into the weirdness of having a key name that can change depending on its semantics - surely a headache for the schema)

For instance, I can't follow the specs regarding having the origin at the ACPC for images and honestly believe the spec just makes unclear something that, by default, was prescribed to be solved by the inherent coordinate system defined by NIfTI (for the case of images).

@alexrockhill introduced `ScanRAS`, #866, (again, only under iEEG, which I don't get why either) and the reading feels like a repetition of the image-based coordinates found below.

To make it even harder, there's a previous modification (#717, [line 94 of the file](https://github.com/bids-standard/bids-specification/commit/f0391a399542fe82c685257c09d113d2734de751#diff-cf6acb04606f962a2fc6aba5e735783f7a686215246eb2ea6e9862394d78b1caR94)) making template-engendered coordinates another option for iEEG.

So, if this PR were to modify anything, I would expect it to modify just the `ScanRAS` definition which seems to be the culprit in the problem sketched in #1025. But, if three other people more acquainted with the specs on coordinate systems see this okay, I'm not going to oppose it, as I said above.

alexrockhill:
I agree that the iEEG specification is messy and needs some work. I think it should be done piecewise in small increments. The current problem addressed by this PR is how are you supposed to know that your template channel positions shouldn't be shared in voxel coordinates of a template NifTI? It doesn't say anywhere in the spec how they should be stored, just that they should be relative to the template. This clarifies that it should be in RAS+ world/scanner coordinates and not in voxels.

I don't really follow your other comments with regards to `ACPC` and `ScanRAS`, perhaps we should open a separate issue and work through those. Again, I would prefer not to discuss them here as I think tackling a large refactoring all at once is preventing any progress from being made.

oesteban:
> I agree that the iEEG specification is messy and needs some work.

I don't know enough to say this. I lean towards the images and templates specs doesn't make much sense.

> This clarifies that it should be in RAS+ world/scanner coordinates and not in voxels.

This is not true. It makes it more unclear. If the problem is with templates used with (i)EEG, this clarification should be somewhere under the item added in #717 regarding templates.

> It doesn't say anywhere in the spec how they should be stored, just that they should be relative to the template. 

Sure, and for some reason, instead of stating it where the specs say that they can be relative to a template, this PR modifies the section about templates itself. Which, as I was saying, is already unclear and unprecise enough.

> I don't really follow your other comments with regards to `ACPC` and `ScanRAS`, perhaps we should open a separate issue and work through those.

It doesn't really apply to my criticism of this PR, so we can just leave it there. I was just trying to say that I don't fully understand this file of the specification. That said, the ACPC origin and the `ScanRAS` keywords were given as separate examples and putting them together just makes everything more cloudy.

alexrockhill:
Ok, @oesteban, is this better?

oesteban:
> Ok, @oesteban, is this better?

Yes, in the sense that it doesn't make the templates section more unclear. Thanks for that.

For (i)EEG, MEG, I don't really know enough to say anything else - so I can't really say yay or nay here.

alexrockhill:
Okay, sounds good, seems like an improvement by me from the iEEG perspective

sappelhoff (commented):
it would be great to get @robertoostenveld's opinion on this revised proposal. From my side this looks fine, but we need an additional (i)MEEG person who has regularly worked with such data to agree as well.

sappelhoff (approved):

sappelhoff:
@dorahermes @robertoostenveld are you fine with this change?

robertoostenveld (approved):
I agree with the suggested change.

PR #1033, opened 2022-03-21
[ENH] Allow README file extensions.

### LINKED ISSUE #972: Allow extension for the README file
GitLab cannot render markdown if the file doesn't have an extension. Currently, BIDS Validator complains that there is no `README` file although `README.md` is there. I think this is due to the following rule:

https://github.com/bids-standard/bids-specification/blob/febae4fb9b2c9f055dc1392e0283c85f05618de0/src/schema/rules/top_level_files.yaml#L2:L5

Could we change that and support file extensions for the `README` file? I know it would require a few changes in the documentation and maybe the validator but if we could work out the list, I'm willing to work on it.

mateuszpawlik:
Closes  #972

Tasks to complete:

- [x] Modify the README description in documentation https://github.com/bids-standard/bids-specification/blob/master/src/03-modality-agnostic-files.md#readme
- [x] Modify the description in the schema https://github.com/bids-standard/bids-specification/blob/ba11771e76978264445cfc35d52eda14d1ca30fc/src/schema/objects/top_level_files.yaml#L24-L26
- [x] Add new file extensions https://github.com/bids-standard/bids-specification/blob/ba11771e76978264445cfc35d52eda14d1ca30fc/src/schema/rules/top_level_files.yaml#L5
- [x] https://github.com/bids-standard/bids-validator/issues/1431 - I think the validator should be updated before the specification changes are merged in. (PR here: https://github.com/bids-standard/bids-validator/pull/1432)

I kept the README description in the specification and in the schema equal. Maybe a macro could help not to need to update such definitions in two places.


src/03-modality-agnostic-files.md
```diff
@@ -143,7 +143,7 @@ Example:
 
 ### `README`
 
-Every BIDS dataset SHOULD come with a free form text file (`README`) describing the dataset in more detail.
+A REQUIRED free-form text file, `README` (with an OPTIONAL extension `.md`, `.rst`, or `.txt`), SHOULD describe the dataset in more detail.
```
sappelhoff, line 146: I think we should add a short snippet of text (with links) that these extensions refer to "markdown", "restructured text", and "generic text" (or similar) respectively, and explicitly state that BIDS does not make recommendations with regards to the "flavor" (e.g. Markdown flavor) and does not validate md/rst syntax.

Instead, I think we should RECOMMEND that READMEs are written from a mindset that somebody reading the file in an un-rendered state should still be able to perfectly understand everything.

more opinions by @effigies @yarikoptic perhaps?

mateuszpawlik: Shall I simply implement your comments or we're waiting for more feedback?

sappelhoff: Both are fair approaches, the prior might result in extra work, but also potentially easier/better reviews by the reviewers :-)

tsalo: This kind of makes me think we should add extensions to the schema, so that the description could just link to the glossary instead of describing each of the extensions itself.

Remi-Gau:
> I agree that we probably shouldn't update LICENSE or CHANGES given that they are more restrictive than README. If someone requests them in the future, we can revisit.

I 1000% agree with this

src/03-modality-agnostic-files.md
mateuszpawlik: I don't know what's the rule. Shall I rephrase the text now and add the links later or wait for the glossary to be ready? I could add the links now if I new how they'll be structured.

sappelhoff: I suggest you add a recommendations along the line of

> Instead, I think we should RECOMMEND that READMEs are written from a mindset that somebody reading the file in an un-rendered state should still be able to perfectly understand everything.

And then we can merge this PR (if other people agree) -- @tsalo (or me) can then add the links in a follow up PR (or in #1063 directly) once #1063 is merged.

src/schema/objects/top_level_files.yaml
```diff
@@ -21,9 +21,18 @@ LICENSE:
...
   name: README
   description: |
-    In addition a free form text file (`README`) describing the dataset in more details SHOULD be
-    provided.
-    The `README` file MUST be either in ASCII or UTF-8 encoding.
+    A REQUIRED free-form text file, `README`, SHOULD describe the dataset in more detail.
+    The `README` file MUST be either in ASCII or UTF-8 encoding and MAY have one of the extensions:
+    `.md` ([Markdown](https://www.markdownguide.org/)),
+    `.rst` ([reStructuredText](https://docutils.sourceforge.io/rst.html)),
+    or `.txt` (unstructured text).
```
yarikoptic, line 28: just to make sure - the version with absent extension (as in current BIDS) corresponds to `.txt` (unstructured text) format.  I wonder there is a value from making it explicit -- probably not, but thought to mention. WDYT?

yarikoptic (commented):
mostly just non-consequential comments, with one concern -- why not to restrict .md to basic?  I bet openneuro and other portals would prefer to render those .md files. If flavor is not regulated -- it would cause troubles, thus reducing the benefit from having .md to start with.

src/schema/objects/top_level_files.yaml
```diff
@@ -21,9 +21,18 @@ LICENSE:
 README:
   name: README
   description: |
-    In addition a free form text file (`README`) describing the dataset in more details SHOULD be
-    provided.
-    The `README` file MUST be either in ASCII or UTF-8 encoding.
+    A REQUIRED free-form text file, `README`, SHOULD describe the dataset in more detail.
```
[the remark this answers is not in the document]
mateuszpawlik: I also can't find a better term. It's free-form in terms of content, not format or syntax.

src/schema/objects/top_level_files.yaml
mateuszpawlik: From my experience, some software may not try to render a file without extension. Other than that, I personally don't see any advantage.

src/03-modality-agnostic-files.md
```diff
@@ -143,8 +143,16 @@ Example:
...
 
-Every BIDS dataset SHOULD come with a free form text file (`README`) describing the dataset in more detail.
-The `README` file MUST be either in ASCII or UTF-8 encoding.
+A REQUIRED free-form text file, `README`, SHOULD describe the dataset in more detail.
+The `README` file MUST be either in ASCII or UTF-8 encoding and MAY have one of the extensions:
+`.md` ([Markdown](https://www.markdownguide.org/)),
+`.rst` ([reStructuredText](https://docutils.sourceforge.io/rst.html)),
+or `.txt` (unstructured text).
+BIDS does not make any recommendations with regards to the
+[Markdown flavor](https://www.markdownguide.org/extended-syntax/#lightweight-markup-languages)
```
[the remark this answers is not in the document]
mateuszpawlik: I think the goal was not to validate the syntax. Therefore, no recommendation for the flavor. In my opinion basic would be too restrictive. For example tables and task lists are useful but not included in Basic syntax.

sappelhoff (approved):

src/schema/objects/top_level_files.yaml
sappelhoff: Some people may write in markdown syntax in a `README` file, so I wouldn't make it seem like the lack of an extension somehow means that the contents should be unstructured.

src/schema/objects/top_level_files.yaml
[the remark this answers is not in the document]
sappelhoff:
```suggestion
    A REQUIRED text file, `README`, SHOULD describe the dataset in more detail.
```

I suggest to remove "free-form" as it's ambiguous. A couple of sentences later, the text points to a guideline of what a good `README*` file should include.

src/03-modality-agnostic-files.md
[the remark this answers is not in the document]
sappelhoff: > I think the goal was not to validate the syntax.

indeed, this has been discussed in several threads and in two (or so) online meetings, notes can be found here:

- https://github.com/bids-standard/bids-validator/pull/1281#discussion_r630118909
- https://github.com/bids-standard/bids-specification/issues/972
- Maintainer Biweekly meeting notes (links to a specific comment therein): https://docs.google.com/document/d/1MuwZt6ToZmsPgH6naFVlPDFtNUz61DzlaZmdNHIAotg/edit?disco=AAAAUZdMt60

> In my opinion basic would be too restrictive. 

agreed. Regarding your concern:

1. People will write their README.md file with flavors such that it renders **at least**  on the platform(s) of their choice. For all other platforms, our recommendation kicks in: READMEs SHOULD be understandable/parsable even in unrendered form
2. READMEs will most likely not include very very advanced MD syntax, and most flavors have significant overlap, which furthermore reduces the problem

I therefore stand by the current proposal to not recommend any flavor.

sappelhoff:
before I forget it @mateuszpawlik could you please add yourself to the contributors list here? https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors

We sync that wiki entry with the corresponding page in the BIDS specification before every new release.

src/03-modality-agnostic-files.md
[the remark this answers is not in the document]
sappelhoff: 1. many people already use MD and have to put their README in .bidsignore to make their dataset valid
2. other people add a txt extension because they are used to having extensions, and are then surprised that this does not work
3. as I mention above: the currently proposed solution will work well for almost all cases: Those cases where people use basic markdown syntax and perhaps a markdown table ... for cases where the syntax cannot be properly rendered due to a flavor incompatibility between README and rendering host, we still have ensured the readability of the README (albeit slightly more ugly; but note: this will happen rarely)

Overall I believe that the current solution (no flavor recommendation, no validation) is a good tradeoff between utility gains for users and validation/specification burden for BIDS. 

> even in the minority I am, I still feel wrong to give up on this

Please feel free to tag more people if you want to discuss this more broadly -- I hope that we can find a consensus.

src/schema/objects/top_level_files.yaml
```diff
@@ -21,9 +21,18 @@ LICENSE:
...
   name: README
   description: |
-    In addition a free form text file (`README`) describing the dataset in more details SHOULD be
-    provided.
-    The `README` file MUST be either in ASCII or UTF-8 encoding.
+    A REQUIRED text file, `README`, SHOULD describe the dataset in more detail.
+    The `README` file MUST be either in ASCII or UTF-8 encoding and MAY have one of the extensions:
+    `.md` ([Markdown](https://www.markdownguide.org/)),
+    `.rst` ([reStructuredText](https://docutils.sourceforge.io/rst.html)),
+    or `.txt`.
```
yarikoptic, line 28: it is somewhat implied but may be worth explicit something like:
```suggestion
    or `.txt`.
    Only a single README file (with or without extension) should be present.
```
?

src/schema/rules/top_level_files.yaml
```diff
@@ -3,6 +3,9 @@ README:
   required: true
   extensions:
   - None
+  - .md
+  - .rst
+  - .txt
```
yarikoptic, line 8: FWIW -- inspired me to file https://github.com/bids-standard/bids-specification/issues/1047

src/schema/objects/top_level_files.yaml
sappelhoff: agreed, though I would phrase it:

> a BIDS dataset MUST NOT contain more than one README file (with or without extension) at its root directory.

mateuszpawlik: Fixed.

sappelhoff (approved):

Remi-Gau (approved):

sappelhoff:
Thanks for this contribution @mateuszpawlik!

mateuszpawlik:
Thank you all for making it happen. That was a nice experience :-) I'm looking forward to more.

PR #1046, opened 2022-03-29
[SCHEMA] Add a new `file_relative` format

tsalo:
Closes None, but is relevant for #1023.

Changes proposed:
- Add a `file_relative` format to `schema/objects/formats.yaml`.
- Add tests for the new format.


sappelhoff:
Thanks @tsalo

Issue #884, opened 2021-09-19
Use a coherent schema definition syntax


erdalkaraca:
Currently, the schema consists of a mix of ad-hoc definitions and syntax.
This makes it hard to implement a tool to generically validate a given BIDS dataset.

In the ancpbids project (see below for link) we experimented with various schema definition languages, such as:
- XSD (XML Schema Definition): industry standard with mature support in many programming languages
- YAMALE: a community effort to formalize data structures using YAML format

Since the BIDS community is more comfortable with JSON/YAML formats, I would like to propose using a more formal schema definition language such as YAMALE to map the BIDS specification onto a machine readable schema.

Examples to follow as additional comments...

Links:
ancpbids: https://github.com/ANCPLabOldenburg/ancp-bids
XSD: https://www.w3.org/TR/xmlschema11-1/
Yamale: https://github.com/23andMe/Yamale

erdalkaraca:
An example of how a coherent schema might look like using Yamale:

```yaml
Dataset:
  .extends: Folder
  subjects: list(include('Subject'), required=False)
  dataset_description: include('DatasetDescriptionFile')
  README: include('File', required=False)
  CHANGES: include('File', required=False)
  LICENSE: include('File', required=False)
  genetic_info: include('JsonFile', required=False)
  samples: include('JsonFile', required=False)
  participants_tsv: include('ParticipantsTsvFile', required=False)
  participants_json: include('TsvSidecarFile', required=False)
  code: include('Folder', required=False)
  derivatives: include('DerivativeFolder', required=False)
  sourcedata: include('Folder', required=False)
  stimuli: include('Folder', required=False)

---
JSONSchemaType:
  map()

MetadataFieldDefinition:
  name: str()
  description: str()
  type: include('JSONSchemaType')

SuffixDefinition:
  name: str()
  description: str()
  type: include('JSONSchemaType')

EntitiyDefinition:
  key: str()
  name: str()
  entity: str()
  description: str()
  type: include('JSONSchemaType')
```

erdalkaraca:
Example definitions of entities:

```yaml
- key: subject
  name: Subject
  entity: sub
  description: |
    A person or animal participating in the study.
  type: label
- key: run
  name: Run
  entity: run
  description: |
    If several scans with [...]
  type: index
- key: mtransfer
  name: Magnetization Transfer
  entity: mt
  description: |
    If files belonging [...]
  type:
    enum:
      - "on"
      - "off"
- key: part
[...]

PR #1048, opened 2022-03-30
[SCHEMA] Define the types present in the schema in a new YAML file

tsalo:
Closes None, but it _might_ be related to #884?

Changes proposed:

- Add a new `schema/objects/types.yaml` file with definitions of the valid "object types" (e.g., suffixes, metadata fields) within the schema. This gives each a name and description.
- Additionally, try to define valid fields for different object types.
    - For example, columns and metadata fields may have a number of fields describing the value types, such as "boolean" or "number", while associated data only have names and descriptions.

To do:

- [ ] I still need to figure out how to actually validate the schema based on the new rules.


src/schema/objects/types.yaml
```diff
@@ -0,0 +1,143 @@
...
+    A regular expression defining valid values of different types.
+  definition:
+    name: string
+    description: string
+    pattern: string
+metadata:
+  name: Metadata Field
+  description: |
+    A field that may be present in a BIDS sidecar JSON file.
+  definition:
```
tsalo, line 85: This is the most extensive "definition," so I think it's the best one to focus on as we determine if this structure is any good.

effigies:
Feels out of scope for the current sprint. We will not likely parse this in any first implementations, but might be worth revisiting once the schema is finalized.

src/schema/meta/types.yaml
```diff
@@ -0,0 +1,151 @@
...
+    name: string
+    display_name: string
+    description: string
+    type:
+    - string:
+        supplementary_fields:
+        - format
+        - unit
+        - enum
+        - pattern
```
tsalo, line 25: Should I define these and other fields in this file as well?

Issue #1047, opened 2022-03-30
schema: Define extensions exclusivity/composition? 


yarikoptic:
looking at https://github.com/bids-standard/bids-specification/pull/1033/files#diff-e1391ae7ff69f13355ee975c7fafc3414020f8ed7f3d26bfa92a4381429f51a0L4 and making a comment https://github.com/bids-standard/bids-specification/pull/1033/files#r838617368 where, as in many other places, we have "alternative" extensions for a file (so only one extension should be used) I also saw
```
participants:
  required: false
  extensions:
  - .tsv
  - .json
```
where it is allowed to have multiple or even "worse" -- having `.json` makes sense only if there is `.tsv` (unless for the inheritance we have .json on top level for some .tsv's down the hierarchy), and also looked up curious cases like
```
src/schema/rules/datatypes/anat.yaml:  extensions:
src/schema/rules/datatypes/anat.yaml-  - .nii.gz
src/schema/rules/datatypes/anat.yaml-  - .nii
src/schema/rules/datatypes/anat.yaml-  - .json
```
where, I think, it is not allowed to have both `.nii.gz` and `.nii` (while similarly to above ok to accompany with `.json`).

I wondered on how we could encode all that in the schema?  initial thinking about separating into a dedicated `sidecar_extensions`, so would look like

```
src/schema/rules/datatypes/anat.yaml:  extensions:
src/schema/rules/datatypes/anat.yaml-  - .nii.gz
src/schema/rules/datatypes/anat.yaml-  - .nii
src/schema/rules/datatypes/anat.yaml-  sidecar_extensions:
src/schema/rules/datatypes/anat.yaml-  - .json
```
and thus making it implied that for `extensions` -- it is "one of" and then `sidecar_extensions` is those which could accompany, iff a file with `extensions:` is available.

WDYT (attn @bids-standard/schema -- the Team I just initiated, feel welcome to invite/add more people if I forgot anyone)?

effigies:
It could be worth abstracting the idea of sidecar that could apply to any file. BEP-027 is proposing that any file could have a .prov.jsonld sidecar.

yarikoptic:
> It could be worth abstracting the idea of sidecar that could apply to any file. BEP-027 is proposing that any file could have a .prov.jsonld sidecar.

good idea IMHO!

> Are there any cases where a non-JSON data file _can't_ have a sidecar JSON file?

<details>
[...]

PR #1049, opened 2022-03-31
ENH: allow for .png and .tif in eeg/ieeg/meg as allowed for micr

yarikoptic:
Discovered while working on https://github.com/bids-standard/bids-specification/issues/1047

I guess it could be argued against but I feel that such unification makes sense.  There might be other common suffixes where we might want to introduce similar unification. 

- [ ] validator support implemented: https://github.com/bids-standard/bids-validator/pull/1443

### DIFF
--- a/src/schema/rules/datatypes/eeg.yaml
+++ b/src/schema/rules/datatypes/eeg.yaml
@@ -66,6 +66,8 @@
   - photo
   extensions:
   - .jpg
+  - .png
+  - .tif
   entities:
     subject: required
     session: optional
--- a/src/schema/rules/datatypes/ieeg.yaml
+++ b/src/schema/rules/datatypes/ieeg.yaml
@@ -67,6 +67,8 @@
   - photo
   extensions:
   - .jpg
+  - .png
+  - .tif
   entities:
     subject: required
     session: optional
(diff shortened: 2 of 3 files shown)


sappelhoff (commented):
I don't really remember why this was restricted to JPG to begin with. Probably to limit the amount of possible extensions ... as we also did with e.g., EEG file formats.

I don't think the suggestion here would hurt anyone.

effigies:
Might be good to check in with all the *EG BEP leads.

robertoostenveld (approved):
I think this makes sense, also for consistency.

Remi-Gau (approved):
makes sense to me too

sappelhoff (commented):
we'd need to open an associated PR to the validator to expand the regex rules, for example here: https://github.com/bids-standard/bids-validator/blob/9c27673c4f35a2b364d09f381149dc34f6c3e12c/bids-validator/bids_validator/rules/file_level_rules.json#L380

sappelhoff (approved):

effigies:
Looks like lots of approval and no dissent. I agree that there doesn't seem to be much benefit in keeping this restricted, as all proposed formats are well-supported (though TIFF is more a container than a codec, I think we can assume people aren't doing really wild things with it).

effigies:
15 days since last change, 7 days since 2 approvals, no objections since then. Merging.

PR #1054, opened 2022-04-07
[SCHEMA] Add MEG, func, and fmap metadata tables

tsalo:
Related to #1014.

Metadata tables for the task fMRI, MEG, and fmap sections.

### DIFF
--- a/src/schema/rules/sidecars/func.yaml
+++ b/src/schema/rules/sidecars/func.yaml
@@ -40,6 +40,7 @@ MRIFuncTimingParameters:
         DelayAfterTrigger: RECOMMENDED
 
 # The mutual exclusion table, spread across 5 definitions
+# NOTE: This introduces a PROHIBITED level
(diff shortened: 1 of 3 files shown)


src/schema/rules/sidecars/func.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+MRIFuncTimingParametersMutualExclusion2:
+    selectors:
+    - modality == "MRI"
+    - datatype == "func"
+    - sidecar contains "SliceTiming"
+    - sidecar contains "VolumeTiming"
+    fields:
+        RepetitionTime: PROHIBITED
+        DelayTime: PROHIBITED
```
tsalo, line 60: This is probably worth discussing. New PROHIBITED option.

tsalo: Actually I'll just add a note to the HackMD.

PR #1059, opened 2022-04-12
Nift header schema

rwblair:
Defines the shape of nifti headers so that we can be explicit about what fields are ok to call in selector or check.

### DIFF
--- a/src/schema/rules/nifti_header.yaml
+++ b/src/schema/rules/nifti_header.yaml
@@ -1,0 +1,39 @@
+- nifti_header:
+  dim_info:
+    freq:
+        type: int
+    phase:
+        type: int
+    slice:
+        type: int
+  dim:
+    type: array
+    minItems: 8
+    maxItems: 8
+    items:
+        type: int
+   pixdim:
+      type: array
+      minItems: 8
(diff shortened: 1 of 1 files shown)


src/schema/rules/nifti_header.yaml
```diff
@@ -0,0 +1,40 @@
+---
+- nifti_header:
+  dim_info:
+    freq:
+      type: int
+    phase:
+      type: int
+    slice:
+      type: int
```
effigies, line 9: Maybe helpful hints for implementors; should not try to make these parseable.

```suggestion
    freq:
      type: int
      # value = dim_info & 0x03
    phase:
      type: int
      # value = (dim_info >> 2) & 0x03
    slice:
      type: int
      # value = (dim_info >> 4) & 0x03
```

src/schema/rules/nifti_header.yaml
```diff
@@ -0,0 +1,40 @@
+---
+nifti_header:
```
tsalo, line 2: Should these all be defined in the objects folder instead? Perhaps we need a new `schema/objects/data_file_headers.yaml` or something.

PR #1061, opened 2022-04-12
SCHEMA: Initial shot at EEG channels and electrodes

effigies:
New ideas:

* `initial_columns` - a subset of columns defining an order; this assumes that you never have, e.g., the second column fixed while the first is variable
* Overriding term enums to restrict options
* `unknown_column_metadata` - indicates whether additional columns MUST have definitions in a sidecar JSON

### DIFF
--- a/src/schema/rules/tabular_data/eeg.yaml
+++ b/src/schema/rules/tabular_data/eeg.yaml
@@ -1,0 +1,55 @@
+---
+EEGChannels:
+    selectors:
+    - modality == "eeg"
+    - suffix == "channels"
+    - extension == ".tsv"
+    initial_columns: [name__channels, type, units]
+    columns:
+        name__channels: REQUIRED
+        type__channels:
+            level: REQUIRED
+            enum:
+            - AUDIO
+            - EEG
+            - EOG
+            - ECG
+            - EMG
+            - EYEGAZE
+            - GSR
+            - HEOG
+            - MISC
+            - PPG
+            - PUPIL
+            - REF
+            - RESP
+            - SYSCLOCK
+            - TEMP
+            - TRIG
+            - VEOG
+        units: REQUIRED
+        description: OPTIONAL
+        sampling_frequency: OPTIONAL
+        reference: OPTIONAL
+        low_cutoff: OPTIONAL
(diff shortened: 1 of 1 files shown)


src/schema/rules/tabular_data/eeg.yaml
```diff
@@ -0,0 +1,55 @@
+---
+EEGChannels:
+    selectors:
+    - modality == "eeg"
+    - suffix == "channels"
+    - extension == ".tsv"
+    initial_columns: [name__channels, type, units]
```
tsalo, line 7:
```suggestion
    initial_columns: [name__channels, type__channels, units]
```

src/schema/rules/tabular_data/eeg.yaml
```diff
@@ -0,0 +1,55 @@
...
+            - HEOG
+            - MISC
+            - PPG
+            - PUPIL
+            - REF
+            - RESP
+            - SYSCLOCK
+            - TEMP
+            - TRIG
+            - VEOG
```
tsalo, line 29: I think that defining a separate object for each modality might be better here.

tsalo: E.g., `type__channels_eeg`.

src/schema/rules/tabular_data/perf.yaml
```diff
@@ -0,0 +1,8 @@
+---
+M0Scan:
+    selectors:
+    - datatype == "perf"
+    - suffix == "m0scan"
```
effigies, line 5:
```suggestion
ASLContext:
    selectors:
    - datatype == "perf"
    - suffix == "aslcontext"
```

src/schema/objects/columns.yaml
```diff
@@ -466,7 +466,31 @@ trigger:
     continuous measurement of the scanner trigger signal
   type: number
 # type column in channels.tsv files
-type__channels:
+type__eeg_channels:
```
tsalo, line 469: The column tables that used `type__channels` will need to be updated as well.

tsalo (approved):
This looks good to me.

PR #1062, opened 2022-04-12
Schema for blood tsv

rwblair:


src/schema/rules/tabular_data/pet.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - sidecar.PlasmaAvail
+    columns:
+        plasma_radioactivity: REQUIRED
+
+BloodMetabolite:
+    selectors:
+    - modality == "pet"
+    - suffix == "blood"
+    - extension == ".tsv"
+    - sidecar.MetaboliteAvail
```
tsalo, line 29: Is this checking if `MetaboliteAvail` is true or just defined?

rwblair: Intent was to check its value.

PR #1063, opened 2022-04-13
[SCHEMA] Correlating existing validator issues to schema entries

bendhouseart:
Adding existing code's from [Validator Issues](https://docs.google.com/spreadsheets/d/181vtw2bt2nHcUgw__G_gFSkKGyuxOxUNr5wHrZXY_uI/edit#gid=0) to entities in schema where `REQUIRED if` is defined.

e.g.
```yaml
# where entity once looked like the following
MRIASLCommonMetadataFieldsM0Type:
    selectors:
    - modality == "MRI"
    - datatype == "perf"
    - suffix == "asl"
    - sidecar.M0Type == "Estimate"
    fields:
        M0Estimate: REQUIRED
```

It now contains a reference to the existing code in the validator:

```yaml
MRIASLCommonMetadataFieldsM0Type:
    selectors:
    - modality == "MRI"
    - datatype == "perf"
    - suffix == "asl"
    - sidecar.M0Type == "Estimate"
    fields:
        M0Estimate:
            level: REQUIRED
            code:
                M0ESTIMATE_NOT_DEFINED: |
                You must define 'M0Estimate' for this file, in case 'M0Type' is defined as 'Estimate'. 
                'M0Estimate' is a single numerical whole-brain M0 value (referring to the M0 of blood), 
                only if obtained externally (for example retrieved from CSF in a separate measurement).
```


src/schema/rules/sidecars/asl.yaml
```diff
@@ -31,7 +31,23 @@ MRIASLTextOnlySliceTiming:
     - suffix == "asl"
     - sidecar.MRAcquisitionType == "2D"
     fields:
-        SliceTiming: REQUIRED
+        SliceTiming:
+            level: REQUIRED
+            code:
+                SLICE_TIMING_NOT_DEFINED_2D_ASL: |
+                    "You should define SliceTiming', in case 'SequenceType' is set to a
```
tsalo, line 38: I realize it would be more verbose, but I kind of prefer something like
```
code:
    name: SLICE_TIMING_NOT_DEFINED_2D_ASL
    message: |
        ....
```

bendhouseart: Yeah, I was going to ask about that. My thoughts were just to collect the code keys and the message would follow (or be empty). I'm happy as long as the code name and message are contained w/ in the code key as there's nothing wrong with being verbose/explicit. @ericearl can you take a look at this and potentially update your code entries to what @tsalo suggested.

src/schema/rules/sidecars/asl.yaml
```diff
@@ -31,7 +31,24 @@ MRIASLTextOnlySliceTiming:
...
+                    is not slice order -- rather, it is a list of times containing the
+                    time (in seconds) of each slice acquisition in relation to the beginning
+                    of volume acquisition. The list goes through the slices along the slice
+                    axis in the slice encoding dimension (see below). Note that to ensure the
+                    proper interpretation of the `SliceTiming` field, it is important to check
+                    if the OPTIONAL `SliceEncodingDirection` exists. In particular, if
+                    `SliceEncodingDirection` is negative, the entries in `SliceTiming` are
+                    defined in reverse order with respect to the slice axis, such that the
+                    final entry in the `SliceTiming` list is the time of acquisition of slice 0.
+                    Without this parameter slice time correction will not be possible."
```
effigies, line 51:
```suggestion
            code: SLICE_TIMING_NOT_DEFINED_2D_ASL
            message: |
                You should define SliceTiming', in case 'SequenceType' is set to a
                2D sequence. 'SliceTiming' is the time at which each slice was
                acquired within each volume (frame) of the acquisition. Slice timing
                is not slice order -- rather, it is a list of times containing the
                time (in seconds) of each slice acquisition in relation to the beginning
                of volume acquisition. The list goes through the slices along the slice
                axis in the slice encoding dimension (see below). Note that to ensure the
                proper interpretation of the `SliceTiming` field, it is important to check
                if the OPTIONAL `SliceEncodingDirection` exists. In particular, if
                `SliceEncodingDirection` is negative, the entries in `SliceTiming` are
                defined in reverse order with respect to the slice axis, such that the
                final entry in the `SliceTiming` list is the time of acquisition of slice 0.
                Without this parameter slice time correction will not be possible.
```

src/schema/rules/sidecars/asl.yaml
```diff
@@ -31,7 +31,24 @@ MRIASLTextOnlySliceTiming:
     - suffix == "asl"
     - sidecar.MRAcquisitionType == "2D"
     fields:
-        SliceTiming: REQUIRED
+        SliceTiming:
+            level: REQUIRED
+            code:
+                name: SLICE_TIMING_NOT_DEFINED_2D_ASL
```
effigies, line 37:
```suggestion
            issue:
                code: SLICE_TIMING_NOT_DEFINED_2D_ASL
```

bendhouseart: 👍

src/schema/rules/sidecars/asl.yaml
bendhouseart: NO!

src/schema/rules/sidecars/asl.yaml
bendhouseart: All fixed.

effigies (approved):

PR #1064, opened 2022-04-13
[RTM] SCHEMA: Encode requirement levels for dataset_description and genetic_info

effigies:
Two open questions I see:

1) How do we want to encode, e.g. NO_AUTHORS, where we effectively promote dataset_description.Authors from OPTIONAL to REQUIRED?
2) Do we want to specify nested fields, such as `Genetics.Database` directly? Or have some way of specifying a nested structure as its own object?

### DIFF
--- a/src/schema/rules/dataset_metadata.yaml
+++ b/src/schema/rules/dataset_metadata.yaml
@@ -6,7 +6,7 @@ dataset_description:
         BIDSVersion: REQUIRED
         DatasetType: RECOMMENDED
         License: RECOMMENDED
-        Authors: OPTIONAL
+        Authors: RECOMMENDED
         Acknowledgements: OPTIONAL
         HowToAcknowledge: OPTIONAL
         Funding: OPTIONAL
--- a/src/schema/rules/dataset_metadata.yaml
+++ b/src/schema/rules/dataset_metadata.yaml
@@ -26,9 +26,9 @@ dataset_description:
     GeneratedBy[].Description: OPTIONAL
     GeneratedBy[].CodeURL: OPTIONAL
     GeneratedBy[].Container: OPTIONAL
-    GeneratedBy[].Container.Type: RECOMMENDED
-    GeneratedBy[].Container.Tag: RECOMMENDED
-    GeneratedBy[].Container.URI: RECOMMENDED
+    GeneratedBy[].Container{}.Type: RECOMMENDED
+    GeneratedBy[].Container{}.Tag: RECOMMENDED
+    GeneratedBy[].Container{}.URI: RECOMMENDED
     SourceDatasets: RECOMMENDED
 
 derivative_description:
@@ -44,15 +44,15 @@ dataset_description_with_genetics:
   - dataset.files contains "/genetic_info.json"
   fields:
     Genetics: REQUIRED
-    Genetics.Dataset: REQUIRED
+    Genetics{}.Dataset: REQUIRED
       level: REQUIRED
       issue:
         name: NO_GENETIC_DATABASE
         message: |
           A genetic_info.json file is present but no Database field present in Genetics
           object in dataset_description.json.
-    Genetics.Database: OPTIONAL
-    Genetics.Descriptors: OPTIONAL
+    Genetics{}.Database: OPTIONAL
+    Genetics{}.Descriptors: OPTIONAL
 
 genetic_info:
   selectors:
--- a/src/schema/rules/dataset_metadata.yaml
+++ b/src/schema/rules/dataset_metadata.yaml
@@ -1,53 +1,53 @@
 dataset_description:
-    selectors:
-    - path == "/dataset_description.json"
-    fields:
-        Name: REQUIRED
-        BIDSVersion: REQUIRED
-        DatasetType: RECOMMENDED
-        License: RECOMMENDED
-        Authors: RECOMMENDED
-        Acknowledgements: OPTIONAL
-        HowToAcknowledge: OPTIONAL
-        Funding: OPTIONAL
-        EthicsApprovals: OPTIONAL
-        ReferencesAndLinks: OPTIONAL
-        DatasetDOI: OPTIONAL
-        GeneratedBy: RECOMMENDED
-        GeneratedBy[].Name: REQUIRED
-        GeneratedBy[].Version: RECOMMENDED
-        GeneratedBy[].Description: OPTIONAL
-        GeneratedBy[].CodeURL: OPTIONAL
-        GeneratedBy[].Container: OPTIONAL
-        GeneratedBy[].Container.Type: RECOMMENDED
-        GeneratedBy[].Container.Tag: RECOMMENDED
-        GeneratedBy[].Container.URI: RECOMMENDED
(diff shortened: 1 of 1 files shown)


src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,40 @@
...
+        EthicsApprovals: OPTIONAL
+        ReferencesAndLinks: OPTIONAL
+        DatasetDOI: OPTIONAL
+        GeneratedBy: RECOMMENDED
+        SourceDatasets: RECOMMENDED
+
+derivative_description:
+    path: /dataset_description.json
+    selectors:
+        dataset.DatasetType == "derivative"
```
tsalo, line 21: How do you envision the `dataset` variable being structured? What would be available in it?

effigies: Right. Unclear. I was assuming that it would basically be the `dataset_description.json` object, or a semantically useful subset of it, a la @rwblair's nifti_header subset.

tsalo: I think we also have used it for lists of modalities present in the dataset in other PRs.

tsalo:
>1. How do we want to encode, e.g. NO_AUTHORS, where we effectively promote dataset_description.Authors from OPTIONAL to REQUIRED?

I think since NO_AUTHORS is a warning, it would be upgrading from OPTIONAL to RECOMMENDED, right? In any case, what if we had a "publication" level for the validator? The checks in this level wouldn't have to be derived from the specification, but would be necessary for data sharing.

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
+dataset_description:
+    selectors:
+    - path == /dataset_description.json
+    fields:
+        Name: REQUIRED
+        BIDSVersion: REQUIRED
+        DatasetType: RECOMMENDED
+        License: RECOMMENDED
+        Authors: OPTIONAL
```
effigies, line 9: Implement #1068:

```suggestion
        Authors: RECOMMENDED
```

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
+dataset_description:
+    selectors:
+    - path == /dataset_description.json
```
effigies, line 3: We should probably quote these things.

```suggestion
    - path == "/dataset_description.json"
```

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+        DatasetDOI: OPTIONAL
+        GeneratedBy: RECOMMENDED
+        GeneratedBy[].Name: REQUIRED
+        GeneratedBy[].Version: RECOMMENDED
+        GeneratedBy[].Description: OPTIONAL
+        GeneratedBy[].CodeURL: OPTIONAL
+        GeneratedBy[].Container: OPTIONAL
+        GeneratedBy[].Container.Type: RECOMMENDED
+        GeneratedBy[].Container.Tag: RECOMMENDED
+        GeneratedBy[].Container.URI: RECOMMENDED
```
tsalo, line 24: For an object, would the convention then be `GeneratedBy.Name` or `GeneratedBy{}.Name` (in case we want to allow metadata keys with periods in them without assuming they're dictionaries)?

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+        SourceDatasets: RECOMMENDED
+
+derivative_description:
+    selectors:
+    - path == /dataset_description.json
+    - dataset.dataset_description.DatasetType == "derivative"
```
tsalo, line 30: In this case, wouldn't the `DatasetType` be available through the targeted file (i.e., the dataset_description)?

effigies: Sure, but what would that look like?

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+    fields:
+        GeneratedBy: REQUIRED
+
+dataset_description_with_genetics:
+    selectors:
+    - path == /dataset_description.json
+    - dataset.files contains "/genetic_info.json"
```
tsalo, line 37: If we're searching for files via paths, would we want to include search patterns here? E.g., `dataset.files contains `/sub-{subject}/*/pet/*` to determine if the subject has PET data?

src/schema/rules/dataset_metadata.yaml
tsalo: data.DatasetType I guess?

tsalo: `json.DatasetType`

src/schema/rules/dataset_metadata.yaml
effigies: I don't think that level of flexibility is called for at this point.

src/schema/rules/dataset_metadata.yaml
effigies: I assumed that `.` would do the job, but `GeneratedBy{}.Name` would be more explicit. Happy to have more opinions here.

effigies: To make your suggestion explicit, I think

```suggestion
        GeneratedBy[].Name: REQUIRED
        GeneratedBy[].Version: RECOMMENDED
        GeneratedBy[].Description: OPTIONAL
        GeneratedBy[].CodeURL: OPTIONAL
        GeneratedBy[].Container: OPTIONAL
        GeneratedBy[].Container{}.Type: RECOMMENDED
        GeneratedBy[].Container{}.Tag: RECOMMENDED
        GeneratedBy[].Container{}.URI: RECOMMENDED
```

tsalo: That looks good to me.

src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+dataset_description_with_genetics:
+    selectors:
+    - path == "/dataset_description.json"
+    - dataset.files contains "/genetic_info.json"
+    fields:
+        Genetics: REQUIRED
+        Genetics.Dataset: REQUIRED
+        Genetics.Database: OPTIONAL
+        Genetics.Descriptors: OPTIONAL
```
effigies, line 42: @tsalo Would this correspond to these names in the metadata.yaml as well?

```suggestion
        Genetics{}.Dataset: REQUIRED
        Genetics{}.Database: OPTIONAL
        Genetics{}.Descriptors: OPTIONAL
```

tsalo: Well, I want to move those names _into_ the `Genetics` object in metadata.yaml anyway, so I'm not too concerned about `Genetics.Dataset`, etc., in that file. We just need a new function to make tables of subobjects for that to work, since there's a table with the Genetics subobjects in the specification.

One random thought though- do we need the dot after the braces?

PR #1066, opened 2022-04-13
Replace `addendum` with `description_addendum`

tsalo:
To reduce confusion between `addendum` and `level_addendum`.


tsalo:
@bendhouseart is this clearer than "addendum" and "level_addendum"?

PR #1069, opened 2022-04-14
SCHEMA: Create context schema in meta/

effigies:
Builds on @rwblair's work in #1059.


src/schema/meta/context.yaml
```diff
@@ -0,0 +1,137 @@
...
+# The subject namespace is constructed once per subject, and is available when
+# visiting all files within that subject.
+#
+# All other (current) namespaces are defined on individual files.
+# Sidecar metadata and file associations are built according to the inheritance
+# principle.
+#
+---
+context:
+    dataset:
```
tsalo, line 21: I believe we also use `dataset.modalities`, but I guess we might want to access that through `subject` or `session`...

src/schema/meta/context.yaml
```diff
@@ -0,0 +1,188 @@
...
+            type: str
+        modality:
+            description: "Modality of current file, e.g., MRI"
+            type: str
+        sidecar:
+            description: "Sidecar metadata constructed via the inheritance principle"
+            type: map
+        associations:
+            description: |
+                Associated files, indexed by suffix, selected according to the inheritance principle
```
tsalo, line 103: Should this include other files in a file collection, such as the other `echo-*` files for a ME-EPI file?

tsalo, line 103: Should this include other files in a file collection, such as the other `echo-*` files for a ME-EPI file?

tsalo, line 103: Should this include other files in a file collection, such as the other `echo-*` files for a ME-EPI file?

src/schema/meta/context.yaml
```diff
@@ -0,0 +1,188 @@
...
+            type: map
+            properties:
+                events:
+                    description: "Path to associated events file"
+                    type: str
+                aslcontext:
+                    description: "Path to associated aslcontext file"
+                    type: str
```
tsalo, line 111: Do we want to hardcode associations in this file? There are a lot of them, and they all depend on the original file's suffix and/or entities.

tsalo, line 111: Do we want to hardcode associations in this file? There are a lot of them, and they all depend on the original file's suffix and/or entities.

tsalo, line 111: Do we want to hardcode associations in this file? There are a lot of them, and they all depend on the original file's suffix and/or entities.

effigies: @tsalo See latest file. I try to make clear that these are only ones that are used in rules.

src/schema/meta/context.yaml
effigies: Are there rules that this would help us enforce?

PR #1074, opened 2022-04-14
SCHEMA: Convert datatype lists to objects

effigies:
This allows us to have an unambiguous name for each rule, such as `schema.rules.datatypes.anat.nonparametric`.

### DIFF
--- a/src/schema/rules/datatypes/anat.yaml
+++ b/src/schema/rules/datatypes/anat.yaml
@@ -1,6 +1,6 @@
 ---
-# Nonparametric
-- suffixes:
+nonparametric:
+  suffixes:
   - T1w
   - T2w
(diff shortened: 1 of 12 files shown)


tsalo:
I think the changes you need to make to `make_entity_table` and `make_filename_template` are pretty small. Both functions just loop through the list, so you just need loop through the dict instead.

tsalo:
I think the changes you need to make to `make_entity_table` and `make_filename_template` are pretty small. Both functions just loop through the list, so you just need loop through the dict instead.

tsalo:
I think the changes you need to make to `make_entity_table` and `make_filename_template` are pretty small. Both functions just loop through the list, so you just need loop through the dict instead.

Issue #841, opened 2021-08-04
`Genetics` subobject in the schema


Remi-Gau:
The `genetics` object in dataset_description has 3 sub-objects: 
- `Dataset` 
- `Database`
- `Descriptors`

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/08-genetic-descriptor.html#dataset-description

Currently the schema has a single yaml file for each sub-object but there is no yaml file for the `Genetics`.
Moreover the nested objects is not captured by the schema.
And the way the sub-objects are mentioned in the rendred table involve a dot notation `Genetics.Dataset` that appears nowhere else in the spec.

### Suggestion

- create yaml file for `Genetics`
- rewrite the part of the the Genetics to remove the dot notation from the table

### Possible issues

Not sure how to formalize the nested structure in the schema itself:

Should the sub-objects of `Genetics` be defined in the yaml of `Genetics`?

Should we keep the yaml files for each sub-object, mention in the README that this dot notation means they are sub-objects?

I don't think it makes sense to have pieces of metadata just called `Dataset`, `Database`,  `Descriptors` when they only relate to `Genetics`.

Tagging @tsalo and @CPernet for ideas and feedback on this.

CPernet:
{
  "Name": "Human Connectome Project",
  "BIDSVersion":  "1.3.0",
  "License": "CC0",
  "Authors": ["1st author", "2nd author"],
  "Funding": ["P41 EB015894/EB/NIBIB NIH HHS/United States"],
  "Genetics": {
     "Dataset": "https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/study.cgi?study_id=phs001364.v1.p1",
     "Database": "https://www.ncbi.nlm.nih.gov/gap/",
     "Descriptors": ["doi:10.1016/j.neuroimage.2013.05.041"]
     }
  "DatasetDOI": "doi:myimagaing data"
}

could be changed without subfields if that's the direction we are taking ; whatever is easier 

{
  "Name": "Human Connectome Project",
  "BIDSVersion":  "1.3.0",
  "License": "CC0",
  "Authors": ["1st author", "2nd author"],
  "Funding": ["P41 EB015894/EB/NIBIB NIH HHS/United States"],
  "GeneticsDataset": ["https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/study.cgi?study_id=phs001364.v1.p1"],
[...]

PR #1075, opened 2022-04-14
State of the schema sprint

effigies:
Mostly doing this to get an RTD pull request build going as we update.


src/schema/rules/dataset_metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+    DatasetDOI: OPTIONAL
+    GeneratedBy: RECOMMENDED
+    GeneratedBy[].Name: REQUIRED
+    GeneratedBy[].Version: RECOMMENDED
+    GeneratedBy[].Description: OPTIONAL
+    GeneratedBy[].CodeURL: OPTIONAL
+    GeneratedBy[].Container: OPTIONAL
+    GeneratedBy[].Container{}.Type: RECOMMENDED
+    GeneratedBy[].Container{}.Tag: RECOMMENDED
+    GeneratedBy[].Container{}.URI: RECOMMENDED
```
Remi-Gau, line 32: Those dot, brackets and curly braces will MATLAB cry: the only special characters allowed in keys by matlab are `_` 

Related to 
https://github.com/bids-standard/bids-specification/pull/1082

effigies: We need some way to mark the requirement levels of sub objects.

Remi-Gau: See also: https://github.com/bids-standard/bids-specification/issues/841

tsalo: #841 is really just an artifact at this point. Once someone (probably me) writes a function to render subobjects in tables, we can get rid of the `Genetics.*` fields.

effigies:
Three live :+1:s. Merging to avoid accumulating merge conflicts.

PR #1072, opened 2022-04-14
[SCHEMA] Define common derivatives rules

rwblair:
cherry-picked commits from schema/derivatives branch

PR #1081, opened 2022-04-17
SCHEMA: Load global schema namespace to permit cross-file references

effigies:
Started by trying out an example inspired by #1072. Needed to create a `Namespace` object. Need to start loading multiple files and testing out fully qualified references.

### DIFF
--- a/tools/schemacode/schemacode/schema.py
+++ b/tools/schemacode/schemacode/schema.py
@@ -3,6 +3,7 @@ import logging
 import os
 from copy import deepcopy
 from pathlib import Path
+from collections.abc import Mapping
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


rwblair:
This is really clever. Playing around with this I was able to get derivatives datatypes working in load_schema with two changes. Moving the yaml files from `datatypes/derivatives` up a level to `datatypes/`, then when iterating over the nested rules glob, don't attempt to dereference the dictionary, instead store as is:
```
        rule[rule_group_file.stem] = dict_
```
Then do a single dereference step outside loop before returning the results:
```
    dereference_yaml(Namespace.build(schema), schema)
    return schema
```

PR #1082, opened 2022-04-18
SCHEMA: Update references to use _ref instead of $ref

effigies:
It looks like using `$ref` will cause issues with the MATLAB YAML reader (or maybe just MATLAB). I think `_ref` is very unlikely to hit collisions while being friendly to various languages.

### DIFF
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -316,14 +316,14 @@ ExampleField:
 ```
 
 Furthermore, if the data type description of a field is reused across fields,
-then it may be defined in a separate field and referenced in each target field with the `$ref` keyword.
-Here is an example of a field definition using `$ref`:
+then it may be defined in a separate field and referenced in each target field with the `_ref` keyword.
+Here is an example of a field definition using `_ref`:
 ```yaml
 ExampleField:
     name: ExampleField
     description: |
         The description of "ExampleField".
-    $ref: _ReferenceField
+    _ref: _ReferenceField
 ```
 The target field (`_ReferenceField` in this case) MUST be located in
 the metadata YAML file, and MUST contain all information
@@ -341,7 +341,7 @@ _ReferenceField:
         - PossibleValue2
 ```
 
-The result of including `$ref` makes the `ExampleField` above equivalent to the following format:
+The result of including `_ref` makes the `ExampleField` above equivalent to the following format:
 ```yaml
 ExampleField:
     name: ExampleField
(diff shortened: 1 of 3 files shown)


tsalo (approved):
LGTM!

Remi-Gau:
> It looks like using `$ref` will cause issues with the MATLAB YAML reader (or maybe just MATLAB). I think `_ref` is very unlikely to hit collisions while being friendly to various languages.

I hate to be that guy so let me bring in some self-deprecation into the mix...

Depends if you consider MATLAB a language...

> MATLAB YAML reader

If only there was such a thing... 

YAML pretty much has to be converted to JSON for MATLAB

AFAIK

> `_ref` is very unlikely to hit collisions

Well loading a JSON that with such key in MATLAB will very likely lead to problems

Added a section in the starter kit because it may matter for some aspects of BIDS events file: 
https://bids-standard.github.io/bids-starter-kit/folders_and_files/metadata.html#interoperability-issues

the TL;DR for here is:

- JSON (YAML in this case) keys must start with a letter

Remi-Gau (changes requested):
> It looks like using `$ref` will cause issues with the MATLAB YAML reader (or maybe just MATLAB). I think `_ref` is very unlikely to hit collisions while being friendly to various languages.

I hate to be that guy so let me bring in some self-deprecation into the mix...

Depends if you consider MATLAB a language...

> MATLAB YAML reader

If only there was such a thing... 

YAML pretty much has to be converted to JSON for MATLAB

AFAIK

> `_ref` is very unlikely to hit collisions

Well loading a JSON that with such key in MATLAB will very likely lead to problems

Added a section in the starter kit because it may matter for some aspects of BIDS events file: 
https://bids-standard.github.io/bids-starter-kit/folders_and_files/metadata.html#interoperability-issues

the TL;DR for here is:

- JSON (YAML in this case) keys must start with a letter

effigies:
Maybe instead of this, we should write a `schema2json` tool that will take our YAML schema with references in it and produce a fully rendered schema with no references. If a conversion is already necessary, why implement dereferencing in another language?

Remi-Gau:
> Maybe instead of this, we should write a `schema2json` tool that will take our YAML schema with references in it and produce a fully rendered schema with no references. If a conversion is already necessary, why implement dereferencing in another language?

uuuuhhh I like that :+1: 

Also would make the fully rendered schema more explicit (albeit more verbose, no?) which could be good for new users.

effigies:
Okay, let's discuss this tomorrow.

PR #1091, opened 2022-04-22
ENH: use __file__ instead of getcwd to determine HERE, provide informative message on # of files

yarikoptic:
I had to spend some time to figure out why running that script did nothing for me
but failed in CI for the PR.  Apparently it was because I had to run it while
in tools/ directory.  IMHO it is suboptimal.

With a short message on how many files were analyzed user at least could
get a hint in case if no files were read at all

### DIFF
--- a/tools/no-bad-latin.py
+++ b/tools/no-bad-latin.py
@@ -11,14 +11,17 @@ import re
 import argparse
 from pull_files import filter_files
 
-HERE = os.getcwd()
-ABSOLUTE_HERE = os.path.dirname(HERE)
+ABSOLUTE_HERE = os.path.dirname(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


effigies (approved):
Looks sensible.

tools/no-bad-latin.py
```diff
@@ -12,13 +12,17 @@
 from pull_files import filter_files
 
 HERE = os.getcwd()
```
sappelhoff, line 14: this variable is no longer used anywhere

```suggestion
```

yarikoptic: THANKS! removed, force-pushed

sappelhoff (approved):

sappelhoff:
Thanks @yarikoptic

PR #1098, opened 2022-05-06
[FIX] Standardized tabular_metadata.yaml syntax to conform to: f04b47ff

TheChymera:
https://github.com/bids-standard/bids-specification/commit/f04b47ff makes a number of things much more interesting, but if we're doing this, probably best done everywhere.

@effigies

### DIFF
--- a/src/schema/rules/tabular_metadata.yaml
+++ b/src/schema/rules/tabular_metadata.yaml
@@ -1,17 +1,15 @@
 ---
-# scans.tsv
-- suffixes:
+scans:
+  suffixes:
   - scans
   extensions:
   - .tsv
   - .json
   entities:
     subject: required
-    # session is required if session is present in the dataset.
-    session: optional
-# sessions.tsv
-# This file may only exist if session is present in the dataset.
-- suffixes:
+    session: optional # session is required if session is present in the dataset.
+sessions: # This file may only exist if session is present in the dataset.
+  suffixes:
   - sessions
   extensions:
   - .tsv


tsalo (approved):
Thanks for catching this. These changes look good to me.

effigies (approved):

src/schema/rules/tabular_metadata.yaml
```diff
@@ -1,17 +1,15 @@
 ---
-# scans.tsv
-- suffixes:
+scans.tsv:
```
effigies, line 2:
```suggestion
scans:
```

src/schema/rules/tabular_metadata.yaml
```diff
@@ -1,17 +1,15 @@
...
   - .json
   entities:
     subject: required
-    # session is required if session is present in the dataset.
-    session: optional
-# sessions.tsv
-# This file may only exist if session is present in the dataset.
-- suffixes:
+    session: optional # session is required if session is present in the dataset.
+sessions.tsv: # This file may only exist if session is present in the dataset.
```
effigies, line 11:
```suggestion
sessions: # This file may only exist if session is present in the dataset.
```

effigies (commented):
Actually, I think we probably don't want `.`s in our keys.

TheChymera:
@effigies fixed?

TheChymera:
@effigies could merge this? I'd need it for updating the validator loading since right now this file needs the old loading model whereas all the others need the new one.

PR #1107, opened 2022-05-26
SCHEMA: Unify metadata

### LINKED ISSUE #1088: SCHEMA: Proposals to standardize schema/objects/*.yaml formats
From conversations in #1086 the key values and name fields are not used entirely consistently in the schema objects. 
columns.yaml uses the `name` field value as the actual header column name we look for in tsv headers.
entities.yaml uses the `entity` field value as the literal entity that we look for in file names
suffixes.yaml uses the entries key as the literal suffix that we look for in file names.

Except for columns.yaml `name` is used as pretty printable/user friendly display name for the object. 

The key used for an object is always how we refer to an object within the specification.

suffix and columns should have the values that we look for in datasets moved from their respective places into a new field.

Proposal 1 is to follow entity suite and have a special key-name unique to that object type. So from columns.yaml this:
```yaml
acq_time__scans:
  name: acq_time
  description: |
    Acquisition time refers to when the first data point in each run was acquired.
```
Becomes:
```yaml
acq_time__scans:
  name: Acquisition Time
  column: acq_time
  description: |
    Acquisition time refers to when the first data point in each run was acquired.
```

Proposal 2 would be to standardize on a single key name across object to represent in dataset values. If we followed columns suite and used `name` for this, pretty print values could shift to `display_name`:
```yaml
# suffix
TwoPE:  # some unambiguous reference that works for MATLAB and ancpBIDS
  name: 2PE  # The actual value used in the specification
  display_name: 2-photon excitation microscopy  # The pretty name
  description: |
    2-photon excitation microscopy imaging data

# entity
inversion:  # also used for variable names in pybids
  name: inv
  display_name: Inversion Time
  description: |
    individual files.
    This entity represents the `"InversionTime` metadata field. Please note that the `<index>`
    denotes the number/index (in the form of a nonnegative integer), not the `"InversionTime"`
    value which needs to be stored in the field `"InversionTime"` of the separate JSON file.
  type: string
  format: index

# column
type__electrodes:
  name: type
  display_name: Electrode Type
  description: |
    Type of the electrode (for example, cup, ring, clip-on, wire, needle).
  type: string
```

Instead of `name` there may be some other more appropriate key name to use, some floated during the 21.04.2022 schema meeting were `alias`, `value`, `shorthand`, etc but none felt quite right.

I would like to solicit input on the proposals and other potential solutions.

effigies:
Two formats to rule them all.

Key/value pairs implemented. Next come named values. This will break things.

Closes #1088.


src/schema/objects/columns.yaml
```diff
@@ -30,41 +34,48 @@ acq_time__sessions:
   format: datetime
 age:
   name: age
+  display_name: Subject age
```
yarikoptic, line 37: do we document anywhere why/what 3 "things" are there to describe the same thing. It seems they are the 

```yaml
{id for a computer}:
  name: {something which is typically just an id but some times different... not sure why different at times, why cannot be the same as id}
  display_name: {a little more expanded for a human being but some times feels superficial ;) }
```

and should be 
```suggestion
  display_name: Participant age
```
to be more politically correct IMHO

src/schema/objects/columns.yaml
```diff
@@ -94,6 +107,7 @@ group:
   - type: number
 handedness:
   name: handedness
+  display_name: Subject handedness
```
yarikoptic, line 110:
```suggestion
  display_name: Participant handedness
```

src/schema/objects/columns.yaml
```diff
@@ -123,6 +137,7 @@ handedness:
   - n/a
 hemisphere:
   name: hemisphere
+  display_name: Electrode hemisphere
```
yarikoptic, line 140: AFAIK electrodes have no hemispheres... 

```suggestion
  display_name: Electrode placement hemisphere
```

src/schema/objects/columns.yaml
```diff
@@ -173,29 +193,34 @@ low_cutoff:
...
   type: string
 mapping:
   name: mapping
+  display_name: Label mapping
   description: |
     Corresponding integer label in the standard BIDS label lookup.
   type: integer
 material:
   name: material
+  display_name: Electrode material
```
yarikoptic, line 209: probably unrelated to this PR: all of the above examples adding "Participant", "Electrode" and other qualifiers suggest eventual possible collision. E.g. couldn't material be eventually be a legit column for some kind of "injector" or "sample"... dunno.  I see many `id`s below include their "domain".  i.e. here to have `electrode_material` ?

src/schema/objects/columns.yaml
```diff
@@ -79,12 +90,14 @@ duration:
     - n/a
 filename:
   name: filename
+  display_name: Filename
   description: |
     Relative paths to files.
   type: string
   format: participant_relative
-group:
+channel_group:
```
tsalo, line 98: Our pattern has generally been `group__channel`, but I don't know if we need to adhere to that after reorganizing.

src/schema/objects/associated_data.yaml
```diff
@@ -4,21 +4,21 @@
 # Information about whether these directories are required or not is present in
 # `rules/associated_data.yaml`.
 code:
-  name: Code
+  display_name: Code
```
effigies, line 7: @rwblair My disruption of the schema demanded that I create a merge conflict with your PR. I offer my humblest apologies.

effigies:
Preserving @yarikoptic's comment:

> do we document anywhere why/what 3 "things" are there to describe the same thing. It seems they are the 
> 
> ```yaml
> {id for a computer}:
>   name: {something which is typically just an id but some times different... not sure why different at times, why cannot be the same as id}
>   display_name: {a little more expanded for a human being but some times feels superficial ;) }
> ```

---

Yes. Where do we document these things in the schema? I would put it this way:

```yaml
{schema key, should be identifier friendly, so [a-zA-z][a-zA-Z0-9]*}:
  [name: {for name/value pair metadata, the name}]
  [value: {for controlled vocabulary metadata, the value}]
  display_name: {name for rendering in specs when name or value is not more appropriate}
```

src/schema/objects/columns.yaml
effigies:
```suggestion
group__channel:
```

src/04-modality-specific-files/04-intracranial-electroencephalography.md
```diff
@@ -294,7 +294,7 @@ and a guide for using macros can be found at
 {{ MACROS___make_columns_table(
    {
       "reference__ieeg": "OPTIONAL",
-      "group": ("OPTIONAL", "Note that any groups specified in `_electrodes.tsv` must match those present here."),
+      "channel_group": ("OPTIONAL", "Note that any groups specified in `_electrodes.tsv` must match those present here."),
```
effigies, line 297:
```suggestion
      "group__channel": ("OPTIONAL", "Note that any groups specified in `_electrodes.tsv` must match those present here."),
```

effigies:
@tsalo @TheChymera Forgot on Thursday this still isn't merged. Could I get a final review when you have a chance?

TheChymera:
Looks good to me in principle as well, could you run the tests again, though? I think it will break some of the built-in validation. Or you can just merge and I can fix in a separate PR if you don't want to bother. I'm pretty sure I query the shorthands via the old nomenclature somewhere.

PR #1109, opened 2022-05-30
[ENH] Skipping tests which requre network access via env variable

TheChymera:
Since these tests require network access, there should be a way to skip them.
This PR introduces the `SCHEMACODE_TESTS_NONETWORK` environment variable which is evaluated to determine whether networktests should be used.

@yarikoptic I tried to do the same thing seen in dandi-cli with `skipif.no_network()` which allows skipping in the fixture (which is the thing that's actually requiring network access), but it did not work :( happy to try to solve this differently.


TheChymera:
@tsalo @effigies interested in this? would really help with the eventual distribution of the package.

yarikoptic:
> but it did not work

even pytest guru @jwodder  wouldn't be able to help with such condensed presentation of the problem ;)  May be try again and provide for a review/state *how* it doesn't work (crashes, doesn't exclude, ...?)

bendhouseart (approved):
Sure why not.

sappelhoff:
> May be try again and provide for a review/state how it doesn't work (crashes, doesn't exclude, ...?)

agreed, could you perhaps provide a more detailed description of what you tried with the skipif decorator? And perhaps a link to the source where it does work (dandi-cli)? Perhaps we don't need this env variable at all.

(However, happy to include it if the skipif decorator really turns out to not work)

yarikoptic:
> , but it did not work :(

since that approach you took was not shared for review, it is impossible to provide recommendation on how to fix it so it would work.  But FWIW, those skip fixtures/marks are "dynamically" produced in https://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/skip.py .

yarikoptic (approved):
Since only a single location, ok as is to me.  But if one more comes up, there should be a proper centralized helper defined.

PR #1079, opened 2022-04-15
DOC: Update README with proposed conventions

effigies:

PR #1112, opened 2022-06-02
DOC: Update README with proposed schema conventions

effigies:
This PR is a significant rewrite of the `src/schema/README.md`. One small change to the structure of the schema is preemptively described here. It is implemented in #1291.

Reviving https://github.com/bids-standard/bids-specification/pull/1079.

### DIFF
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -160,8 +160,7 @@ references (the cases in which they are used will be presented later):
 
 ### Expressions
 
-In order to define a rule, we describe a limited language for boolean expressions,
-that is, having values of `true` or `false`.
+Rules definitions make use of a limited language of expressions that always evaluate to `true` or `false`.
 These expressions may be used as `selectors`, determining whether a rule applies,
 or `checks`, determining whether a rule is satisfied.
 
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -11,25 +11,23 @@ Any changes to the specification should be mirrored in the schema.
 
 At the time of this writing, the schema has the following file layout:
 
-```
-src/schema
-├── BIDS_VERSION
-├── meta
-│   └── context.yaml
-├── objects
-│   ├── associated_data.yaml
-│   ├── ...
-│   └── top_level_files.yaml
-├── rules
-│   ├── associated_data.yaml
-│   ├── checks
-│   │   ├── asl.yaml
-│   │   ├── ...
-│   │   └── mri.yaml
-│   ├── ...
-│   └── top_level_files.yaml
-└── SCHEMA_VERSION
-```
+    src/schema
+    ├── BIDS_VERSION
+    ├── meta
+    │   └── context.yaml
+    ├── objects
+    │   ├── associated_data.yaml
+    │   ├── ...
+    │   └── top_level_files.yaml
+    ├── rules
+    │   ├── associated_data.yaml
+    │   ├── checks
+    │   │   ├── asl.yaml
+    │   │   ├── ...
+    │   │   └── mri.yaml
+    │   ├── ...
+    │   └── top_level_files.yaml
+    └── SCHEMA_VERSION
 
 The top-level organization includes `objects`, where terms are defined;
 `rules`, where constraints (such as valid filenames or required metadata fields)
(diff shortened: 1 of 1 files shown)


src/schema/README.md
```diff
@@ -481,6 +489,102 @@ Specifically, files in the first group may have `task`, `run`, `processing`, and
 while files in the second group may not.
 Also, when files in the second group have the `acq` entity, the associated value MUST be `crosstalk`.
 
+### `sidecars/*.yaml`
+
+Sidecar files introduce the idea of "selectors" to determine whether a set of fields is added
+to the sidecar schema that will be applied to the sidecar.
+From here, the `fields` property lists will be accumulated to set requirement levels.
```
tsalo, line 496: This is confusingly worded to me. Perhaps

```suggestion
From here, the metadata fields listed in the `fields` property will be accumulated across dictionaries to set requirement levels after evaluating all selectors.
```

That's not really any better, actually.

tsalo (dismissed):
This looks really good, but there are a couple of small gaps.

sappelhoff:
I reconsidered and don't think we absolutely need it for the release

effigies:
@tsalo While looking through objects, I noticed some weirdness. We use `name` in the sense of field name (name in name/value pair) in columns/entities/metadata. However in common_principles, it's just a copy of `display_name` and in `top_level_files` it's the file path. I would suggest we drop it altogether in these cases.

The `top_level_files` use to indicate the filename can be replaced with the `rules.top_level_files` (especially in light of #1202).

WDYT?

src/schema/README.md
```diff
@@ -87,371 +400,197 @@ reference__eeg:
...
-    description: |
-        The description of "ExampleField".
-    anyOf:
-        - type: number
-        - type: array
-            items:
-            type: number
-```
+A significant portion of BIDS is devoted to the naming of files, and almost all file names consist
+of entities, a suffix, an extension, and a data type. Exceptions will be noted below.
```
effigies, line 558: This documents the current state.

I would like to rename `rules.datatypes` to `rules.filenames` and include `rules.associated_data` and `rules.top_level_files` inside this.

src/schema/README.md
```diff
@@ -483,35 +624,229 @@ Specifically, files in the first group may have `task`, `run`, `processing`, and
...
+
+### One-off rules
+
+-   `rules.modalities` - The keys in this file are the modalities, the values objects with the following field:
+    | Field       | Description                           |
+    | ----------- | ------------------------------------- |
+    | `datatypes` | List of datatypes mapping to modality |
+
+-   `rules.dataset_metadata` - These are similar to `rules.sidecars.*`, for JSON files at the root level.
+    This is likely to go away in favor of other approaches.
```
effigies, line 845: I do not like this, so I didn't bother fleshing it out. I think what we want is to fully define these documents inside the `objects.*` namespace, and find some way in the rules to say something like "a file named `dataset_description.json` contains the object `objects.metadata.DatasetDescription`".

effigies (commented):
Okay. I think this is basically ready to go. Would appreciate critical reads. 

cc @bids-standard/schema-users

src/schema/README.md
```diff
@@ -7,70 +7,382 @@ Currently the portions of the specification that rely on this schema are
 the entity tables, entity definitions, filename templates, and metadata tables.
 Any changes to the specification should be mirrored in the schema.
```
rwblair, line 9: This file is getting long enough that a ToC might be warranted.

src/schema/README.md
```diff
@@ -7,70 +7,382 @@ Currently the portions of the specification that rely on this schema are
...
+└── SCHEMA_VERSION
+```
 
-The schema is divided into two parts: the object definitions and the rules.
+The top-level organization includes `objects`, where terms are defined;
+`rules`, where constraints (such as valid filenames or required metadata fields)
+are defined;
+and `meta`, where definitions useful for interpreting the schema are defined.
+
+Each file is a YAML structure, most often an *object*.
```
rwblair, line 39:
```suggestion
Each file is made up of YAML data, most often an *object*.
```

src/schema/README.md
effigies: Thought about it. Would be its own thing to maintain. Note that GitHub has something built-in when viewing the file, though sadly not the diff:

image

rwblair (commented):
I use the expressions section of this document all the time and would like to see it merged. 

Who do we imagine the intended audience for this document is? I held back on suggestions that might be too introductory or better served by a doc specifically aimed at bep authors or new contributors.

src/schema/README.md
rwblair: I'm waffling on whether the very specific sections of each of the top level namespaces should be split into their own README in that subdirectory.

src/schema/README.md
```diff
@@ -7,70 +7,382 @@ Currently the portions of the specification that rely on this schema are
...
--   top-level files,
--   and non-BIDS associated directories.
+Because the same description may be used in multiple locations, a mechanism is needed
+to ensure that the correct path is discovered to render the description in each location.
+To do this, the path should follow the form `SPEC_ROOT/path/within/source.md#anchor`.
+For example, to link to the
+[Definitions](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#definitions)
+section of
+[Common principles](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html),
+use the path `SPEC_ROOT/02-common-principles.md#definitions`, e.g.,
```
effigies, line 92:
```suggestion
use the path `SPEC_ROOT/02-common-principles.md#definitions`:
```

src/schema/README.md
```diff
@@ -87,371 +400,197 @@ reference__eeg:
...
 
-       -   `unit` (standard units),
+BIDS follows RFC 2119 and has three requirement levels: OPTIONAL, RECOMMENDED and REQUIRED.
+In the schema, we use `optional`, `recommended` and `required`. A rule interpreter (validator)
+is expected to treat missing REQUIRED data/metadata as an error, missing RECOMMENDED
+data/metadata as a warning, and silently pass over missing OPTIONAL data.
 
-       -   `dataset_relative` (relative paths from dataset root),
+BIDS also defines a level `DEPRECATED`, rendered in the schema as `deprecated`, and corresponding
+to a warning if the data/metadata is present.
```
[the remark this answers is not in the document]
effigies: It does not exist so far. I'd like to avoid it until it becomes definitely needed.

src/schema/README.md
effigies: Not a bad idea. Would rather not do it right now.

src/schema/README.md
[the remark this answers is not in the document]
bendhouseart: Can't that case be covered with a `!(ImProhibited in blah)`?

src/schema/README.md
```diff
@@ -7,70 +7,382 @@ Currently the portions of the specification that rely on this schema are
...
-Specifically, each version of the object must have its own definition within the relevant file.
-However, since object files are organized as dictionaries, each object must have a unique key.
-Thus, we append a suffix to each re-used object's key in order to make it unique.
-For objects with `CamelCase` names (for example, metadata fields), the suffix will start with a single underscore (`_`).
-For objects with `snake_case` names, two underscores must be used.
+In a few cases, two objects with the same name appear multiple times in the specification.
+When this happens, it is preferred to find a common definition, and clarify it in the rules (see below).
+However, in some cases, the object description and permissible values differ, and it needs to be defined
+as two separate objects.
+The following conventions are used:
```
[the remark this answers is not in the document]
effigies: Done

src/schema/README.md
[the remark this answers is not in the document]
effigies: That would be a check, not a requirement level. So yes, it can be done so far. Maybe there will be a case where we'd have to bend over backwards to avoid it and it makes more sense to go ahead. Haven't seen it yet.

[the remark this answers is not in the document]
bendhouseart: @rwblair maintains the validator, so I could go either way on this :)

bendhouseart:
>I use the expressions section of this document all the time and would like to see it merged.

Yes, please it's great!

bendhouseart (approved):

src/schema/README.md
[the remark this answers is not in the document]
rwblair: There's a pr some where with prohibited as a requirement level, I think its useful for the mutual exclusion cases.

effigies:
Thanks!

PR #1113, opened 2022-06-02
ENH: Add namespace object that allows attribute and key access to fields

effigies:
Reviving #1081


effigies:
@rwblair Took a stab at loading the schema through the new object. I'm able to get it to load in IPython, but haven't tracked down what's different from the old method that's causing issues.

Won't be able to get back to this today, but wanted to push so you can play if you're so inclined.

PR #1120, opened 2022-06-08
[ENH] Validator inheritance support

TheChymera:
As discussed 2 weeks ago, the XS validator does not cover inheritance, a BIDS feature which may or may not cause more issues than it's worth.

Should work but needs more work, draft for now.


TheChymera:
@tsalo interested in reviewing this as well?

TheChymera:
@tsalo better?

TheChymera:
@effigies interested in merging? :)

effigies:
Haven't had a chance to review yet.

tools/schemacode/schemacode/validator.py
```diff
@@ -658,6 +675,77 @@ def log_errors(validation_result):
...
+    """
+    Generate regex strings applying BIDS inheritance expansion to an input string.
+
+    Parameters
+    ----------
+    regex_string : str
+        String representing the regex to which inheritance expansion should be applied.
+    datatype : str, optional
+        Datatype string to remove as part of inheritance expansion.
+    debug : bool, optional
```
yarikoptic, line 692: let's not introduce `debug` options to functions.  There is `lgr.debug` which could be "unconditionally" used to log desired debug information, and that is already used in this module

```shell
$> git grep '\.debug' -- tools/schemacode/schemacode/
tools/schemacode/schemacode/schema.py:        lgr.debug(f"Loading {object_group_file.stem} objects.")
tools/schemacode/schemacode/schema.py:        lgr.debug(f"Loading {rule_group_file.stem} rules.")
tools/schemacode/schemacode/schema.py:        lgr.debug(f"Loading {rule_group_file.stem} rules.")
tools/schemacode/schemacode/tests/conftest.py:        lgr.debug("Cloning %r into %r", url, path)
```

there is also env var which can adjust (not sure why we have two invocations since render probably imports schema anyways)

```
$> git grep set_logger_level
tools/schemacode/schemacode/render.py:utils.set_logger_level(lgr, os.environ.get("BIDS_SCHEMA_LOG_LEVEL", logging.INFO))
tools/schemacode/schemacode/schema.py:utils.set_logger_level(lgr, os.environ.get("BIDS_SCHEMA_LOG_LEVEL", logging.INFO))
tools/schemacode/schemacode/utils.py:def set_logger_level(lgr, level):
```

tools/schemacode/schemacode/validator.py
```diff
@@ -658,6 +675,77 @@ def log_errors(validation_result):
...
+        {
+            "regex": [
+                r".*?(?P<remove>sub-\(\?P<subject>\(\[0\-9a\-zA\-Z\]\+\)\)/).*?",
+                r".*?(?P<remove>sub-\(\?P=subject\))",
+                r".*?/(?P<remove>\(\|ses-\(\?P<session>\(\[0\-9a\-zA\-Z\]\+\)\)/\)\(\|_ses-\("
+                r"\?P=session\)\)_).*?",
+            ],
+            "replace": ["", "", ""],
+        },
+    ]
```
yarikoptic, line 715: so it is only subject and session entities which are "eroded" or some other entities as well?   E.g. example in https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#the-inheritance-principle would erode/remove `_acq-` from the sidecar file to consider top level task .json file to inherit.

tools/schemacode/schemacode/validator.py
```diff
@@ -228,13 +228,19 @@ def load_top_level(
 
 def load_entities(
     my_schema,
+    inheritance_regex=r".*?\\\.json.*?",
```
yarikoptic, line 231: https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#the-inheritance-principle talks also abut .bvec and .tsv but those aren't to be implemented in this PR I would assume.  Let's then add a TODO comment

```suggestion
    # TODO: support for .tsv and .bvec files when validator starts to care about their content
    inheritance_regex=r".*?\\\.json.*?",
```

or because we are not here to check for content etc, may be right away add support for those .bvec and .tsv? (I guess should be taken from schema regexes for where those files already used).  But then I am not sure if it is worth adding it as an option to the function -- do you see it getting some other values?

yarikoptic: but also why there is `.*?` trailer in regex -- could it be `.json.gz` or alike?

tools/schemacode/schemacode/tests/test_validator.py
```diff
@@ -221,6 +221,47 @@ def test__add_suffixes():
...
+        "(|_dir-(?P<direction>([0-9a-zA-Z]+)))"
+        "(|_run-(?P<run>([0-9]*[1-9]+[0-9]*)))"
+        "(|_echo-(?P<echo>([0-9]*[1-9]+[0-9]*)))"
+        "_phase(\\.nii\\.gz|\\.nii|\\.json)$",
+    ]
+
+    inheritance_expanded_entries = _inheritance_expansion(base_entry, datatype="func")
+    assert inheritance_expanded_entries == expected_entries
+
+
```
yarikoptic, line 264: I think it would have been nice if there were somewhat more straightforward tests on some listings of files (like the ones in the spec at https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#the-inheritance-principle) which would check what files would be considered ok (or not ok) through "inheritance" principle. E.g. those from bids specification:

ok:
- `task-rest_bold.json`
- `sub-01/ses-test/sub-01_ses-test_task-overtverbgeneration_bold.json`

not ok (I guess ;)):
- `task-rest.json`
- `sub-01/sub-01_ses-test_task-overtverbgeneration_bold.json`
- `sub-01/ses-test/task-overtverbgeneration_bold.json`
- ... become creative ... ;)

yarikoptic (changes requested):
Initial quick review

tools/schemacode/schemacode/validator.py
TheChymera: `_acq-` is already optional, so doesn't need to be eroded explicitly.

tools/schemacode/schemacode/validator.py
TheChymera: fixed formats, and addressed `.*?` ambiguity in [9d0b4e3](https://github.com/bids-standard/bids-specification/pull/1120/commits/9d0b4e336149f039fa9435ff1a4e414cfa60aca4).

tools/schemacode/schemacode/tests/test_validator.py
TheChymera: It might seem more straightforward on the surface, but we're not creating file names, rather regex patterns. This could be done, but it would mean an additional function to check whether two filenames match the different regex patterns created, it would be a util function for the tests, not actually used by the implementation, and I would argue it would make the inspection more intransparent.

If you're concerned about testing for false positives in inheritance validation, this would be better tested and verify the actual validation code if we had a dataset with the broken inheritance filenames you listed above. We can add that to our new bids-error-examples repo, I made an issue: https://github.com/bids-standard/bids-error-examples/issues/1

tools/schemacode/schemacode/validator.py
TheChymera: Might have been better as a separate PR since it's not specific to inheritance support... but given that PRs are a bit slow, might as well add it in here :)

tools/schemacode/schemacode/tests/test_validator.py
yarikoptic: as long as this particular validations are operating on filenames only, it is IMHO much easier/lightweight just to create test cases in the code which populate list of files in bids dataset which are given to the functions to be tested.  It makes it easier and faster then to create parametric tests etc, instead of relying on instantiation in real datasets as in a repo.

tools/schemacode/schemacode/validator.py
```diff
@@ -396,32 +402,26 @@ def validate_all(
...
-    if debug:
-        itemwise_results = []
+    itemwise_results = []
     matched = False
     match_listing = []
     for target_path in paths_list:
-        if debug:
-            print(f"Checking file `{target_path}`.")
-            print("Trying file types:")
+        lgr.debug(f"Checking file `{target_path}`.")
```
yarikoptic, line 409: here and elsewhere use %-interpolation instead of f-strings to avoid possible errors later on if e.g. `target_path` does contain some `%` and logger would start %-interpolating, thus

```suggestion
        lgr.debug("Checking file `%s`.", target_path)
```

tools/schemacode/schemacode/tests/test_validator.py
TheChymera: > faster then [...] relying on instantiation in real datasets
this is true, but with regard to that we can simply optimize download so that we only fetch what we need: https://github.com/bids-standard/bids-specification/pull/1143

The problem is that the only valid reference for “true” BIDS data are the bids repositories, everything else can get out of sync under the radar.

We support validation for a list of paths, and we test that here: https://github.com/bids-standard/bids-specification/blob/9fd2a9951c45a9038079be257c44f6751ebf5e45/tools/schemacode/schemacode/tests/test_validator.py#L332
but note that this also generates the file list from the bids-examples repo.

In as far as we have hard-coded test input, this is for unit tests where we need the specific strings in the absence of the loaders which generate them, this is the case for testing this function, and we provide the regexes it operates on. Working with real filenames is more of an integration test case... We could do it if we really want to, but again, this would be a wrapper function which applies the patterns to hard-coded values, which might cause both false positives and false negatives as it would be additional logic used nowhere else but in the tests.

TheChymera: if you insist :)

yarikoptic:
for codespell I have submitted separate https://github.com/bids-standard/bids-specification/pull/1148 to fix it asap.

tools/schemacode/bidsschematools/validator.py
```diff
@@ -714,9 +780,9 @@ def _get_directory_suffixes(my_schema):
 
 def validate_bids(
     bids_paths,
+    dummy_paths=False,
```
yarikoptic, line 783: I don't like this name -- there is nothing "dummy" about them...  I would like to make separation of two modes to be made more "explicit", not a flag in API.
I was to propose a change, but let me first clarify - do you foresee to validation of a collection of individual folders, e.g. `['sub-01', 'sub-02']` -- would that be legit? 
I kinda like an idea to be able to validate a specific subject[/session] but I wonder if that would be feasible in the long run as more checks would be added -- would it mean to not check anything about e.g. `rawdata/` or `participants.tsv` at the top level?

tools/schemacode/bidsschematools/validator.py
```diff
@@ -728,6 +794,9 @@ def validate_bids(
     ----------
     paths : str or list of str
```
yarikoptic, line 795: I do like just `paths` better than some `bids_paths` which is in the signature of the function.  Please unify one way or another though.

yarikoptic (changes requested):
Let's possibly figure out better API for having two different modes

Also sent https://github.com/TheChymera/bids-specification/pull/1 with some suggestions/enhancements spotted while reviewing code

tools/schemacode/bidsschematools/validator.py
```diff
@@ -683,6 +680,75 @@ def log_errors(validation_result):
...
+        expansions.insert(
+            0,
+            {
+                "regex": f".*?(?P<replace1>{datatype}/).*?",
+                "replace1": "",
+            },
+        )
+
+    expanded_regexes = []
+    lgr.debug("Applying inheritance expansion to:\n`%s`", regex_string)
```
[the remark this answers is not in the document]
TheChymera: Honestly I think this messes things up more, but I haven't used logging with grep too often, so maybe your consideration is more important :)

tools/schemacode/bidsschematools/validator.py
```diff
@@ -714,9 +780,9 @@ def _get_directory_suffixes(my_schema):
 
 def validate_bids(
     bids_paths,
+    dummy_paths=False,
     schema_reference_root="{module_path}/data/",
     schema_version=None,
-    debug=False,
     report_path=False,
     suppress_errors=False,
```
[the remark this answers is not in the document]
TheChymera: you're commenting on a line which was removed :3

tools/schemacode/bidsschematools/validator.py
yarikoptic: > > there is nothing "dummy" about them
> 
> I think the name is perfectly accurate: [“dummy, noun. An imitation of a real or original object, intended to be used as a practical substitute.”](https://www.wordnik.com/words/dummy)

it doesn't matter if they are *real* or *substitute* since this particular validator does not care to look into content of the files. They can as well/better be *real* paths! The critical difference IMHO that they are paths to files, and thus are not recursively navigated to pick up more paths via `_get_paths` (note that name `_get_paths` is not descriptive of what it means to get_paths from the paths it gets... may be `_get_files_recursively` would have been a bit more descriptive).  

So, in principle, if all `paths` point to files not directories, there should be no point in calling `_get_paths`, right?  So then one solution could be to do `isdir` on each path and if a directory, and only if dir, expand into a list of files under it.  And then this flag could be avoided entirely.

But I don't like this suggestion of mine because it makes it somewhat "implicit" and thus in some cases would consider some `sub-01` user-provided folder to be a filename depending on the status of filesystem/invocation.  That is why I would have preferred a more explicit separation:

There could be `_check_filenames` and `validate_bids` (already there, entry point function for a user) where 
- `_check_filanemes` - would expect to operate only on filenames relative to the top of the dataset (e.g. `sub-01/func/sub-01_bold.nii.gz'), and not require any content/physical presence. Consider it the test for `NOT_INCLUDED` (1) of current bids-validator errors.  I used `_check` to initiate collection of `_checks` which validator should slowly grow. E.g. you could right away add a `_check_valid_json` which would check for `JSON_INVALID` error condition.
- `validate_bids` would call `_get_paths` to expand/filter, figure out the top of the dataset, convert to relative paths and then call `_check_filenames` or whatever other checks it would know about.

tools/schemacode/bidsschematools/validator.py
TheChymera: ok

tools/schemacode/bidsschematools/validator.py
TheChymera: > it doesn't matter if they are real or substitute since this particular validator does not care to look into content of the files.

it matters, because while the validator does not check the files (yet) it needs to check whether they are files or directories.

> They can as well/better be real paths!

yes, the can, and ideally should. This entire thing was introduced because you wanted us to test non-existent paths for validator inheritance. It is however also useful for the DANDI API evaluation, as well as other similar use cases people might want to use the library for, so it seems worthwhile in addition to the original rationale.

> So, in principle, if all `paths` point to files not directories, there should be no point in calling `_get_paths`, right?

Yes, except when they are dummy paths, i.e. not real paths, in that case you cannot check either isfile or isdir

> in some cases would consider some sub-01 user-provided folder to be a filename

yes, that is the issue with just checking for isdir and assuming that everything which is not a dir, including a dummy path, is actually a file.

I'm not entirely sure why we should separate this into a separate top-level function, in effect what we are doing is simply ignoring `NOT_INCLUDED` errors... perhaps the better plan would be to parameterize ignoreable errors once they are codified (though that could get convoluted). In any case, for the time being, since we operate on a list of paths in the end anyway, that simply means skipping the “smart” path acquisition bits, which is what we now do.

yarikoptic: > it matters, because while the validator does not check the files (yet) it needs to check whether they are files or directories.

so - "not yet", and is that check a part of regex? if not -- could be a separate `_check_pathtypes` check AFAIK. If part of regex (directories must have trailing `/`) -- then again no need to incorporate `isdir` within `_check_filenames` (might better become `_check_paths` then) - assignment of trailing `/` would be done by `_get_paths` based on the "type" (dir vs file)

> ... , in that case you cannot check either isfile or isdir

again -- you are not doing that discrimination "yet" or at all in the regex based checking.  I think it is valuable to keep design/implementation clean as if you have checks which are based purely on regexes - keep it that way, do not mix in all kinds of other checks.  If you had some other implementation, we would consider different design.  ATM I am just asking about cleaner separation of regex matching functionality which does not need any real path/directory.

> I'm not entirely sure why we should separate this into a separate top-level function

because then you can unittest that function much better by itself, instead of testing the entire larger beast, and require instantiation of real paths etc.  Moreover later we would be able to annotate checks either they need real paths/directories or not, thus making it easy to go through checks which do not require real paths if we do not have them.

Note -- for DANDI archive we would indeed need entry into validation which would provide just the paths and not just to checking by regexes. So we indeed might need another function or options to `validate_bids` to operate  solely on a list of paths, and then use only checks which do not need real files.  But we better arrive to that point, and again I would not call those paths "dummy" -- they would be real paths for real files but "on the archive" i.e. just not available locally.

TheChymera: > because then you can unittest that function much better by itself, instead of testing the entire larger beast, and require instantiation of real paths etc.

But passing a parameter which controls whether directory walk is invoked prevents us from testing those parts of the “larger beast”. Speaking of larger beasts, there are nevertheless parts of the function, such as bids schema dir loading and reporting which would need to remain not equal (repeated code), but identical (splitting off a function for the commonalities between the two top-level functions). That would make everything more nested overall, and I think would contravene rather than support “cleaner separation”.

> keep design/implementation clean as if you have checks which are based purely on regexes

All checks are always based purely on regexes, there is no exception and I don't think there should ever to be one, even when we end up checking file contents — that would be a subsequent check after we get the regex match paths.

> they would be real paths for real files but "on the archive"

That is specific to the DANDI use case, as far as the validator is concerned those paths are just strings with no meaningful parent directories and therefore no known location in reality :) For the validator, the main utility of such an option would be automatically generated or hard-coded strings for testing (as are already present in this PR), in which case they are purely theoretical dummy paths.

In any case I think we have long departed the realm of inheritance support, and I would need some of these features to fix active bugs (e.g. https://github.com/bids-standard/bids-specification/issues/1155 ). If you insist on the refactoring, could you perhaps open an issue about that? We could also use this opportunity to get broader feedback.

yarikoptic (approved):
I still don't like `dummy_paths` and question some other options naming, but others approved and we can't keep this PR hostage - will be accumulating conflicts etc. Could be RFed later (but soonish).

TheChymera:
@effigies interested in approving this? I can merge it myself now, but I heard 2 approvals before go is best practice :)

TheChymera:
@effigies interested in approving this as well? I knew there was something I wanted to bring up in the meeting today :)

TheChymera:
@sappelhoff interested in merging this? :)

sappelhoff:
thanks @TheChymera

PR #1125, opened 2022-06-14
[FIX] Clarify MEG empty-room recommendations

robertoostenveld:
@mcvinding noted to me that the recommendation for empty-room MEG recordings was confusing, as it confounds two practices. 

Some labs record an empty-room recording every morning to be shared, that is then to be used for all subjects/sessions recorded during that day.

Other labs record an empty-room recording prior to each experimental session, those are not shared.

The text specified that it is "RECOMMENDED to perform an empty-room recording for each experimental session" (i.e., non -shared) and subsequently it explained how to organize the daily shared empty-room recordings in a `sub-emptyroom/ses-YYYYMMDD` structure.

The updated text splits the two and does not prescribe the recommended `ses-YYYYMMDD` practice for empty-room recordings done prior to each experimental session, as those are more logically shared with the actual subject and session.

### DIFF
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -610,7 +610,7 @@ It should be possible to query empty-room recordings just like usual subject
 recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
 SHOULD be present as well.
 
-Example:
+Example 1:
 
 <!-- This block generates a file tree.
 A guide for using macros can be found at
@@ -627,6 +627,31 @@ A guide for using macros can be found at
             "sub-emptyroom_ses-20170801_task-noise_meg.ds": "",
             "sub-emptyroom_ses-20170801_task-noise_meg.json": "",
             "sub-emptyroom_ses-20170801_task-noise_channels.tsv": "",
+            }
+         }
+      },
+   }
+) }}
+
+
+Example 2:
+
+<!-- This block generates a file tree.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_filetree_example(
+   {
+   "sub-control01": {
+      "ses-01": {
+         "sub-01_ses-01_scans.tsv": "",
+         "meg": {
+            "sub-control01_ses-01_task-rest_meg.ds": "",
+            "sub-control01_ses-01_task-rest_meg.json": "",
+            "sub-control01_ses-01_task-rest_channels.tsv": "",
+            "sub-control01_ses-01_task-noise_meg.ds": "",
+            "sub-control01_ses-01_task-noise_meg.json": "",
+            "sub-control01_ses-01_task-noise_channels.tsv": "",
             }
          }
       },
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -589,7 +589,7 @@ has to be updated, then for MEG it could be considered to be a new session.
 ## Empty-room MEG recordings
 
 Empty-room MEG recordings capture the environmental and recording system's noise.
-In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session.
+In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session (see `AssociatedEmptyRoom` field in the `*_meg.json` sidecar file).
 
 Empty-room recordings may be collected once per day, where a single empty-room recording may be shared between multiple subjects and/or sessions (see example 1).
 Empty-room recordings can also be collected for each individual experimental session (see example 2).
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -592,7 +592,7 @@ Empty-room MEG recordings capture the environmental and recording system's noise
 In the context of BIDS it is RECOMMENDED to perform an empty-room recording for each experimental session.
 
 If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions.
-For example, they may be collected once a day. 
+For example, they may be collected once a day.
 In that case it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`.
 The label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be set to `noise`.
 If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`.
@@ -622,7 +622,6 @@ A guide for using macros can be found at
       },
    }
 ) }}
-
 
 Example 2:
 
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -592,15 +592,15 @@ Empty-room MEG recordings capture the environmental and recording system's noise
 In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session.
 
 Empty-room recordings may be collected once per day, where one empty-room recording may be used with multiple subjects and/or sessions (see example 1).
-Empty-room recordings can also be collected for each individual experimental session (see example 2). 
+Empty-room recordings can also be collected for each individual experimental session (see example 2).
 
-In the case of empty-room recordings being associated with multiple subjects and/or sessions, it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`. 
-If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`. 
+In the case of empty-room recordings being associated with multiple subjects and/or sessions, it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`.
+If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`.
 The `scans.tsv` file containing the date and time of the acquisition SHOULD also be included.
 The rationale is that this naming scheme will allow users to easily retrieve the empty-room recording that best matches a particular experimental session, based on date and time of the recording.
 It should be possible to query empty-room recordings just like usual subject recordings, hence all metadata sidecar files (such as the `channels.tsv`) file SHOULD be present as well.
 
-In the case of empty-room recordings being collected for the individual experimental session, it is recommended to store the empty-room recording along with that subject and session. 
+In the case of empty-room recordings being collected for the individual experimental session, it is recommended to store the empty-room recording along with that subject and session.
 
 In either case, the label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be set to `noise`.
 
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -606,7 +606,7 @@ In either case, the label for the [`task-<label>`](../99-appendices/09-entities.
 
 Example 1:
 
-One noise recording per day, applying to all subjects for that day.
+One empty-room recording per day, applying to all subjects for that day.
 
 <!-- This block generates a file tree.
 A guide for using macros can be found at
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -588,26 +588,17 @@ has to be updated, then for MEG it could be considered to be a new session.
 
 ## Empty-room MEG recordings
 
-Empty-room MEG recordings capture the environmental and recording system's
-noise.
-In the context of BIDS it is RECOMMENDED to perform an empty-room recording for
-each experimental session.
+Empty-room MEG recordings capture the environmental and recording system's noise.
+In the context of BIDS it is RECOMMENDED to perform an empty-room recording for each experimental session.
 
-If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions. For example, they may be collected once a day. 
-In that case it is RECOMMENDED to store the empty-room recording inside a subject directory
-named `sub-emptyroom`.
-The label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be
-set to `noise`.
-If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the
-empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`.
-The `scans.tsv` file containing the date and time of the acquisition SHOULD
-also be included.
-The rationale is that this naming scheme will allow users to easily retrieve the
-empty-room recording that best matches a particular experimental session, based
-on date and time of the recording.
-It should be possible to query empty-room recordings just like usual subject
-recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
-SHOULD be present as well.
+If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions.
+For example, they may be collected once a day. 
+In that case it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`.
+The label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be set to `noise`.
+If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`.
+The `scans.tsv` file containing the date and time of the acquisition SHOULD also be included.
+The rationale is that this naming scheme will allow users to easily retrieve the empty-room recording that best matches a particular experimental session, based on date and time of the recording.
+It should be possible to query empty-room recordings just like usual subject recordings, hence all metadata sidecar files (such as the `channels.tsv`) file SHOULD be present as well.
 
 Example 1:
 
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -589,7 +589,8 @@ has to be updated, then for MEG it could be considered to be a new session.
 ## Empty-room MEG recordings
 
 Empty-room MEG recordings capture the environmental and recording system's noise.
-In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session (see `AssociatedEmptyRoom` field in the `*_meg.json` sidecar file).
+
+It is RECOMMENDED to explicitly specify which empty-room recording should be used with which experimental run(s) or session(s). This can be done via the [`AssociatedEmptyRoom`](SPEC_ROOT/99-appendices/14-glossary.md#associatedemptyroom-metadata) field in the `*_meg.json` sidecar files.
 
 Empty-room recordings may be collected once per day, where a single empty-room recording may be shared between multiple subjects and/or sessions (see example 1).
 Empty-room recordings can also be collected for each individual experimental session (see example 2).
--- a/src/04-modality-specific-files/02-magnetoencephalography.md
+++ b/src/04-modality-specific-files/02-magnetoencephalography.md
@@ -590,7 +590,7 @@ has to be updated, then for MEG it could be considered to be a new session.
 
 Empty-room MEG recordings capture the environmental and recording system's noise.
 
-It is RECOMMENDED to explicitly specify which empty-room recording should be used with which experimental run(s) or session(s). This can be done via the [`AssociatedEmptyRoom`](SPEC_ROOT/99-appendices/14-glossary.md#associatedemptyroom-metadata) field in the `*_meg.json` sidecar files.
+It is RECOMMENDED to explicitly specify which empty-room recording should be used with which experimental run(s) or session(s). This can be done via the [`AssociatedEmptyRoom`](../99-appendices/14-glossary.md#associatedemptyroom-metadata) field in the `*_meg.json` sidecar files.
 
 Empty-room recordings may be collected once per day, where a single empty-room recording may be shared between multiple subjects and/or sessions (see example 1).
 Empty-room recordings can also be collected for each individual experimental session (see example 2).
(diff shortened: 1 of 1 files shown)


guiomar:
I'm very happy that we finally adopt this practice. While creating the specs many years ago this idea (i.e. having the associated empty rooms inside each subjects folder, when recorded especifically for each subject/session) was discarted, and I felt this produced a folder structure and code to match these files that was unnecessarily complicated. 

For example, we experienced that while organizing the OMEGA repository with hundreds of participants many of them with multiple sessions, and each of them with a emptyroom recording that was dumped into the emptyroom subject mixing them up and making it less intuitive.

@ftadel @Moo-Marc @rcassani I tag you here in case we finally merge this PR, so you are aware for the brainstorm BIDS importer and reorganization of the OMEGA dataset.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -592,7 +592,10 @@ Empty-room MEG recordings capture the environmental and recording system's
 noise.
 In the context of BIDS it is RECOMMENDED to perform an empty-room recording for
 each experimental session.
-It is RECOMMENDED to store the empty-room recording inside a subject directory
+
+If empty-room recordings are not done prior to each individual experimental session, but
+on a daily schedule, one empty-room recording may have to be used with multiple subjects and/or sessions. In that case
+it is RECOMMENDED to store the empty-room recording inside a subject directory
```
hoechenberger, line 598: Store it inside … which subject's directory? The `sub-emptyroom`? or `sub-participant`? I find this now confusing in combination with the two examples provided.

hoechenberger: Ooooh!! I actually missed these last two words! Sorry about the confusion!!

hoechenberger (approved):

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -630,4 +633,29 @@ A guide for using macros can be found at
    }
 ) }}
 
+
+Example 2:
+
```
guiomar, line 638: Sorry I did my suggestions as a commit, instead of as a review.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -592,7 +592,10 @@ Empty-room MEG recordings capture the environmental and recording system's
 noise.
 In the context of BIDS it is RECOMMENDED to perform an empty-room recording for
 each experimental session.
-It is RECOMMENDED to store the empty-room recording inside a subject directory
+
+If empty-room recordings are not done prior to each individual experimental session, but
```
Moo-Marc, line 596: "If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions. For example, they may be collected once a day."

Moo-Marc (changes requested):
Do we want to comment on the possibility of mixing the two approaches?  It could be useful in a large repository mixing data from various sources, or if a lab does both approaches.  The daily recordings could be used by a lab tech, and the subject-specific ones for analysis.  If we allow it, we could merge the two examples.

guiomar:
Yes @Moo-Marc I think it would be a good idea that the two approaches could be merged together

alexrockhill (commented):
This seems very reasonable +1 from me.

If you have 100 subjects and 50 empty room sessions with 2 subjects collected on each day, is there a way not to duplicate each empty room session twice? Maybe to use the Example 1 method but add an `empty_ses` tag to the task data in a sidecar (rest in the example)? Or is the best solution to just store the duplicate? And, if so, can you make a symbolic link and still have it pass BIDS? As I'm writing this, I'm thinking that answers to these questions might be nice to have in the specification also.

hoechenberger:
> Yes @Moo-Marc I think it would be a good idea that the two approaches could be merged together

But then we'd have an ambiguity if both are present. Which one should a software use?

hoechenberger:
> If you have 100 subjects and 50 empty room sessions with 2 subjects collected on each day, is there a way not to duplicate each empty room session twice? Maybe to use the Example 1 method but add an `empty_ses` tag to the task data in a sidecar (rest in the example)? Or is the best solution to just store the duplicate? 

At NeuroSpin, we've done it such that when you convert the recorded data to BIDS, basically everything would be symlinks, so you could "duplicate" stuff a million times if you wanted to and still wouldn't run out of disk space :) We use the `symlink` parameter of MNE-BIDS's `[write_raw_bids()](https://mne.tools/mne-bids/stable/generated/mne_bids.write_raw_bids.html)`.

alexrockhill:
> > If you have 100 subjects and 50 empty room sessions with 2 subjects collected on each day, is there a way not to duplicate each empty room session twice? Maybe to use the Example 1 method but add an `empty_ses` tag to the task data in a sidecar (rest in the example)? Or is the best solution to just store the duplicate?
> 
> At NeuroSpin, we've done it such that when you convert the freshly recorded data to BIDS, basically everything would be symlinks, so you could "duplicate" stuff a million times if you wanted to and still wouldn't run out of disk space :) We use the `symlink` parameter of MNE-BIDS's [`write_raw_bids()`](https://mne.tools/mne-bids/stable/generated/mne_bids.write_raw_bids.html).
> 
> Also makes the conversion much faster ;)

That sounds nice and related to this. I know symlinks are probably mentioned other places in the BIDS spec but my 2 cents would be that it's nice having the recommended strategy all in the same place.

src/04-modality-specific-files/02-magnetoencephalography.md
robertoostenveld: thanks, I used your text.

robertoostenveld:
My proposal is to keep this pull request as just a small clarification of the current text without any extensions to it. It is possible to come up with further improvements for empty room data, but I think that would better be a separate PR that follows this.

hoechenberger (approved):

hoechenberger:
> My proposal is to keep this pull request as just a small clarification of the current text without any extensions to it. It is possible to come up with further improvements for empty room data, but I think that would better be a separate PR that follows this.

Absolutely, this was just a bit of an off-topic discussion :)

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -630,4 +623,28 @@ A guide for using macros can be found at
    }
 ) }}
 
+Example 2:
+
+<!-- This block generates a file tree.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_filetree_example(
```
alexrockhill, line 632: My suggestion is just to add a note here that the empty room recordings can be symlinked and note that it might be helpful in the case where two empty room recordings correspond to the same subject to save data space

alexrockhill: Not everyone might know who is reading this spec without looking it up that symlinks are allowed in BIDS

hoechenberger: @alexrockhill I'd be against adding this here, this is really an "implementation detail" that won't even be platform-agnostic and portable. If a Linux user zips a BIDS structure containing symlinks and sends it to their Windows friends, they'll often into issues (depending on how the zip file was created)

Let's open a separate discussion on the very issue of symlinks, no? :)

alexrockhill: Sounds good

robertoostenveld: are symbolic links allowed? I searched the specification, the word "symlink" or "symbolic" is nowhere to be found. Symbolic links would indeed not work on all file systems and operating systems. Great if they work for your use-case, but  I am not sure whether they should be included in the specification. 

Please do start a separate issue so that non-MEG folks can chime in as well. It would not surprise me if this has already been discussed in the past and that someone can share insights from the past.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,19 @@ has to be updated, then for MEG it could be considered to be a new session.
...
-The `scans.tsv` file containing the date and time of the acquisition SHOULD
-also be included.
-The rationale is that this naming scheme will allow users to easily retrieve the
-empty-room recording that best matches a particular experimental session, based
-on date and time of the recording.
-It should be possible to query empty-room recordings just like usual subject
-recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
-SHOULD be present as well.
+Empty-room MEG recordings capture the environmental and recording system's noise.
+In the context of BIDS it is RECOMMENDED to perform an empty-room recording for each experimental session.
```
ftadel, line 592: Do really recommend to acquire noise recordings before each **session**, i.e. a subject does 3 runs x 3 tasks, you record 9 noise sessions? Or does "experimental session" refer to "all the contiguous sessions of a subject"?
Is this how you do it at the Donders?

I had the impression that at many sites, this is more of a daily thing, done by the lab manager also as a way to track the technical health of the MEG system.

ftadel (changes requested):

src/04-modality-specific-files/02-magnetoencephalography.md
sappelhoff: yes, would be important to be precise here also with regards to what ["session" means in BIDS](https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#definitions)

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,19 @@ has to be updated, then for MEG it could be considered to be a new session.
...
 
-Example:
+If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions.
+For example, they may be collected once a day.
+In that case it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`.
```
ftadel, line 596: "In that case": does that imply that in the former case, it is RECOMMENDED to store it in the subject folder together with the other tasks? If this is the case, maybe it should be made more explicit?

Would it make sense to separate the two typical cases even more clearly, with headers: 
 - case 1: one noise recordings per session folder
 - example 1
 - case 2: one noise recordings per day
 - example 2
?

src/04-modality-specific-files/02-magnetoencephalography.md
hoechenberger: > Do really recommend to acquire noise recordings before each **session**, i.e. a subject does 3 runs x 3 tasks, you record 9 noise sessions?

My understanding is that one "session" is one visit to the lab

src/04-modality-specific-files/02-magnetoencephalography.md
hoechenberger: > Would it make sense to separate the two typical cases even more clearly, with headers?

I like this idea.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,19 @@ has to be updated, then for MEG it could be considered to be a new session.
...
-on date and time of the recording.
-It should be possible to query empty-room recordings just like usual subject
-recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
-SHOULD be present as well.
+Empty-room MEG recordings capture the environmental and recording system's noise.
+In the context of BIDS it is RECOMMENDED to perform an empty-room recording for each experimental session.
 
-Example:
+If empty-room recordings are not collected for each individual experimental session, one empty-room recording may be used with multiple subjects and/or sessions.
+For example, they may be collected once a day.
```
dnacombo, line 595: In case of conflict (a sub-emptyroom exists, and a noise recording exist inside the session of interest), the noise recording inside the session folder should supersede the sub-emptyroom.

src/04-modality-specific-files/02-magnetoencephalography.md
dnacombo: An example 3 could illustrate the case when there is a conflict between a local noise recording in the session, and a sub-emptyroom. Automatically assigning the correct noise recording to a given task recording needs arbitration.

A policy like "if there is no noise recording in a session, and several noise recordings are present in sub-emptyroom for a given day, then the noise recording closest in time should be assigned to the session's recordings".

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,19 @@ has to be updated, then for MEG it could be considered to be a new session.
...
+In that case it is RECOMMENDED to store the empty-room recording inside a subject directory named `sub-emptyroom`.
+The label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be set to `noise`.
+If a [`session-<label>`](../99-appendices/09-entities.md#ses) entity is present, its label SHOULD be the date of the empty-room recording in the format `YYYYMMDD`, that is `ses-YYYYMMDD`.
+The `scans.tsv` file containing the date and time of the acquisition SHOULD also be included.
+The rationale is that this naming scheme will allow users to easily retrieve the empty-room recording that best matches a particular experimental session, based on date and time of the recording.
+It should be possible to query empty-room recordings just like usual subject recordings, hence all metadata sidecar files (such as the `channels.tsv`) file SHOULD be present as well.
+
+Example 1:
```
dnacombo, line 603: One noise recording per day applying to all subjects for that day.

src/04-modality-specific-files/02-magnetoencephalography.md
ftadel: > My understanding is that one "session" is one visit to the lab

Yes, obviously. I got distracted... Please ignore this previous comment about the session.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -630,4 +623,28 @@ A guide for using macros can be found at
    }
 ) }}
 
+Example 2:
```
dnacombo, line 626: One recording per session, stored within the session folder.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,19 @@ has to be updated, then for MEG it could be considered to be a new session.
 
 ## Empty-room MEG recordings
 
-Empty-room MEG recordings capture the environmental and recording system's
-noise.
-In the context of BIDS it is RECOMMENDED to perform an empty-room recording for
```
dnacombo, line 593: A single empty room recording should exist for any given session. Not more.

Moo-Marc: I disagree with this.  Are you saying that if I collect multiple noise recordings, I can't have them all in my BIDS dataset?  See also my comment below about the AssociatedNoiseRecording field.

src/04-modality-specific-files/02-magnetoencephalography.md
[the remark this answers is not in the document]
robertoostenveld: There are two recommendations, one on the lab practice and one on the data organization, and the two are not really aligned. These came out of [BEP008](https://bids.neuroimaging.io/bep008), see page 15, which happened to be the first big BEP that was merged. It might have been better had it been captured in 2018 as not being optimal.

The two practices are both being used, and I personally recommend that everyone does in their lab whatever works for them. I think it would be better if BIDS would restrict itself to data organization, and not to lab procedures. But simply scrapping the "empty-room recording per session" from the spec might lead to the impression that that is not recommended, whereas it actually was the recommended strategy (by consensus) in BEP006.

Should we rephrase, say that both are possible, that "empty-room recording per session" is recommended (to stay in line with BEP006) and give an example for both? Actually, compared to the current text in this PR the only change would then be to also give an example for the "empty-room recording per session".

src/04-modality-specific-files/02-magnetoencephalography.md
robertoostenveld: BIDS uses the 80/20 Pareto Principle: we don't have to deal with all possible situations that might arise (as then we'll continue discussing indefinitely), but only with those 20% of the cases that result in 80% of the value. I think your case 3 falls outside the 80% of the value.

src/04-modality-specific-files/02-magnetoencephalography.md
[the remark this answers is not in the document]
Moo-Marc: I agree with @robertoostenveld.  Still, how would you feel about changing a single word:  "In the context of BIDS it is RECOMMENDED to **associate** an empty-room recording for each experimental session."  This implicitly maintains the idea that it should be performed, while bringing it back to the organization context of BIDS, and the statement now works for both practices.

hoechenberger:
> But I would leave it for MEG BIDS conversion tools to address this question, instead of prescribing it in BIDS itself.

Generally I agree but since I'm also a co-developer of such a tool I know first-hand that too many rules and exceptions in the standard can lead to a mess in the software that aims to implement that standard 😅 

So I'm not veto'ing here, but I want to emphasize that exceptions and multiple ways of doing the same thing WILL lead to problems and bugs.

Moo-Marc:
My automated conversion tool for our newly acquired MEG data currently doesn't fill the AssociatedEmptyRoom field.  (I fill it semi-manually for our OMEGA dataset.)  I feel that if there are multiple noise recordings in a session, they are usually not equivalent.  From my experience, here's why there may be multiple noise recordings in one session:

- One short noise recording to quickly look at the spectrum and evaluate the state of the system.  Another longer one to save for later analysis.
- A problem with one or more sensors was seen in the first recording, requiring defluxing or tuning.  Another noise recording was performed after.
- Environmental conditions changed during the recording session and noted.  A noise recording was performed at the end to capture this.  
- Multiple subjects for a single study are run back to back and one noise recording is collected between each.

In many of these cases, only the user will be able to tell which noise recording should be used with each experimental recording.  This is why I'd prefer not to specify general priority rules for this association in BIDS.  

Stepping back though, I guess the 80/20 rule would also apply here.  In most cases there would only be a single matching noise recording, so additional rules would not be required.

src/04-modality-specific-files/02-magnetoencephalography.md
robertoostenveld: thanks for the suggestion for splitting up the examples @dnacombo. With commit 0aa328e85799aa5f0a951545c8838fddc7ee1eba I have made the two cases more explicit in the text as well.

src/04-modality-specific-files/02-magnetoencephalography.md
robertoostenveld: this has been addressed in 0aa328e85799aa5f0a951545c8838fddc7ee1eba

sappelhoff (approved):
I think this is a nice clarification. Thanks a lot!

robertoostenveld:
> No reference to the recommended "AssociatedEmptyRoom" field of the json sidecar?

There is now the sentence "In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session". That would indeed work by specifying `AssociatedEmptyRoom`. But I don't think that we need to spell that out more explicitly.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,25 @@ has to be updated, then for MEG it could be considered to be a new session.
...
+The rationale is that this naming scheme will allow users to easily retrieve the empty-room recording that best matches a particular experimental session, based on date and time of the recording.
+It should be possible to query empty-room recordings just like usual subject recordings, hence all metadata sidecar files (such as the `channels.tsv`) file SHOULD be present as well.
+
+In the case of empty-room recordings being collected for the individual experimental session, it is recommended to store the empty-room recording along with that subject and session.
+
+In either case, the label for the [`task-<label>`](../99-appendices/09-entities.md#task) entity in the empty-room recording SHOULD be set to `noise`.
+
+Example 1:
+
+One noise recording per day, applying to all subjects for that day.
```
hoechenberger, line 609:
```suggestion
One empty-room recording per day, applying to all subjects for that day.
```

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,25 @@ has to be updated, then for MEG it could be considered to be a new session.
...
-empty-room recording that best matches a particular experimental session, based
-on date and time of the recording.
-It should be possible to query empty-room recordings just like usual subject
-recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
-SHOULD be present as well.
+Empty-room MEG recordings capture the environmental and recording system's noise.
+In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session.
 
-Example:
+Empty-room recordings may be collected once per day, where one empty-room recording may be used with multiple subjects and/or sessions (see example 1).
```
hoechenberger, line 594:
```suggestion
Empty-room recordings may be collected once per day, where a single empty-room recording may be shared between multiple subjects and/or sessions (see example 1).
```

hoechenberger (commented):
Two tiny suggestion re phrasing

hoechenberger:
... also because AssociatedEmptyRoom happens on a per-run and not on a per-session level ...

hoechenberger:
> can you make a suggested edit to clarify it?
> 
> 

I can try tonight, currently I'm only on my phone :)

Moo-Marc:
Thanks all, this looks good to me.  Regarding the `AssociatedEmptyRoom` field, how about this:

"In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session (see `AssociatedEmptyRoom` field in the [*_meg.json sidecar file](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#sidecar-json-_megjson))".

Adjusting or removing the link as required - I'm not sure what's the convention.

src/04-modality-specific-files/02-magnetoencephalography.md
```diff
@@ -588,26 +588,25 @@ has to be updated, then for MEG it could be considered to be a new session.
...
-The `scans.tsv` file containing the date and time of the acquisition SHOULD
-also be included.
-The rationale is that this naming scheme will allow users to easily retrieve the
-empty-room recording that best matches a particular experimental session, based
-on date and time of the recording.
-It should be possible to query empty-room recordings just like usual subject
-recordings, hence all metadata sidecar files (such as the `channels.tsv`) file
-SHOULD be present as well.
+Empty-room MEG recordings capture the environmental and recording system's noise.
+In the context of BIDS it is RECOMMENDED to associate an empty-room recording to each experimental session (see `AssociatedEmptyRoom` field in the `*_meg.json` sidecar file).
```
hoechenberger, line 592: I rewrote this. But I'm not sure how to produce the correct link. Could sb please help?

```suggestion

It is RECOMMENDED to explicitly specify which empty-room recording should be used with which experimental run(s) or session(s). This can be specified via the [`AssociatedEmptyRoom`](99-appendices/14-glossary.html#associatedemptyroom-metadata) field in the `*_meg.json` sidecar files.
```

sappelhoff: You had it almost right, we just need to:

- add a "SPEC_ROOT/" in front of internal links
- use `.md` as extensions, instead of html

sappelhoff: ah no ... "SPEC_ROOT" was wrong, ... `../` was needed instead. We only need "SPEC_ROOT" when using internal links from within schema files.

PR #1131, opened 2022-06-20
[FIX] Fix MEG datatype for tabular data

rob-luke:
Small change I found while preparing the fNIRS BEP


sappelhoff (approved):
LGTM, thanks for the catch!

tsalo (approved):
Thanks for this!

PR #1135, opened 2022-07-07
[INFRA] Rename schema tools package and build with SCHEMA_VERSION

effigies:
Rather than using versioneer and tying the schema package to the repository version, this proposes using `SCHEMA_VERSION`. This will allow us to upload to pypi without relying on the specification release schedule, and the `SCHEMA_VERSION` semver rules should be more informative.

Thoughts?


sappelhoff (commented):
sounds reasonable to me 🤔 what about the `/tools/schemacode/schemacode/_version.py` file though? Shouldn't that be removed as well?

effigies:
Should drop Python 3.6, and I guess we'll need to require `importlib_resources` for Python <3.9.

tools/schemacode/schemacode/__init__.py
```diff
@@ -1,12 +1,19 @@
 """A Python package for working with the BIDS schema."""
 from . import render, schema, utils
 
+try:
+    from importlib.resources import as_file, files
+except ImportError:  # PY<3.9
+    from importlib_resources import files, as_file
+
+from . import render, schema, utils
+
```
tsalo, line 10: I think this can be removed given the imports in line 2.

tsalo (commented):
All of the changes make sense to me. I am surprised that the docs-building CI can't find the SCHEMA_VERSION file though. Any clue what's causing that?

TheChymera (approved):

PR #1137, opened 2022-07-08
logic update addressing literals being moved under the `value` key

TheChymera:
schema representation changed again, adapted logic to continue to populate list of suffixes which are directories.


tsalo (approved):
LGTM. Thanks!

sappelhoff (approved):
xref: follow up to https://github.com/bids-standard/bids-specification/pull/1107

PR #1138, opened 2022-07-08
[SCHEMA] Record which version of BIDS specification was used for validation.

TheChymera:
Note that this is not `SCHEMA_VERSION` which is a syntax version for schema compatibility. This records the specific `bids_schema_version`, which should identify the schema contents rather than the syntax.


tools/schemacode/schemacode/validator.py
```diff
@@ -763,6 +763,20 @@ def validate_bids(
...
+    # Record schema version.
+    # Not sure whether to incorporate in validation_result.
+    if bids_schema_dir == os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "data",
+        "schema",
+    ):
+        # Declare we are using live version,
+        # string will evaluate as larger than numbered versions.
+        schema_version = "99999.0.0"
```
sappelhoff, line 775: could we also declare `"dev"` here? the 99999 looks a bit unfamiliar

TheChymera:
Regarding the two lines not covered:
We don't currently ship any library of past BIDS versions, nor do we support a lot of historical versions, so there's no way to check for correct schema specifications other than live at this point.

TheChymera:
@yarikoptic want to review? I need 2 reviews to merge ^^

TheChymera:
@effigies @tsalo interested? The 2 failing tests are unrelated to this PR.

TheChymera:
@effigies this adds the BIDS version file to the schema directory, and loads it in the validator, as discussed last Thursday.
Currently this file is not edited for release, best for a separate PR I think.

tools/schemacode/bidsschematools/tests/test_validator.py
```diff
@@ -343,6 +368,15 @@ def test_bids_datasets(bids_examples, tmp_path):
...
     assert len(result["path_tracking"]) == 0
 
+    # Is the schema version recorded correctly?
+    schema_path = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "../data/schema",
+    )
+    with open(os.path.join(schema_path, "BIDS_VERSION")) as f:
+        expected_version = f.readline().rstrip()
+    assert result["bids_version"] == expected_version
```
effigies, line 378: We seem to do this a lot just to validate. I think we could do the following in `__init__.py`:

```Diff
 version_file = files("bidsschematools.data") / "schema" / "SCHEMA_VERSION"
 with as_file(version_file) as path:
     __version__ = path.read_text().strip()
+
+bids_version_file = files("bidsschematools.data") / "schema" / "BIDS_VERSION"
+with as_file(bids_version_file) as path:
+    bids_version = path.read_text().strip()
```

It's done once on startup, allows access to the package and any downstream packages, and this block becomes:

```suggestion
    import bidsschematools as bst
    assert result["bids_version"] == bst.bids_version
```

TheChymera:
@effigies addressed points raised.

tools/schemacode/bidsschematools/tests/test_validator.py
```diff
@@ -343,6 +343,15 @@ def test_bids_datasets(bids_examples, tmp_path):
     # Have all files been validated?
     assert len(result["path_tracking"]) == 0
 
+    # Is the schema version recorded correctly?
+    schema_path = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "../data/schema",
```
yarikoptic, line 349: if using `os.path` then make it 
```suggestion
        os.pardir, "data", "schema",
```
since not sure if would mix ok on windows

tools/schemacode/bidsschematools/__init__.py
```diff
@@ -13,5 +13,9 @@
 ]
 
 version_file = files("bidsschematools.data") / "schema" / "SCHEMA_VERSION"
-with as_file(version_file) as path:
-    __version__ = path.read_text().strip()
+with as_file(version_file) as f:
+    __version__ = f.read_text().strip()
```
yarikoptic, line 17: I don't think `SCHEMA_VERSION` should be the version of the library , but that is separate issue from this PR

src/schema/BIDS_VERSION
```diff
@@ -0,0 +1 @@
+v1.7.1-dev
```
yarikoptic, line 1: who / how would produce this file?  It should become part of the release/post-release process of bids-specification or otherwise it just would get outdated and thus reporting an incorrect version to the user!  I wonder if @bids-maintenance folks would allow it being hardcoded here for that reason.  And if they do , http://github.com/bids-standard/bids-specification/blob/HEAD/Release_Protocol.md should reflect the need to adjust this file as well.

tools/schemacode/bidsschematools/tests/test_schema.py
```diff
@@ -1,7 +1,30 @@
...
 
-from bidsschematools import schema
+from bidsschematools import __bids_version__, schema
+
+
+def test__get_bids_version(tmp_path):
+    # Is the version being read in correctly?
+    schema_path = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        "../data/schema",
```
yarikoptic, line 13:
```suggestion
        os.pardir, "data", "schema",
```
as elsewhere

yarikoptic (changes requested):

tools/schemacode/bidsschematools/__init__.py
TheChymera: yep, we have a dedicated issue for that

src/schema/BIDS_VERSION
TheChymera: Yes, the plan is for it to be auto-generated for release, but within the actual git tracking it's equivalent to https://github.com/bids-standard/bids-specification/blob/c80582efbee220b6f16e20ca795d25179bd95285/mkdocs.yml#L1

TheChymera: @yarikoptic ok, I added this to the release protocol.

yarikoptic (approved):
In general I hate duplication and thought to suggest some approach which would avoid it, but then it would require coding of that logic in every tool which would decide to work with BIDS schema content directly.  That would be suboptimal. So, despite duplication and "avoidable" complication of Release Protocol I approve this PR.

PR #1143, opened 2022-07-14
[INFRA] Reduced BIDS example download sizes

TheChymera:
Reduces test suite duration from 38s to about 6s locally.

### DIFF
--- a/tools/schemacode/bidsschematools/tests/conftest.py
+++ b/tools/schemacode/bidsschematools/tests/conftest.py
@@ -14,13 +14,13 @@ lgr = logging.getLogger()
 # good proportion modalities would be represented, as well as datasets exemplifying
 # tricky edge-cases, such as directory pseudo-files.
 BIDS_SELECTION = [
-    "asl003", # anat, perf, _asl, _T1w
-    "eeg_cbm", # eeg
-    "hcp_example_bids", # anat, fmap
-    "micr_SEM", # micr, SEM
-    "micr_SPIM", # micr, SPIM, .ome.tif
-    "pet003", # pet, anat
-    "qmri_tb1tfl", # fmap, _TB1TFL
+    "asl003",  # anat, perf, _asl, _T1w
+    "eeg_cbm",  # eeg
+    "hcp_example_bids",  # anat, fmap
+    "micr_SEM",  # micr, SEM
+    "micr_SPIM",  # micr, SPIM, .ome.tif
+    "pet003",  # pet, anat
+    "qmri_tb1tfl",  # fmap, _TB1TFL
 ]
 
 
--- a/tools/schemacode/bidsschematools/tests/conftest.py
+++ b/tools/schemacode/bidsschematools/tests/conftest.py
@@ -9,9 +9,20 @@ from bidsschematools import schema, utils
 
 lgr = logging.getLogger()
 
+BIDS_SELECTION = [
+    "asl003",
+    "eeg_cbm",
+    "hcp_example_bids",
+    "micr_SEM",
+    "micr_SPIM",
+    "pet001",
+    "pet003",
+    "qmri_tb1tfl",
+]
+
 
 @pytest.mark.no_network
-def get_gitrepo_fixture(url):
+def get_gitrepo_fixture(url, whitelist):
     @pytest.fixture(scope="session")
     def fixture():
         path = tempfile.mktemp()  # not using pytest's tmpdir fixture to not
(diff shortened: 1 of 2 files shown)


sappelhoff:
@TheChymera could you please submit future PRs from a branch on your fork, instead of from a branch off this repo? That'd help us keeping branches tidy here.

TheChymera:
@sappelhoff sure, sorry for the mix-up.

tsalo:
Running tests on a subset of available example datasets seems like a good idea to me. Are there specific reasons why you chose these particular datasets?

tools/schemacode/schemacode/tests/conftest.py
```diff
@@ -9,19 +9,45 @@
...
 
+BIDS_SELECTION = [
+    "asl003",
+    "eeg_cbm",
+    "hcp_example_bids",
+    "micr_SEM",
+    "micr_SPIM",
+    "pet001",
+    "pet003",
+    "qmri_tb1tfl",
```
tsalo, line 20: Can you add in-line comments explaining each of the choices? Something like `# asl, t1w, longitudinal` would probably be sufficient in most cases.

TheChymera: There isn't really any particular reason, other than me trying to cover a wide range of modalities and avoiding some of the heavier datasets... I guess I could drop one of the `pet*` ones and add something else instead. Also I'll probably want to add the OME.ZARR example once it gets merged, as it also allows us to test whether directory pseudo-files are handled correctly.

tsalo: I don't have an issue with the `pet` datasets. 6 seconds is _really_ fast!

Still, having some idea of what parts of the spec each example dataset covers will help us identify any gaps in the list.

sappelhoff: > Also I'll probably want to add the OME.ZARR example once it gets merged

That can be done once this is addressed: https://github.com/bids-standard/bids-examples/pull/315#issuecomment-1183408044

TheChymera: @tsalo ok, added comments, not entirely sure the inline ones really help that much, since it's mostly in the archive name, but they don't bother much either :)

TheChymera:
@effigies @tsalo rebased, remaining test failures are (1) typo fixed [elsewhere](https://github.com/bids-standard/bids-specification/pull/1145/commits/ca8016873bf0e1dbff87fb5fd7b5b4f1a4676d2b) and (2) linkchecker --- both are unrelated to this PR.

TheChymera:
diff coverage is 100%, percentage change is artificial.

effigies (approved):

sappelhoff (approved):

TheChymera:
@sappelhoff is it fixed, should I rebase?

sappelhoff:
not yet fixed unfortunately, I don't even know what went wrong. Probably the problem was that #1000 was not rebased prior to merge, so the positive checks were misleading. If you want to look into a fix, that'd be welcome.

sappelhoff:
@TheChymera I solved one issue in #1153, however the remaining issue seems valid:

> /tmp/tmp203jv8xe/eeg_matchingpennies/sourcedata/sub-06/eeg/sub-06_task-matchingpennies_eeg.xdf` file was not matched by any regex schema entry

note that the XDF file that triggers the error is in `sourcedata`, where we may place any file formats.

TheChymera:
>  in `sourcedata`, where we may place any file formats.

This is not yet implemented by the validator, best for a different PR, made an issue to track that https://github.com/bids-standard/bids-specification/issues/1155 .

For the time being the fix is to revert to the original PR data listing, not sure how/why `eeg_cbm` got replaced with `eeg_matchingpennies` 🤔. Though adding one of the datasets which have such directories will be good as a test case once support is in.

sappelhoff:
> For the time being the fix is to revert to the original PR data listing, not sure how/why eeg_cbm got replaced with eeg_matchingpennies thinking. Though adding one of the datasets which have such directories will be good as a test case once support is in.

I did that in https://github.com/bids-standard/bids-specification/pull/1143/commits/b82ae523f484bc377d248a3e86070e44f5d5bb48 for precisely the reason that eeg_cbm is very minimal and eeg_matchingpennies is a more complete example that uses more BIDS features.

But now that you opened an issue to track one problem we identified, I am fine with merging this. Thanks!

PR #1145, opened 2022-07-14
[ENH] Filtering child datasets from parent validation

TheChymera:
Formerly, the validator would index child datasets (i.e. derivatives) automatically when pointed at a parent dataset, and complain about having found more than one `README`, `dataset_description.json`, etc.

### DIFF
--- a/tools/schemacode/bidsschematools/validator.py
+++ b/tools/schemacode/bidsschematools/validator.py
@@ -37,8 +37,6 @@ def _get_paths(
     * Figure out how to return paths from BIDS root.
     * Deduplicate paths (if input dirs are subsets of other input dirs), might best be done at the
         very end.
-    * Specifying file paths currently breaks because they are truncated based on the bids_paths
-        input.
     """
     exclude_subdirs = [
         rf"{os.sep}.dandi",
@@ -55,12 +53,19 @@ def _get_paths(
     ]
 
     path_list = []
+    bids_root_found = False
     for bids_path in bids_paths:
         bids_path = os.path.abspath(os.path.expanduser(bids_path))
         if os.path.isfile(bids_path):
             path_list.append(bids_path)
             continue
         for root, dirs, file_names in os.walk(bids_path, topdown=True):
+            if "dataset_description.json" in file_names:
+                if bids_root_found:
+                    dirs[:] = []
+                    file_names[:] = []
+                else:
+                    bids_root_found = True
             if any(root.endswith(i) for i in pseudofile_suffixes):
                 # Add the directory name to the validation paths list.
                 path_list.append(f"{root}/")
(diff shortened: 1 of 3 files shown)


effigies (approved):

TheChymera:
@effigies resolved rename conflict, there was apparently also a typo in master which the test suite warned me about.

tools/schemacode/bidsschematools/validator.py
```diff
@@ -55,12 +53,19 @@ def _get_paths(
...
         if os.path.isfile(bids_path):
             path_list.append(bids_path)
             continue
         for root, dirs, file_names in os.walk(bids_path, topdown=True):
+            if "dataset_description.json" in file_names:
+                if bids_root_found:
+                    dirs[:] = []
+                    file_names[:] = []
+                else:
+                    bids_root_found = True
```
effigies, line 68: I think this is probably doing the right thing under normal cases, but it feels a bit loose.

For example, if we pass multiple BIDS root directories as `bids_paths`, the first one will be descended, but the others will not. If we pass the parent directory of a BIDS directory, `dataset_description.json` will not be found at the first level, but the second. If we pass the parent of multiple BIDS directories, we will index the first BIDS dataset passed to us by `os.walk()`.

I wonder if we want something a bit less flexible (and hopefully more predictable under weird inputs):

```suggestion
    path_list = []
    for bids_path in bids_paths:
        bids_path = os.path.abspath(os.path.expanduser(bids_path))
        if os.path.isfile(bids_path):
            path_list.append(bids_path)
            continue
        description = os.path.join(bids_path, "dataset_description.json")
        if not os.path.exists(description):
            raise FileNotFoundError(f"Could not find {description}. Is this a BIDS dataset?")
        first = True
        for root, dirs, file_names in os.walk(bids_path, topdown=True):
            if not first and "dataset_description.json" in file_names:
                dirs[:] = []
                file_names[:] = []
            first = False
```

TheChymera: @effigies thanks for the feedback.

> if we pass multiple BIDS root directories as `bids_paths`, the first one will be descended, but the others will not.
In such a case a notification would be issued prior to the code logic reaching this function:
https://github.com/bids-standard/bids-specification/blob/1cd8cad58d8b31257c0c8c6edffb73ce3948ba3f/tools/schemacode/bidsschematools/validator.py#L601

> If we pass the parent directory of a BIDS directory, `dataset_description.json` will not be found at the first level, but the second.
This seems like ok behaviour, no?

> If we pass the parent of multiple BIDS directories, we will index the first BIDS dataset passed to us by os.walk()
This also seems correct.

The idea I think is that validation with one consistent BIDS version requires either zero (fallback to newest BIDS version) or one (BIDS dataset validation) `dataset_description.json` file. I agree that the logic could be improved, perhaps by having `select_schema_dir()` walk all the child paths and issue a warning as part of a different PR. But in any case, given that we descend the tree here as part of finding files, this stop would need to happen again here. Otherwise we'd need a refactoring which merges `select_schema_dir()` and `_get_paths()`... but I'm not sure that would actually simplify logic, even if at first glance it appears it might.

PR #1114, opened 2022-06-02
[SCHEMA] Add "common.yaml" standard error schema rules

effigies:
Reviving #1067, which got closed because I deleted the `schema-sprint` branch.

PR #1146, opened 2022-07-14
[SCHEMA] Common error messages for rules that have no check in schema.

rwblair:
Modification of  #1114

### DIFF
--- a/src/schema/rules/errors.yaml
+++ b/src/schema/rules/errors.yaml
@@ -4,77 +4,70 @@
 # BIDS Validator Original Issue Code #26
 NiftiHeaderUnreadable:
     code: NIFTI_HEADER_UNREADABLE
-    description: |
+    message: |
         We were unable to parse header data from this NIfTI file.
         Please ensure it is not corrupted or mislabeled.
     level: error
-    validation: shallow
     selectors:
     - extension in [".nii", ".nii.gz"]
 
 # BIDS Validator Original Issue Code #27
 JsonInvalid:
     code: JSON_INVALID
-    description: |
+    message: |
         Not a valid JSON file.
     level: error
-    validation: shallow
     selectors:
     - extension == ".json"
 
 # BIDS Validator Original Issue Code #28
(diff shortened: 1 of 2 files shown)


src/schema/rules/errors.yaml
```diff
@@ -0,0 +1,252 @@
...
+
+# BIDS Validator Original Issue Code #26
+NiftiHeaderUnreadable:
+    code: NIFTI_HEADER_UNREADABLE
+    message: |
+        We were unable to parse header data from this NIfTI file.
+        Please ensure it is not corrupted or mislabeled.
+    level: error
+    selectors:
+    - extension in [".nii", ".nii.gz"]
```
effigies, line 12: Does this syntax need to be updated?

rwblair: yes

effigies (approved):

src/schema/rules/errors.yaml
```diff
@@ -0,0 +1,252 @@
+---
+# This file describes rules for broad "standard" errors in the specification.
+
+# BIDS Validator Original Issue Code #26
+NiftiHeaderUnreadable:
+    code: NIFTI_HEADER_UNREADABLE
```
yarikoptic, line 6: Me curious: Why code isn't just the key of the entry?

rwblair: The idea was towards consistency, there are places in the schema that have custom error messages use the `code` and `message` keys:
https://github.com/bids-standard/bids-specification/blob/a93d4f2a27b2b79ba569a1d02f1e7f560af3525f/src/schema/rules/sidecars/asl.yaml#L27

Though those are nested in that `issue` key. I don't know if we want to consider lifting the `code` and `message` there a level higher or not. The schema needs its own schema.

sappelhoff (approved):

sappelhoff:
I think I should have rebased this before merging. The changes lead to a broken CI in `master` now -- can you have a look @rwblair?

PR #1148, opened 2022-07-18
[FIX] fix typo in PR title which got into src/CHANGES.md

yarikoptic:
That leads to failing CI as in
https://github.com/bids-standard/bids-specification/pull/1120/files#file-src-changes-md-L13

Ideally we should add spell checking for PR title, but those would keep changing,
would need separate workflow and react to changes to titles and then have no way
to easily say "ignore". May be src/CHANGES.md should be excluded... separate issue would
that be


tsalo:
Won't the changelog updater automatically fix it since @effigies corrected the PR title, the next time there's a merged PR to trigger it?

yarikoptic:
> Won't the changelog updater automatically fix it since @effigies corrected the PR title, the next time there's a merged PR to trigger it?

I hope that changelog updated considers only relevant PRs since the last release and not all PRs historically since that would be growing burden to avoid. ... indeed looking at https://github.com/bids-standard/bids-specification/blob/HEAD/.circleci/config.yml#L176  it considers only the last merge commit, so no retroactive fixes

yarikoptic:
> It will happily modify any entry since the first GitHub version if we go and change the titles.

how would it do that if it works only on the latest one (`git log -1`)??? point blind me to the code

effigies:
https://github.com/bids-standard/bids-specification/blob/e347f304b18ef5c230c8d223a85e15eac18c0909/.circleci/config.yml#L108-L120

PR #1159, opened 2022-07-23
[INFRA] Added error tests

TheChymera:
We should also test whether the validation fails when it should fail. This PR does that, leveraging https://github.com/bids-standard/bids-error-examples

### DIFF
--- a/tools/schemacode/bidsschematools/tests/test_validator.py
+++ b/tools/schemacode/bidsschematools/tests/test_validator.py
@@ -3,6 +3,7 @@ import shutil
 
 import pytest
 
+from .conftest import BIDS_ERROR_SELECTION
 
 def test__add_entity():
     from bidsschematools.validator import _add_entity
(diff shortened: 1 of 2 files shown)


sappelhoff (approved):

effigies (approved):

tools/schemacode/bidsschematools/tests/test_validator.py
```diff
@@ -362,3 +364,23 @@ def test_broken_json_dataset(bids_examples, tmp_path):
...
         report_path=True,
     )
+
+
+def test_error_datasets(bids_error_examples):
+    from bidsschematools.validator import validate_bids
+
+    schema_path = "{module_path}/data/schema/"
+
+    for i in os.listdir(bids_error_examples):
```
yarikoptic, line 374: Ideally (so we see on which example it fails and which not instead of failing on the first one) this looping should be done via `@pytest.mark.parametrize` over this test but I am not sure if that is feasible to merry with `bids_error_examples` fixture.  @jwodder -- do you know if that is possible?

effigies: What about:

```Python
from bidsschematools.conftest import BIDS_ERROR_SELECTION

@pytest.mark.parametrize("dataset", BIDS_ERROR_SELECTION)
def test_error_datasets(bids_error_examples, dataset):
    ...
```

jwodder: @effigies *Iff* `BIDS_ERROR_SELECTION` equals the set of non-dot entries in `bids_error_examples`, then that would work.

yarikoptic: short term this `BIDS_ERROR_SELECTION` should work but IMHO there should be no "whitelisting" here -- all metadata pertinent to what datasets are good or bad should be within those respective example repositories. That would lead for those `*_SELECTION` variables to disappear and thus this solution no longer work.  But I guess for now it is better than `for` loop indeed ;)

yarikoptic (approved):

yarikoptic:
It is green and plenty of approvals and dealing with tests, so cannot break anything. I will just merge, please alert me if I shouldn't use such super-powers ;)

PR #1162, opened 2022-07-25
[INFRA] Using label string as in contemporary schema

TheChymera:
Historical left-over, pointed out by @yarikoptic

PR #1164, opened 2022-07-26
[FIX] clarify alphanumeric

yarikoptic:
since `non-`, regex should have `^`

and place 0-9 first for consistency (complimentary to #1162 by @TheChymera )

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -2986,7 +2986,7 @@ TaskName:
     Name of the task.
     No two tasks should have the same name.
     The task label included in the file name is derived from this `"TaskName"` field
-    by removing all non-alphanumeric (`[a-zA-Z0-9]`) characters.
+    by removing all non-alphanumeric (`[^0-9a-zA-Z]`) characters.
     For example `"TaskName"` `"faces n-back"` will correspond to task label
     `facesnback`.
   type: string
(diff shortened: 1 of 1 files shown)


TheChymera:
I'm not entirely sure that this is how the sentence is supposed to be read, you *remove* all *non*-alphanumeric characters, and you are left with `[0-9a-zA-Z]`. 

I think those parentheses were simply there to demonstrate what this entry looks like, and not to rephrase the sentence in regex-jargon (in which case it would still not be very clear since `[^0-9a-zA-Z]` sounds like it you remove all the alphanumeric chars).

src/schema/objects/metadata.yaml
```diff
@@ -2986,7 +2986,7 @@ TaskName:
     Name of the task.
     No two tasks should have the same name.
     The task label included in the file name is derived from this `"TaskName"` field
-    by removing all non-alphanumeric (`[a-zA-Z0-9]`) characters.
+    by removing all non-alphanumeric (`[^0-9a-zA-Z]`) characters.
```
sappelhoff, line 2989: Perhaps this would be a better clarification?

```suggestion
    by removing all non-alphanumeric characters (that is, all except those matching `[0-9a-zA-Z]`).
```

sappelhoff (approved):

Issue #1155, opened 2022-07-20
Do not attempt to validate associated data.


TheChymera:
Directory names can be found in `associated_data.yaml`.

TheChymera:
datasets to test this:

```
chymera@decohost ~/src/bids-examples $ find -name "stimuli"
./eeg_ds003654s_hed_longform/stimuli
./eeg_ds000117/stimuli
./ieeg_visual_multimodal/stimuli
./eeg_rishikesh/stimuli
./eeg_ds003654s_hed/stimuli
./ieeg_visual/stimuli
./eeg_matchingpennies/stimuli
./eeg_ds003654s_hed_inheritance/stimuli
./ieeg_filtered_speech/stimuli
./ds000117/stimuli
chymera@decohost ~/src/bids-examples $ find -name "derivatives"
./qmri_qsm/derivatives
./qmri_mese/derivatives
./ieeg_motorMiller2007/derivatives
./ieeg_epilepsy/derivatives
./qmri_mp2rageme/derivatives
./qmri_mtsat/derivatives
./synthetic/derivatives
./ieeg_epilepsy_ecog/derivatives
./qmri_mpm/derivatives
./ieeg_visual/derivatives
./eeg_rest_fmri/derivatives
./qmri_vfa/derivatives
./ds000248/derivatives
./qmri_mp2rage/derivatives
./qmri_irt1/derivatives
./qmri_sa2rage/derivatives
./ds000117/derivatives
```

PR #1169, opened 2022-07-28
Print BIDSVersion warnings only once per dataset

TheChymera:
Preventing log spamming when passing in file list (as seen in https://github.com/dandi/dandi-cli/issues/989 ).


TheChymera:
The fact that we were getting this to begin with was a design fluke, not sure whether it's worthwhile to have a test for whether or not this unaesthetic reporting ever comes back.

tools/schemacode/bidsschematools/validator.py
```diff
@@ -600,55 +600,57 @@ def select_schema_dir(
...
-                raise ValueError(
-                    "You have selected files belonging to 2 different datasets. "
-                    "Please run the validator once per dataset."
+        if dataset_description not in dataset_descriptions:
+            dataset_descriptions.append(dataset_description)
+    if len(dataset_descriptions) > 1:
+        raise ValueError(
+            f"You have selected files belonging to {len(dataset_descriptions)} "
+            "different datasets. Please run the validator once per dataset."
+        )
```
yarikoptic, line 609: what if there is a BIDS subdaset under `rawdata/` or `derivatives/` -- wouldn't we find multiple dataset_descriptions?

TheChymera: whelp, your're right... we do filter those out but we filter them *after* we detect the schema in the path getting function... might end up having to merge this PR with https://github.com/bids-standard/bids-specification/issues/1155 --- which I've also been meaning to tackle for a while.

TheChymera: @yarikoptic ok, so this is not actually happening, as the dataset_descriptor finding simply crawls up until if finds a file. So https://github.com/bids-standard/bids-specification/blob/73db3bafe3317645d1b72d1798308f85a5efd238/tools/schemacode/bidsschematools/validator.py#L602 will only ever find one or zero files per iteration.

TheChymera:
Don't mind the lint error, it's from the master branch, solved here → https://github.com/bids-standard/bids-specification/pull/1176

tools/schemacode/bidsschematools/validator.py
yarikoptic: so it would error out only if I point with the `paths` to multiple BIDS datasets, e.g. `[".", "rawdata"]` right?

then may be adjust docstring for `bids_paths` to say that they all must be within the same BIDS dataset, i.e. should not e.g. point to main BIDS dataset and derivatives etc.

TheChymera: @yarikoptic done.

effigies (approved):

PR #1184, opened 2022-08-04
Downstream fixes

TheChymera:
A number of small fixes we worked out in DANDI which currently (though ideally not for much longer) also bundles the bidsschematools validator.


sappelhoff (commented):
@TheChymera why is this a draft? Are there more fixes to come?

PR #1190, opened 2022-08-08
[SCHEMA] add task definition to schema

### LINKED ISSUE #1186: Update task definition in appendix.
The one on the entities page in the appendix does not match the one in the common principle, unlike for subject, session... 

https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#task

https://bids-specification.readthedocs.io/en/stable/02-common-principles.html

Doing so will require updating the schema.

Remi-Gau:
closes #1186

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -361,7 +361,7 @@ task:
     (with different sets of instructions) they will be considered one (combined) task.
     The `task-<label>` MUST be consistent across subjects and sessions.
 
-    Files with the `task-<label>` entity SHOULD have an associated 
+    Files with the `task-<label>` entity SHOULD have an associated
     [events file](SPEC_ROOT/04-modality-specific-files/05-task-events.md#task-events),
     as well as certain metadata fields in the associated JSON file.
 
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -351,18 +351,19 @@ task:
   description: |
     A set of structured activities performed by the participant.
     Tasks are usually accompanied by stimuli and responses, and can greatly vary in complexity.
-    For the purpose of this specification we consider the so-called "resting state" a task.
+    For the purpose of this specification we consider the so-called "resting state" a task,
+    although events files are not expected for resting state data.
+    Additionally, a common convention in the specification is to include the word "rest" in
+    the `task` label for resting state files (for example, `task-rest`).
+
     In the context of brain scanning, a task is always tied to one data acquisition.
     Therefore, even if during one acquisition the subject performed multiple conceptually different behaviors
     (with different sets of instructions) they will be considered one (combined) task.
     The `task-<label>` MUST be consistent across subjects and sessions.
 
-    1. Files with a filename containing the `task-<label>` entity SHOULD have an associated `*events.tsv` file
-       and certain metadata fields MUST or SHOULD be present in the associated JSON file
-       depending on the datatype.
-    2. For resting state task:
-      - a RECOMMENDED convention is to use labels beginning with `rest`.",
-      - `*events.tsv` files are not necessary.
+    Files with the `task-<label>` entity SHOULD have an associated 
+    [events file](SPEC_ROOT/04-modality-specific-files/05-task-events.md#task-events),
+    as well as certain metadata fields in the associated JSON file.
 
   type: string
   format: label
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -356,6 +356,14 @@ task:
     Therefore, even if during one acquisition the subject performed multiple conceptually different behaviors
     (with different sets of instructions) they will be considered one (combined) task.
     The `task-<label>` MUST be consistent across subjects and sessions.
+
+    1. Files with a filename containing the `task-<label>` entity SHOULD have an associated `*events.tsv` file
+       and certain metadata fields MUST or SHOULD be present in the associated JSON file
+       depending on the datatype.
+    2. For resting state task:
+      - a RECOMMENDED convention is to use labels beginning with `rest`.",
+      - `*events.tsv` files are not necessary.
+
   type: string
   format: label
 tracer:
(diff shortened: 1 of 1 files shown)


src/schema/objects/entities.yaml
```diff
@@ -349,10 +349,13 @@ task:
   name: task
   display_name: Task
   description: |
-    Each task has a unique label that MUST only consist of letters and/or
-    numbers (other characters, including spaces and underscores, are not
-    allowed).
```
Remi-Gau, line 354: I believe this part of the rule section of the schema and I have therefore removed it even if it is in the common principles sections

src/schema/objects/entities.yaml
```diff
@@ -349,10 +349,13 @@ task:
...
-    Those labels MUST be consistent across subjects and sessions.
+    A set of structured activities performed by the participant.
+    Tasks are usually accompanied by stimuli and responses, and can greatly vary in complexity.
+    For the purpose of this specification we consider the so-called "resting state" a task.
+    In the context of brain scanning, a task is always tied to one data acquisition.
+    Therefore, even if during one acquisition the subject performed multiple conceptually different behaviors
+    (with different sets of instructions) they will be considered one (combined) task.
+    The `task-<label>` MUST be consistent across subjects and sessions.
```
Remi-Gau, line 358: Tempted to remove this last line as this is also more of a rule, but since other entities definition have similar rules, I have kept it for now.

sappelhoff (approved):

src/schema/objects/entities.yaml
```diff
@@ -349,10 +349,13 @@ task:
   name: task
   display_name: Task
   description: |
-    Each task has a unique label that MUST only consist of letters and/or
-    numbers (other characters, including spaces and underscores, are not
-    allowed).
-    Those labels MUST be consistent across subjects and sessions.
+    A set of structured activities performed by the participant.
+    Tasks are usually accompanied by stimuli and responses, and can greatly vary in complexity.
```
tsalo, line 353: Is there a distinction to be made between the task definition under common principles and the task entity? This part seems more related to the common principle than the entity.

Remi-Gau: I don't think we should make a distinction unless we want to make things extra confusing.

Are we making such distinctions for subject or session?

tsalo: There are small differences at the moment. The session common principle has a sentence about PET data that the entity definition doesn't have, but I think that's more about information drift between duplicated text.

There _is_ a meaningful difference between the subject common principle and entity, in that the common principle explicitly mentions the term "Participant".

I see what you mean though- perhaps there's no reason to give them distinct definitions.

tsalo: I do think it's worth mentioning the following though:

1. Having the `task` entity present means that having an associated events file is recommended, and there are certain metadata fields that are also recommended or required.
2. It is recommended to use `task-rest` for resting-state scans, and events files are not necessary for resting-state data.

Remi-Gau: >     1. Having the `task` entity present means that having an associated events file is recommended, and there are certain metadata fields that are also recommended or required.
> 
>     2. It is recommended to use `task-rest` for resting-state scans, and events files are not necessary for resting-state data.

I agree that those should be in, but is the "definition" section the best place (not rhetorical). I am ok putting them there but it feels likes this is also blurring the  object and rule delineation we have in the schema, no?

Remi-Gau: Good point. If we find ways to best disentangle objects and rules later we'll cross that bridge when we get to it but for now let's add it as is.

src/schema/objects/entities.yaml
```diff
@@ -349,10 +349,21 @@ task:
...
+    Therefore, even if during one acquisition the subject performed multiple conceptually different behaviors
+    (with different sets of instructions) they will be considered one (combined) task.
+    The `task-<label>` MUST be consistent across subjects and sessions.
+
+    1. Files with a filename containing the `task-<label>` entity SHOULD have an associated `*events.tsv` file
+       and certain metadata fields MUST or SHOULD be present in the associated JSON file
+       depending on the datatype.
+    2. For resting state task:
+      - a RECOMMENDED convention is to use labels beginning with `rest`.",
+      - `*events.tsv` files are not necessary.
```
Remi-Gau, line 365: @tsalo 
an attempt at wording your suggestions

tsalo, line 365:
```suggestion
    Files with the `task-<label>` entity SHOULD have an associated 
    [events file](SPEC_ROOT/04-modality-specific-files/05-task-events.md#task-events),
    as well as certain metadata fields in the associated JSON file.
```

src/schema/objects/entities.yaml
```diff
@@ -349,10 +349,21 @@ task:
   name: task
   display_name: Task
   description: |
-    Each task has a unique label that MUST only consist of letters and/or
-    numbers (other characters, including spaces and underscores, are not
-    allowed).
-    Those labels MUST be consistent across subjects and sessions.
+    A set of structured activities performed by the participant.
+    Tasks are usually accompanied by stimuli and responses, and can greatly vary in complexity.
+    For the purpose of this specification we consider the so-called "resting state" a task.
```
tsalo, line 354:
```suggestion
    For the purpose of this specification we consider the so-called "resting state" a task,
    although events files are not expected for resting state data.
    Additionally, a common convention in the specification is to include the word "rest" in
    the `task` label for resting state files (for example, `task-rest`).

```

src/schema/objects/entities.yaml
tsalo: I made a couple of suggestions that integrate the changes more organically. Feel free to reject them if you like the bullet list though.

tsalo (commented):
Just a few thoughts regarding the phrasing.

Remi-Gau:
> Just a few thoughts regarding the phrasing.

Thanks. :rocket:

sappelhoff (approved):
@tsalo do you approve?

tsalo (approved):
LGTM!

PR #1058, opened 2022-04-11
[SCHEMA] Add internal validation test for schema

tsalo:
References #1037.

Changes proposed:

- Add a test that checks that all objects referenced in the schema rules are also defined as schema objects.

To do:

- [ ] Add test for order of entity definitions in `objects/entities.yaml` (#1071).

Issue #1037, opened 2022-03-22
Add schema validation tests to schemacode


tsalo:
I'd like to add a few tests to schemacode to ensure that the schema is "valid". We should probably define rules for the schema here to ensure that they make sense.

One test should check that all objects referred to in the rules portion of the schema are actually defined in the objects portion of the schema.

PR #1193, opened 2022-08-09
[SCHEMA] Order entities and add workflow to check entity order

### LINKED ISSUE #1071: [SCHEMA] use the schema to organize the schema
Occurred to me that the rule files that describes each datatypes list the entities for each suffix group in an order that seems arbitrary.

See for example: https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/datatypes/anat.yaml

There is a note about this in the README of the schema and that the actual entity order should read from `rules/entities.yaml`

I do believe that this file should be the final arbiter of the entity order (because it can tell you how to order entities for suffix groups that are not yet in BIDS), however...

Could we possibly "lint" the schema (automatically using the schema itself?) so that the entities in each suffix group are actually always ordered as they should appear in a filename? This would definitely limit the number of errors from new users of the schema who might have missed those 2 lines in the README.

May also help tools down the line that this way do not have to always look up the `entities.yml` file to know how to order entities.

### LINKED ISSUE #1189: Sort entities for each suffix group in schema
Currently the list of entities associated with each group of suffix for a given datatype can be listed in any ways and do not have to match the  canonical BIDS entity order. 

https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/entities.yaml

That file should be the final arbiter on bids entities order, but it would be more convenient for schema user and downstream tools if the list of entities were properly ordered for each suffix group. 

I suggest adding some CI workflow for the schema to check that entities in suffix groups are in the right order.

Remi-Gau:
closes #1189 
closes #1071

## unrelated

- fixes some typos and use of abbreviations in schema


sappelhoff (approved):

tsalo:
I think this is a good thing to do, though I don't know if it needs a separate Action. Perhaps a more general one meant to validate schema files would be good. I proposed adding tests for validating the schema in #1037, and I think this fits within that overall goal.

Remi-Gau:
Should we merge both and then into a different PR refactor the code and convert this action into one that runs all schema tests ?

tsalo (approved):
@Remi-Gau that's a fair point. In that case, this LGTM.

Remi-Gau:
waiting for another positive review of #1058 and then we will merge both.

@sappelhoff 

I can take care of reorganizing the code after that.

sappelhoff:
@Remi-Gau I just approved, please go ahead :-)

PR #1196, opened 2022-08-12
[ENH] add task metadata to PET

### LINKED ISSUE #1188: Task metadata for PET
There seems to be no mention of the typical metadata usually associated with the task  entity for PET (TaskName, Instructions, TaskDescription...). 

Cannot recall if this was intentional or overlooked?

@melanieganz @mnoergaard

Any idea?

Remi-Gau:
closes #1188 

To view the final render:

https://bids-specification--1196.org.readthedocs.build/en/1196/04-modality-specific-files/09-positron-emission-tomography.html#task


src/04-modality-specific-files/09-positron-emission-tomography.md
```diff
@@ -276,6 +276,24 @@ and a guide for using macros can be found at
...
+
+<!-- This block generates a metadata table.
+The definitions of these fields can be found in
+  src/schema/objects/metadata.yaml
+and a guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_metadata_table(
+   {
+      "TaskName": ("RECOMMENDED", "If used to denote resting scans, a RECOMMENDED convention is to use labels beginning with `rest`.")
```
Remi-Gau, line 289: Usually this is REQUIRED for almost all datatypes (except for behavioral), for backward compatibility I suspect it is better to keep it RECOMMENDED.

Remi-Gau: Added an extra comment that tries to make it clear that eventhough task is not required for PET for resting scans, if one decided to use it, it would be better to stick to the convention other datatypes use.

src/schema/rules/sidecars/pet.yaml
```diff
@@ -181,3 +181,22 @@ PETReconstruction:
             level: recommended
         SinglesRate:
             level: recommended
+
+# PET task information
+PETTask:
+    selectors:
+    - modality == "pet"
```
tsalo, line 188: Should `"task" in entities` (or whatever the appropriate syntax is) be included here? Essentially, is this task-related metadata recommended when there is no task entity in the filename?

src/04-modality-specific-files/09-positron-emission-tomography.md
```diff
@@ -276,6 +276,27 @@ and a guide for using macros can be found at
 
 All reconstruction-specific parameters that are not specified, but one wants to include, should go into the `ReconMethodParameterValues` field.
 
+#### Task
+
+If the OPTIONAL [`task-<label>`](../99-appendices/09-entities.md#task) is used,
```
tsalo, line 281: What does not having `task` imply? Just that the scan is resting-state, or something more specific? Like it's meant as a static, rather than functional, scan or something?

src/schema/rules/sidecars/pet.yaml
Remi-Gau: ah yeah you are right: those metadata are conditional on task entity being present.
let me check how this expressed in the schema.

src/04-modality-specific-files/09-positron-emission-tomography.md
Remi-Gau: Given that it's PET it could just be that you are measuring binding potential of some ligand I suspect. Some not exactly static because you have a temporal component but not functional for sure. @mnoergaard ?

src/schema/rules/sidecars/pet.yaml
Remi-Gau: fixed.

src/04-modality-specific-files/09-positron-emission-tomography.md
mnoergaard: Not having a task means rest (to lie still in the scanner and note move), and this is how most of PET scans are carried out. Both for static and dynamic scans.

src/04-modality-specific-files/09-positron-emission-tomography.md
mnoergaard: I agree that it is best to keep recommended, and great that you added the extra comment.

mnoergaard (approved):
LGTM - but this will also require a few updates to the validator, right?

Remi-Gau:
Better but it seems that dumping extra metadata in the json did not trigger any failure in the bids examples repo, so let's open an issue to remember to do it but hell should not break loose if we don't do it right away. 

https://github.com/bids-standard/bids-examples/pull/330

sappelhoff:
> Better but it seems that dumping extra metadata in the json did not trigger any failure in the bids examples repo, so let's open an issue to remember to do it but hell should not break loose if we don't do it right away.

That is because `AdditionalProperties` is not declared in the JSON schema ... and if not declared, it defaults to "true" -- so you may add any metadata right now.

- http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties
- https://github.com/bids-standard/bids-validator/blob/d96d6ff2ee30a6620d91d1d881fdc6e5db9023e9/bids-validator/validators/json/schemas/pet.json#L1

Still, the metadata should be added to the json schema in the validator

sappelhoff (approved):
approved, provided we add "task" to https://github.com/bids-standard/bids-validator/blob/d96d6ff2ee30a6620d91d1d881fdc6e5db9023e9/bids-validator/validators/json/schemas/pet.json#L1

PR #1201, opened 2022-08-15
[SCHEMA] Ross' serious sidecar rewrite for validator prototype dev

rwblair:
Opening this to see what all local changes I made to make the JS validator prototype work.


rwblair:
Historically we have been working from the assumption that sequential rules in sidecars/*.yaml will overwrite previous key's requirement levels if the proper selectors are hit. This is problematic  in that a validator would have to check all the rules in a file before attempting to validate a context. This PR is a proof of concept for writing rules that are mutually exclusive so that each rule could be considered atomic. An entry shouldn't be optional in one applicable rule and then required in another or vice versa.

I have readded the level_addendums to the entries that match the addedum text. It seems as though we only move from lower requirement levels to higher requirement levels, but this may not always be true.

rwblair:
@tsalo I didn't add a related rules, I think we still could but it may be premature until there is a render for these files.

src/schema/rules/sidecars/asl.yaml
```diff
@@ -16,20 +16,22 @@ MRIASLTextOnly:
...
 
-MRIASLTextOnlySliceTiming:
+MRIASLTextOnlySliceTimingRec:
+    selectors:
+    - modality == "MRI"
+    - datatype == "perf"
+    - suffix == "asl"
+    - sidecar.MRAcquisitionType == "2D"
+    fields:
+        SliceTiming: recommended
```
tsalo, line 28:
```suggestion
        SliceTiming: required

MRIASLTextOnlySliceTimingRec:
     selectors:
     - modality == "MRI"
     - datatype == "perf"
     - suffix == "asl"
     - sidecar.MRAcquisitionType != "2D"
     fields:
         SliceTiming: recommended
```

Is this what you mean by atomic rules?

I think this will make rendering tables harder, since the original rules that have level addenda can't be place in the original tables because the selectors are different. I think we'll also need tests that check that all rules are mutually exclusive, or reviewers will have trouble identifying any flaws in contributors' logic.

src/schema/rules/sidecars/asl.yaml
```diff
@@ -16,20 +16,22 @@ MRIASLTextOnly:
...
-            level: recommended
-            level_addendum: required if `LookLocker` is `true`
 
-MRIASLTextOnlySliceTiming:
+MRIASLTextOnlySliceTimingRec:
+    selectors:
+    - modality == "MRI"
+    - datatype == "perf"
+    - suffix == "asl"
+    - sidecar.MRAcquisitionType == "2D"
```
tsalo, line 26:
```suggestion
    - sidecar.MRAcquisitionType != "2D"
```

src/schema/rules/sidecars/asl.yaml
```diff
@@ -51,12 +53,20 @@ MRIASLTextOnlySliceTiming:
                     final entry in the `SliceTiming` list is the time of acquisition of slice 0.
                     Without this parameter slice time correction will not be possible.
 
-MRIASLTextOnlyFlipAngle:
+MRIASLTextOnlyFlipAngleRec:
+    selectors:
+    - modality == "MRI"
+    - datatype == "perf"
+    - suffix == "asl"
+    - sidecar.LookLocker == false
```
tsalo, line 61: Since LookLocker is optional, I don't think `sidecar.LookLocker == false` is sufficient. We either need `sidecar.LookLocker != true`, an or between  `sidecar.LookLocker == false` and `'"LookLocker in sidecar'`, or two separate rules.

rwblair: Ah this relies on a JS idiosyncrasy where `undefined == false` resolves to True. I'll make it more explicit

src/schema/rules/sidecars/asl.yaml
```diff
@@ -51,12 +53,20 @@ MRIASLTextOnlySliceTiming:
                     final entry in the `SliceTiming` list is the time of acquisition of slice 0.
                     Without this parameter slice time correction will not be possible.
 
-MRIASLTextOnlyFlipAngle:
+MRIASLTextOnlyFlipAngleRec:
+    selectors:
+    - modality == "MRI"
+    - datatype == "perf"
+    - suffix == "asl"
+    - '!(LookLocker in sidecar) || sidecar.LookLocker == false'
```
rwblair, line 61: @tsalo how do you feel about this? or would you prefer two separate rules?

tsalo: I like having one rule, and this is pretty readable.

PR #1202, opened 2022-08-15
[ENH] breaking part of rwblair's PR #1108 out into this smaller addition

bendhouseart:
Adds type descriptor to folder entities per @rwblair 's #1108, but in a more bite-sized way.


src/schema/rules/top_level_files.yaml
```diff
@@ -1,34 +1,43 @@
...
   required: true
+  pattern: 'README'
   extensions:
   - None
   - .md
   - .rst
   - .txt
 CHANGES:
-  required: false
+  required: true
```
effigies, line 13:
```suggestion
  required: false
```

effigies:
I think this will be superseded by #1288.

PR #1203, opened 2022-08-16
[FIX] Fix schema test checking rules against objects

tsalo:
Closes None.

Changes proposed:

- Move `test_rules.py` into the appropriate folder.
- Replace extension values with extension keys in the rules files.
    - For example, the key for TSV files is "tsv", while the value is ".tsv".
- Replace suffix values with suffix keys in the rules files.
- Fix several column names in rules files. Some were wrong/mistyped.
- Rename `rules/sidecars/entities.yaml` to `rules/sidecars/entity_rules.yaml`.
    - The problem is that test_rules looks for any dictionaries (including files) that have a key matching an object type, such as "suffixes", "metadata", or "entities". The name "entities" can thus be considered reserved.


tsalo:
One limitation of the test is that it doesn't recursively look through subfolders, so the `derivatives` folders are not being tested.

Also, `labelgii` and `dlabelnii` aren't defined in `objects/extensions.yaml` yet.

TheChymera:
@tsalo so this is an easy fix, which would just need to go deeper in the nested dictionary structure [after this line](https://github.com/bids-standard/bids-specification/blob/b59550c3d934c788f0a4829293bf19e583e2b109/tools/schemacode/bidsschematools/validator.py#L158). I can fix it, but I still don't think this extension reorganizations is a bad idea.... If I'm the only one, we can go ahead, but maybe explicitly asking for comments somehow would be a good idea. This just adds imho unnecessary higher-level categorizations which will make schema editing more intransparent and enable all sorts of exciting discussions about how we divvy up strings/regexes among these abstraction categories.

src/schema/rules/datatypes/anat.yaml
```diff
@@ -14,9 +14,9 @@ nonparametric:
   - FLASH  # deprecated
   - PD  # deprecated
   extensions:
-  - .nii.gz
-  - .nii
-  - .json
+  - niigz
```
TheChymera, line 17: can we maybe not do this? =^.^=

see https://github.com/bids-standard/bids-specification/pull/1203#issuecomment-1216762725

TheChymera (changes requested):

src/schema/rules/datatypes/anat.yaml
tsalo: I mean, I still think we should, but it's up to the maintainers/steering group now.

tools/schemacode/bidsschematools/tests/test_rules.py
```diff
@@ -54,6 +54,7 @@ def test_rule_objects(schema_obj):
 
         for type_instance in type_instances_in_rules:
             path, instance = type_instance
```
effigies, line 56: Nit-pick, btw:

```suggestion
        for path, instance in type_instances_in_rules:
```

tsalo: Thanks!

PR #1211, opened 2022-08-16
Metadata improvements and moved pandas and numpy to optional deps

TheChymera:
Ideally we might want to consider whether render can work without numpy/pandas altogether... but probably too big a task at this point.

@tsalo @yarikoptic

### DIFF
--- a/tools/schemacode/bidsschematools/__init__.py
+++ b/tools/schemacode/bidsschematools/__init__.py
@@ -4,12 +4,6 @@ try:
 except ImportError:  # PY<3.9
     from importlib_resources import as_file, files
 
-__all__ = [
-    "render",
-    "schema",
-    "utils",
-]
-
 version_file = files("bidsschematools.data") / "schema" / "SCHEMA_VERSION"
 with as_file(version_file) as f:
     __version__ = f.read_text().strip()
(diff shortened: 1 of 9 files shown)


tools/schemacode/bidsschematools/__init__.py
```diff
@@ -1,6 +1,4 @@
 """A Python package for working with the BIDS schema."""
-from . import render, schema, utils
```
effigies, line 2: If we're going to remove these, then we should remove them from `__all__`.

tools/schemacode/setup.cfg
```diff
@@ -21,8 +21,6 @@ classifiers =
 [options]
 python_requires = >=3.7
 install_requires =
-    numpy
-    pandas
     tabulate
```
effigies, line 24: I would move this to `render`.
```suggestion
```

tools/schemacode/setup.cfg
```diff
@@ -34,6 +32,8 @@ zip_safe = false
 doc =
     sphinx>=1.5.3
     sphinx_rtd_theme
+render =
+    pandas
```
effigies, line 36:
```suggestion
    pandas
    tabulate
```

effigies (approved):
Otherwise looks fine.

PR #1224, opened 2022-08-17
[SCHEMA] add schema and spec version to object dumped by bidsschematools

rwblair:
These files may be passed around and have lives outside of readthedocs url. Having basic provenance info may save someone a headache in the future.


effigies (approved):

rwblair:
https://bids-specification--1224.org.readthedocs.build/en/1224/schema.json

PR #1226, opened 2022-08-17
Use static testdata if present

TheChymera:
We should support static testdata now that we have releases.


TheChymera:
@effigies --- regarding decreasing coverage, unsure whether we should fully execute this as part of the test suite.

It's kind of a special case for releases. If we were to link it to some specific test data archive, that would actually be expected to sometimes fail between releases as meaningful differences between static archives and the live example repos accumulate.

effigies:
We should exclude the `tests/` directory from coverage in any case. A path not hit there is not indicative of anything.

tools/schemacode/pyproject.toml
```diff
@@ -29,3 +29,8 @@ multi_line_output = 3
 markers = [
   "validate_schema: tests that validate the schema itself",
 ]
+
+[coverage:run]
+parallel = True
+omit =
+    bidsschematools/tests/*py
```
effigies, line 36: TOML syntax.

```suggestion
parallel = true
omit = ["*/tests/*"]
```

Also, I think we need to update the test dependencies in setup.cfg to use `coverage[toml]` to ensure this gets picked up.

tools/schemacode/setup.cfg
```diff
@@ -36,7 +36,7 @@ render =
     pandas
 tests =
     codecov
-    coverage<5.0
+    coverage[toml]<5.0
     flake8<5.0
```
effigies, line 40: While we're fiddling around here, can we try relaxing these? I suspect they were needed to wait for other tooling to catch up with an API-breaking change, but I hope we won't need them forever.

```suggestion
    coverage[toml]
    flake8
```

effigies (approved):

TheChymera:
@effigies don't pull yet, codecov is still iffy :/

effigies:
Oh, do we just need to tell codecov to ignore these as well?

https://docs.codecov.com/docs/codecovyml-reference#ignore

effigies:
Sorry, I meant I think we need a `codecov.yml` file in addition to configuring coverage.

TheChymera:
@effigies I have no idea why this works, but there we have it :3

TheChymera:
@effigies no, it doesn't work,,, it does this weird thing where it pretends it worked for a few minutes and then says it didn't -.-

TheChymera:
@effigies ok, this works, project-wide coverage is down not due to the addition of uncovered code, but the removal of valid but irrelevant code from coverage.

PR #1230, opened 2022-08-18
[ENH] Prettier; Updating yaml formatting to play nicely w/ yamllint, .editorconfig, and prettier. (Warning: compromises made)

bendhouseart:
This PR updates the options passed to the yamllinter to make it more amenable to code formatting tools such as prettier. Additionally settings relating to formatting are stored in an `.editorconfig` file manage prettier and text editor behavior.

One side effect of this is to eliminate the non-nestedness that was enforced in our 2-space formatted yaml files, e.g.:
```yaml
MyCoolExample:
  CheckOutThisCooList:
  - a
  - b
  - c
```
is now displayed as:
```yaml
MyCoolExample:
  CheckOutThisCooList:
    - a
    - b
    - c
```

Be horrified.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+[*.yaml]
+max_line_length = 120
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+quote_type = single
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# 2022-08-19 12:59:40 -0400 - markiewicz@stanford.edu - [git-blame-ignore-rev] prettified schema files.
+aaccb5ffc10e7f9460bbe9eb8c246872de167543
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,13 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.0.1
     hooks:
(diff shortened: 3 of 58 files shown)


sappelhoff:
drive-by note --> if we merge "formatting/style" commits like this that touch many files, we may want to use a `.git-blame-ignore-revs` file, to keep our "git blame" clean. See e.g., https://github.com/bids-standard/pybv/blob/4af6bbab86e6151a3a64d759d4eca4a15e9b253a/.git-blame-ignore-revs#L1

effigies:
If we include `[git-blame-ignore-rev]` in the commit messages for commits we want to ignore, we can use this to generate an ignore list for us:

```Shell
git log --grep "\[git-blame-ignore-rev\]" --pretty=format:"# %ad - %ae - %s%n%H" > .git-blame-ignore-revs
```

Will look something like:

```
# Fri Jul 22 22:38:37 2022 -0400 - markiewicz@stanford.edu - STY: isort
7572472ccbaba85565e98bf1b5def0e6957a7d0d
# Fri Jul 22 22:26:02 2022 -0400 - markiewicz@stanford.edu - STY: blue
9911244fd3eed7b68f2f8688ff9c7ffe4bd5ad48 
```

sappelhoff:
The nice thing about the file is, that GitHub parses it and applies it to the blame you can view in the web UI

bendhouseart:
Suggestions added per requests, additionally I abandoned the idea of adding this to actions and making it automatic, set up a make file instead.

effigies:
I've managed to make all of your work look tiny: https://github.com/bids-standard/bids-specification/commit/b6eec94bb53218299f95eee9482ec5d29315f4f0

tsalo (approved):
Glad to see some automated formatting!

effigies (approved):

Issue #1232, opened 2022-08-18
Metadata tables to build from schema tables


effigies:
The following were not completed in #1225

1) [qMRI appendix](https://bids-specification.readthedocs.io/en/stable/99-appendices/11-qmri.html#metadata-requirements-for-qmri-maps)
2) [Tabular files](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#tabular-files) - this is probably going to need to be a subobject table instead of a sidecar table
3) [dataset_description.json](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#dataset_descriptionjson) (and [here](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#derived-dataset-and-pipeline-description)) - these may need a new renderer
4) [Phenotypic and assessment data](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#phenotypic-and-assessment-data) - same as tabular
5) [Genetic Information](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/08-genetic-descriptor.html#genetic-information) - subobject

_Originally posted by @effigies in https://github.com/bids-standard/bids-specification/issues/1225#issuecomment-1219793893_

PR #1225, opened 2022-08-17
ENH: Add make_sidecar_table macro that accepts a table name

effigies:
Starting to convert our in-text tables to references to our in-schema tables.

PR #1233, opened 2022-08-18
SCHEMA: Make subobject tables from qualified object names

effigies:
Continuation of #1225, focusing on subobjects. Seeing what I can knock out of #1232.


src/schema/objects/metadata.yaml
```diff
@@ -1053,18 +1078,39 @@ GeneratedBy:
...
+          RECOMMENDED if `Name` is `"Manual"`.
         type: string
       CodeURL:
+        name: CodeURL
+        description: URL where the code used to generate the dataset may be found.
         type: string
         format: uri
       Container:
+        name: Container
+        description: Used to specify the location and relevant attributes of software container image used to produce the dataset. Valid keys in this object include `Type`, `Tag` and [`URI`][uri] with [string][] values.
```
bendhouseart, line 1108:
```suggestion
        description: Used to specify the location and relevant attributes of software container image 
        used to produce the dataset. Valid keys in this object include `Type`, `Tag` and [`URI`][uri] with [string][] 
        values.
```

effigies:
```suggestion
        description: |
            Used to specify the location and relevant attributes of software container image 
            used to produce the dataset. Valid keys in this object include `Type`, `Tag` and
            [`URI`][uri] with [string][] values.
```

src/schema/objects/metadata.yaml
```diff
@@ -466,6 +466,31 @@ Columns:
   type: array
   items:
     type: string
+ColumnDescription:
```
effigies, line 469: How do people feel about this being in `objects.metadata`? This describes an object in a sidecar to a TSV file. For example, if my TSV has a column `handedness`, my JSON could have:

```JSON
{
  "handedness": {
    "Description": "Edinburgh handedness inventory",
  }
}
```

This introduces `name: '.*'` to indicate that the column name could be anything. But perhaps we want something else, like `objects.structures` or `objects.types`.

src/schema/objects/metadata.yaml
```diff
@@ -1053,18 +1078,39 @@ GeneratedBy:
   minItems: 1
   items:
     type: object
+    required_fields: [Name]
+    recommended_fields: [Version]
```
effigies, line 1082: This PR introduces the convention of `required_fields` and `recommended_fields`. This was the rough consensus of the recent schema sprint.

src/schema/objects/metadata.yaml
```diff
@@ -1073,6 +1119,26 @@ GeneratedBy:
           URI:
             type: string
             format: uri
+GeneticInfo:
```
effigies, line 1122: `GeneticInfo` is the full contents of `/genetic_info.json`. Is this metadata? Should this be somewhere else? I think it is still good to have it somewhere in `objects.` Could be `objects.documents`, including `dataset_description.json`.

effigies (commented):
@bids-standard/schema-users Some comments on things I'd like your input on.

effigies:
@bendhouseart @tsalo I pared this down to bits that shouldn't need much debate.

src/schema/objects/metadata.yaml
bendhouseart: groovy.

src/schema/rules/dataset_metadata.yaml
```diff
@@ -22,14 +22,6 @@ dataset_description:
...
     DatasetDOI: optional
     GeneratedBy: recommended
-    GeneratedBy[].Name: required
-    GeneratedBy[].Version: recommended
-    GeneratedBy[].Description: optional
-    GeneratedBy[].CodeURL: optional
-    GeneratedBy[].Container: optional
-    GeneratedBy[].Container{}.Type: recommended
-    GeneratedBy[].Container{}.Tag: recommended
-    GeneratedBy[].Container{}.URI: recommended
```
bendhouseart, line 32: Great these looked terrible, ty ty.

tools/schemacode/bidsschematools/render/tables.py
```diff
@@ -500,55 +499,48 @@ def make_metadata_table(schema, field_info, src_path=None, tablefmt="github"):
...
+    object_name: str,
+    src_path: ty.Optional[str] = None,
+    tablefmt: str = "github",
+):
+    """Create a metadata table (markdown) based on the properties of an object
 
     Parameters
     ----------
-    schema : Namespace
+    schema : dict
```
bendhouseart, line 512:
```suggestion
    schema : Namespace
```

Think this got jumbled, did you mean Namespace here instead of dict per line 503?

bendhouseart (changes requested):
Just wanted to clarify that one issue w/ the doc string.

src/modality-specific-files/genetic-descriptor.md
```diff
@@ -33,14 +33,7 @@ The definitions of these fields can be found in
...
 -->
-{{ MACROS___make_subobject_table(
-   ("objects", "metadata", "Genetics"),
-   {
-      "Dataset": "REQUIRED",
-      "Database": "OPTIONAL",
-      "Descriptors": "OPTIONAL",
-   },
-) }}
+{{ MACROS___make_subobject_table("metadata.Genetics") }}
```
tsalo, line 36: This doesn't introduce any problems, but I did notice that this table and the above text don't match. The text says that there is a `.` notation, which we have dropped, so it's actually not clear what the key to the subobject should be (i.e., `Genetics`).

tsalo (approved):
Just one small issue with the Genetics table that wasn't actually introduced in this PR. Not sure if you want to address it here or leave it for another PR.

effigies:
Addressed all comments. Will merge on green unless someone stops me.

bendhouseart (approved):

src/modality-specific-files/genetic-descriptor.md
tsalo: Resolved by d90d36d85647d81f2c88ca19deba61830afa50ec.

Issue #1154, opened 2022-07-20
Add bidsschematools API documentation to website


tsalo:
It looks like there are [several mkdocs plugins](https://github.com/mkdocs/mkdocs/wiki/MkDocs-Plugins#api-documentation-building) that can build API documentation. Does anyone have a preference?

effigies:
On balance, I'd probably prefer not to have it directly on RTD. What if we set it up as a standard sphinx page and push to https://bids-standard.github.io/bidsschematools/?

tsalo:
I figured we'd bury it in the appendix, but I'm happy with a separate site as well, as long as we link to it in the schema documentation.

effigies:
Should we be using https://github.com/bids-standard/bids-schema? I know that @TheChymera did not want to call the Python package `bids-schema` on the grounds that that should be a data package, but that repository does have a `tools/` subdirectory that the package could ultimately wind up in. In which case we would maybe push to https://bids-standard.github.io/bids-schema/api/?

yarikoptic:
> Ideally that should contain 0% Python code and therefore have no API @yarikoptic

yes/most-likely ;)  At least ATM while both schema content, its `SCHEMA_VERSION` and `bidsschematools` are still in very active development, IMHO it is best to keep them close by in this repo, and use that `bids-schema` only for export of various versions of BIDS (as expressed in schema).  But later, who knows -- we might move `bidsschematools` into a separate repo. I do not see indeed a reason to "pair" archives of schema releases with the code in one repo really.

effigies:
@bendhouseart Got things running over at https://bidsschematools.readthedocs.io/.

PR #1235, opened 2022-08-18
[STY] Improve object formatting in the glossary

tsalo:
Closes None. I'm hoping that these changes will make the glossary more useful, since we'll be linking to it more and more as time goes by.

Changes proposed:

- Include object type-specific fields in glossary entries, such as "Allowed values" for enums.
- Bold categories in entries.
- Add format field.
- Prettify the schema information.


tools/schemacode/bidsschematools/render.py
```diff
@@ -149,14 +170,16 @@ def make_glossary(schema, src_path=None):
             new_name = f"{new_name} ({group})"
             all_objects[new_name] = {}
             all_objects[new_name]["key"] = f"objects.{group}.{key}"
+            all_objects[new_name]["type"] = TYPE_CONVERTER.get(group, group)
```
bendhouseart, line 173: cheeky 👍

bendhouseart (approved):

Remi-Gau (approved):
Looks good to me but we should probably link to the schema README if people want more info on how to understand the extra info, no?

BTW: we should probably update that schema README, no? It feels like the format info could be refactored into its own section

tsalo:
I agree that we should link to the README more, but I'm not sure how to do that. I guess I could link to the `master` schema README, but I think it would be better to link to the appropriate commit's version.

I'm sure we'll update the README at the end of the sprint or some time next week.

tsalo:
@Remi-Gau WDYT about pushing that off until we have #1154 dealt with? That way, we can just link to another page, instead of trying to hunt down the appropriate version on GitHub?

Remi-Gau:
Totally fine with me.

tsalo:
Cool! I'll merge this now then.

PR #1237, opened 2022-08-19
[SCHEMA] Change modalities to instruments

### LINKED ISSUE #593: Rename modalities.yaml in schema
The file [modalities.yaml](https://github.com/bids-standard/bids-specification/blob/5a7c1d75a92b4e1b4489a6cb01082720c268467f/src/schema/modalities.yaml) describes the technologies supported in the specification (e.g., MRI, EEG). Given that "modality" has its own meaning, and will soon have a distinct definition in **Common principles** (see #592), the naming of the file in the schema is misleading. Some viable alternatives might be "technologies.yaml" or "instruments.yaml".

### LINKED ISSUE #1141: add 'modality' to glossary
I've seen the word 'modality' used in different context, for instance /anat and /func defined as 2 (MRI) modalities. Within BIDS, these are not since under 'Modality specific files', what is listed is different data types based on technology. 

Proposal: a modality corresponds to a technology-driven data type (MRI, MEG, EEG, Behaviour, Eye tracking, etc) and does not separate within modality (e.g. anat vs. func).

tsalo:
Closes #593 and possibly closes #1141. The problem that this PR attempts to address is that the definition of "modality" in Common Principles conflicts with its use in the schema.

Changes proposed:

- Rename `schema/objects/modalities.yaml` to `schema/objects/instruments.yaml`.
- Rename `schema/rules/modalities.yaml` to `schema/rules/instruments.yaml`.
- Add a new "Instrument" element to the Common Principles.

### DIFF
--- a/src/schema/rules/instruments.yaml
+++ b/src/schema/rules/instruments.yaml
@@ -7,11 +7,9 @@ mri:
   - fmap
   - func
   - perf
-eeg:
+bioamp:
   datatypes:
   - eeg
-ieeg:
-  datatypes:
   - ieeg
 meg:
   datatypes:
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -36,8 +36,8 @@ context:
                 ignored:
                     description: "Set of ignored files"
                     type: array
-                modalities:
-                    description: "Modalities present in the dataset"
+                instruments:
+                    description: "Instruments used in the dataset"
                     type: array
                 subjects:
                     description: "Collections of subjects in dataset"
--- a/src/schema/objects/common_principles.yaml
+++ b/src/schema/objects/common_principles.yaml
@@ -83,6 +83,18 @@ index:
   description: |
     A nonnegative integer, possibly prefixed with arbitrary number of 0s for consistent indentation,
     for example, it is `01` in `run-01` following `run-<index>` specification.
+instrument:
+  name: Instrument
+  display_name: Instrument
+  description: |
+    The class of technology used to record data to a file.
+    For example, both EEG and iEEG are data types using biopotential amplification.
+
+    Instrument is not directly reflected in the BIDS filesystem,
+    though the specification itself generally nests **data type**-specific sections within
+    **instrument**-specific chapters.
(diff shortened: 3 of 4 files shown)


src/schema/rules/instruments.yaml
```diff
@@ -1,17 +1,15 @@
...
   - fmap
   - func
   - perf
-eeg:
+bioamp:
   datatypes:
   - eeg
-ieeg:
-  datatypes:
   - ieeg
```
tsalo, line 13: This is a core change- now EEG and iEEG are both considered "biopotential amplification". At some point it would be nice to reorganize the specification to reflect this, but I don't think it's necessary.

tsalo:
@nellh noted that OpenNeuro uses the term "Modality", and keeps EEG and iEEG separate. If we change it to "Instrument" and combine EEG and iEEG under "Biopotential Amplification", OpenNeuro would likely be impacted, and users probably want to search for _either_ EEG _or_ iEEG, not both.

CPernet:
1) I think https://github.com/bids-standard/bids-specification/issues/1141 makes the point that modality is not well defined in BIDS, and as such is a separate issue of how OpenNeuro uses it.
2) I agree that people want to search iEEG, MEG, EEG on their own, and @robertoostenveld said so too. @nellh could rely on the `datatype` rather than `modality` and everything should keep running as before
3) I'd have approved those changes

effigies:
> 2) I agree that people want to search iEEG, MEG, EEG on their own, and @robertoostenveld said so too. @nellh could rely on the `datatype` rather than `instrument` (previous `modality`) and everything should keep running as before

Yes, OpenNeuro can maintain its own mapping. It has been nice to rely on BIDS validation outputs directly, as it reduces maintenance tasks when updating software for new BEPs and avoids getting out of sync.

I wonder if we could keep modality as an unexposed component of the validator that OpenNeuro (and other tools that want a more intuitive, albeit less logically consistent, grouping) could rely on, so that this information does stay single sourced. So it would get emitted by the validator JSON but not be exposed in UI or used in schema rules.

PR #1241, opened 2022-08-19
[SCHEMA] Reorganize raw dataset rules

rwblair:
This will be an unpopular reorganizaton. Opening PR to track discussion on other ways we might want to separate out derivatives and raw rules


tsalo:
Are there any rules that can only be applied in raw datasets, and never in derivative datasets? It feels like any rules labeled "raw" should _only_ apply to raw datasets.

rwblair:
Closing in favor of #1288

PR #1246, opened 2022-08-22
[INFRA] auto ignore dependencies update from change log

Remi-Gau:
I suspect we will ignore from the changelog most the dependabot PRs rather than the other way around, so this should do the trick.


sappelhoff (approved):
nice :+1:

PR #1252, opened 2022-08-24
CI: Package bids-examples and bids-error-examples into Python distributions on schema-* tags

TheChymera:
@bendhouseart something went wrong with the other branch and actions fail to start up https://github.com/orgs/community/discussions/26746

exploring whether the “make a new branch” fix works


TheChymera:
@bendhouseart ok, solved that issue

TheChymera:
@effigies @bendhouseart 

ok, so at long last this is done. To recap:
* Depending on the `BIDSSCHEMATOOLS_RELEASE` envvar being set, one of our tests will create a testdata copy directory which is not torn down (this was already in).
* Now, however, the directory for this archive is `bidsschematools/tests/data/bundled`.
* We set this variable depending on whether or not the first seven characters of the tag are `schema-`
* If the variable is set we build the package — including testdata, which is less than 1MB compressed and 2.7 MB uncompressed.
* If the variable is set we test the upload to test.pypi
* If the variable is set and the test upload succeeded we upload to pypi
* we now have a script which should make testing for and development of tag-based hooks easier inside (`scripts/test_tag`)

For this to actually do anything at all, an additional 2 commits will be required by someone who knows what our secrets are called, which:
1. set tokens:
  * replaces `secrets.BIDS_TOKEN_TEST` in the “Real Upload” section to whatever the BIDS pypi token is called.
  * replaces `secrets.BIDS_TOKEN_TEST` in the “Test Upload” section to whatever the BIDS test.pypi token is called.
  * **alternatively** you can give me permissions to upload dedicated tokens for the both
  * **optionally** we can remove the `autodata2` branch entry from the head of [`.github/workflows/schemacode_ci.yml`](https://github.com/bids-standard/bids-specification/pull/1252/files#diff-ea378ca5441718caedc3fb2b818149ec3f0875b5be85ee7a1a99b291410288a5)
2. trigger release:
  * remove `-dev` from `src/schema/SCHEMA_VERSION`
  * add a tag called `schema-0.4.0`

TheChymera:
@effigies should be good to go. Thanks for your help. If you squash you can also add @bendhouseart as co-author since he helped as well.

.github/workflows/schemacode_ci.yml
```diff
@@ -4,52 +4,82 @@ on:
   push:
     branches:
       - "master"
-    paths:
-      - "tools/schemacode/**"
-      - "src/schema/**"
+      - "autodata2"
```
effigies, line 7:
```suggestion
```

effigies (approved):
LGTM. Just need to drop the extra branch before merging.

.github/workflows/schemacode_ci.yml
TheChymera: ah, yes, I had said optionally in the comment and thought for some reason you wanted to keep it, but sure, dropping it makes sense.

effigies:
0.4.0 is up on PyPI.

TheChymera:
Amazing, just tested it, works system-wide via Gentoo and passes all tests :3

https://ppb.chymera.eu/5a3281

bendhouseart:
> @bendhouseart ok, solved that issue

That looked nasty.

Issue #1080, opened 2022-04-16
SCHEMA: Expression syntax


effigies:
In order to express many rules that are found in BIDS, it's necessary to be able to evaluate arithmetic and logical expressions on arrays and objects (or their equivalents in other languages).

There are two primary types of expression:

* Selectors: boolean expressions that indicate whether a rule applies to a specific context
* Assertions: boolean expressions that the file (or entire dataset) must satisfy in order not to emit a warning or error

We have identified a number of operations that are necessary:

* Parentheses
* (Sub)string comparison
* Arithmetic and numeric comparison
* Contains/is-contained-by
* Negation
* Array indexing
* Object property lookup (`obj.property` or `obj["property"]`)
* Compute result on entire array
  * e.g., `max(arr) > 0` or `any(arr > 0)`
* Find length of array
  * Possibly also string length and shape of multi-dimensional arrays
  * If needed we could make these properties, so looking up `obj.size` or `obj.shape[0]`
* Find type of object
  * Could be made an `obj.type` property

We will need a limited language that either is commonly implemented or could be fairly easily implemented in any programming language.

One possibility would be the [Common Expression Language](https://github.com/google/cel-spec). That currently has implementations in Go and Python. We could consider a restricted subset of this language to make reimplementation simpler.

Happy to hear other suggestions.

effigies:
@tsalo made some notes on the sorts of data we need to operate over and operations they support.

The "keywords" are a namespace of values available to an expression.

----

* Keywords:
  * dataset
    * Contains dataset_description fields in the object.
    * Summary lists of object types in the dataset?
      * E.g., a list of the modalities under `dataset.modalities`.
  * sidecar
    * The associated JSON file's contents, as an object.
  * data?
    * The contents of the data file. Can have many different forms, I guess.
  * associated
    * An object of associated files, primarily used in the selectors. The keys of "associated" are generally referenced directly as keywords in the checks.
* File types:
  * tsv: The contents of a TSV data file.
    * Attributes:
      * .shape: A two-element tuple with the number of rows and columns.
    * Sidecar Attributes:
      * sidecar.Columns: Gets populated automatically, to make it easier to validate tsv and tsv.gz files in the same manner.
  * image: A NIfTI file or something similar.
    * Attributes:
      * .shape
      * .ndim
  * json: A JSON file (sidecar will be a json). This file type is loaded as an "object". 
    * See "field types: object" for info about attributes and operators.
* Field types:
  * object
[...]

PR #1254, opened 2022-08-25
SCHEMA: Uniformize rules

effigies:
This PR aims to achieve consistency throughout the schema.

* [x] Use consistent syntax for issues (`{issue: {code: ..., message: ..., level: ...}}`)
* [x] Use requirement level as a bare string on fields sidecar rules unless there are additional subfields
* [x] Eliminate redundant checks when encountered
* [x] Conform modality strings (currently have `"mri"` and `"MRI"`
* [x] Conform all selector and check expressions according to the current state of #1080


src/schema/meta/context.yaml
```diff
@@ -36,6 +39,9 @@ context:
         ignored:
           description: 'Set of ignored files'
           type: array
+        datatypes:
+          description: 'Data types present in the dataset'
+          type: array
```
effigies, line 44: Two changes to the context were required. Is this version-bump justifying?

effigies:
I believe this is good to go. Nothing currently backwards incompatible, but some additions motivated by a couple of checks.

effigies:
Bump @rwblair.

rwblair:
Haven't merged any of the function injections into the context eval space so most of these will fail in schema prototype validator. No need to put the cart in front of the horse if our expression language is done changing, I'd be fine with merging these right now.

Should there be a version bump?

effigies:
> Haven't merged any of the function injections into the context eval space so most of these will fail in schema prototype validator. No need to put the cart in front of the horse if our expression language is done changing, I'd be fine with merging these right now.

However works best for you. I could factor out the expression changes, if you'd rather those be a separate PR.

> Should there be a version bump?

Probably. I'll do that.

rwblair (approved):

sappelhoff:
Will the NIRS branch need to be updated with changes introduced here? If yes, would you mind going the extra mile and taking that on @effigies?

effigies:
`master` merged cleanly into `bep030`. Looking at the PR, I don't see any of the violations I was cleaning up in here.

Issue #1198, opened 2022-08-12
Add legend below filename templates


Remi-Gau:
Small suggestion to add a legend in the macro that generate the filename templates.

May help new comers.

It would add:

**Legend**

- entities between square brackets (for example `[_ses-<label>]`) are OPTIONAL
- for a definition of `label` and `index`, please refer to the relevant section in the [common principles](https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#definitions)

tsalo:
I think you could link to the label and index format glossary entries instead of the common principles. Also, maybe a link to the entities appendix could be included in the legend.

PR #1219, opened 2022-08-17
[ENH] Add filename template legend

Remi-Gau:
closes #1198
fixes #1220  

### unrelated
- update install of bidsschema tool via requirements
- add black, import sort and flake8 to pre-commit config
- run black and import sort on non schema code python
- add setup.cfg with flake8 config in root folder to tell pre-commit what config to use

PR #1259, opened 2022-08-29
[ENH] Add filename template legend

Remi-Gau:
closes https://github.com/bids-standard/bids-specification/issues/1198
replaces #1219

let's see how this one goes

### DIFF
--- a/tools/schemacode/bidsschematools/render.py
+++ b/tools/schemacode/bidsschematools/render.py
@@ -487,7 +487,8 @@ def append_filename_template_legend(
     md = MarkdownIt().disable("image").enable("table")
 
     legend = """
-- Filename entities or folders between square brackets (for example, `[_ses-<label>]`) are OPTIONAL.
+- Filename entities or folders between square brackets 
+  (for example, `[_ses-<label>]`) are OPTIONAL.
 - Some entities may only allow specific values,
   in which case those values are listed in `<>`, separated by `|`.
 - `_<suffix>` means that there are several (>6) valid suffixes for this filename pattern.
--- a/tools/schemacode/bidsschematools/render.py
+++ b/tools/schemacode/bidsschematools/render.py
@@ -476,14 +476,7 @@ def make_filename_template(
     return codeblock
 
 
-def append_filename_template_legend(
-    text=None,
-    pdf_format=False,
-):
-
-    if text is None:
-        return
-
+def append_filename_template_legend(text, pdf_format=False):
     md = MarkdownIt().disable("image").enable("table")
 
     legend = """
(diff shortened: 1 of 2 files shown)


tools/schemacode/bidsschematools/render.py
```diff
@@ -469,11 +470,55 @@ def make_filename_template(
...
+def append_filename_template_legend(
+    text=None,
+    pdf_format=False,
+):
+
+    if text is None:
+        return
+
+    md = MarkdownIt().disable("image").enable("table")
+
```
effigies, line 488: Let's move this below.

```suggestion
```

tools/schemacode/bidsschematools/render.py
```diff
@@ -469,11 +470,55 @@ def make_filename_template(
...
+        text += f"""
+
+<details>
+<summary><strong>Legend:</strong></summary>
+<ul>
+{md.render(legend)}
+</ul>
+</details>
+
+"""
```
effigies, line 517: The disable/enable things don't seem to change the code at all. Also, `<ul>` is part of the render.

```suggestion
    else:
        md = MarkdownIt()
        text += f"""
<details>
<summary><strong>Legend:</strong></summary>
{md.render(legend)}
</details>
"""
```

tools/schemacode/bidsschematools/render.py
```diff
@@ -469,11 +470,55 @@ def make_filename_template(
...
 
 
+def append_filename_template_legend(
+    text=None,
+    pdf_format=False,
+):
+
+    if text is None:
+        return
+
```
effigies, line 486: I don't see a point in allowing `text` to be `None` and having a special carveout. We take a `str`, return a `str`.

```suggestion
def append_filename_template_legend(text, pdf_format=False):
```

effigies (commented):
Thank you for paring it down! It is much easier to think about when it's just the heart of it. Some small suggestions that should have no impact on the render but tighten up the function a bit.

tools/schemacode/bidsschematools/render.py
Remi-Gau: ha yeah this me wrongly assuming I was the cause of this [bug](https://github.com/bids-standard/bids-specification/pull/1219#issuecomment-1217769662) and trying things around but forgetting to delete them.

Remi-Gau: FYI I a have no idea what this means:

> having a special carveout

effigies: A carveout is a case that needs to be handled specifically; here we had an invalid value being checked for but also made the default.

Apologies, this is probably an Americanism, as it is often used to describe exemptions that are written into laws.

effigies (approved):

sappelhoff (approved):

tools/schemacode/bidsschematools/render.py
Remi-Gau: No worries
I thought it was some pythonic or well known coding lingo and google was not helpful there.

tsalo (approved):
💯 😄

sappelhoff:
Thanks @Remi-Gau

Issue #1194, opened 2022-08-10
[DISCUSSION] Redundancy in the specification


Remi-Gau:
Open-ended discussion to see how people feel about the presence of redundancy in the specification.

For a bunch of reasons (how the specifcation got built, how each datatype section aims to be more "standalone"...) there is a certain amount of redundancy in the specification. That comes with pros and cons (that may be different for readers and maintainers).

That being said, I wonder if some redundancy could not be removed.

One example from https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#task-including-resting-state-imaging-data

The following passage just re-explains the inheritance principle without mentioning.

> If this information is the same for all participants, sessions and runs it can be provided in task-<label>_bold.json (in the root directory of the dataset). However, if the information differs between subjects/runs it can be specified in the sub-<label>/func/sub-<label>_task-<label>[_acq-<label>][_run-<index>]_bold.json file. If both files are specified fields from the file corresponding to a particular participant, task and run takes precedence.

I would be tempted this paragraph because of its redundant character. Others may feel otherwise as it gives an in-context short concrete "example" of the inheritance principle. However other datatypes do not get similar example so it feels very uneven on the whole and feels like it'd be simpler to not have this example at all.

This is just one random snippet of something that occurs in several places in the spec.

Would be curious how people feel about this?

Issue #1126, opened 2022-06-14
[SCHEMA] Refactor channel types


sappelhoff:
- We have lists of **coordinate systems** for different modalities
- For example: MEG, EEG
- Coordinate systems from EEG may also be used in MEG
- This is nicely expressed in the schema, see the `$ref` items: https://github.com/bids-standard/bids-specification/blob/24dbe58010d96c0ce4d4c775230cef94605c6068/src/schema/objects/metadata.yaml#L60-L72

- The situation is similar for channel types: we have types defined in different modalities (search for `type__ieeg_channels`, `type__eeg_channels`, `type__meg_channels`)
- For example types like "EEG", "RESP", or "PUPIL" -- these are the same, regardless for which modality they are used (that is, specifying PUPIL in an EEG dataset is the same as PUPIL in an MEG dataset)
- Currently these are organized as such: https://github.com/bids-standard/bids-specification/blob/24dbe58010d96c0ce4d4c775230cef94605c6068/src/schema/objects/columns.yaml#L471-L509
- Meaning we have lots of duplication

Should we just have a single `type__channels` metadata entry? that is shared across the spec?

Or should we keep them modality specific in "overarching variables", but then have a `type__channels` defined as

```yaml
type__channels:
  name: ....
  description: ...
  anyOf:
  - $ref: _type__meg_channels
  - $ref: _type__eeg_channels
  - $ref: _type__ieeg_channels
  - $ref: ...
```

sappelhoff:
closed in #919

PR #1260, opened 2022-08-29
[STY] Link to glossary in tables with many valid values

tsalo:
Closes None. Stems from https://github.com/bids-standard/bids-specification/pull/802#discussion_r957695260.

Changes proposed:

- In cases where metadata fields or columns allow only certain values (i.e., `enum`s) and there are a large number of those valid values (>15), link to the associated glossary entry instead.

### DIFF
--- a/tools/schemacode/bidsschematools/render.py
+++ b/tools/schemacode/bidsschematools/render.py
@@ -17,9 +17,9 @@ lgr = utils.get_logger()
 utils.set_logger_level(lgr, os.environ.get("BIDS_SCHEMA_LOG_LEVEL", logging.INFO))
 logging.basicConfig(format="%(asctime)-15s [%(levelname)8s] %(message)s")
 
-ENTITIES_PATH = "SPEC_ROOT/99-appendices/09-entities.html"
-GLOSSARY_PATH = "SPEC_ROOT/99-appendices/14-glossary.html"
-GLOSSARY_PATH_MD = "SPEC_ROOT/99-appendices/14-glossary.md"
+# Remember to add extension (.html or .md) to the paths when using them.
+ENTITIES_PATH = "SPEC_ROOT/99-appendices/09-entities"
+GLOSSARY_PATH = "SPEC_ROOT/99-appendices/14-glossary"
 TYPE_CONVERTER = {
     "associated_data": "associated data",
     "columns": "column",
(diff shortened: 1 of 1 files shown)


sappelhoff (commented):
looks like a good idea to me! However when clicking on a link to send me to the glossary, I find headings such as `#type-sense-1-columns` when I would have expected something like `EEG channel type keywords`

tsalo:
I'm happy to revisit how the glossary is structured- anything to make it more useful to readers. However, I'm not sure how we'd get "EEG channel type keywords"? The current name, "type _sense 1_ (columns)" comes from the fact that the column name is "type", but there are multiple column definitions with that column name, so _sense 1_ is referring to the first definition.

We currently use the schema dictionary's _key_ to dissociate different senses. While the keys have to be unique, and we recommend making the unique element (i.e., the part after the `__`) informative, we don't actually use that unique element in the glossary at all. I don't want to add a new field to the dictionary, so here are two ideas for how we could make the glossary entry names more interpretable:

1. Make the `display_name` field more specific. 
    - We don't enforce unique `display_name`s though, and if we want to use `display_name` as part of the glossary key, then we need to ensure that it's unique.
2. Use the unique element of the key in the glossary entry.
    - We'd need to make sure the unique element is meaningful, though, and we'd probably need a convention to break the unique element into distinct words.

sappelhoff:
Thanks for the explanation! I think it's "type" in particular that tripped me, because it sounds so generic ... if I had been looking at "sampling frequency *sense 1* (columns)", I don't think I would have been confused.

We should at some point refactor the channel types (#1126), also related to:

- #1194 because the channel types (eeg, meg, ieeg, soon nirs) are typically shared anyhow

rwblair (approved):

sappelhoff:
Thanks @tsalo

sappelhoff:
@tsalo not sure what went wrong, but on the fNIRS PR, the linking doesn't work so well: 

https://bids-specification.readthedocs.io/en/bep030/04-modality-specific-files/11-functional-near-infrared-spectroscopy.html

Try clicking on the _"For a list of valid values for this field, see the ..."_ link in `FiducialsCoordinateSystem`

it leads to: https://bids-specification.readthedocs.io/en/bep030/99-appendices/14-glossary.html#objects.FiducialsCoordinateSystem

that link doesn't exist ... however this one does:

https://bids-specification.readthedocs.io/en/bep030/99-appendices/14-glossary.html#fiducialscoordinatesystem-metadata

tsalo:
It should go to `objects.metadata.FiducialsCoordinateSystem`. I'll take a look now.

Issue #612, opened 2020-09-17
Remove duplicate entity definitions in specification text


tsalo:
In #568 we created a new appendix with the entity definitions. Whenever the entities are referenced in the text, a hyperlink is used to point to the appropriate section of the appendix. The appendix definitions should be considered the "canonical" definitions for these entities, with text in the rest of the specification used only to provide modality-specific context and examples. We did not, however, edit the specification text much in #568, so there are a lot of duplicate definitions and elaborations that could be removed or at least cut down.

Remi-Gau:
Was thinking I could do some of that.

Just to be sure: any mention of any of the entities from the appendix should have a hyperlink?

But only if it is "on its own" and not in an example of a "path".

For example:

Add a hyperlink to `sub-<label>` in the example below.

```
The purpose of this RECOMMENDED file is to describe properties of participants
such as age, sex, handedness.
In case of single-session studies, this file has one compulsory column
`participant_id` that consists of `sub-<label>`, followed by a list of optional
columns describing participants.
Each participant MUST be described by one and only one row.
```

But obviously not `sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.json`

```
In this particular case, the sidecar JSON file
`sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_phasediff.json`
MUST define the time of two echos used to map the phase and finally calculate
the phase-difference map.
```

tsalo:
I _wish_ we could render code snippets with hyperlinks, but I don't think it's possible in Markdown (it was something I looked into for #610). 

In addition to the linking, there are actual definitions that could be eliminated or reduced, such as this one in MEG:

> The proc-<label> entity is analogous to rec-<label> for MR and denotes a variant of a file that was a result of particular processing performed on the device. This is useful for files produced in particular by Elekta’s MaxFilter (for example, sss, tsss, trans, quat, mc), which some installations impose to be run on raw data because of active shielding software corrections before the MEG data can actually be exploited.

I don't know what the updated text should be, but since the description in the text is a complete duplicate of the definition of the entity, it would be great to reduce it. I think the main text should provide some section-specific context, rather than fully defining the entity. Plus, duplications in the text will make it harder to update entity definitions, since we'll have multiple places we have to change.

tsalo:
Here is a gist that may make it possible to add hyperlinks to code blocks: https://gist.github.com/jesstelford/cb4dd6fafc18221ce27250e84fd19327

[...]

PR #1265, opened 2022-08-30
[ENH] Remove redundant entity definitions on MRI page

tsalo:
Related to #612.

This PR removes any paragraphs describing entities, wherein the description does not differ appreciably from, or expand upon, the official entity definition in the schema, from the MRI page.


tools/schemacode/bidsschematools/render.py
```diff
@@ -477,7 +477,10 @@ def make_filename_template(
 
 
 def append_filename_template_legend(text, pdf_format=False):
+    """Append a legend to filename templates."""
     legend = """
+- For more information about filename elements (for example, entities, suffixes, extensions),
+  follow the links embedded in the filename template.
```
effigies, line 483: This will probably be confusing in the PDF build.

tsalo: Done!

Remi-Gau (approved):
LGTM

Thanks for this refactoring @tsalo 

Had a quick look I think that there may be a paragraph or 2 we could similarly remove from PET or microscopy in another PR.

tsalo:
>Had a quick look I think that there may be a paragraph or 2 we could similarly remove from PET or microscopy in another PR.

Absolutely! I definitely want to work through the rest of the spec to remove anything unnecessary.

sappelhoff (approved):

Remi-Gau (approved):

PR #1271, opened 2022-09-01
[FIX] Use correct glossary path in links generated by #1260

tsalo:
Closes None, but fixes a bug that I introduced in #1260.


sappelhoff:
Thanks for the quick fix! I also don't know how that issue sneaked its way into the PR ... I remember successfully testing before approving :shrug: anyhow, seems to work now.

PR #967, opened 2022-01-07
[ENH] BEP034 Computational modelling

MichaelSchirner:
Dear BIDS community,

this PR replaces [#850](https://github.com/bids-standard/bids-specification/pull/850) on BEP034 Computational modelling.

Briefly, in the original PR the new modality `comp` with datatypes
1.   time series (and their derivatives)
2.   connectivities
3.   model descriptions (equations and parameters)
4.   coordinates
5.   code

was proposed.

In this new PR we address feedback from the maintainer meeting:

-    We now removed the datatype "time series" from the PR and instead will join efforts with the group around [issue 713 "Convergence on time series and events"](https://github.com/bids-standard/bids-specification/issues/713) towards making "time series" a BIDS-wide datatype that can also be used in other modalities. 
-    Furthermore, we will join efforts with [BEP017 Generic BIDS connectivity data schema](https://docs.google.com/document/d/1ugBdUF6dhElXdj3u9vw0iWjE6f_Bibsro3ah7sRV0GA/edit#heading=h.mqkmyp254xh6) and add the use case "structural connectivity for computational model simulation" to BEP017 to achieve a BIDS-wide connectivity datatype that can be used for different modalities.  
-   Instead, this new PR now introduces only one new data type "model" with the filename suffixes "_eq" and "_param" and the file extension ".xml" to store the equations and parameters of computational models in [LEMS](http://lems.github.io/LEMS) format.
-    Lastly, we propose to store the source codes and the machine codes that were used to create computational results in the modality agnostic folder `code`.

The new PR will be discussed **on Jan 11 at 17 CET** in the [NFDI-Neuro Webinar](https://nfdi-neuro.de/event/nfdi-neuro-webinar-ritter-et-al-bids/), please join the discussion.

PR #1280, opened 2022-09-12
WIP: Proposal for `model-<modelname>` directory 

arokem:
As an auxilliary to work being done on BEP16 (DWI derivatives), we are proposing a change to the common derivatives specification to allow the creation of a sub-directory into which model parameters will be saved, together with a `models.tsv` metadata file to describe all the different models that are fit within a pipeline.

### DIFF
--- a/src/05-derivatives/01-introduction.md
+++ b/src/05-derivatives/01-introduction.md
@@ -28,6 +28,75 @@ in [Derived dataset and pipeline description][derived-dataset-description].
     to ensure that the values are still relevant and appropriate to the type of
     output data.
 
+## Model-based grouping of derivatives 
+
+-  Here, a *model* is defined as a process via which parameters that describe
+   the data are fit/derived. This could include, but is not limited to things 
+   like: 
+    
+    1. The beta weights for different kinds of events in an event-related design 
+       in fMRI. 
+    1. The six components of the diffusion tensor in DTI. 
+    1. Etc. 
+
+   Model-based derivatives SHOULD be saved in a directory named 
+   `model-<modelname>` that is placed under the datatype from which the model 
+   was derived. A metadata file, `models.tsv` is OPTIONAL, accompanied by 
+   a `models.json` file that is REQUIRED, if and only `models.tsv` is present.
+   The `models.json` is a dictionary for the `models.tsv` columns. 
+   The purpose of the RECOMMENDED `models.tsv` file is to describe properties of the models 
+   such as the datatype from which they are derived and a human-readable description of the 
+   model. An additional metadata file, `model-<modelname>.json` is REQUIRED and describes the metadata for model fitting procedures, etc. for each model. 
+   
+   For example, if the model is principally derived from DWI data. 
+   It would go into the following structure: 
+   
+    ```Text
+    <pipeline_name>/
+        sub-<label>/
+            <datatype>/
+                model-<modelname>/
+                    <model files>
+                    <model metadata files>
+                model-<modelname>_model.json
+                models.tsv
+                models.json
+    ```
+
+    The contents of the `model-<modelname>/` folders can differ widely between models 
+    and modalities and are described in the relevant modality-specific derivative 
+    specifications. 
+
+    For a concrete example, consider the following multi-modal dataset:
+
+    ```Text
+    pipelineush-v2022a/
+        models.tsv
+        models.json
+        sub-01/
+            dwi/
+                model-DTI/
+                    sub-01_model-DTI_param-tensor_model.nii.gz
+                    sub-01_model-DTI_param-tensor_model.json
+                    sub-01_model-DTI_param-S0_model.nii.gz
+                    sub-01_model-DTI_param-S0_model.json
+                model-DTI_model.json
+            func/
+                model-VGG16/
+                    sub-01_model-VGG16_param-all_model.h5
+                    sub-01_model-VGG16_param-all_model.json
+                model-VGG16.json
+
+
+    ```
+    In this case, the `models.tsv` file could contain the following content: 
+
+    ```Text
+    model_id    datatype    description
+    DTI	dwi Diffusion tensor model.
+    VGG16	func    A VGG convolutional neural network.
+    ```
+
 ## File naming conventions
 
 -   Filenames that are permissible for a raw BIDS data type have a privileged
@@ -68,7 +137,7 @@ in [Derived dataset and pipeline description][derived-dataset-description].
 -   When naming files that are not yet standardized, it is RECOMMENDED to use
     names consistent with BIDS conventions where those conventions apply.
     For example, if a summary statistic is derived from a given task, the file
-    name SHOULD contain [`_task-<label>`](../appendices/entities.md#task).
+    name SHOULD contain [`_task-<label>`](../appendices/entities.md#task). 
 
 <!-- Link Definitions -->
 


arokem:
See https://github.com/bids-standard/bids-bep016/issues/50 for relevant context.

effigies:
I haven't really integrated this whole proposal yet, but I want to jot down some quick notes about the direction we took in FitLins, which is at least superficially quite different. Not sure if it can apply here at all...

In FitLins, we have hierarchical models (e.g., run-level -> subject-level -> dataset-level), such that you could potentially have multiple dataset-level models that build off the results in a common subject-level model. Instead of nesting models more deeply inside modalities, I've been pulling them to the top:

```
fitlins_derivatives/
    dataset_description.json
    node-run/
        sub-01/
        sub-02/
        ...
    node-subject/
        sub-01/
        sub-02/
        ...
    node-group/
        group_results
```

I imagined the `node-*` becoming derivative datasets in their own right, and then `fitlins_derivatives` becoming a super-dataset (a concept still to be defined). This would give the flexibility to, say, distribute first-level results as a standalone dataset for image-based meta-analysis.

effigies:
@arokem If you want to fix up the markdown, use `npm install remark; npx remark <markdown-file> -o`.

arokem:
Thanks for the quick comments here @effigies! Are these proposals really in conflict? I can imagine having a `model-<modelname>` directory within every one of the `node-` levels that you specified. Seems complementary to me.

arokem:
> @arokem If you want to fix up the markdown, use npm install remark; npx remark <markdown-file> -o.

Thanks! Unfortunately, working in codespaces because UT's wifi won't let me pull stuff down from GitHub :-/ 

I'll let @oesteban do the rest of these fixes :-)

oesteban:
@effigies 

```
$ npm install remark; npx remark src/05-derivatives/01-introduction.md -o src/05-derivatives/01-introduction.md
npm WARN saveError ENOENT: no such file or directory, open '/Users/oesteban/workspace/bids-specification/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/Users/oesteban/workspace/bids-specification/package.json'
npm WARN bids-specification No description
npm WARN bids-specification No repository field.
npm WARN bids-specification No README data
npm WARN bids-specification No license field.

+ remark@14.0.2
updated 1 package and audited 94 packages in 0.764s

42 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

npx: installed 55 in 2.831s
command not found: remark
```

oesteban:
the npx remark command doesn't seem to work

effigies:
I ran it. Apparently it doesn't fix everything, though...

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,76 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+It would go into the following structure:
+
+```Text
+<pipeline_name>/
+    sub-<label>/
+        <datatype>/
+            model-<modelname>/
+                <model files>
+                <model metadata files>
+            model-<modelname>_model.json
```
oesteban, line 63:
```suggestion
```
I will sleep better tonight if we don't get into prescribing anything specific about the model itself.

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,76 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+                sub-01_model-DTI_param-tensor_model.nii.gz
+                sub-01_model-DTI_param-tensor_model.json
+                sub-01_model-DTI_param-S0_model.nii.gz
+                sub-01_model-DTI_param-S0_model.json
+            model-DTI_model.json
+        func/
+            model-VGG16/
+                sub-01_model-VGG16_param-all_model.h5
+                sub-01_model-VGG16_param-all_model.json
+            model-VGG16.json
```
oesteban, line 90: Make this BIDS compatible:
```suggestion
            model-VGG16_model.json
```

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,76 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+1.  Etc.
+
+Model-based derivatives SHOULD be saved in a directory named
+`model-<modelname>` that is placed under the datatype from which the model
+was derived. A metadata file, `models.tsv` is OPTIONAL, accompanied by
+a `models.json` file that is REQUIRED, if and only `models.tsv` is present.
+The `models.json` is a dictionary for the `models.tsv` columns.
+The purpose of the RECOMMENDED `models.tsv` file is to describe properties of the models
+such as the datatype from which they are derived and a human-readable description of the
+model. An additional metadata file, `model-<modelname>.json` is REQUIRED and describes the metadata for model fitting procedures, etc. for each model.
```
oesteban, line 51: I would leave the model metadata open for definition after the folder idea is accepted. I agree that we have at BEP016 a proposal that potentially generalizes across modalities, but would not slow down this new folder proposal:
```suggestion
model.
```

PeerHerholz:
I was wondering how this fits with the [computational model BEP](https://github.com/bids-standard/bids-specification/pull/967) (if it's related/needs to)?

oesteban:
> I was wondering how this fits with the [computational model BEP](https://github.com/bids-standard/bids-specification/pull/967) (if it's related/needs to)?

As per that BEP: **BIDS Computational Model files store the mathematical and computational descriptions of
computational models as well as *simulation results*.** Based on that I'd say the only overlap would be the **simulation results** bit, and not entirely because this PR shies away from specifying models themselves. I believe that BEP has a very large overlap with the BIDS-statsmodel.

Instead, we want this PR to gauge the interest/position of the BIDS community in the possibility that we add the new `model-<model_id>/` level to the directory hierarchy of derivatives.

There are some deeply related aspects of this PR that we intendedly left for the future:
- Multi-datatype models (e.g., a model that generates derivatives from both `func/` and `dwi/` resources). We were entertaining ideas like allowing a folder `func+dwi/` to exist, or have a more general `multi/` possibility. In any case, we did not want to get into that issue here, but will come up almost immediately after this one is decided.
- The actual specification of models. There's a pretty much developed proposal, very comprehensive, in BEP016 -- thanks to the persistence, attention to detail, and overall titanic effort of @Lestropie.

Those are critical elements that affect this PR, so it's good to have them in mind (and it's a likely possibility that a majority of people push towards having one or both issues covered here too, should it be going towards an accepted decision). 

It'd be great to have critical reviews from @rwblair and @tsalo, and anyone else involved in the BIDS schema feat. Does this complicate things on that front?

Historically, this PR comes from bids-standard/bids-bep016#50. It does not itself fully resolve the inheritance issue described there, but it basically evolves from Option 3 there, which indirectly resolves the problem to the extent it is necessary within BEP016.

oesteban:
@arokem @Lestropie I have allowed myself to remove the WIP label from the title. I have made some edits to use macros and make the text more consistent with the rest of the spec.

For a render of this PR at this point, go to https://output.circle-artifacts.com/output/job/ecd7481c-1064-41ad-a51b-e362395503e4/artifacts/0/dev_docs/01-introduction.html

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,75 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+    For a concrete example, consider the following multi-modal dataset:
+
+    ```Text
+    pipelineush-v2022a/
+        models.tsv
+        models.json
+        sub-01/
+            dwi/
+                model-DTI/
+                    sub-01_model-DTI_param-tensor_model.nii.gz
```
francopestilli: @arokem @oesteban one question that came up today about this is why the suffix is `_model.nii.gz` given that both the folder and the file already contain the word `model` in `func` the suffix could be `_bold.nii.gz` perhaps in dw it could be `_dwi.nii.gz`?

oesteban: As I commented above, this PR does not intend to get into that question. I have updated the template above to be agnostic to this detail.

As for this particular example, I'd be fine by removing this example, but perhaps indicating that the suffix here is fictional and does not represent any existing specification would be sufficient.

Finally, a discussion about the BEP016 proposal about this suffix would be necessary.

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,158 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+      "models.json": "",
+      "sub-01": {
+         "dwi": {
+            "model-DTI": {
+                "sub-01_model-DTI_param-tensor_model.nii.gz": "",
+                "sub-01_model-DTI_param-tensor_model.json": "",
+                "sub-01_model-DTI_param-S0_model.nii.gz": "",
+                "sub-01_model-DTI_param-S0_model.json": ""
+            },
+            "model-DTI_model.tsv": ""
```
arokem, line 165:
```suggestion
            "model-DTI_model.json": ""
```

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,158 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+         },
+         "func": {
+            "model-VGG16": {
+                "sub-01_model-VGG16_param-all_model.h5": "",
+                "sub-01_model-VGG16_param-all_model.json": ""
+            },
+            "model-VGG16_model.tsv": ""
```
arokem, line 172:
```suggestion
            "model-VGG16_model.json": ""
```

arokem (commented):
These will be json files, right?

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,158 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+
+   {
+   "<pipeline_name>": {
+      "sub-<label>": {
+         "<datatype>": {
+            "model-<label>": {
+                "<model files>": ""
+            },
+            "models.tsv": "",
+            "models.json": ""
```
tsalo, line 79:
```suggestion
            "sub-<label>_models.tsv": "",
            "sub-<label>_models.json": ""
```

tsalo (commented):
A few initial thoughts:

1. The schema doesn't have mechanism for the additional folder layer you're proposing, so we'd need to extend it. I think it would be helpful to have a general concept associated with this level beyond "models". If this _is_ a layer the community wants to use for models, then presumably it will be applicable to other types of derivatives.
2. I think you need a `models` suffix in `schema/objects/suffixes.yaml`, as well as rules for it- perhaps those could go in `schema/rules/tabular_metadata.yaml`.

src/schema/objects/suffixes.yaml
```diff
@@ -669,6 +669,17 @@ meg:
   description: |
     Unprocessed MEG data stored in the native file format of the MEG instrument
     with which the data was collected.
+models:
```
arokem, line 672: Should this be `model` in singular?

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,159 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+
+   {
+   "<pipeline_name>": {
+      "sub-<label>": {
+         "<datatype>": {
+            "model-<label>": {
+                "<model files>": ""
+            },
+            "sub-<label>_models.tsv": "",
+            "sub-<label>_models.json": ""
```
arokem, line 80: Here, in singular?

```suggestion
            "sub-<label>_model.tsv": "",
            "sub-<label>_model.json": ""
```

arokem (commented):
Is the suffix `model` or `models`?

src/schema/objects/suffixes.yaml
oesteban: I don't think so, the file is called `_models.tsv`.

As opposed to that, there will be a new `model` entry in this file in the context of #1282.

src/05-derivatives/01-introduction.md
oesteban: (I don't think so, see above)

nbeliy:
Dear all, by chance I stumbled on this BEP, and as I'm working on various BIDS-related software, I wanted to put my 5 cents.

My concern is sub-folders like `dwi/model-DTI` may confuse BIDS tools (and make difficult to develop new) as it deviates from standard BIDS naming convention (`sub-<>_<entities>_<suffix>.<extension>`). 
From implementation point of view we will need to add a lot of tests into dataset parcing in order to index such files.

I would propose to put models into separate data-type folder together with the sidecar json and summury tsv file:
```
sub-abc/ses-def/models/sub-abc_ses-def_mod-DTI_model.d
sub-abc/ses-def/models/sub-abc_ses-def_mod-DTI_model.json
sub-abc/ses-def/models/sub-abc_ses-def_models.tsv
sub-abc/ses-def/models/sub-abc_ses-def_models.json
```
This way the model data is introduced exactly the same way as any other data file in the dataset, simplifying parsing.
The `d` extension is here just to point it's a folder.
With carefull selection of entities (for ex for parameter adjustments) you can see which folder correspond to what model:
```
sub-abc_mod-DTI_par1-00_par2-00_model.d
sub-abc_mod-DTI_par1-00_par2-01_model.d
sub-abc_mod-DTI_par1-00_par2-02_model.d
```
With convention that `00` means nominal/default value, and non-zero index the corresponding parameter adjustment.
Or alternatively using one entity for different parameter sets with meaningful naming.

For the group level models `models` folder can be moved into root directory:
```
root/
    models/acq-DTI_model.d
    sub-abc/
```

Cheers,
Nikita

oesteban:
> My concern is sub-folders like `dwi/model-DTI` may confuse BIDS tools (and make difficult to develop new) as it deviates from standard BIDS naming convention (`sub-<>_<entities>_<suffix>.<extension>`).

I don't think there's anything in this PR that is not supported by the current specifications. It's true that the BIDS Validator may require some extra heuristic, but that's more a problem of being able to formalize BIDS rules into a schema that can be used directly by the validator.

Since this only affects BIDS-Derivatives (and that's clearly stated at a couple of places in this PR) I don't think this would break any existing software either, basically because there's no specification to break at this point.

Finally, I'll also mention that this `model-<label>/` folder is basically specified like `sub-<label>` + `participants.tsv` file or the `ses-<label>/` folder. So it doesn't seem to break any existing conventions for folder names.

> I would propose to put models into separate data-type folder together with the sidecar json and summury tsv file:

I think these suggestions fit better in #1282. Can you check that out?

Thanks a lot for the input, happy to clarify further if the argument is not sufficiently convincing.

nbeliy:
> Since this only affects BIDS-Derivatives (and that's clearly stated at a couple of places in this PR) I don't think this would break any existing software either, basically because there's no specification to break at this point.

Indeed, the derivatives in BIDS are not regulated.
But even so, deviating from BIDS style of names can provide difficulties to use such derivatives in the analysis scripts.
If you take and example of [bids-matlab](https://github.com/bids-standard/bids-matlab) (free publicity!),
it allow to parse files even if they are not in the official BIDS schema.
But the `dwi/model-DTI` folder will be ignores as it doesn't starts with `sub-`.

Imagine you want a script that compare different models inside the same dataset.
If these models folders are supported, then you just need to  query all files with suffix `model`,
and you have all of them.

In opposite, you need to do recursive search for the folders containing `model-<name>`,
and then parse the subject/session names retrieve corresponding images.
It's feasible, but requires additional code.

I'm speaking less from human and more from machine perspective.
Bids software know how to deal with bids-formatted names and structure,
so any deviations will add just more complexity.

>  I think these suggestions fit better in https://github.com/bids-standard/bids-specification/pull/1282. Can you check that out?

My point was about `model-<name>` sub-folder, not about the name of suffix.
The example I give can be changed in suffixes, entities and extensions.
But I fight for data-type folder and `sub-` prefix)

oesteban:
> But even so, deviating from BIDS style

How this proposal deviates from BIDS style specifically?

> If you take and example of [bids-matlab](https://github.com/bids-standard/bids-matlab) (free publicity!),
> it allow to parse files even if they are not in the official BIDS schema.
> But the dwi/model-DTI folder will be ignores as it doesn't starts with sub-.

BIDS should remain agnostic to the decisions made by the software developer. This idea comes from the fact that scientists use folders typically to cluster the outputs of important models in their analysis. They typically group by subject first, and that's why `sub-<label>` is at the top. It seemed logical to offer the opportunity to encode models this way. That said, you don't need to have a model directory if you don't want (the same way you may not have the session folder).

> Imagine you want a script that compare different models inside the same dataset.
> If these models folders are supported, then you just need to query all files with suffix model,
> and you have all of them.

Human: goes to the `_models.tsv` file and sees all the models with relevant info.
Software: e.g., pybids could do this tomorrow with a simple edit of the bids schema inside:
```Python
>>> layout.get_models(datatype="func")
[ list of all model identifiers under datatype func ]
```

nbeliy:
From [BIDS/common principles](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-name-structure)

> For a data file that was collected in a given session from a given subject, the filename MUST begin with the string sub-<label>_ses-<label>. If the session level is omitted in the folder structure, the filename MUST begin with the string sub-<label>, without ses-<label>.
>
>Note that sub-<label> corresponds to the subject entity because it has the sub- "key" and<label> "value", where <label> would in a real data file correspond to a unique identifier of that subject, such as 01. The same holds for the session entity with its ses- key and its <label> value.

Folder `model-<name>` is collected from one subject (being placed in `sub-` folder) but don't contain `sub-` entity.

> Software: e.g., pybids could do this tomorrow with a simple edit of the bids schema inside:

But if it works without any edit, isn't better? 
If all models for a subject are collected in datatype folder, with same naming conventions,
they could be treated as normal data and retrieving with simple query:
```
layout.get_data(datatype="models", type="func")
```
I don't use pybids so I don't know how to query individual entities.

Anyway it's just a suggestion, to take or to leave.

oesteban:
> From [BIDS/common principles](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-name-structure)
> 
> > For a data file ...

This is not a data file, this is a folder. Folders do not apply the entity concatenation principle.

> But if it works without any edit, isn't better?

Your proposal (or any derivative of it) will require a change in PyBIDS of exactly the same scale of changes as the one here (because you are proposing new suffices). From this perspective, there's no difference, IMHO.

> I don't use pybids so I don't know how to query individual entities.

I'm not saying here we should cater to any specific project -- I don't use bids-matlab so for me the argument goes the other way around.

src/05-derivatives/01-introduction.md
```diff
@@ -52,6 +52,7 @@ that is placed under the datatype from which the model was derived.
       "sub-<label>": {
          "<datatype>": {
             "model-<label>": {
+                "model_description.json": "",
```
francopestilli, line 55: This seem helpful. as discussed in person

src/05-derivatives/01-introduction.md
```diff
@@ -63,6 +64,23 @@ that is placed under the datatype from which the model was derived.
...
+
+Mirroring the special role of the `dataset_description.json` in any BIDS and BIDS Derivatives
+dataset, a `model_description.json` file MUST be found at the top level of every model folder
+(see template above).
+
+Example:
+
+```JSON
+{
+  "Name": "Diffusion Tensor Model (DTI)",
```
francopestilli, line 77: Diffusion Tensor Model or perhaps better Diffusion Tensor Imaging (DTI)?

src/05-derivatives/01-introduction.md
```diff
@@ -74,6 +92,7 @@ Template:
       "sub-<label>": {
          "<datatype>": {
             "model-<label>": {
+                "model_description.json",
```
francopestilli, line 95: Looks good

francopestilli (changes requested):

src/05-derivatives/01-introduction.md
arokem:
```suggestion
                "model_description.json": "",
```

src/05-derivatives/01-introduction.md
```diff
@@ -28,6 +28,180 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+-->
+{{ MACROS___make_columns_table(
+   {
+      "model_id": ("REQUIRED", "There MUST be exactly one row for each model."),
+      "datatype": "RECOMMENDED",
+      "description__models": "RECOMMENDED",
+   }
+) }}
+
+In this case, the `models.tsv` file could contain the following content:
```
arokem, line 141: This example metadata file refers to a file/folder layout that is introduced further down, so this should probably also go further down.

arokem: Resolved in https://github.com/bids-standard/bids-specification/pull/1280/commits/2e4e8ae8a4288048821419d3dc9e101a51b184db

src/05-derivatives/01-introduction.md
arokem: OIC. I got confused. The description makes it sound like there is one `models.tsv` file for the dataset. Here, we introduce the notion that different subjects (And different datatypes?) can have their own `sub-<label>_models.tsv` file (and `<datatype>_models.tsv` is also allowable?). We might want to describe this before providing an example.

src/05-derivatives/01-introduction.md
oesteban: Okay, it was also cheap to replace `_model` with `_<suffix>` because it is totally true this example is a bit distracting (hopefully that will be fixed by #1282).

src/05-derivatives/01-introduction.md
oesteban: Right, when I added the `sub-<>_` prefix I actually disallowed a `/models.tsv` implicitly. I'll send some changes to make this explicit.

francopestilli (approved):

Lestropie:
Making "model" a datatype is an interesting suggestion. It might cause more ambiguity than it solves, but let me follow that train of thought, if I can maybe poke and prod at the concerns about BIDS "conformity" in the most general sense.

You can conceptualise the current filesystem structure as following this hierarchy:
1. Single-entity-based directories (`sub-<label>`, `ses-<label>`);
2. One non-entity directory (datatype);
3. Files (must replicate entities from 1. and have a suffix).

What's being proposed can violate this structure in a couple of different ways:
1. Having eg. `model-<label>/` as a sub-directory of eg. `dwi/` breaks the ordering between 1 and 2;
2. Having directory names replicate entities from parent directories has no precedent.

Point 1 is what is resolved by having `model` as its own datatype. Point 2 is a consequence of us going for Option 3 in https://github.com/bids-standard/bids-bep016/issues/50.

Hoping that might help rather than hinder seeing where each side is coming from.

francopestilli:
hi @oesteban @Lestropie @arokem @PeerHerholz @poldrack @adelavega @effigies 

I finally had the time to look at all these proposals [organized by @Lestropie ](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit) I am grateful for that. 

Also, I looked over what @oesteban has proposed [here](https://github.com/bids-standard/bids-bep016/issues/69), thank you @oesteban. 

Furthermore, @PeerHerholz pointed out that the term model (and 'models') is already being used in the [Stats Model Specification](https://bids-standard.github.io/stats-models/index.html) and by the [Computational Models](https://github.com/bids-standard/bids-specification/pull/967) which impinge on each others.

I am starting from the principles and on the fact that we should not fix BIDS but develop a description for connectivity data: 
*Less is more*
target easier and doable goals, finish less but get it done!

*The 80/20 rule*
tackle solutions that address 80% of most commonly used cases

*Identify & focus on*
 - core principles not extreme/rare cases
 - main ideas not specific details
 - creating the backbone of the BEP, avoid focus on theoretical aspects


It seems to me that we are not going to find a solution that will **not** have problems. Each solution will either require changes to BDIS, or (many of them) break the inheritance principle, or overlap with other datatypes or the Stats Model Spec.

I would like to suggest that instead of trying to solve this problems, the 8 of us. We advance with a BEP. We pick one option that will have problems, that it will not be perfect and bring the option to the wider community.

I also suggest that we bring it with the hope that the broader community accepts it. This is because we can gain much from having a poor BIDS spec on this and little from continuing to find a perfect way.

I am aware that what I am writing means that we will most likely later need to make changes. Yet, I would love to use the 80/20 or even a 70/30 rule here and just move forward with a proposal that will be sent to the community for feedback.

Looking at all the proposals [described by @Lestropie here](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit#) and [the option described by @oesteban here](https://github.com/bids-standard/bids-bep016/issues/69), we attempted to make another example for functional model [here](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit#bookmark=id.dk273skz9kbj)

The solution seems to be OK to us. Very similar to what a [Stats Model Specification](https://bids-standard.github.io/stats-models/index.html) but reasonable, not perfect. This approach should also allow to rewrite the [Dimensionality reduction network](https://docs.google.com/document/d/1GTWsj0MFQedXjOaNk6H0or6IDVFyMAysrJ9I4Zmpz2E/edit?usp=sharing) and not affect the [Connectivity Matrix description](https://docs.google.com/document/d/1ugBdUF6dhElXdj3u9vw0iWjE6f_Bibsro3ah7sRV0GA/edit?usp=sharing) which uses an approach similar to the original DWI BEP16 approach.

I am sure we can find cases in which this fails. But 7030 ok to y'all? Can we keep this one and send it to the community, please?

arokem:
@francopestilli : just making sure that I understand what you are suggesting: you want to incorporate the content of this PR into the [BEP16 repo](https://github.com/bids-standard/bids-bep016/) and close it here? Basically, kicking the can of the discussion on the different alternatives that @Lestropie detailed in [the google doc](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit) to the discussion on the BEP? Is the idea that the other BEPs that depend on this decision wait to see how BEP16 shakes out?

adelavega:
I would suggest we should not merge it with BEP 16, but I also don't think we can make this decision in haste, as it has potential implications.

Alternatively, this becomes its own BEP, and BEP16 and related efforts "depend" on the `model` BEP. That would give people more time to review this proposal without necessarily holding anything up.

adelavega:
+1 for `node`. I think `model` is a very loaded word with more conceptual implications, whereas `node` can be more generic.

francopestilli:
> @francopestilli : just making sure that I understand what you are suggesting: you want to incorporate the content of this PR into the [BEP16 repo](https://github.com/bids-standard/bids-bep016/) and close it here? Basically, kicking the can of the discussion on the different alternatives that @Lestropie detailed in [the google doc](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit) to the discussion on the BEP? Is the idea that the other BEPs that depend on this decision wait to see how BEP16 shakes out?

Hi @arokem actually I went over each one if the options @Lestropie proposed in the Google doc. I added also examples at the bottom of the doc of how this proposal would save func data. 

I was not able to find a solution in any of the options @Lestropie reviewed and presented. Did i get that wrong? Was there any option in the doc that was a potential solution / a way forward?

Just to clarify i did follow everything rob did

oesteban:
> We also are using `node` in the sense of a node in a computational graph, and I am for using it that way here. (As an aside, I note that BEP 017 is using it to refer to nodes in a connectivity graph defined by a connectivity matrix...)

I think we are talking about different *nodes* in a computational graph. I'm fairly confident that your `node-run`, or `node-subject` have internally more *nodes* of a computational subgraph inside. Here, *node* is used as an *atomic* entity that doesn't have a computational sub-graph underneath (obviously, this is mostly theoretical and arbitrary because these models will certainly be decomposable in smaller pieces).

Regarding the overlap with the connectivity BEP - I have less arguments to maintain `node`. Perhaps, we could go back to the root of the PR and, instead of enabling these folders to encode computational nodes (or a particular model's outcomes), think of these as a utility for the user to create "results sets" (which is the common practice of researchers this BEP is trying to enable):

```
└─ <pipeline_name>/
   └─ sub-<label>/
      └─ <datatype>/
         ├─ set-<label>/
         │  ├─ resultset_description.json 
         │  └─ <node files> 
         ├─ sub-<label>_sets.tsv 
         └─ sub-<label>_sets.json 
```

As you can see, I'm shying away from putting this higher in the hierarchy because that will open another can of worms. A less voluminous problem we intendedly avoided was multi-datatype analyses. It seems reasonable to propose allowing `set-<label>` folders as siblings of `<datatype>/`, having a new `multi/` datatype folder where these resultsets can be stored or allowing composition `<datatype1>+<datatype2>/`. However, that is left for future PRs and debate or otherwise this one might not see a decision ever. For the same reasons, I would avoid stating in this particular PR that the folders can be used at any level of the tree structure.

francopestilli:
> ?

Re "the other BEPs that depend on this" it looks like we have had several of the community discussing this issue. Would it be fair to hope that this can be accepted quickly?

-- Yes, we we would have BEPs waiting for this and one BEP (the dimensionality reduction one) has already used the 'model-' specification (it would need to be updated to the final version though).

oesteban:
> BEPs waiting for this and one BEP (the dimensionality reduction one) has already used the 'model-' specification (it would need to be updated to the final version though).

I would encourage everyone to try not to need this feature -- I see it more as a resource to the user than as a key to unlock the progress at the complicated points of other BEPs. Otherwise, we risk rushing important changes that we will not be able to roll back (backward compatibility) and I agree with @adelavega that we do want the input of as many as possible before proceeding with this idea.

arokem:
I like "result" instead of "model" ("resultset" also fine!), to avoid the already crowded space around that word. I think that "node" might confuse both because of the overloading of the term with analysis of connectivity, and because I suspect that there are many that don't think of the results they obtain in terms of nodes in a computational graph, even if that's a technically accurate description (I don't usually, unless someone/something forces me to). 

@francopestilli : I am still not sure what you suggest. That we move on with the proposal in this PR? The example you added at the bottom of the google doc is not based on this PR, but based on a different variety of this idea ("alternative 3"), which does not introduce the new "results_description.json".  

IIUC, one difference between "alternative 3" and the current PR is whether result/resultset metadata are stored in multiple files that are all called "resultset_description.json" (or similar) or whether we require that this be saved in files that have different names. Though formally I think that this is a minor difference, and both seem acceptable to me, I think that there might be some practical benefits to varying names across results/resultsets in minimizing confusion. 

So, with minimal changes from the proposal in https://github.com/bids-standard/bids-specification/pull/1280#issuecomment-1255269988:

```
└─ <pipeline_name>/
   └─ sub-<label>/
      └─ <datatype>/
         ├─ set-<label>/
         │  ├─ set-<label>.json  
         │  └─ <resultset files> 
         ├─ set-<label>_<suffix>.json 
         ├─ sub-<label>_sets.tsv 
         └─ sub-<label>_sets.json 
```

And with some additional variations (adding a suffix, pulling it up one directory)

```
└─ <pipeline_name>/
   └─ sub-<label>/
      └─ <datatype>/
         ├─ set-<label>_<suffix>/
         │  └─ <resultset files> 
         ├─ set-<label>_<suffix>.json 
         ├─ sub-<label>_sets.tsv 
         └─ sub-<label>_sets.json 
```

Could potentially go even further:  

```
└─ <pipeline_name>/
   └─ sub-<label>/
      └─ <datatype>/
         ├─ sub-<label>_set-<label>_<suffix>/
         │  └─ <resultset files> 
         ├─ sub-<label>_set-<label>_<suffix>.json 
         ├─ sub-<label>_sets.tsv 
         └─ sub-<label>_sets.json 
```

Where the `sub-<label>` entity could then further down the line be replaced by a `group-<label>` entity if that makes sense, but that's probably for a future PR.

arokem:
@francopestilli : You [said](https://github.com/bids-standard/bids-specification/pull/1280#issuecomment-1255267461): 

> I was not able to find a solution in any of the options @Lestropie reviewed and presented. Did i get that wrong? Was there any option in the doc that was a potential solution / a way forward?

Is this about functional derivatives? Is there something special about the functional derivatives that makes it impossible to store them in the way proposed in this PR or in the other alternatives in the [google doc](https://docs.google.com/document/d/12V2Q4y-JhKdMm6-kwFo6PeIyoKL8lKUwJs-MG6JUprw/edit) (I'll keep linking to it, so that we don't get confused about which doc)? Or -- alternatively- in one of the ways in my recent [comment](https://github.com/bids-standard/bids-specification/pull/1280#issuecomment-1255438301)? I might be missing something, not being as familiar with the functional derivatives.

oesteban:
Thanks for the rebase.

There seems to be more agreement with the "result-set" concept, instead of model, WDYT?

arokem:
I think we want to avoid that dash, right? So either "results" or "resultset"?

poldrack:
+1 for "results" as it's the common usage (no one ever uses "resultset" in real life)

oesteban:
> I think we want to avoid that dash, right? So either "results" or "resultset"?

Yes, definitely.



> +1 for "results" as it's the common usage (no one ever uses "resultset" in real life)

There are two things I like about "resultset":
1. Feels natural to use `set-<label>` as entity, while `res-<label>` is already there (for resolution in derivatives) and `result-<label>` has double the length.
2. It stresses the fact that there might be other "sets" of results under the same derivatives

However, the same way we have "participants" and `sub-<label>`, we could have "results" and `set-<label>`, and I'd be happy with this.

francopestilli:
I agree. Result sounds good. Even though in my mind result sounds more scientific that these derivatives will be.

arokem:
> I agree. Result sounds good. Even though in my mind result sounds more scientific that these derivatives will be.

I see what you mean. I think this is important if we can foresee wanting to use "result" for some future derivatives that are closer to the scientific conclusions. OTOH, something even closer to the "results" section of a paper might fit into the concepts laid out here too.

arokem:
I've tried to implement something along these lines in the most recent [45467f8](https://github.com/bids-standard/bids-specification/pull/1280/commits/45467f808198298cfc845a2a13b97d1a41654c9e). I don't think that we can entirely get rid of the notion of a model and the related `model_id`, so I didn't here. The question is whether this provides the clarity and generality we would hope for.

PeerHerholz:
Hi everyone,

I would like to revive this thread and the corresponding discussion. 

After a meeting with @arokem concerning the [BIDS connectivity](https://pestillilab.github.io/bids-connectivity/)-related [BEPs](https://pestillilab.github.io/bids-connectivity/scope_approach/), we wondered about the current state of this proposal. Given its scope and aims, we consider it a very useful and important addition. For example, within the [connectivity-related BEPs](https://pestillilab.github.io/bids-connectivity/scope_approach/), [BEP16](https://github.com/bids-standard/bids-bep016/blob/53f74843b89f1a23251dda60101f7f895861590a/src/derivatives/05-diffusion-derivatives.md) and [Dimensionality reduction-based networks](https://docs.google.com/document/d/1GTWsj0MFQedXjOaNk6H0or6IDVFyMAysrJ9I4Zmpz2E/edit?usp=sharing) would both profit from a coherent way of describing what `model` was applied in which manner and what files were derived from that (e.g. mentioned [here](https://github.com/bids-standard/bids-bep016/issues/68)).

However, we are also aware that it will introduce quite a few changes regarding multiple core aspects and `BEP`s.

oesteban:
Thanks @PeerHerholz. @bids-standard/steering - perhaps some mechanism to initiate a reflection period and a deadline for making a decision would be helpful to preempt PR stagnation.

effigies:
@arokem I'm going to propose a fix for the failing table, if you want to hold off on pushes for a couple minutes.

effigies:
Okay, I'm done.

arokem:
For the time being, I have rebased this on top of current master to facilitate more discussion of this proposal. 

@tsalo : regarding your [previous comment](https://github.com/bids-standard/bids-specification/pull/1280#pullrequestreview-1106245851), should I add a new file called "models.yaml" [in here](https://github.com/bids-standard/bids-specification/tree/master/src/schema/rules/tabular_data) and try to follow the example of the existing files in that folder?

PR #1285, opened 2022-09-15
update documentation regarding cap manufacturing.

helenacockx:
- this includes changing the examples as proposed by @rob-luke [here](https://github.com/bids-standard/bids-specification/pull/802#discussion_r903334061)
- -the remark of @robertoostenveld regarding not making the string `custom` required for the field CapManufacturer (see [here]( https://github.com/bids-standard/bids-specification/pull/802#discussion_r904838368)) is not yet addressed by this pull request


src/04-modality-specific-files/11-functional-near-infrared-spectroscopy.md
```diff
@@ -58,33 +58,35 @@ the `*_nirs.json` file.
...
+-   If an Artinis Medical Systems cap with custom positions,
+as may be done by cutting custom holes in the cap,
+was used:
     ```JSON
-    "CapManufacturer": "EasyCap",
-    "CapManufacturersModelName": "custom",
-    "NIRSPlacementScheme": "n/a"
+    "CapManufacturer": "Artinis Medical Systems",
+    "CapManufacturersModelName": "headcap with print, size L, it was modified by adding holes for the optodes according to the NIRSPlacementScheme and optode_layout.pdf",
+    "NIRSPlacementScheme": "see optode_layout.pdf: 2 groups over the left and right dlPFC, 2 groups over the left and right PPC, 1 group over the left M1 and PMC"
```
sappelhoff, line 82: if a PDF file is recommended in this example, I think it should be mentioned in the text that this file needs to be listed in .bidsignore

effigies: Or `sourcedata/`?

sappelhoff (commented):
> -the remark of @robertoostenveld regarding not making the string custom required for the field CapManufacturer (see https://github.com/bids-standard/bids-specification/pull/802#discussion_r904838368) is not yet addressed by this pull request

@robertoostenveld argues that a cap manufacturer is always defined, and "custom" hides that information. Instead, we should be specific with that field and, for example, put the name of the tailor that sewed the cap there.

I agree with that in principle, however, I would argue that we should still leave `CapManufacturer` as "custom" then, and provide descriptions in an additional field like `CapManufacturerDescription`. That's because analysis software can then more easily parse this field: That is, the software may know that "custom" is unparsable, but not that "tailor XYZ" is **not** a standard model that should be added to the software.

src/04-modality-specific-files/11-functional-near-infrared-spectroscopy.md
robertoostenveld: Alternative graphical depictions of the cap could be used, for example a photo, or a bitmap drawing, or a word document. I think that those documents themselves should not be part of the spec (and hence must be listed in the bidsignore for the validator)

helenacockx:
I suggested two adaptations of the documentation regarding the discussion here above:
- I added the explanation that you can refer to a file with a graphical depiction of the cap in the 'NIRSPlacementScheme' and explained that this file should be listed in the .bidsignore file
- I improved the wordings to explain what 'CapManufacturer' = custom signifies, by changing the order of the sentences and being more explicit about what this means (i.e. no (modified) commercially available cap)

effigies:
Sorry, I made a drive-by comment earlier without very clearly considering the situation here, and I'm afraid we've wasted your time.

`.bidsignore` is a validator hack (or detail, if you want to be charitable). The validator is mentioned a couple times in the spec, but not as an arbiter of what you should do, and I think it's a bad idea to make the specification depend on the details of one implementation. If we start specifying what should go in the `.bidsignore`, we're going to need to specify the format of `.bidsignore`. And of course, one of the problems with `.bidsignore` is that it is used to abuse the BIDS validator and allow non-BIDS data to masquerade as nearly-empty BIDS datasets, which severely harms data reusability. So it should remain use-at-your-own risk, and not be blessed by the spec.

The specification does provide for four directories that are essentially shielded from inspection by the validator because they are so broad: `sourcedata/`, `derivatives/`, `stimuli/` and `code/`. In MRI, I believe if someone wanted to store an exam card, they would put it in `sourcedata/`, which is why I suggested it. I don't know if that would be acceptable for NIRS users, but if they're willing to use the `.bidsignore` hack, it's not clear why they would object to that.

If `sourcedata/` is unsatisfactory, we could introduce a more formal (again shielded) directory called `documentation/` (or `docs/` or `documents/`) which would be the go-to recommendation whenever this comes up in the spec.

rob-luke:
I am very pleased with the changes @helenacockx suggests here. However, how do we resolve the bidsignore vs sourcedata discussion? either way would work for me.

robertoostenveld:
good point by @effigies. I consider adding a  `/documentation`folder an overkill if it is just for (f)NIRS, as it impacts all modalities and would have to be discussed more widely. For now I would be fine with suggesting in the specification that informative photos can be stored in `/sourcedata`.

rob-luke:
Great! Lets go with `/sourcedata` then. @helenacockx can you make the changes and I will hit merge! Thank you all

src/04-modality-specific-files/11-functional-near-infrared-spectroscopy.md
```diff
@@ -56,35 +56,39 @@ be directly attached to the scalp with adhesive. To facilitate description of
 the wide variety of possible configurations, several fields are RECOMMENDED within
 the `*_nirs.json` file.
 Additionally, in certain situations, reserved keywords MUST be used.
-When a custom made cap is used, the reserved keyword `custom` MUST be used in the
-`CapManufacturer` field.
-Similarly, when a custom cap is used or custom modifications are made to a cap,
+When custom modifications are made to a commercially available cap or a custom cap is used,
 then the reserved keyword `custom` MUST be used for the `CapManufacturersModelName` field.
+When a custom made cap is used, i.e. no (modified) commercially available cap,
```
sappelhoff, line 61:
```suggestion
When a custom made cap is used, that is, no (modified) commercially available cap,
```

src/04-modality-specific-files/11-functional-near-infrared-spectroscopy.md
```diff
@@ -56,35 +56,39 @@ be directly attached to the scalp with adhesive. To facilitate description of
...
+the reserved keyword `custom` MUST be used in the `CapManufacturer` field.
 If no cap is used, the reserved keyword `none` MUST be used in the `CapManufacturer`
 and `CapManufacturersModelName` field.
+The use of 'NIRSPlacementScheme' is RECOMMENDED, certainly when no cap or a costomized cap is used,
```
sappelhoff, line 65:
```suggestion
The use of `NIRSPlacementScheme` is RECOMMENDED when no cap or a customized cap is used,
```

src/04-modality-specific-files/11-functional-near-infrared-spectroscopy.md
```diff
@@ -56,35 +56,39 @@ be directly attached to the scalp with adhesive. To facilitate description of
...
+When custom modifications are made to a commercially available cap or a custom cap is used,
 then the reserved keyword `custom` MUST be used for the `CapManufacturersModelName` field.
+When a custom made cap is used, that is, no (modified) commercially available cap,
+the reserved keyword `custom` MUST be used in the `CapManufacturer` field.
 If no cap is used, the reserved keyword `none` MUST be used in the `CapManufacturer`
 and `CapManufacturersModelName` field.
+The use of `NIRSPlacementScheme` is RECOMMENDED when no cap or a customized cap is used,
+and describes the positioning of the optodes. This field may also contain a reference to a file providing
+a graphical depiction of the cap, for example a PDF file, a photo, or a bitmap drawing. Note that this file
+should be placed in the [`/sourcedata`](../02-common-principles.md#source-vs-raw-vs-derived-data) directory, as it cannot be recognized by the validator.
```
effigies, line 68: Here's some language that leaves room for us to specify some options in the future:

```suggestion
and describes the positioning of the optodes.
This field may also contain a reference to a file providing a graphical depiction of the cap,
for example a PDF file, a photo, or a bitmap drawing.
If the referred file is not specified in BIDS, it MAY be placed in the
[`/sourcedata`](../02-common-principles.md#source-vs-raw-vs-derived-data) directory.
```

I don't want to suggest adding it to the BEP at this late date, but *EG and microscopy all define the `_photo` suffix, which might cover the use cases that don't require a PDF containing multiple images. That could be an easy extension after 1.8.0.

robertoostenveld: A difference is that in \*EG the "landmark and/or fiducial" photos are taken at the individual participant level and serve as "data", whereas here it (probably) is a one-time photo of a cap on a dummy head or a schematic drawing that serves to document the cap design.

So summarizing: we have photos for the purpose of \*EG coregistration, and we can have a  picture or drawing of the NIRS cap. We could in principle define/document both the other way around as well (i.e. custom EEG or MEG-OPM cap design picture, and NIRS landmark photos), but I don't see the immediate need for that. I suggest to do that with a future small PR once the need arises (i.e. upon request).

rob-luke:
Thanks @helenacockx @robertoostenveld @effigies @sappelhoff 
Merging 🎉

PR #1287, opened 2022-09-16
SCHEMA: Drop unstandardized name field from objects.top_level_files

effigies:
Proposed in #1112 comments. `name` is neither the file path nor a consistently human-readable string (but `display_name` is this). I think we can drop this without consequence.


bendhouseart (approved):
Agreed, name doesn't seem to add much to these merge away.

PR #1288, opened 2022-09-16
SCHEMA: Reorganize file objects and rules

effigies:
This is a significant refactoring of the schema. I'm not sure if it justifies a minor version bump, since it will break things, but we haven't released a minor release since the last breakage.

Builds off of #1287, which was separated out from these drastic changes.

Summary:

* `objects.files`
   * Combines old `objects.top_level_files` and `objects.associated_data`
   * Removes `name` field, adds `file_type`, which is either `regular` or `directory`
* `rules.files`
   * Contains `common`, `raw` and `deriv`
   * `rules.files.common`
      * Adds `level` field, replacing `required: true`
      * `rules.files.common.core` combines `rules.top_level_files` and `rules.associated_data`, minus participants and samples
      * `rules.files.common.tables` combines `rules.tabular_metadata` and participants and samples
   * `rules.files.raw`
      * Mostly copies of `rules.datatypes.*`
      * Separates out associated data that spans more than one modality
         * `rules.files.raw.channels` - contains channels, coordsystem, electrodes (will contain optodes with nirs)
         * `rules.files.raw.photo` - contains photos
         * `rules.files.raw.task` - contains events and timeseries
   * `rules.files.deriv`
      * `rules.files.deriv.preprocessed_data` - was `rules.datatypes.derivatives.common_derivatives`, but removed a lot of derivatives of things where `space` wouldn't apply (such as events and photos). We can consider re-adding these with `desc-`, but I'm not sure how much these will ever occur. They're also not "data files" in the way that derivatives discusses, so it's not clear that they're specifically provided for, either.
      * `rules.files.deriv.imaging` - was `rules.datatypes.derivatives.common_imaging_derivatives`
         * This could probably be cleaned up further. Some of the things are actually supersets of what's in `preprocessed_data`. I've run out of energy in this effort, though, and it's already pretty big.

This will break tests. I'll get to them, but I wanted to get comments. @bids-standard/schema-users


tsalo (approved):
I like the changes.

Just going to make this note to refer back to whenever the `file_type` thing confuses me:

> A Linux system, just like UNIX, makes no difference between a file and a directory, since **a directory is just a file containing names of other files.**

tools/schemacode/bidsschematools/render/text.py
```diff
@@ -209,6 +209,7 @@ def value_key_table(namespace):
 
 
 def make_filename_template(
+    file_type,
```
tsalo, line 212: Should this be "dataset_type"? "file_type" makes me think "directory" vs. "regular" based on the other changes in this PR.

tsalo (commented):
Just have one thought about the new `make_filename_template` parameter.

tools/schemacode/bidsschematools/render/text.py
effigies: The idea was that the rules are for files more than for datasets, but that's a fair point. Happy to change to `dataset_type` unless someone has a better/clearer suggestion.

src/schema/rules/files/raw/photo.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - .tif
+  datatypes:
+    - eeg
+    - ieeg
+    - meg
+    - nirs
+  entities:
+    subject: required
+    session: optional
+    acquisition: optional
```
effigies, line 17: @rob-luke @sappelhoff I didn't follow the discussion of `photo.*` closely. Was there a reason that nirs photos were missing the optional acquisition flag? I can break this back out into a separate rule, but I wanted to check that this wasn't an oversight first.

effigies: Actually, now I'm seeing that the rule was never rendered, and there's no mention in the text. Should I just remove it?

rwblair:
Some entries in rules.files.* have corresponding entries in objects (everything in common) and others don't (everything in raw and deriv). This is not a problem. What do we imagine happening if a directory becomes a valid entry in a datatype directory? And then we need to specify rules for can be in that directory? I suppose there's no reason that any given raw or deriv rule couldn't have a corresponding entry in objects if it really needs to. Not something that I'd like to write for every single thing.

Also with an eye towards the idea of something like `group-<label>` becoming a valid directory in the root directory of derivatives, how would we specify that in the rules.deriv folder? Could add a key to the rules specifying that the file should be at root `location: '/'` or that its parent is a specific entry in objects (or another rule). Under the latter regime `dataset_description.json` might get a new key in its rule entry like`parent: $ref.objects.directories.root`.

None of this is blocking. All the rules in this PR are interpretable and make a lot of things easier as a consumer of the schema. (My secret agenda is to get rules specifying that subject directories are legal at the root of the dataset. and whats legal inside of them and so on.)

tools/schemacode/bidsschematools/validator.py
```diff
@@ -360,20 +306,16 @@ def load_all(
     -------
     all_regex : list of dict
         A list of dictionaries, with keys including 'regex' and 'mandatory'.
-    my_schema : list of dict
+    my_schema : Mapping
         Nested dictionaries representing the full schema.
     """
```
effigies, line 311: @TheChymera Do you want to look over the changes to the validator code at all?

PR #1289, opened 2022-09-16
[ENH] use schema to mention which "top directories" are allowed

### LINKED ISSUE #1284: What directories can be at the top level of a BIDS Dataset?
Is there some place in the BIDS specification that states what directories under the data set root are valid?

I know that the following are okay:  `derivatives`, `stimuli`, `code`, `sourcedata`.... but I couldn't find anywhere in the spec where it clearly stated that.   Could someone point me there if I missed it?

I also know that putting random directories under the data set root doesn't validate.  Is there a list somewhere?

Remi-Gau:
closes #1284 

- [x] update macro doc
- [x] add test


tools/mkdocs_macros_bids/macros.py
```diff
@@ -347,3 +347,20 @@ def define_common_principles(src_path=None):
...
+
+    text = ""
+
+    for key in schema_obj["objects"]["associated_data"]:
+
+        required = schema_obj["rules"]["associated_data"][key]["required"]
+        status = "REQUIRED" if required == True else "OPTIONAL"
+        description = schema_obj["objects"]["associated_data"][key]["description"]
+
+        text += f"- `{key}`: [{status}] {description}"
```
Remi-Gau, line 364: @tsalo do you think the core of this macro should actually go into the render module of the schema tools?

src/schema/objects/associated_data.yaml
```diff
@@ -6,10 +6,10 @@
 code:
   display_name: Code
   description: |
-    A directory in which to store any code used to generate the derivatives from the
-    source data.
+    A directory in which to store any code used to generate the derivatives
+    from the source data.
```
Remi-Gau, line 10: This description feels too restrictive.

I am pretty sure we have datasets in the examples that contain also the code used to run the fMRI, EEG, ... experiments.

Also, shouldn't this way "the raw data from the source data"?

Remi-Gau:
@VisLab 

Is that what you had in mind?

https://bids-specification--1289.org.readthedocs.build/en/1289/02-common-principles.html#other-top-level-directories

src/02-common-principles.md
```diff
@@ -109,6 +109,20 @@ data type as defined above.
 A data type directory SHOULD NOT be defined if there are no files to be placed
 in that directory.
 
+### Other top level directories
+
+In addition of the subject sub-directories, the root directory of a BIDS dataset
+can also contain the following directories:
+
```
Remi-Gau, line 116: Definitely not sure about the wording here: suggestions more than welcome.

src/schema/objects/associated_data.yaml
Remi-Gau: may be better to reuse the text we already have:

https://bids-specification.readthedocs.io/en/latest/03-modality-agnostic-files.html#code

Remi-Gau:
Another would be useful would be to add internal links to the sections of specs that relate to each directory.

tools/mkdocs_macros_bids/macros.py
tsalo: I think that would be preferable

Remi-Gau: okidoki

will give this a go... but tomorrow... now is 💤 time...

VisLab:
Perfect!

tools/mkdocs_macros_bids/macros.py
Remi-Gau: Done

Remi-Gau:
> Another would be useful would be to add internal links to the sections of specs that relate to each directory.

Done

src/schema/objects/associated_data.yaml
Remi-Gau: Updated the text in the schema and added a link to the code section.

tools/schemacode/bidsschematools/render.py
```diff
@@ -1133,3 +1133,37 @@ def define_common_principles(schema, src_path=None):
     string = string.replace("SPEC_ROOT", get_relpath(src_path))
 
     return string
+
+
+def define_allowed_top_directories(schema, src_path=None) -> str:
```
Remi-Gau, line 1138: @tsalo not sure it makes much sense to add a test for that

effigies:
@Remi-Gau Is this high priority, as in you want to get it in for the release? Trying to decide when to review this.

Remi-Gau:
not high priority IMHO
@VisLab do you agree?

VisLab:
No it's not a high priority, but I think it should go in at some point.
Thanks...

tools/schemacode/bidsschematools/render.py
tsalo: The problem is that RTD will still build if the rendering function is broken, so any kind of test would be helpful. Even just a smoke test to make sure that a string is returned (without checking the contents) would be useful.

src/schema/objects/associated_data.yaml
```diff
@@ -6,24 +6,26 @@
...
   description: |
-    A directory in which to store any code used to generate the derivatives from the
-    source data.
+    A directory in which to store any code
+    (for example the one used to generate the derivatives from the raw data).
+    See the [relevant section](SPEC_ROOT/03-modality-agnostic-files.md#code)
+    for more information.
 derivatives:
-  display_name: Derivative data
+  display_name: Derivative data.
```
tsalo, line 14:
```suggestion
  display_name: Derivative data
```

tsalo (commented):
It would be nice to ultimately have all of the info about each folder in the schema, so we don't need to link out to specific sections, but I think linking is fine for now. I just have one small suggestion.

Remi-Gau:
I think that anyway @effigies has a pr in the pipeline that will wreck a lot of how this work even if this does not change the final output so I am waiting for this one to be in and adapt from there.

sappelhoff:
> I think that anyway effigies has a pr in the pipeline that will wreck a lot of how this work even if this does not change the final output so I am waiting for this one to be in and adapt from there.

converting to draft then -- please link to that "PR in the pipeline" if you know where it exists :)

Remi-Gau:
> please link to that "PR in the pipeline" if you know where it exists :)

done

Remi-Gau:
> @Remi-Gau Conflicting #1288 has been merged.

thanks for the ping

will update accordingly

Remi-Gau:
Still need to add test but other than that we are back on track.

src/02-common-principles.md
sappelhoff: I'd use MAY instead of "can", other than that, it sounds good.

But it also made me notice that we sometimes use sub-directories and other times subdirectories in the spec. What should we use? my browser marks the word without hyphen as wrong, but Google shows me that both can be used. (very minor stuff here)

sappelhoff (approved):
+1 to merge, pending the test.

Remi-Gau:
> sometimes use sub-directories and other times subdirectories in the spe

Should we do a separate PR for this?

sappelhoff:
> Should we do a separate PR for this?

yes, I can do it, do you agree to always use the hyphen?

src/02-common-principles.md
sappelhoff:
```suggestion
### Other top level directories

In addition of the subject subdirectories, the root directory of a BIDS dataset
can also contain the following directories:

```

src/schema/objects/files.yaml
```diff
@@ -130,27 +130,29 @@ code:
   display_name: Code
   file_type: directory
   description: |
-    A directory in which to store any code used to generate the derivatives from the
-    source data.
+    A directory in which to store any code
+    (for example the one used to generate the derivatives from the raw data).
+    See the [relevant section](SPEC_ROOT/03-modality-agnostic-files.md#code)
+    for more information.
```
effigies, line 136: Two things.

1) Is `objects` the right place to put these user-friendly hints to render? I've generally thought of terms as trying to be clear with minimal reference to the spec.

2) Style-wise I think it's better to use more specific names in URL text when possible:

```suggestion
    See the [Code section](SPEC_ROOT/03-modality-agnostic-files.md#code)
    for more information.
```

src/02-common-principles.md
```diff
@@ -109,6 +109,17 @@ data type as defined above.
 A data type directory SHOULD NOT be defined if there are no files to be placed
 in that directory.
 
+### Other top level directories
+
+In addition of the subject subdirectories, the root directory of a BIDS dataset
+can also contain the following directories:
```
effigies, line 115:
```suggestion
In addition to the subject directories, the root directory of a BIDS dataset
MAY also contain the following directories:
```

tools/schemacode/bidsschematools/render/text.py
```diff
@@ -480,3 +480,37 @@ def define_common_principles(schema, src_path=None):
...
+
+            level = schema["rules"]["files"]["common"]["core"][key]["level"]
+
+            description = schema["objects"]["files"][key]["description"]
+
+            string += f"- `{key}`: [{level.upper()}] {description}"
+
+    string = string.replace("SPEC_ROOT", utils.get_relpath(src_path))
+
+    return string
```
effigies, line 516: 1) Level seems redundant (and a bit distracting in the render). Using `MAY` above should do the trick.
2) A lot of redundant lookups in the schema.

Using the schema as a namespace and dropping level, we can do this more tersely:

```suggestion
    string = ""

    for dirname, definition in schema.objects.files.items():
        if definition.file_type == "directory":
            string += f"- `{dirname}`: {definition.description}"

    return string.replace("SPEC_ROOT", utils.get_relpath(src_path))
```

effigies (commented):
Mostly aesthetic comments.

tools/schemacode/bidsschematools/render.py
Remi-Gau: @tsalo added a very tiny smoke test

src/schema/objects/files.yaml
Remi-Gau: > 1. Is `objects` the right place to put these user-friendly hints to render? I'  generally thought of terms as trying to be clear with minimal reference to the spec.

Yes I agree that this is suboptimal though I am even less of a fan of the alternative which is to have a macro call for each subdirectory.

tools/schemacode/bidsschematools/render/text.py
Remi-Gau: ha thanks I had not yet looked up how to use the namespaces you have implemented in the schema.

concerning "level" I do not feel strongly about it, I think I was prematurely optimizing for some weird cases where the level value would not be optional.

effigies: > I had not yet looked up how to use the namespaces you have implemented in the schema.

The short version is "you can use it as a dict, or an object; if the result is a dict, it will also be a namespace".

Remi-Gau: did you put this as TL;DR in the readme :-p ?

tsalo (approved):
LGTM. Thanks!

VisLab:
@Remi-Gau Thanks for seeing this through.

PR #1291, opened 2022-09-17
SCHEMA: Refactor filename template rendering, normalize requirement level in entity overrides

effigies:
In #1112, I noticed that the overrides of entity formats in MEG used `requirement` instead of the `level` we use elsewhere. They also redundantly include `type: string`, as all entities are string-typed.

When working on the rendering fixes, it seemed we were repeating ourselves a bit and one thing led to another. The fundamental changes:

1) Instead of constructing a `paragraph` string bit by bit, we now build entire lines and create a `lines` list that can be joined once at the end.
2) The functionality in `_format_entity()` function was written several times, so that's abstracted.
3) Constructing the filename line is made simpler by collecting the entity stem, set of suffixes and set of extensions, and iterating at the last moment.

### DIFF
--- a/src/schema/rules/datatypes/meg.yaml
+++ b/src/schema/rules/datatypes/meg.yaml
@@ -38,8 +38,7 @@ calibration:
     subject: required
     session: optional
     acquisition:
-      requirement: required
-      type: string
+      level: required
       enum:
         - calibration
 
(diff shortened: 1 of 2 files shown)


tsalo (approved):
It looks good to me!

PR #1292, opened 2022-09-20
[SCHEMA] Add click to bidsschematools requirements

jwodder:
The `bst` command provided by the bidsschematools package requires click in order to run, yet click is not listed as a dependency in `setup.cfg`.  This PR fixes that.


effigies (approved):
Good catch, thanks!

Issue #1294, opened 2022-09-20
[BUG] test_inheritance_examples is not Windows friendly


yarikoptic:
### Describe your problem in detail.

https://github.com/conda-forge/staged-recipes/pull/20486 suggests feedstock for conda distribution of bidsschematools. It runs tests but one fails on windows only:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=571246&view=logs&j=240f1fee-52bc-5498-a14a-8361bde76ba0&t=2fb5b0a7-737c-5f5c-8f3f-f6db6174bacf&l=2569

```
2022-09-20T16:32:19.5192530Z __________________________ test_inheritance_examples __________________________
2022-09-20T16:32:19.5193712Z 
2022-09-20T16:32:19.5194437Z     def test_inheritance_examples():
2022-09-20T16:32:19.5196130Z         from bidsschematools.validator import validate_bids
2022-09-20T16:32:19.5197371Z     
2022-09-20T16:32:19.5198219Z         correct_inheritance = [
2022-09-20T16:32:19.5198999Z             "/lala/sub-01/ses-test/sub-01_ses-test_task-sometask_bold.json",
2022-09-20T16:32:19.5200696Z             "/lala/sub-01/sub-01_task-sometask_bold.json",
2022-09-20T16:32:19.5201396Z             "/lala/task-sometask_bold.json",
2022-09-20T16:32:19.5201971Z         ]
2022-09-20T16:32:19.5202490Z         incorrect_inheritance = [
2022-09-20T16:32:19.5203140Z             "/lala/sub-01/sub-01_ses-test_task-sometask_bold.json",
2022-09-20T16:32:19.5203810Z             "/lala/ses-test_task-sometask.json",
2022-09-20T16:32:19.5204352Z         ]
2022-09-20T16:32:19.5204911Z     
2022-09-20T16:32:19.5205419Z         result = validate_bids(
2022-09-20T16:32:19.5206227Z             correct_inheritance + incorrect_inheritance,
2022-09-20T16:32:19.5206909Z             accept_dummy_paths=True,
2022-09-20T16:32:19.5207448Z         )
2022-09-20T16:32:19.5207985Z     
2022-09-20T16:32:19.5209379Z >       assert result["path_tracking"] == incorrect_inheritance
[...]

PR #1295, opened 2022-09-20
[INFRA] test bidsschematools  against osx and windows as well

yarikoptic:
Inspired by https://github.com/bids-standard/bids-specification/issues/1294 and thus should fail one test on windows ATM.

attn @TheChymera . I am ok if fix for #1294 would simply absorb this PR

### DIFF
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -51,8 +51,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        os: ["ubuntu-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        include:
+          - os: macos-latest
+            python-version: 3
+          - os: windows-latest
+            python-version: 3
     name: ${{ matrix.os }} with Python ${{ matrix.python-version }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/schemacode_ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["ubuntu-latest"]
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
```
effigies, line 55: Given that it's all filesystem manipulation and not weird numerical edge cases, it seems like overkill to test the whole Python range on all three platforms. I would suggest:

```suggestion
        os: ["ubuntu-latest"]
        python-version: ["3.7", "3.8", "3.9", "3.10"]
        include:
          - os: macos-latest
            python-version: 3
          - os: windows-latest
            python-version: 3
```

TheChymera (approved):
As per [the other discussion regarding windows](https://github.com/bids-standard/bids-specification/issues/1294) I really think we should consider deprecating this OS entirely and just pointing people to [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux). Windows causes too much trouble with trivial details such as permissions and paths.

*However* given that we seem to have downstream demand, as you mentioned in that discussion, we might be forced to do this for the moment :( What I'd suggest is to keep pushing WSL and revisit this again regularly to avoid having this dangling test constraint in perpetuity. It's a bit strange that we try to push for best data practices, try to get people to use our specific custom standards and software, and don't put an equal or greater amount of effort into getting them to use the far more established standards and software of GNU/linux....

yarikoptic:
ok -- this one is good, but I think we should not merge it since fails on Windows. @TheChymera - please base your fix on top of this PR branch, meanwhile I will just close it (branch should still lurk there)

PR #1297, opened 2022-09-21
[SCHEMA] Test schema code on Windows/OSX, fix small POSIX assumption

TheChymera:
Supersedes: https://github.com/bids-standard/bids-specification/pull/1295#event-7426140754

### DIFF
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -53,6 +53,11 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
+        include:
+          - os: macos-latest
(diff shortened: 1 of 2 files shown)


TheChymera:
@effigies could you also add a patch increment release tag for BST after this is merged? Or can I also do that?

effigies (approved):
If I'm understanding correctly, `accept_dummy_paths` is just used for testing? This seems good, since tests on example datasets are passing.

TheChymera:
@effigies it was originally introduced for testing, but it finds some niche usage in DANDI as well for parsing entities in paths returned from objects without needing to download the files or even the symlinks (many datasets have tons of files, so even just creating placeholder files would take a while). I'm thinking it could have broader use as well in that paradigm.

PR #1299, opened 2022-09-22
[FIX][SCHEMA] Add conditionals for PET ReconMethod* and ReconFilter

bendhouseart:
Adds conditionals to change required PET fields to optional when "parent" fields are unknown or "none", specifically:

* add conditionals for ReconMethod Metadata

* add min and max values to scatter fraction

* add conditional for ReconFilterSize

Intention is to avoid users adding unnecessary place holders to pass validation, e.g. [0, 0], ["none", "none"], hopefully avoids ambiguity instead of increasing it.

Arguments could be made that fields should always be required, but several recent use cases have prompted this minor change to the spec. Additionally, it's more straight forward (to this user) to toggle optional than to define `null` placeholders for values and datatypes.


src/schema/objects/metadata.yaml
```diff
@@ -2660,6 +2660,7 @@ ScatterFraction:
     type: number
     minimum: 0
     maximum: 100
+
```
tsalo, line 2663: Is this new line accidental?

src/schema/rules/sidecars/pet.yaml
```diff
@@ -176,6 +180,25 @@ PETReconstruction:
     SinglesRate: recommended
     RandomRate: recommended
 
+# set required reconstruction fields to optional when "none" is provided as a parameter label
+EntitiesReconMethodMetadata:
+  selectors:
+    - modality == "pet"
+    - suffix == "pet"
+    - count(sidecar.ReconMethodParameterLabels, "none") >= 1
```
tsalo, line 188:
```suggestion
    - "none" in sidecar.ReconMethodParameterLabels
```

I _think_ this is the preferred way of saying "X contains 'string'"

src/schema/rules/sidecars/pet.yaml
```diff
@@ -150,12 +150,16 @@ PETReconstruction:
       description_addendum: This partly matches the DICOM Tag (0054,1103) `Reconstruction Method`.
     ReconMethodParameterValues:
       level: required
+      level_addendum: |
+        Required if `ReconMethodParameterLabels` does not contain 'none', optional if
+        no labels exist.
```
tsalo, line 155: I think this needs to be reflected in the selectors- i.e., we need two mutually exclusive sets of selectors. One with `"none" in sidecar.ReconMethodParameterLabels` and one with `!("none" in sidecar.ReconMethodParameterLabels)`.

tsalo: Also, shouldn't the level_addendum be applied to `ReconMethodParameterUnits` as well?

src/schema/rules/sidecars/pet.yaml
```diff
@@ -150,12 +150,16 @@ PETReconstruction:
...
       description_addendum: This partly matches the DICOM Tag (0054,1103) `Reconstruction Method`.
     ReconFilterType:
       level: required
       description_addendum: This partly matches the DICOM Tag (0018,1210) `Convolution Kernel`.
     ReconFilterSize:
       level: required
+      level_addendum: Required if `ReconFilterType` is not 'none'.
       description_addendum: This partly matches the DICOM Tag (0018,1210) `Convolution Kernel`.
```
tsalo, line 163: I think this needs to be split out into a separate rule as well.

tsalo (dismissed):
I think the rules need to be reconfigured a bit.

src/schema/objects/metadata.yaml
bendhouseart, line 2663:
```suggestion
```

bendhouseart: Yes.

src/schema/rules/sidecars/pet.yaml
bendhouseart: I had originally done it that way, but changed it after reading over Chris' updated schema readme. It wasn't clear to me whether or not a an element in an array was a subfield or not. I felt that using count made it much more clear that "none" was an element in an array. Perhaps I just need to update the documentation clarify that the `in` operator works this way on arrays.

src/schema/rules/sidecars/pet.yaml
bendhouseart: I'm sorry, I don't understand what you mean by that. Is that not what is defined below in `EntitiesReconFilterData`?

src/schema/rules/sidecars/pet.yaml
bendhouseart: Fits with the logic, I've never seen them have units so I don't pay them much mind.

bendhouseart: >I think this needs to be reflected in the selectors- i.e., we need two mutually exclusive sets of selectors. One with "none" in sidecar.ReconMethodParameterLabels and one with !("none" in sidecar.ReconMethodParameterLabels).

Are you suggesting the below?

```
RuleOne:
  selectors: 
      - "none" in sidecar.ReconMethodParameterLabels
  fields:
    ReconMethodParameterUnits: optional
    ReconMethodParameterValues: optional

RuleTwo:
   selectors:
       - !("none" in sidecar.ReconMethodParameterLabels)
   fields:
     ReconMethodParameterUnits: required
     ReconMethodParameterValues: required
```

Do we really need mutual exclusion for this? It seems excessive when one rule covers both cases, or is it a technical limitation with the validator or schema tools?

effigies:
@bendhouseart While working through some examples, we've found that it's going to be very difficult to both render and enforce rules that have `required` fields and then relax under certain conditions. So we can do two things:

1) Make the optional default and use `level_addendum` to describe when it's required, and then add a required rule with a selector. Here only the first table will get rendered, and it will say something like `OPTIONAL, but REQUIRED if XYZ`.

```YAML
PETReconstruction:
  selectors:
    - modality == "pet"
    - suffix == "pet"
  fields:
    ReconMethodParameterLabels:
      level: required
    ReconMethodParameterUnits:
      level: optional
      level_addendum: |
        required if `ReconMethodParameterLabels` does not contain 'none'

PETReconstructionReq:
  selectors:
    - modality == "pet"
    - suffix == "pet"
    - '!intersects(sidecar.ReconMethodParameterLabels, ["none"])'
  fields:
    ReconMethodParameterUnits: required
```

2) Split the rule into two separate tables, one `required` and one `optional`, and then combine during the render:

```YAML
PETReconstruction:
  selectors:
    - modality == "pet"
    - suffix == "pet"
  fields:
    ReconMethodParameterLabels:
      level: required

PETReconstructionReq:
  selectors:
    - modality == "pet"
    - suffix == "pet"
    - '!intersects(sidecar.ReconMethodParameterLabels, ["none"])'
  fields:
    ReconMethodParameterUnits:
      level: required
      level_addendum: |
        optional if `ReconMethodParameterLabels` does not contain 'none'

PETReconstructionOpt:
  selectors:
    - modality == "pet"
    - suffix == "pet"
    - intersects(sidecar.ReconMethodParameterLabels, ["none"])
    ReconMethodParameterUnits: optional
```

```Markdown
{{ MACROS___make_sidecar_table(["pet.PETReconstruction". "pet.PETReconstructionReq"]) }}
```

CPernet:
@effigies isn't `OPTIONAL but REQUIRED if ` similar to what is already used for ['NonlinearGradientCorrection'](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#sequence-specifics) which then would make sense to use a  rule for conditional stuff -- although that case this is `RECOMMENDED, but REQUIRED if ` (which would also actually be better for _ReconMethod Parameters_ and _ReconFilter Parameters_

effigies:
Yes, these are equivalent approaches in terms of enforcement. The choice between the two is which way PET maintainers would prefer it to be rendered.

Note that if you make a field RECOMMENDED, there will be warnings when it is missing. (The current validator does not follow this rule, but the schematized one will.)

melanieganz:
@effigies, so we just discussed this again in the OpenNeuroDev meeting and we agreed upon going for  `RECOMMENDED, but REQUIRED if ` . We'd like it to spit out a warning during validation , since we want to encourage people to go look for the values, but we accept that they might be hard to find/dig out.

src/schema/rules/sidecars/pet.yaml
```diff
@@ -146,16 +146,20 @@ PETReconstruction:
       level: required
       description_addendum: This partly matches the DICOM Tag (0054,1103) `Reconstruction Method`.
     ReconMethodParameterUnits:
-      level: required
+      level: recommended
+      level_addendum: Recommended, but required if `ReconMethodParameterLabels` does not contain 'none'.
```
effigies, line 150:
```suggestion
      level_addendum: required if `ReconMethodParameterLabels` does not contain `"none"`
```

src/schema/rules/sidecars/pet.yaml
```diff
@@ -146,16 +146,20 @@ PETReconstruction:
...
       description_addendum: This partly matches the DICOM Tag (0054,1103) `Reconstruction Method`.
     ReconMethodParameterValues:
-      level: required
+      level: recommended
+      level_addendum: |
+        Recommended, but required if `ReconMethodParameterLabels` does not contain 'none'.
```
effigies, line 155:
```suggestion
      level_addendum: required if `ReconMethodParameterLabels` does not contain `"none"`
```

src/schema/rules/sidecars/pet.yaml
```diff
@@ -146,16 +146,20 @@ PETReconstruction:
...
       description_addendum: This partly matches the DICOM Tag (0054,1103) `Reconstruction Method`.
     ReconFilterType:
       level: required
       description_addendum: This partly matches the DICOM Tag (0018,1210) `Convolution Kernel`.
     ReconFilterSize:
-      level: required
+      level: recommended
+      level_addendum: Recommended, but required if `ReconFilterType` is not "none".
```
effigies, line 162:
```suggestion
      level_addendum: required if `ReconFilterType` is not `"none"`
```

effigies (approved):
Looks right. Some styling fixes for the table.

effigies:
Final render: image

effigies:
@melanieganz @CPernet @mnoergaard Any of you up for giving this another review and a green check if you're happy with it?

sappelhoff:
tagging again: @melanieganz @CPernet @mnoergaard

sappelhoff:
Thanks all!

PR #1300, opened 2022-09-23
[SCHEMA] Invert logic of bold onset hints

rwblair:
We only want to generate these messages when the check returns false.


effigies (approved):
Good catch

tsalo (approved):

PR #1301, opened 2022-09-25
[SCHEMA] Refactor namespace

effigies:
While working on #1288 a few changes to namespace became necessary/convenient. However, since these are bug-fixes and enhancements, not backwards-incompatible changes, I wanted to break these changes out into their own PR, to keep that one focused.

The changes are:

1) Making it a mutable mapping makes chains of references more tractable. e.g. if objB refers to objA and objC refers to objB, updating the reference schema during dereferencing becomes necessary to avoid multiple passes.
2) Adding a `level` keyword to items/keys/values allows you to do something like `schema.objects.keys(level=2)` to get things like `metadata.RepetitionTime` and `columns.participant_id`.

### DIFF
--- a/tools/schemacode/bidsschematools/validator.py
+++ b/tools/schemacode/bidsschematools/validator.py
@@ -845,10 +845,13 @@ def validate_bids(
 
     Examples
     --------
-    >>> from bidsschematools import validator
-    >>> bids_paths = '~/.data2/datalad/000026/rawdata'
-    >>> schema_version='{module_path}/data/schema/'
-    >>> validator.validate_bids(bids_paths, schema_version=schema_version)"
+
+    ::
+
+        from bidsschematools import validator
+        bids_paths = '~/.data2/datalad/000026/rawdata'
+        schema_version='{module_path}/data/schema/'
+        validator.validate_bids(bids_paths, schema_version=schema_version)
 
     Notes
     -----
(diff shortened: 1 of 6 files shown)


tools/schemacode/bidsschematools/validator.py
```diff
@@ -845,10 +845,13 @@ def validate_bids(
 
     Examples
     --------
-    >>> from bidsschematools import validator
-    >>> bids_paths = '~/.data2/datalad/000026/rawdata'
-    >>> schema_version='{module_path}/data/schema/'
-    >>> validator.validate_bids(bids_paths, schema_version=schema_version)"
+
+    ::
```
bendhouseart, line 849: I'm not sure if I prefer this or the more verbose `>>>`, but still very cool.

effigies: The `>>>` means the code gets executed in doctests, but this test fails on anybody but @TheChymera's computer.

bendhouseart (approved):
Too clever for me, not too much to add other than I'm glad to see this get broken up into types and abstracted a little bit away from `bidsschematools/schema.py`

tools/schemacode/bidsschematools/validator.py
bendhouseart: Ah, well we aren't all as blessed to own a computer like @TheChymera thanks for adding this.

effigies:
Thanks for the review!

PR #1302, opened 2022-09-25
CI: Consolidate validation jobs into single workflow

effigies:
We have a lot of separate GitHub actions workflows to check that contributions conform to style guidelines. They make sense to me to consolidate.

Also added a concurrency spec, so that jobs will self-cancel if a new commit is pushed to the branch, and rerendered with prettier.


sappelhoff (approved):

sappelhoff (commented):
Ah, I think the badges in our README.md need to be updated accordingly: https://github.com/bids-standard/bids-specification/blob/959b64630d0952924adccade0a5fe58f49dd00be/README.md?plain=1#L1-L4

I am +1 for adding badges for all our workflows, but feel free to do your thing, as long as we don't have non-working badges.

effigies:
@sappelhoff I'd be comfortable merging this without 2 reviews, but as it's mine, I'll leave that decision to you.

PR #1305, opened 2022-09-27
Blanket ignoring dotdirs

TheChymera:
Closes: https://github.com/dandi/dandi-cli/issues/1114


effigies:
Generally fine with this. Since the spec doesn't say anything about dot files, I'd say it's implementation dependent what to do about them.

tools/schemacode/bidsschematools/validator.py
```diff
@@ -42,17 +42,11 @@ def _get_paths(
...
     """
-    exclude_subdirs = [
-        rf"{os.sep}.dandi",
-        rf"{os.sep}.datalad",
-        rf"{os.sep}.git",
-    ]
     # `.bidsignore` is not, in fact, a BIDS file, as per:
     # https://github.com/bids-standard/bids-specification/issues/980
+    # Perhaps this should be parameterized for downstream flexibility and not having to keep track
+    # of downstream nuisance files here.
```
effigies, line 48: I think this is probably the right move in the long run. I think we might want to make the interface `validate(..., ignore=ignore_func)` where the first step when looking at a file is `if ignore_func(rel_path): continue`.

But that doesn't need to be done here.

effigies (approved):

PR #1306, opened 2022-09-30
SCHEMA: Add index_columns metadata, render in tables

### LINKED ISSUE #1304: Primary keys / index columns in tabular metadata files
#1290 highlighted a rule that got added with microscopy that we have not considered in the schema. In [Samples file](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#samples-file):

> The combination of `sample_id` and `participant_id` MUST be unique.

And of course in participants.tsv "There MUST be exactly one row for each participant."

In both of these cases, each row is metadata associated with an object identified by one or more of the columns, making those columns the [primary key](https://en.wikipedia.org/wiki/Primary_key) in relational database terms. The values (or tuple of values, if multiple columns) must be unique because they can be used to look up the row. While we could write rules and come up with expressions that would enforce these constraints, it seems less ad hoc to make this a declarative property of a table.

Here's a proposal:

```YAML
Participants:
  selectors:
    - path == "participants.tsv"
  initial_columns:
    - participant_id
  columns:
    participant_id: require
    species: recommended
    age: recommended
    sex: recommended
    handedness: recommended
    strain: recommended
    strain_rrid: recommended
  index_columns: [participant_id]
  additional_columns: allowed

samples:
  selectors:
    - path == "samples.tsv"
  columns:
    sample_id: required
    participant_id: required
    sample_type: required
    pathology: recommended
    derived_from: recommended
  index_columns: [sample_id, participant_id]
  additional_columns: allowed
```

Tooling would then enforce (and perhaps take advantage of) this constraint.

I use `index_columns` because it feels a little less jargony than primary key, but we could also just use `primary_key` and document its meaning.

Note that there are some TSV files (such as events.tsv) where there is no primary key, so we can't get away with an implicit "first column is primary unless otherwise stated". Or we can, but then we would need to declare some file types as not having them.

effigies:
Closes #1304.

### DIFF
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -742,6 +742,7 @@ rule is applied.
     | `selectors`          | List of expressions; any evaluating false indicate rule does not apply                                         |
     | `columns`            | Object with keys that may be found in `objects.columns`, values either a requirement level or an object        |
     | `initial_columns`    | An optional list of columns that must be the first N columns of a file                                         |
+    | `index_columns`      | A list of columns that uniquely identify a row.                                                                |
     | `additional_columns` | Indicates whether additional columns may be defined. One of `allowed`, `allowed_if_defined` and `not_allowed`. |
 
 The following tables demonstrate how mutual exclusive, required fields, may be
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -742,7 +742,7 @@ rule is applied.
     | `selectors`          | List of expressions; any evaluating false indicate rule does not apply                                         |
     | `columns`            | Object with keys that may be found in `objects.columns`, values either a requirement level or an object        |
     | `initial_columns`    | An optional list of columns that must be the first N columns of a file                                         |
-    | `index_columns`      | A list of columns that uniquely identify a row.                                                                |
+    | `index_columns`      | An optional list of columns that uniquely identify a row.                                                      |
     | `additional_columns` | Indicates whether additional columns may be defined. One of `allowed`, `allowed_if_defined` and `not_allowed`. |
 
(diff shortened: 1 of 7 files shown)


src/schema/README.md
```diff
@@ -742,6 +742,7 @@ rule is applied.
     | `selectors`          | List of expressions; any evaluating false indicate rule does not apply                                         |
     | `columns`            | Object with keys that may be found in `objects.columns`, values either a requirement level or an object        |
     | `initial_columns`    | An optional list of columns that must be the first N columns of a file                                         |
+    | `index_columns`      | A list of columns that uniquely identify a row.                                                                |
```
bendhouseart, line 745: @effigies is this optional or required?  Lines [182](https://github.com/effigies/bids-specification/blob/769c62e5ca63c5d3602ffce3d8b5fd10ec5a763b/tools/schemacode/bidsschematools/render/tables.py#L182) and [218](https://github.com/effigies/bids-specification/blob/769c62e5ca63c5d3602ffce3d8b5fd10ec5a763b/tools/schemacode/bidsschematools/render/tables.py#L218) seem to imply it, if so 
```suggestion
    | `index_columns`      | An optional list of columns that uniquely identify a row.                                                                |
```

bendhouseart (approved):
Sure.

sappelhoff (approved):

tsalo (approved):
I'm disappointed by the lack of an Oxford comma, but the changes look good to me otherwise. Thanks!

effigies:
> I'm disappointed by the lack of an Oxford comma, but the changes look good to me otherwise. Thanks!

It's not clear we'll ever have three index columns, anyway.

PR #1315, opened 2022-10-10
SCHEMA: Use lazy loading

effigies:
This ~~uses~~vendors the [lazy_loader](https://github.com/scientific-python/lazy_loader) package to do two things:

1) Import our submodules so they are accessible in the root namespace without being individually imported. For example:

```Python
import bidsschematools as bst
schema = bst.schema.load_schema()
```

2) Lazy-load non-stdlib modules, so that they only get imported when used. This makes it a bit easier to do imports and selective tests without setting up an environment where we install `bidsschematools[all]`.

The new dependency is extremely lightweight and is very likely to become ubiquitous in the scientific Python ecosystem.

Discussed this with @TheChymera back in August, but wanted to wait until `lazy_loader` had a non-alpha release.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,3 @@
 ignore:
-    - "*/*/tests/*"
     - "**/tests/*"
+    - "*/_lazy.py"
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,3 @@
 ignore:
     - "**/tests/*"
-    - "*/_lazy.py"
+    - "**/_lazy.py"
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: "Install build dependencies"
         run: pip install --upgrade build twine
       - name: "Install test dependencies on tag"
-        run: pip install --upgrade pytest pyyaml pandas tabulate markdown-it-py
+        run: pip install --upgrade pytest
         if: ${{ startsWith(github.ref, 'refs/tags/schema-') }}
       - name: "Build archive on tag"
         run: pytest tools/schemacode/bidsschematools -k make_archive
--- a/tools/schemacode/bidsschematools/render/text.py
+++ b/tools/schemacode/bidsschematools/render/text.py
@@ -2,8 +2,10 @@
 import logging
 import os
 
-import yaml
-from markdown_it import MarkdownIt
+import lazy_loader as lazy
+
+yaml = lazy.load("yaml")
+markdown_it = lazy.load("markdown_it")
 
 from bidsschematools.render import utils
 from bidsschematools.schema import Namespace, filter_schema, load_schema
@@ -438,7 +440,7 @@ def append_filename_template_legend(text, pdf_format=False):
 
 """
     else:
-        md = MarkdownIt()
+        md = markdown_it.MarkdownIt()
         text += f"""
 <details>
 <summary><strong>Legend:</strong></summary>
--- a/tools/schemacode/bidsschematools/render/text.py
+++ b/tools/schemacode/bidsschematools/render/text.py
@@ -2,10 +2,10 @@
 import logging
 import os
 
-import lazy_loader as lazy
+from .. import _lazy
 
-yaml = lazy.load("yaml")
-markdown_it = lazy.load("markdown_it")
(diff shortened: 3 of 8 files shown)


tools/schemacode/bidsschematools/render/text.py
```diff
@@ -438,7 +440,7 @@ def append_filename_template_legend(text, pdf_format=False):
 
 """
     else:
-        md = MarkdownIt()
+        md = mdit.MarkdownIt()
```
tsalo, line 443: Should this be `markdown_it`?

.github/workflows/schemacode_ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: "Install build dependencies"
         run: pip install --upgrade build twine
       - name: "Install test dependencies on tag"
-        run: pip install --upgrade pytest pyyaml pandas tabulate markdown-it-py
+        run: pip install --upgrade pytest lazy_loader
```
tsalo, line 36: Why install it here if it's a core dependency in `setup.cfg`?

tsalo (commented):
This seems like a good thing to have. I just have a couple of questions.

TheChymera:
@effigies well, the reason I hadn't followed up is because I concluded that we might be digging ourselves deeper in the dependency proliferation hole.

> is very likely to become ubiquitous 

I don't think this is currently the case, I had looked it up back in August, and it doesn't seem to have changed: https://repology.org/project/python:lazy-loader/versions
In particular it's missing on Debian and all the downstream distros from it... As far as Gentoo goes, I can package it quite easily by the looks of it, but I feel it's always a bad sign for software that I'm working on (other than esoteric data analysis stuff) to need yet another package...
As always, if you think it's a good idea, don't let me get in your way, but maybe then we should also ping @yarikoptic re packaging it for debian?

effigies:
We could vendor it as `bidsschematools/_lazy_loader.py` for now. It is already being used by scikit-image and networkx.

.github/workflows/schemacode_ci.yml
effigies: Because this doesn't install, it just builds. There's a weird build step where we use pytest to bundle the state of the BIDS examples into the repository so it gets pulled in by `python -m build`.

tools/schemacode/bidsschematools/render/text.py
yarikoptic: it was defined above as `mdit = lazy.load("markdown_it")` so I guess it is technically correct here/now, BUT may be to avoid such confusions it should be
- `import lazy_loader` (to not need to think either it is `lazy_loader` vs `lazy`)
- `markdown_it = lazy.load("markdown_it")` (to similarly minimize cognitive load here)
?

tsalo: Okay, I was definitely not caffeinated enough when I reviewed this PR. Thank you for explaining!

.github/workflows/schemacode_ci.yml
tsalo: I'm not sure I understand the build step, but thanks for clarifying.

yarikoptic:
confirming that it is indeed not packaged for debian and nobody even issued an ITP (intend to package) for it. not sure I want to take the burden on my shoulders, so I issued RFP (request for packaging) "just in case": https://bugs.debian.org/1021619

tools/schemacode/bidsschematools/render/text.py
yarikoptic: 1c: oh well, I think we know that conventions are not *gods given* and could change. I understand somewhat wide spread, frequently used and "concise" import conventions  such as `np`, or `sp` but abbreviating not so long into not so short and not so frequently used (once or twice in a file) -- mneh, no point. anyways -- want to keep `lazy` - fine with me.  `markdown_it` -> `mdit` for few uses imho makes little sense to me personally.

effigies: I did change it to `markdown_it`...

.codecov.yml
```diff
@@ -1,3 +1,3 @@
 ignore:
-    - "*/*/tests/*"
     - "**/tests/*"
+    - "*/_lazy.py"
```
effigies, line 3:
```suggestion
    - "**/_lazy.py"
```

PR #1328, opened 2022-10-17
[ENH] Generalize rules that labels must be used consistently

### LINKED ISSUE #1314: Clarification of task entity in the appendix
### Clarification of task entity in appendix

In the current description of [`task-<label>`](../appendices/entities.md#task) it states wrt definition of task: 

task
Full name: Task
Format: task-<label>
Definition: Each task has a unique label that MUST only consist of letters and/or numbers (other characters, including spaces and underscores, are not allowed). Those labels MUST be consistent across subjects and sessions.

@satra commented in [PR 1185](https://github.com/bids-standard/bids-specification/pull/1185) on the last sentence here not being clear, see [here](https://github.com/bids-standard/bids-specification/pull/1185#discussion_r987993984). Specifically, the question was what does this mean if half the subjects do the task and half don't? @Remi-Gau suggested to update this in general, but since it has broader implications, I made this separate issue for it.

Now my understanding is similar to @Remi-Gau's,  meaning label task-foo for subject 1 must mean the same thing as task-foo for subject 2. There is no restrictions of the type "all participants must have done all the tasks." 

So I propose to extend the last sentence to say:
"Those labels MUST be consistent across subjects and sessions, meaning label task-foo for subject X must mean the same thing as task-foo for subject Y."

Opinions or objections?

Remi-Gau:
closes #1314

### DIFF
--- a/src/schema/objects/common_principles.yaml
+++ b/src/schema/objects/common_principles.yaml
@@ -93,6 +93,9 @@ label:
     for example, it is `rest` in `task-rest` following `task-<label>` specification.
     Note that labels MUST not collide when casing is ignored
     (see [Case collision intolerance](SPEC_ROOT/02-common-principles.md#case-collision-intolerance)).
+    For a given data type, labels MUST be consistent across subjects and sessions.
+    For example label `acq-hires` for the `anat` data of a given subject
+    must mean the same thing as the label `acq-hires` used for the `anat` of any other subject.
 modality:
   name: Modality
   display_name: Modality
--- a/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
+++ b/src/04-modality-specific-files/01-magnetic-resonance-imaging-data.md
@@ -590,7 +590,7 @@ MUST be used to distinguish both images.
 In such a case, two files could have the following names:
 `sub-01_acq-singleband_dwi.nii.gz` and `sub-01_acq-multiband_dwi.nii.gz`.
 The user is free to choose any other label than `singleband` and
-`multiband`, as long as they are consistent across subjects and sessions.
+`multiband`.
 
 ### REQUIRED gradient orientation information
 
--- a/src/04-modality-specific-files/07-behavioral-experiments.md
+++ b/src/04-modality-specific-files/07-behavioral-experiments.md
@@ -24,7 +24,6 @@ rather than `_events.tsv`.
 Each task has a unique label that MUST only consist of letters and/or numbers
 (other characters, including spaces and underscores, are not allowed) with the
 [`task-<label>`](../appendices/entities.md#task) entity.
-Those labels MUST be consistent across subjects and sessions.
 
 The OPTIONAL [`acq-<label>`](../appendices/entities.md#acq) entity corresponds to a custom label to
 distinguish different conditions present during multiple runs of the same task.
--- a/src/04-modality-specific-files/09-positron-emission-tomography.md
+++ b/src/04-modality-specific-files/09-positron-emission-tomography.md
@@ -80,7 +80,6 @@ Volumes MUST be stored in chronological order (the order they were acquired in).
 
 The OPTIONAL [`task-<label>`](../appendices/entities.md#task) is used to
 indicate a task subjects were asked to perform in the scanner.
-Those labels MUST be consistent across subjects and sessions.
 For task based PET, a corresponding [`task events`](./05-task-events.md) file MUST be provided
 (please note that this file is not necessary for resting scans).
 
@@ -91,8 +90,6 @@ but REQUIRED to distinguish between tracers if multiple are used.
 The label used is arbitrary and each file requires a separate JSON sidecar
 with details of the tracer used (see below).
 Examples are `trc-18FFDG` for fludeoxyglucose or `trc-11CPIB` for Pittsburgh compound B.
-Other labels are permitted, as long as they are consistent across subjects and sessions
-and consist only of the legal label characters.
 
 If more than one run of the same task and acquisition (tracer) are acquired during
 the same session, the [`run-<index>`](../appendices/entities.md#run) entity MUST be used:
--- a/src/04-modality-specific-files/10-microscopy.md
+++ b/src/04-modality-specific-files/10-microscopy.md
@@ -133,8 +133,7 @@ MAY be used to distinguish a different set of parameters used for acquiring the 
 For example, two images of the same sample acquired by bright-field microscopy (BF) in PNG format at
 different magnification of 40x and 60x.
 In such case two files could have the following names: `sub-01_sample-01_acq-40x_BF.png` and
-`sub-01_sample-01_acq-60x_BF.png`, however the user is free to choose any other label as long as
-they are consistent across subjects and sessions.
+`sub-01_sample-01_acq-60x_BF.png`.
 
 The [`stain-<label>`](../appendices/entities.md#stain) entity MAY be used to distinguish
 image files from the same sample using different stains or antibodies for contrast enhancement.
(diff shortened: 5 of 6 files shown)


src/schema/objects/common_principles.yaml
```diff
@@ -93,6 +93,9 @@ label:
     for example, it is `rest` in `task-rest` following `task-<label>` specification.
     Note that labels MUST not collide when casing is ignored
     (see [Case collision intolerance](SPEC_ROOT/02-common-principles.md#case-collision-intolerance)).
+    For a given data type, labels MUST be consistent across subjects and sessions.
+    For example label `acq-hires` for the `anat` data of a given subject
+    must mean the same thing as the label `acq-hires` used for the `anat` of any other subject.
```
Remi-Gau, line 98: This is the main bit.

I am restricting "consistency" to within a datatype (anat, func, eeg...) because:
- most previous mentions of consistency appeared within the page of a given modality
- I could imagine that `acq-highRes` could easily mean different things in MRI and EEG for example.

Remi-Gau:
Something I am not sure about:
- no idea how to validate for this
- this rule may be too restrictive

For example if we have 2 anat files

- sub-01_acq-highRes_T1w.nii
- sub-01_acq-highRes_T2w.nii

It may be the case that `highRes` refers to different things for those different types of suffixes.

src/schema/objects/common_principles.yaml
```diff
@@ -93,6 +93,9 @@ label:
     for example, it is `rest` in `task-rest` following `task-<label>` specification.
     Note that labels MUST not collide when casing is ignored
     (see [Case collision intolerance](SPEC_ROOT/02-common-principles.md#case-collision-intolerance)).
+    For a given data type, labels MUST be consistent across subjects and sessions.
+    For example label `acq-hires` for the `anat` data of a given subject
```
tsalo, line 97:
```suggestion
    For example, an `acq` entity label `hires` used on the `anat` data of a given subject
```

src/schema/objects/common_principles.yaml
tsalo, line 98:
```suggestion
    MUST mean the same thing as `acq-hires` in `anat` files of any other subject.
```

sappelhoff (commented):
I think we would make our lives easier to require that labels are unique throughout the dataset, not only in individual modalities 🤔 it seems like a huge source of confusion for little benefit otherwise

src/schema/objects/common_principles.yaml
tsalo: What is the level of consistency required? It seems like it's up to the curator. E.g., `acq-hires` could be 0.5x0.5x0.5 in one subject, and 0.6x0.6x0.6 in another, as long as there's some other anatomical scan in each that is lower-resolution, right? Or should they be differentiated in that case (e.g., `acq-hires` and `acq-`reallyhires`)?

Remi-Gau:
Probably a PR to revisit early 2024

sappelhoff (approved):
I stand by my point in https://github.com/bids-standard/bids-specification/pull/1328#pullrequestreview-1152271940 (require labels to be consistent throughout the dataset, rather than only modalities) ... BUT I think the present changes are more in the spirit of the current spec, and thus I deem this PR mergeable as a nice "refactoring".

(Note: not sure if the `[ENH]` is fitting, therefore ... maybe `[MISC]`?)

src/schema/objects/common_principles.yaml
```diff
@@ -87,6 +87,9 @@ label:
     for example, it is `rest` in `task-rest` following `task-<label>` specification.
     Note that labels MUST not collide when casing is ignored
     (see [Case collision intolerance](SPEC_ROOT/common-principles.md#case-collision-intolerance)).
+    For a given data type, labels MUST be consistent across subjects and sessions.
+    For example, an `acq` entity label `hires` used on the `anat` data of a given subject
+    MUST mean the same thing as `acq-hires` in `anat` files of any other subject.
```
Remi-Gau, line 92:
```suggestion
    Labels MUST be consistent across subjects and sessions and data types.
    For example, an `acq` entity label `hires` used on the `anat` data of a given subject
    MUST mean the same thing as `acq-hires` in `func` files of any other subject.
```

Remi-Gau:
> I stand by my point in [#1328 (review)](https://github.com/bids-standard/bids-specification/pull/1328#pullrequestreview-1152271940) (require labels to be consistent throughout the dataset, rather than only modalities) ... BUT I think the present changes are more in the spirit of the current spec, and thus I deem this PR mergeable as a nice "refactoring".

Updated to have consistency across datatypes too.
 
> (Note: not sure if the `[ENH]` is fitting, therefore ... maybe `[MISC]`?)

Done

tsalo (approved):
The changes look good to me. One thing I do wonder, though, is if we should make it clear that "consistency" is at the discretion of the dataset curator. The distinction between `lores` and `hires` might be different for different curators.

For example, one might have T1ws at 1 mm3 and 0.8 mm3, and T2ws at 1 mm3 and 0.6 mm3, and I think it's up to the curator to decide if they was `acq-lores` (1 mm3), `acq-hires` (0.8 mm3), and `acq-veryhires` (0.6 mm3) or just `acq-lores` (1 mm3) and `acq-hires` (0.8 mm3 and 0.6 mm3).

effigies:
I overall approve of the centralization and the spirit of this, but this can lead to a problem:

1) Validation of this would be computationally painful, even if it were well defined. However, without metadata tags tied to each entity, how would we validate this?
2) If a tool reads this "MUST", assumes it means something is validated and can be depended on, and decides to optimize by not rereading metadata it thinks it knows, you could end up with invalid processing.

effigies:
> Maybe less constraining: turning this into an admonition to tell users they should be careful about consistency (so no SHOULD or MAY but try to bring attention to this in general).

Rereading this after a couple months (looking through the 1.10.0 milestone), I think I agree with this.

src/modality-specific-files/microscopy.md
```diff
@@ -102,7 +102,8 @@ Consistent with other data types in BIDS, the session entity is optional.
 
 The [`sample-<label>`](../appendices/entities.md#sample) entity is REQUIRED for
 Microscopy data and is used to distinguish between different samples from the same subject.
-The label MUST be unique per subject and is RECOMMENDED to be unique throughout the dataset.
+Contrary to other labels, the `sample-<label>` MUST be unique per subject
+and is RECOMMENDED to be unique throughout the dataset.
```
yarikoptic, line 106: I do not see why I am to be disallowed to use `sample-1`, `sample-2` and so on across my subjects.
It seems that the intent here to demand some UUIDs,  but I do not see really why here and why only for `sample`s.

Moreover, as we ATM have mandatory `sub-` leading folder anyways, I do not understand really what "unique per subject" means really.

src/schema/objects/common_principles.yaml
```diff
@@ -87,6 +87,9 @@ label:
     for example, it is `rest` in `task-rest` following `task-<label>` specification.
     Note that labels MUST not collide when casing is ignored
     (see [Case collision intolerance](SPEC_ROOT/common-principles.md#case-collision-intolerance)).
+    Labels MUST be consistent across subjects and sessions and data types.
+    For example, an `acq` entity label `hires` used on the `anat` data of a given subject
+    MUST mean the same thing as `acq-hires` in `func` files of any other subject.
```
yarikoptic, line 92: MUST is IMHO too strong of a word here since we do not even define the meaning of "mean" in this case -- is it  EXACTLY the same resolution? the same shimming? clearly it cannot be "the same everything" since then functional must be an anatomical ;)  Other example could be `_desc` for which we provision `descriptions.tsv` per each subject/session hence allowing for different "meaning"s (e.g. exact filtering parameters) for the same overall descriptor `filt`.

Hence

```suggestion
    Labels SHOULD be consistent across subjects and sessions and data types.
    For example, an `acq` entity label `hires` used on the `anat` data of a given subject
    SHOULD mean the same thing as `acq-hires` in `func` files of any other subject.
```

yarikoptic: might be additional factor -- IIRC bids-validator would issue a WARNING whenever "same" (e.g. `_task-movie_run_1`) files across subjects have different "qualities" (e.g. length), not ERROR.  If changed to `MUST`, it then MUST issue an ERROR rendering some previously valid BIDS datasets invalid, breaking change.

PR #1332, opened 2022-10-20
INFRA: Drop numeric XX- from markdown file names

### LINKED ISSUE #1323: Drop numeric `XX-` from markdown file names
### Your idea

We've already done this with appendices in #1256. The numbers required placing appendices in the order they were added to the spec, not following any kind of logical or aesthetic order.

I think the modality-specific files in particular are suffering from this ordering: There are some groupings by similarity (volumetric MRI + PET, detector array *EG + NIRS) that would make sense to reflect in the section ordering. Additionally, when we have multiple BEPs running, they can clash on section numbering, which is simple to resolve but unclear why we should need to.

effigies:
Closes #1323.


sappelhoff (approved):
argh, I introduced merge conflicts by merging some PRs (I think). Sorry. +1 on merging this as soon as conflicts are resolved.

A potential re-ordering of sections can then follow in another PR that won't be as prone to conflicts.

sappelhoff:
- note that although the CI "checkmark" for your last commit https://github.com/bids-standard/bids-specification/pull/1332/commits/3427934e73c91d351d49be7bef36f8bbda3af792 is green, there were failures that I tried to fix with the following commits
- however, the build_docs_pdf check is stubborn, and I can't see what's wrong: locally the PDF builds well ... when logged into circleci via ssh, the problem is the same (perhaps obviously)

sappelhoff:
Ah, I see one consequence of dropping the `XX` prefixes now. The PDF build gets screwed up in its order:

https://github.com/bids-standard/bids-specification/blob/341a944a4fb427e8a068e600368c6ce2826b544f/pdf_build_src/pandoc_script.py#L23-L32

☝️ as can be seen in this snippet, the build was relying on the order being determined via the file names.

sappelhoff (changes requested):
PDF build is broken (see comments above)

effigies:
Seems this needs more thought.

sappelhoff:
probably we could parse `mkdocs.yml`'s "Nav": https://github.com/bids-standard/bids-specification/blob/6b09a667a43f3c4699a73d6f71729f7e76a16d85/mkdocs.yml#L52


... then we wouldn't need to have the loop that's dependent on file names, and we wouldn't have to maintain a duplicate list of files just for the PDF build

PR #1333, opened 2022-10-20
[ENH] Introduce GIFTI formats in derivatives

### LINKED ISSUE #1317: Add new derivative file extensions
### Your idea

In https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#file-formation-specification we define NIfTI, TSV and JSON as generally valid extensions.

For derivatives, I think it makes sense to introduce CIFTI-2 and GIFTI files, along with their extensions (e.g., `.dtseries.nii`, `.surf.gii`) to establish these as acceptable file types. We could use a limited subset, if there are some that are never seen in practice and we don't want to endorse without a use-case, or we could just say "If it follows these specs, then it must be fine".

Thoughts? Any other file types to consider?

effigies:
Closes #1317.


src/05-derivatives/01-introduction.md
```diff
@@ -70,10 +70,76 @@ in [Derived dataset and pipeline description][derived-dataset-description].
...
+| `NIFTI_INTENT_CONNECTIVITY_DENSE`              | `.dconn.nii`    |
+| `NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES`       | `.dtseries.nii` |
+| `NIFTI_INTENT_CONNECTIVITY_PARCELLATED`        | `.pconn.nii`    |
+| `NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SERIES` | `.ptseries.nii` |
+| `NIFTI_INTENT_CONNECTIVITY_DENSE_TRAJECTORY`   | `.dtraj.nii`    |
+| `NIFTI_INTENT_CONNECTIVITY_DENSE_SCALARS`      | `.dscalar.nii`  |
+| `NIFTI_INTENT_CONNECTIVITY_DENSE_LABELS`       | `.dlabel.nii`   |
+| `NIFTI_INTENT_CONNECTIVITY_PARCELLATED_SCALAR` | `.pscalar.nii`  |
+
+Unless otherwise stated, bare `.nii` extensions SHOULD not be used.
```
sappelhoff, line 130:
```suggestion
Unless otherwise stated, bare `.nii` extensions SHOULD NOT be used.
```

effigies:
@sappelhoff Thanks for moving this to 1.9.0 and prompting me to revisit.

I would at least like @neurolabusc's input here, particularly because GIFTI as specified and GIFTI as supported are two different things. I do not want to imply that BIDS endorses a to-spec reading of GIFTI, but nor do I think it is the place to attempt to explain GIFTI-as-supported and confine BIDS to that.

My goal here is to say "where appropriate, we will permit CIFTI-2 and GIFTI file structures, and here's where to go for more information when you see these".

neurolabusc:
@effigies in my mind, CIFTI and GIFTI are two separate issues.

**GIFTI** over the last couple month I have spent a bit of time with GIfTI and evaluating different tools. This has become one of the popular interchange standards in our field. Any format has its tradeoffs. GIFTI's limitations are speed, size and the fact that extensibility means that it can be interpreted and used differently. The advantages are that it is human readable, widely adopted and the extensibility means it can be adapted to fill emerging and tool specific niches. 

Many tools implement GIfTI a little differently, and in general they ignore triangle winding. We have recently identified issues with [FSL](https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind2302&L=FSL&P=R34669&X=D9D370C5933DB49DBD&Y), [FSL](https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=FSL;cb1a54da.2303), [nibabel](https://github.com/nipy/nibabel/issues/1198), [FreeSurfer](https://github.com/nipy/nibabel/issues/1101), [FreeSurfer](https://neurostars.org/t/freesurfer-cras-offset/5587), [FreeSurfer](https://neurostars.org/t/transform-freesurfer-mesh-coordinates-to-volume-coordinates/23693), [NiiVue](https://github.com/niivue/niivue/issues/507). However, these issues are fixable (and in many of the identified cases already fixed) and the needs to harmonize implementations should be expected with any format. In one sense, this variability highlights the pervasive popularity of GIFTI in our field. It filled a Darwinian niche and is widely adopted.

**CIFTI-2** From my perspective, this format was created by the connectome group to fill their specific use case. It is not yet widely adopted in our field. This format was not created by a consensus building across the community and is not yet widespread. I do think there is a use case and a niche for a format like CIFTI, but I do think there should be a community discussion of whether CIFTI-2 should be adopted or another more BIDS-like format that can adopt a similar role. Selecting a community format has implications, and we need to consider how this is implemented across languages,  performance, extended and the level of human readability (e.g. JSON vs XML). The fact that CIFTI mixes NIfTI binary headers and an XML extension means it feels unnatural to both those who like the speed/size/explicitness of binary as well as those who like the readability and extensibility of text information. One wonders if a NRRD-like solution or BSON solution might be more harmonious. For example, work by @fangq with [neurojson](https://github.com/fangq/jsonlab) hints at formats that can have simplicity, readability, extensibility, speed and file size. I am not saying we should not adopt CIFTI as a community, but I do think we should carefully consider this choice, evaluate alternatives, and develop a consensus.

effigies:
CIFTI-2 dropped from this PR. Will open a separate discussion.

Remi-Gau (approved):
OK with me
the CI failure is unrelated

sappelhoff (approved):

PR #1336, opened 2022-10-21
[ENH] Add qMRI fieldmap filename templates

### LINKED ISSUE #1335: No filename template for qMRI fieldmaps
I would expect about here. 

https://bids-specification.readthedocs.io/en/latest/appendices/qmri.html#field-maps

They should be easy to generate given that they are in the schema.

https://github.com/bids-standard/bids-specification/blob/5c84df3d0f8455d0a74a560e96a22c8648b06e0f/src/schema/rules/files/raw/fmap.yaml#L41

Remi-Gau:
closes #1335

### DIFF
--- a/src/appendices/qmri.md
+++ b/src/appendices/qmri.md
@@ -193,6 +193,29 @@ Explanation of the table:
 
 #### Field maps
 
+<!--
+This block generates a filename templates.
+The inputs for this macro can be found in the directory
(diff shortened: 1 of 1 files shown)


agahkarakuzu (approved):
The macro lists all fmap related suffixes for both maps and file collections, looks good to me. Thank you!

effigies (approved):

PR #1344, opened 2022-10-28
FIX: Ignore hidden directories correctly in validation

TheChymera:
Minor fix

### DIFF
--- a/tools/schemacode/bidsschematools/tests/test_validator.py
+++ b/tools/schemacode/bidsschematools/tests/test_validator.py
@@ -316,3 +316,20 @@ def test_error_datasets(bids_error_examples, dataset):
     )
     # Are there non-validated files?
     assert len(result["path_tracking"]) != 0
+
+
+def test_gitdir(bids_examples, tmp_path):
+    """Maybe better handled in example data?"""
+    from distutils.dir_util import copy_tree
+
+    from bidsschematools.validator import validate_bids
+
+    selected_dir = os.path.join(bids_examples, BIDS_SELECTION[0])
+    tmp_path = str(tmp_path)
+    copy_tree(selected_dir, tmp_path)
+
+    os.makedirs(os.path.join(tmp_path, ".git"))
+    with open(os.path.join(tmp_path, ".git", "config"), "w") as temp_file:
+        temp_file.write("")
+    result = validate_bids(tmp_path)
+    assert len(result["path_tracking"]) == 0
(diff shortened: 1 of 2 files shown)


effigies (approved):

effigies:
I would recommend just mocking a repository with some dummy files.

TheChymera:
@effigies you mean just create a `.git` subdir via the fixture? I think that might actually over-complicate stuff. I could put it in bids-error-examples, but it's actually not an error, the point is that a bids dataset should not throw warnings about invalid viles if it contains `.git/` — hence I was thinking about bids-examples.

effigies:
I mean write a fixture that creates a minimal dataset.

```Python
@pytest.fixture
def mock_dataset(tmp_path):
    dd = tmp_path / "dataset_description.json"
    dd.write_text(json.dumps({"BIDSVersion": "1.8.0", "Name": "Mock dataset"})
    t1 = tmp_path / "sub-01" / "anat" / "sub-01_T1w.nii.gz"
    t1.parent.mkdir(parents=True)
    t1.write_bytes(b'')
    yield tmp_path
```

That you can then use with:

```Python
def test_gitdir(mock_dataset):
    gitconfig = mock_dataset / ".git" / "config"
    gitconfig.parent.mkdir()
    gitconfig.write_text('')
    validate(mock_dataset)
```

I don't think real examples should be our end-all-be-all of tests.

TheChymera:
Hm, ok, though I still think this *might* be better covered in static test data.
The determinant of that is whether “hidden directories do not matter” is a characteristic of BIDS or a characteristic of the built-in BIDS validator specifically.

If it's the former, then this should be present in the reference static example data, since a `.git` dir not raising errors is a feature of BIDS. The main advantage of having such an example distributed statically, beyond it “making sense” in theory, is that people working on other validators can point them to that repository and see whether they're implementing BIDS correctly — whereas our tests can't really be leveraged by other validators.

If it's the latter case, yeah, though then I would ask why we're hard-coding this behaviour to begin with 🤔 If it's really just a convenience feature for downstream use in DANDI maybe it should be parameterized rather than default behaviour (though that would be a different commit).

effigies:
The BIDS examples are primarily instructive to the user and also need to validate. They are not an exhaustive set of test cases for a validator, and I don't think we should try to make them so. We could create (and I thought @yarikoptic had started to...) a set of cases that are more oriented to covering things that should pass or fail.

yarikoptic:
> We could create (and I thought @yarikoptic had started to...) a set of cases that are more oriented to covering things that should pass or fail.

sorry, for clarity and posterity,  I am not working on any of such set of cases.  FWIW: I do like programmatic/fixture way to test you suggested in https://github.com/bids-standard/bids-specification/pull/1344#issuecomment-1295367182 .

effigies (approved):

PR #1347, opened 2022-11-02
[SCHEMA] Add tsv to extensions for any file with dseg suffix.

rwblair:
These tsv files aren't listed in the filename templates for segmentations but are specified as allowed here:
https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#common-image-derived-labels

### DIFF
--- a/src/schema/rules/files/deriv/imaging.yaml
+++ b/src/schema/rules/files/deriv/imaging.yaml
@@ -94,9 +94,7 @@ anat_parametric_discrete_segmentation:
     density: optional
     description: optional
   extensions:
-    - .nii.gz
-    - .nii
-    - .json
+    - $ref: rules.files.raw.anat.parametric.extensions
     - .tsv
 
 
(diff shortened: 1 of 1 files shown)


src/schema/rules/files/deriv/imaging.yaml
```diff
@@ -93,6 +93,12 @@ anat_parametric_discrete_segmentation:
     resolution: optional
     density: optional
     description: optional
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+    - .tsv
```
tsalo, line 100: If the goal is to extend the extensions with `.tsv`, do you want to reference the existing list of extensions instead?

```suggestion
  extensions:
    - $ref: rules.files.raw.anat.parametric.extensions
    - .tsv
```

rwblair: I do, had forgotten if it worked with extensions, thank you.

rwblair: dang looks like dereferencing might be failing on lists

PR #1349, opened 2022-11-04
[ENH] Recommend gzip header fields be set to empty values

### LINKED ISSUE #136: mismatch between name of a *.nii.gz file and the *.nii file it contains
I posted it somewhere [else](https://github.com/poldracklab/fmriprep/issues/1480#issuecomment-457218760) but I figure this is also a potential BIDS issue.

AFAIK BIDS allow both *.nii and *.nii.gz format but there is no requirement for the *.nii file inside to have the same name as the *.nii.gz file it is in. Moreover several *.nii.gz in a folder can contain *.nii files that despite being different in terms of content have the same name.

For example the files `sub-001_task-MGT_run-01_bold.nii.gz` and `sub-001_task-MGT_run-02_bold.nii.gz` can both contain a file that is called `vol0000_xform-00000.nii`.

Maybe I am missing something on how unzipping works but I think this could sometimes be an "inconvenience" for (mostly matlab/spm) users when the first thing they will do is unzip everything. So maybe it would be good to recommend in the main specs that those names are matched.

kousu:
By removing extraneous metadata, pipelines generating BIDS datasets can avoid making unnecessary changes.

We are motivated to this since we are working on combining BIDS with [datalad](https://www.datalad.org/) in our workflows, and if a pipeline regenerates hundreds of images with no real change, then we rapidly bloat our storage costs and obscures real changes in `git log`. datalad (thanks to its dependency [git-annex](https://git-annex.branchable.com/)) avoids the bandwidth/storage costs of downstream users, since most of our users download once only, or are always downloading fresh HEAD copies to processing nodes, but our internal storage costs are a big factor for us.

This can be implemented with `gzip -n` in most environments, or equally by running [`strip-nondeterminism`](https://salsa.debian.org/reproducible-builds/strip-nondeterminism) over a BIDS dataset before publishing.

closes #136

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -368,7 +368,8 @@ datasets and non-compliant derivatives.
 ### Imaging files
 
 All imaging data MUST be stored using the NIfTI file format. We RECOMMEND using
-compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. Imaging data SHOULD
+compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. If using compressed files,
+the gzip header MUST lack source filenames timestamps. Imaging data SHOULD
 be converted to the NIfTI format using a tool that provides as much of the NIfTI
 header information (such as orientation and slice timing information) as
 possible. Since the NIfTI standard offers limited support for the various image
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -369,7 +369,7 @@ datasets and non-compliant derivatives.
 
 All imaging data MUST be stored using the NIfTI file format. We RECOMMEND using
 compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. If using compressed files,
-the gzip header MUST lack source filenames timestamps. Imaging data SHOULD
+the gzip header SHOULD lack source filenames and timestamps. Imaging data SHOULD
 be converted to the NIfTI format using a tool that provides as much of the NIfTI
 header information (such as orientation and slice timing information) as
 possible. Since the NIfTI standard offers limited support for the various image


effigies:
I don't think we can make this a requirement as it will invalidate many existing datasets. I would support making it a recommendation, which would produce validation warnings, assuming there is a straightforward way to check for this metadata in Javascript.

kousu:
Cool :). I thought you might say something like that. That would be okay, but I wanted to start out as strong as possible.

Before totally giving up on the "MUST", let me point out that `strip-nondeterminism` should be able to rapidly re-validate all those datasets. We've had to deal with breaking BIDS changes in our lab already that involved more complicated fixes than that.

effigies:
Agreed that the fix is easy, and some labs will be willing to apply it. Apart from concern about making BIDS a moving target and generally annoying people who've done their bit, my perspective here is the OpenNeuro one, where we host data archives (so reproducibility of the archive isn't a concern) but don't generally have permission to update datasets on the user's behalf.

It is true that there've been breaks in the past; most breaking changes in the validator involved rules that were previously unenforced. There are some others where an ambiguity had to be resolved in one direction or another. We do honestly try to keep these to a minimum, and will generally try to accommodate existing practice except when it makes interpretation of the dataset ambiguous or incoherent.

For your lab/center's use case, the validator allows warnings to be promoted to errors, so internally you could have more strict validation. OpenNeuro does this for missing authors, as we can't generate DOIs without this metadata.

So my vote remains for a backwards-compatible SHOULD, though I would support moving to MUST in a BIDS 2.0.

xref https://github.com/bids-standard/bids-validator/issues/1400

kousu:
> the validator allows warnings to be promoted to errors

Oh okay! I hadn't noticed that yet. Thank you. That's good enough for me :)

effigies (approved):

DimitriPapadopoulos (approved):

effigies:
Rebase changed apparent commit time, but it was authored on Nov 4, so this has passed the 5 day review period. Merging.

PR #1353, opened 2022-11-10
SCHEMA: Define JSON documents in schema

effigies:
Follow up to #1233. This makes `genetic_info.json` and `dataset_description.json` objects in `objects.documents` instead of `objects.metadata`, which seems more appropriate.

One sticking point is how to indicate conditional requirement levels in an object, rather than a rule.


src/schema/objects/documents.yaml
```diff
@@ -0,0 +1,71 @@
...
+      $ref: objects.metadata.ReferencesAndLinks
+    DatasetDOI:
+      $ref: objects.metadata.DatasetDOI
+    GeneratedBy:
+      $ref: objects.metadata.GeneratedBy
+    SourceDatasets:
+      $ref: objects.metadata.SourceDatasets
+    Genetics:
+      $ref: objects.metadata.Genetics
+      level_addendum: required if `genetic_info.json` is present
```
effigies, line 50: @tsalo It turns out that the validator has been giving an error, so we should reflect that in the spec.

bendhouseart (approved):
#liked, #subscribed.

src/schema/objects/documents.yaml
tsalo: I've opened #1395 about this.

PR #1360, opened 2022-11-17
NF: Add GZIP_HEADER_DATA issue, following gh-1349

effigies:
Adds a `gzip` property to the context object and a check that timestamp, filename and comment are all null.


bendhouseart (approved):
Liked and subscribed.

sappelhoff (approved):

PR #1128, opened 2022-06-15
[ENH] BEP 020 Eye Tracking

mszinte:
Dear BIDS maintainer, 
Here is the specifications of the BEP 020 about eye tracking.
- it follows the main discussion initiated on a google document.
- it includes the different modification the group of maintainers suggested to us during our zoom meeting.
- it includes the macros as used in other modality specific extensions
- it includes links toward dataset examples.

Please let me know if you feel something is wrong.
Best,

---

- [x] implement macros
  - [x] for filename templates ?
  - [x] for examples ?
  - [x] for metadata table
- [x] add contributors to the [wiki](https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors) (so they can be added to the contributors page)
- [ ] end docmentation
- [ ] update examples
- [ ] update validator

PR #1366, opened 2022-12-08
[MISC] Sort subsections within sections alphabetically

yarikoptic:
I was feeling the need for it each time I was navigating BIDS, but seeing
https://github.com/bids-standard/bids-specification/pull/1128/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2 suggesting the same made me propose it.

I think having subsections sorted where there is no really really a requirement for order helps visitor to navigate the document/website instead of doing linear search among all subsections.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -6,47 +6,47 @@ nav:
       - Introduction: introduction.md
       - Common principles: common-principles.md
       - Modality agnostic files: modality-agnostic-files.md
-      - Modality specific files:
-          - Magnetic Resonance Imaging: modality-specific-files/magnetic-resonance-imaging-data.md
-          - Magnetoencephalography: modality-specific-files/magnetoencephalography.md
-          - Electroencephalography: modality-specific-files/electroencephalography.md
-          - Intracranial Electroencephalography: modality-specific-files/intracranial-electroencephalography.md
-          - Task events: modality-specific-files/task-events.md
-          - Physiological recordings: modality-specific-files/physiological-recordings.md
+      - Modality specific files:  # keep sections in sorted order when adding/editing
           - Behavioral experiments (with no neural recordings): modality-specific-files/behavioral-experiments.md
+          - Electroencephalography: modality-specific-files/electroencephalography.md
           - Genetic Descriptor: modality-specific-files/genetic-descriptor.md
-          - Positron Emission Tomography: modality-specific-files/positron-emission-tomography.md
-          - Microscopy: modality-specific-files/microscopy.md
-          - Near-Infrared Spectroscopy: modality-specific-files/near-infrared-spectroscopy.md
-          - Motion: modality-specific-files/motion.md
+          - Intracranial Electroencephalography: modality-specific-files/intracranial-electroencephalography.md
+          - Magnetic Resonance Imaging: modality-specific-files/magnetic-resonance-imaging-data.md
           - Magnetic Resonance Spectroscopy: modality-specific-files/magnetic-resonance-spectroscopy.md
-      - Derivatives:
+          - Magnetoencephalography: modality-specific-files/magnetoencephalography.md
+          - Microscopy: modality-specific-files/microscopy.md
+          - Motion: modality-specific-files/motion.md
+          - Near-Infrared Spectroscopy: modality-specific-files/near-infrared-spectroscopy.md
+          - Physiological recordings: modality-specific-files/physiological-recordings.md
+          - Positron Emission Tomography: modality-specific-files/positron-emission-tomography.md
+          - Task events: modality-specific-files/task-events.md
+      - Derivatives:  # keep sections in sorted order when adding/editing
           - BIDS Derivatives: derivatives/introduction.md
           - Common data types and metadata: derivatives/common-data-types.md
           - Imaging data types: derivatives/imaging.md
       - Longitudinal and multi-site studies: longitudinal-and-multi-site-studies.md
       - Glossary: glossary.md
       - BIDS Extension Proposals: extensions.md
-      - Appendix:
-          - Schema: appendices/schema.md
+      - Appendix:  # keep sections in sorted order when adding/editing
+          - Arterial Spin Labeling: appendices/arterial-spin-labeling.md
           - Contributors: appendices/contributors.md
-          - Licenses: appendices/licenses.md
-          - Entity table: appendices/entity-table.md
+          - Coordinate systems: appendices/coordinate-systems.md
+          - Cross modality correspondence: appendices/cross-modality-correspondence.md
           - Entities: appendices/entities.md
+          - Entity table: appendices/entity-table.md
           - File collections: appendices/file-collections.md
-          - Units: appendices/units.md
           - Hierarchical Event Descriptors: appendices/hed.md
+          - Licenses: appendices/licenses.md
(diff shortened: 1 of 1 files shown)


yarikoptic:
> Fine as is for me but always wondering if we should not put a comment in the yml saying that those sub sections are ordered alphabetrically.

I rebased and added a comment for each of those sections. AFAIK comments are ok for yaml so hopefully it would not break any workflow. Check it out

mkdocs.yml
```diff
@@ -5,44 +5,44 @@ nav:
...
       - Modality agnostic files: modality-agnostic-files.md
-      - Modality specific files:
-        - Magnetic Resonance Imaging: modality-specific-files/magnetic-resonance-imaging-data.md
-        - Magnetoencephalography: modality-specific-files/magnetoencephalography.md
-        - Electroencephalography: modality-specific-files/electroencephalography.md
-        - Intracranial Electroencephalography: modality-specific-files/intracranial-electroencephalography.md
-        - Task events: modality-specific-files/task-events.md
-        - Physiological and other continuous recordings: modality-specific-files/physiological-and-other-continuous-recordings.md
+      - Modality specific files: # keep sections in sorted order when adding/editing
         - Behavioral experiments (with no neural recordings): modality-specific-files/behavioral-experiments.md
```
effigies, line 9: Why do we want these sorted alphabetically? I agree that order of addition to the spec is not ideal, but this feels unintuitive. I would expect EEG/MEG/iEEG/NIRS to be near each other, for example.

mkdocs.yml
```diff
@@ -5,44 +5,44 @@ nav:
...
+        - Intracranial Electroencephalography: modality-specific-files/intracranial-electroencephalography.md
+        - Magnetic Resonance Imaging: modality-specific-files/magnetic-resonance-imaging-data.md
+        - Magnetoencephalography: modality-specific-files/magnetoencephalography.md
         - Microscopy: modality-specific-files/microscopy.md
         - Near-Infrared Spectroscopy: modality-specific-files/near-infrared-spectroscopy.md
-      - Derivatives:
+        - Physiological and other continuous recordings: modality-specific-files/physiological-and-other-continuous-recordings.md
+        - Positron Emission Tomography: modality-specific-files/positron-emission-tomography.md
+        - Task events: modality-specific-files/task-events.md
+      - Derivatives: # keep sections in sorted order when adding/editing
```
effigies, line 20: I guess it's lucky that derivatives were already sorted. But these are also sorted from "most broadly applicable" to "most specific".

mkdocs.yml
```diff
@@ -5,44 +5,44 @@ nav:
...
         - BIDS Derivatives: derivatives/introduction.md
         - Common data types and metadata: derivatives/common-data-types.md
         - Imaging data types: derivatives/imaging.md
       - Longitudinal and multi-site studies: longitudinal-and-multi-site-studies.md
       - Glossary: glossary.md
       - BIDS Extension Proposals: extensions.md
-      - Appendix:
+      - Appendix: # keep sections in sorted order when adding/editing
+        - Arterial Spin Labeling: appendices/arterial-spin-labeling.md
         - Contributors: appendices/contributors.md
```
effigies, line 29: Similarly, I think `Contributors`, `Licenses`, `Entities`, `Units`, etc were at the front for being more broadly useful.

mkdocs.yml
```diff
@@ -5,44 +5,44 @@ nav:
...
         - Quantitative MRI: appendices/qmri.md
-        - Arterial Spin Labeling: appendices/arterial-spin-labeling.md
-        - Cross modality correspondence: appendices/cross-modality-correspondence.md
+        - Units: appendices/units.md
       - Changelog: CHANGES.md
-    - The BIDS Starter Kit:
-      - Website: https://bids-standard.github.io/bids-starter-kit
-      - Tutorials: https://bids-standard.github.io/bids-starter-kit/tutorials/tutorials.html
+    - The BIDS Starter Kit: # keep sections in sorted order when adding/editing
       - GitHub repository: https://github.com/bids-standard/bids-starter-kit
```
effigies, line 43: Why should the GitHub source repo be prioritized over rendered docs?

mkdocs.yml
yarikoptic: In case of ambiguity I prefer "explicitness" of sorted order.  It has been already many times I had to search up for a section because I managed to miss it in some "intuitive" order. 
If there is no "flow" through the sections requiring some to precede another (is there?) I would really prefer explicit alphabetic sorting even if it would not have some "intuitive feel" ;)

mkdocs.yml
yarikoptic: Just the same logic -- explicit ordering, and who knows what I am looking for ? (I often come to RTD to find the original repo behind it). Having said that -- we could though make it into `Sources (GitHub):` instead and bring down the line/make robust to change in hosting provider.

mkdocs.yml
yarikoptic: yeah -- here may be we could leave as is without comment

mkdocs.yml
mszinte: My intuitive feel was different and actually close to the proposal of @yarikoptic ... The first time I rendered the doc for the BEP020 I though my modifications didn't work as i couldn't find it in the top of the list ("Eyetracking" starting with "E")... this is why i changed it first, before being noticed that it couldn't be part of my PR.

sappelhoff (commented):
I think I am fine with alphabetical ordering. Once you know what you are looking for, it'll be easier to find.

The drawback of course is, that new visitors to the spec don't get the subsections in "out subjective order of relevance" (e.g., the "Units" appendix before the ASL appendix).

mkdocs.yml
sappelhoff: +1 for `Sources (GitHub)`

> I often come to RTD to find the original repo behind it

I also do that, but I don't think that's necessarily a representative behavior for BIDS users ... I am still fine with the proposed change of ordering alphabetically as I think it has more advantages than drawbacks.

mkdocs.yml
yarikoptic: well, meanwhile it became 

[Schema](https://bids-specification.readthedocs.io/en/stable/appendices/schema.html)
[Contributors](https://bids-specification.readthedocs.io/en/stable/appendices/contributors.html)
[Licenses](https://bids-specification.readthedocs.io/en/stable/appendices/licenses.html)
[Entity table](https://bids-specification.readthedocs.io/en/stable/appendices/entity-table.html)
[Entities](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html)
[File collections](https://bids-specification.readthedocs.io/en/stable/appendices/file-collections.html)
[Units](https://bids-specification.readthedocs.io/en/stable/appendices/units.html)

so Units are far from the front etc attesting to the point that without clear deterministic order opinions would differ and overall leading to some arbitrary order favoring some and disfavoring others

yarikoptic:
Dear participants and other @bids-maintenance folks!   I have resolved conflicts and commented a little more.  I would appreciate your final comments/decisions/actions: IMHO, especially accounting for changes which happened to the menu items meanwhile, a deterministic order (e.g. sorted) would be beneficial and would reduce cognitive burden in adding new ones and navigating the website where the list of menu items is only growing and current, mostly "evolutional", order does not really make sense.

PR #1368, opened 2022-12-08
RF: allow for any "recording" file to be listed in _scans.tsv not just "neural recording"

yarikoptic:
I think this would be more appropriate since we could have _beh and other files with their acq_time.  I kept it as "recording file" (not e.g. just "data file") to keep better alignment with acq_time name of the column.

So could also be "acquisition time" instead of "recording time".


sappelhoff (approved):
I think this makes sense

PR #1369, opened 2022-12-09
[ENH] add screen parameters metadata

Remi-Gau:
Extracts screen parameters metadata from [BEP20](https://github.com/bids-standard/bids-specification/pull/1128) and adds them to `events.tsv` associated metadata to make them usable by other modalities.

---
**EDIT**

Clarification: those extra metadata are meant to go in the `events.json`

---

Motivation:

The metadata listed below are not eyetracking specific and can in principle be used by any experiment that uses visual stimulation. Those metadata are actually fairly important for anyone who would like to replicate an experiment where visual stimulus relative size matters.

- ScreenDistance
- ScreenRefreshRate
- ScreenResolution
- ScreenSize


Remi-Gau:
@mszinte as discussed I am extracting this part of BEP020.

Remi-Gau:
@effigies 

The requirement level for those metadata is currently recommended.

For eyetracking some will have to be required, so this may require overriding the `StimulusPresentation` subobject for eyetrcaking to be more like this.

```yml
StimulusPresentation:
  name: StimulusPresentation
  display_name: Stimulus Presentation
  description: |
    Object containing key-value pairs related to the software used to present
    the stimuli during the experiment, specifically:
    `"OperatingSystem"`, `"SoftwareName"`, `"SoftwareRRID"`, `"SoftwareVersion"` and
    `"Code"`.
    See table below for more information.
  type: object
  recommended_fields:
    - OperatingSystem
    - ScreenRefreshRate
    - SoftwareName
    - SoftwareRRID
    - SoftwareVersion
    - Code
  required_fields: #   <---------  those fields are required
    - ScreenDistance
    - ScreenResolution
    - ScreenSize
  properties:
    OperatingSystem:
      $ref: objects.metadata.OperatingSystem
    ScreenDistance:
      $ref: objects.metadata.ScreenDistance
    ScreenRefreshRate:
      $ref: objects.metadata.ScreenRefreshRate
    ScreenResolution:
      $ref: objects.metadata.ScreenResolution
    ScreenSize:
      $ref: objects.metadata.ScreenSize
    SoftwareName:
      $ref: objects.metadata.SoftwareName
    SoftwareRRID:
      $ref: objects.metadata.SoftwareRRID
    SoftwareVersion:
      $ref: objects.metadata.SoftwareVersion
    Code:
      $ref: objects.metadata.Code
```

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
+    - type: number
+      unit: cm
+    - type: string
+      enum:
+        - n/a
+ScreenRefreshRate:
+  name: ScreenRefreshRate
+  display_name: Screen Refresh Rate
+  description: |
+    Refresh rate of the screen (when used), in Hertz.
```
sappelhoff, line 2843:
```suggestion
    Refresh rate of the screen (when used), in Hertz (equivalent to frames per second, "FPS").
```

... right?

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
+        minItems: 2
+        maxItems: 2
+    - type: string
+      enum:
+        - n/a
+ScreenSize:
+  name: ScreenSize
+  display_name: Screen Size
+  description: |
+    Screen size in cm
```
sappelhoff, line 2866: as a clarification that only the "active screen" is meant?

```suggestion
    Screen size in cm, excluding potential screen borders
```

sappelhoff (approved):
+1 to include this in the experiment metadata

effigies:
I think we could do either an override or just a separate check for those values.

effigies:
The current description is

> Object containing key-value pairs related to the software used to present
> the stimuli during the experiment, specifically:
> `"OperatingSystem"`, `"SoftwareName"`, `"SoftwareRRID"`, `"SoftwareVersion"` and
> `"Code"`.

If you want to make this include hardware and positioning details, then the description will need to be udpated. Alternately you could create a new metadata field like `PresentationHardware`.

CPernet:
I don't understand why/how this would go into ``events.tsv``? I do get your point that this is a generic type of info.  
The reason why it doesn't make sense to me to put that in ``events.tsv`` is that for 99% of use cases, the screen distance, refresh rate, resolution and size doesn't change - which means replicating the same values over and over again in the tsv file.  
Is is possible to store the info in the ``events.json`` even if the key/values do not correspond to columns of the tsv?

robertoostenveld:
@CPernet I think that the `events.json` is meant by the "metadata for the `events.tsv`".  That would place it along https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html#stimulus-presentation-details

Remi-Gau:
> I don't understand why/how this would go into `events.tsv`? I do get your point that this is a generic type of info. The reason why it doesn't make sense to me to put that in `events.tsv` is that for 99% of use cases, the screen distance, refresh rate, resolution and size doesn't change - which means replicating the same values over and over again in the tsv file. Is is possible to store the info in the `events.json` even if the key/values do not correspond to columns of the tsv?

yup as mentioned by @robertoostenveld I meant the events.json. Will add a correction in the top message. Sorry for the confusion.

CPernet:
I don't think that's what @robertoostenveld meant -- I guess the verification here is 1 - all columns of the tsv should be described in json but 2 - can json have more elements than tsv

Remi-Gau:
> If you want to make this include hardware and positioning details, then the description will need to be udpated. Alternately you could create a new metadata field like `PresentationHardware`.

OK I decided to actually remove from the description the list of fields that this object should include because it felt redundant to list them verbatim and then have a table rendered to list them again.

Remi-Gau:
> 2 - can json have more elements than tsv

Yes they can.

Whether to describe the stimuli database used or some of the presentation details:

- https://bids-specification.readthedocs.io/en/latest/modality-specific-files/task-events.html#stimuli-databases

- https://bids-specification.readthedocs.io/en/latest/modality-specific-files/task-events.html#stimulus-presentation-details

```json
{
    "trial_type": {
        "LongName":   "Emotion image type",
        "Description": "Type of emotional face from Karolinska database that is displayed",
        "Levels": {
            "afraid": "A face showing fear is displayed",
            "angry":  "A face showing anger is displayed",
            "sad":    "A face showing sadness is displayed"
        }
    },
    "identifier": {
        "LongName": "Unique identifier from Karolinska (KDEF) database",
        "Description": "ID from KDEF database used to identify the displayed image"
    },
    "StimulusPresentation": {
        "OperatingSystem": "Linux Ubuntu 18.04.5",
        "SoftwareName": "Psychtoolbox",
        "SoftwareRRID": "SCR_002881",
        "SoftwareVersion": "3.0.14",
        "Code": "doi:10.5281/zenodo.3361717"
    }
}
```

CPernet:
ok good (because of course in tsv that was weird)

CPernet (approved):

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
+  description: |
+    Refresh rate of the screen (when used), in Hertz (equivalent to frames per second, "FPS").
+  type: number
+  unit: Hz
+ScreenResolution:
+  name: ScreenResolution
+  display_name: Screen Resolution
+  description: |
+    Screen resolution in pixel
+    (for example `[1920, 1200]` for a screen of 1920-length by 1080-height pixels),
```
yarikoptic, line 2851:
```suggestion
    (for example `[1920, 1200]` for a screen of 1920-width by 1080-height pixels),
```

not a native speaker here but length is IMHO ambiguous and height typically goes with width

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
+        maxItems: 2
+    - type: string
+      enum:
+        - n/a
+ScreenSize:
+  name: ScreenSize
+  display_name: Screen Size
+  description: |
+    Screen size in cm, excluding potential screen borders
+    (for example `[47.2, 29.5]` for a screen of 47.2-length by 29.5-height cm),
```
yarikoptic, line 2867:
```suggestion
    (for example `[47.2, 29.5]` for a screen of 47.2-width by 29.5-height cm),
```

src/schema/objects/metadata.yaml
Remi-Gau: good catch

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
     minimum: 0
     maximum: 100
+ScreenDistance:
+  name: ScreenDistance
+  display_name: Screen Distance
+  description: |
+    Distance between the participant's eye and the screen. If no screen was used, use `n/a`.
+  anyOf:
+    - type: number
+      unit: cm
```
yarikoptic, line 2835: I understand that cm is the most sensible "range" of the unit to express distances and sizes here, but because we don't actually use unit to annotate the number, IMHO it would be cleaner to use the base unit of `m` so when I encounter 1 for eg width I know it is 1 meter without looking up the standard to see the unit expected to be in the specific type of the width. 

May be someone could point me to prior discussion on this aspect of there was any?

FWIW I don't see cm used in similar implicit unit https://github.com/bids-standard/bids-specification/search?q=cm&type=

src/schema/objects/metadata.yaml
```diff
@@ -2825,6 +2825,57 @@ ScatterFraction:
...
+  display_name: Screen Size
+  description: |
+    Screen size in cm, excluding potential screen borders
+    (for example `[47.2, 29.5]` for a screen of 47.2-width by 29.5-height cm),
+    if no screen use `n/a`.
+  anyOf:
+    - type: array
+      items:
+        type: number
+        unit: cm
```
yarikoptic, line 2873: Especially here `m` would much better unit IMHO... and as such I would prefer all of them to be just meters

src/modality-specific-files/task-events.md
```diff
@@ -268,7 +268,11 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
         "SoftwareName": "Psychtoolbox",
         "SoftwareRRID": "SCR_002881",
         "SoftwareVersion": "3.0.14",
-        "Code": "doi:10.5281/zenodo.3361717"
+        "Code": "doi:10.5281/zenodo.3361717",
+        "ScreenDistance": 180,
```
yarikoptic, line 272: 1.80 of meters here is very sensible

src/schema/objects/metadata.yaml
sappelhoff: I also think that cm is the typical unit for distance from screen (judging from what I know from my own studies and from what I generally see in papers); however that said, I don't feel strongly against making it `m` either.

src/schema/objects/metadata.yaml
sappelhoff: why would meter be much better for screen size? screens are usually smaller than one meter, aren't they?

robertoostenveld: I think meter would be better because it is the SI base unit, i.e., the unit without prefix, as recommended [here](https://bids-specification.readthedocs.io/en/stable/appendices/units.html). That also aligns with the related derived units (and measurements), such as Luminance (cd/m^2).

Remi-Gau: either way I don't feel too strongly about this, though relying on pure unprefixed SI as much as possible has its merits.

@mszinte any opinion?

yarikoptic:
BTW thank you for "extracting" this  as a separate PR. A great example for https://github.com/bids-standard/bids-specification/issues/371

src/modality-specific-files/task-events.md
dorahermes: +1 on using SI base units (meters) following Unit appendix (as @robertoostenveld also noted)

src/schema/objects/metadata.yaml
Remi-Gau: Talked with @mszinte.

Fine with using meters.

Remi-Gau, line 2835:
```suggestion
      unit: m
```

src/schema/objects/metadata.yaml
Remi-Gau, line 2873:
```suggestion
        unit: m
```

Remi-Gau:
Been more than 5 days.

Merging.

src/modality-specific-files/task-events.md
yarikoptic, line 272: @Remi-Gau - even though unit was adjusted to become `m`, examples seems were not like here  and below!... I will send a PR

Remi-Gau: Oops. Thanks for spotting it.

PR #1373, opened 2022-12-25
BF: sizes were decided to be in meters. Fix examples

yarikoptic:
Follow up to https://github.com/bids-standard/bids-specification/pull/1369


Remi-Gau (approved):
Thanks for fixing this.

effigies (approved):

PR #1377, opened 2023-01-06
[ENH] BEP022 - Magnetic Resonance Spectroscopy

### LINKED ISSUE #680: Extend BIDS format to magnetic resonance spectroscopy
I am working on a general format for MRS and MRSI data. I read the BIDS format article; it has been stated that it will support spectroscopy in the future. To the best of my knowledge, it has not happened yet.  but there are some works, for example, this repo(https://github.com/wexeee/spec2nii) uses a JSON sidecar file for storing parameters. However; It needs some collaborations to reach a consensus about open issues. I would suggest extending BIDS to MRS data.

markmikkelsen:
> [!IMPORTANT]  
> This BEP is under community review from June 3 - June 14, 2024. Please participate at https://github.com/bids-standard/bids-specification/discussions/1846.

This pull request is to add BEP022 (magnetic resonance spectroscopy) to the BIDS specification.

Note that some of the MRI metadata descriptions have been modified to generalize to MRS or to make slightly different definitions apply to MRI and MRS.

> [!Tip]
>
> [**HTML preview of this BEP**](https://bids-specification--1377.org.readthedocs.build/en/1377/modality-specific-files/magnetic-resonance-spectroscopy.html)

Closes #680.


markmikkelsen:
I'm not sure why the check_links test is failing. Seems to be a URL leading to a publication, but there isn't anything wrong with the link.

src/schema/objects/metadata.yaml
```diff
@@ -70,6 +70,20 @@ AnalyticalApproach:
...
         type: string
+AnatomicalImage:
+  name: AnatomicalImage
+  display_name: Anatomical Image
+  description: |
+    The path(s) to the anatomical MR image file(s), if present,
+    to which the associated MRS data file corresponds.
+    Contains one or more [BIDS URIs](SPEC_ROOT/common-principles.md#bids-uri).
+    Using forward-slash separated paths relative to the participant subdirectory is
+    [DEPRECATED](SPEC_ROOT/common-principles.md#definitions).
```
effigies, line 81: Is there a reason to write in deprecated functionality to a new extension?

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,89 @@
...
+  fields:
+    InstitutionName: recommended
+    InstitutionAddress: recommended
+    InstitutionalDepartmentName: recommended
+    Manufacturer: recommended
+    ManufacturersModelName: recommended
+    DeviceSerialNumber: recommended
+    StationName: recommended
+    SoftwareVersions: recommended
+    HardcopyDeviceSoftwareVersion: DEPRECATED
```
effigies, line 21: We should not include deprecated fields in new data descriptions. If it contains information that is otherwise unspecified, then it should not be deprecated.

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,89 @@
...
+    SpectrometerFrequency: required
+    SpectralWidth: required
+    EchoTime: required
+
+MRSRecommendedFields:
+  selectors:
+    - modality == "mrs"
+  fields:
+    NumberOfSpectralPoints: recommended
+    NumberOfTransients: recommended for SVS
```
effigies, line 54: Is this optional for non-SVS data, or does it only apply to SVS data?

If it's generally optional, but promoted to recommended for SVS, then we want to make it clear here, and then add a second table that will be used for validation:

```YAML
MRSRecommendedFields:
  selectors:
    - modality == "mrs"
  fields:
    NumberOfSpectralPoints: recommended
    NumberOfTransients:
      level: optional
      level_addendum: recommended for SVS
    ...

SVSRecommendedFields:
  selectors:
    - datatype == "mrs"
    - suffix == "svs"
  fields:
    NumberOfTransients: recommended
```

If it only makes sense for SVS, then we'll want to remove it from this table, and add the second table like so:

```YAML
SVSRecommendedFields:
  selectors:
    - datatype == "mrs"
    - suffix == "svs"
  fields:
    NumberOfTransients:
      level: recommended
      level_addendum: for SVS
```

In this second case, the macro will need to combine the two fields:

```
{{ MACROS___make_sidecar_table(["mrs.MRSRecommendedFields", "mrs.SVSRecommendedFields"]) }}
```

markmikkelsen: It only makes sense for SVS, so the second case is more appropriate. Looping in @martin3141 and @wtclarke to confirm.

wtclarke: Agree mostly with Mark, on his assessment. Though you could sometimes define it for mrsi, it isn't generally a parameter of interest for those types. However, for unlocalised spectroscopy (`unloc`) then again it would be of interest like svs.

markmikkelsen: Good point; it also applies to `*_unloc` data.

effigies: I think it would make sense to just call it `OPTIONAL but RECOMMENDED for SVS data`, which would correspond to option 1.

markmikkelsen: It also applies to `unloc` data, so it should be `RECOMMENDED` for such data. Can we add: `level_addendum: for SVS and unloc`? It shouldn't be `OPTIONAL`, given how important it is to report `NumberOfTransients` in MRS papers (https://doi.org/10.1002/nbm.4484).

effigies: The way we would do this is:

```suggestion
    NumberOfTransients:
      level: optional
      level_addendum: recommended for SVS and unloc
```

This will be used for rendering.

And then add a separate table for enforcing:

```YAML
NumberOfTransientsRecommended:
  selectors:
    - datatype == "mrs"
    - intersects([suffix], ["svs", "unloc"])
  fields:
    NumberOfTransients: recommended
```

markmikkelsen: This sounds like the way forward to me!

Remi-Gau:
Had a quick look. I may be missing something but...

I am wondering if there is any reason to add the matrix size as a metadata when it is already implicitly present in the nifti header?

The description of VolumeAffineMatrix and AcquisitionVoxelSize make it clear that those values can actually be different from that found in the header. But this does not seem to be the case for the MatrixSize?

If so that it may lead to a duplication of information that the validator would need to check for consistency and software would need to know how to handle in case of conflict between the 2 values. So if MatrixSize is allowed to be different than the header value it may be good to use another name and update the description. If it is meant to be the same, then it may be better to remove it.

Sorry if I am missing something obvious.

Remi-Gau:
It seems that this PR for the MRS modality introduces some metadata that could be reused by several if not all the datatypes under the MRI modality.

It may actually be a good idea to extract those from this PR and add them separately to the MRI common metadata in another PR.

This would: 
- help make the MRS PR smaller by integrating a chunck of it early
- ensure (or at least try to maximize) metadata consistency between modalities

I can try to pick some of those metadata and prepare a draft PR.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: Removed.

src/schema/objects/metadata.yaml
markmikkelsen: Removed.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: I've updated the file to reflect the last suggestion.

markmikkelsen:
> Had a quick look. I may be missing something but...
> 
> I am wondering if there is any reason to add the matrix size as a metadata when it is already implicitly present in the nifti header?
> 
> The description of VolumeAffineMatrix and AcquisitionVoxelSize make it clear that those values can actually be different from that found in the header. But this does not seem to be the case for the MatrixSize?
> 
> If so that it may lead to a duplication of information that the validator would need to check for consistency and software would need to know how to handle in case of conflict between the 2 values. So if MatrixSize is allowed to be different than the header value it may be good to use another name and update the description. If it is meant to be the same, then it may be better to remove it.
> 
> Sorry if I am missing something obvious.

@Remi-Gau, you are correct that `MatrixSize` is implicit to the nifti header. We can certainly remove it.

markmikkelsen:
> It seems that this PR for the MRS modality introduces some metadata that could be reused by several if not all the datatypes under the MRI modality.
> 
> It may actually be a good idea to extract those from this PR and add them separately to the MRI common metadata in another PR.
> 
> This would:
> 
> * help make the MRS PR smaller by integrating a chunck of it early
> * ensure (or at least try to maximize) metadata consistency between modalities
> 
> I can try to pick some of those metadata and prepare a draft PR.

@Remi-Gau - agreed on this.

src/schema/objects/metadata.yaml
```diff
@@ -1797,6 +1875,16 @@ MRAcquisitionType:
   enum:
     - 2D
     - 3D
+MRAcquisitionType__mrs:
```
Remi-Gau, line 1878: Not sure it is worth making another metadata entry here when there is so much in common with `MRAcquisitionType`: probably better to make `MRAcquisitionType` more generic and add any extra detail (DICOM tag correspondance) when calling the macro for MRI metadata.

markmikkelsen: I did it this way to allow `1D` to be a third option for MRS only. Is there another way to do this aside from this approach?

Remi-Gau:
maintenance note: added a link to the rendered BEP in the top message of this PR.

src/schema/objects/metadata.yaml
effigies: Is this a valid value for DICOM Tag 0018, 0023 `MR Acquisition Type`? If not, where does this come from?

I would probably be okay with permitting `MRAcquisitionType` to be 1D/2D/3D in general, even though 1D doesn't make sense for MRI. We could put in a special warning in `rules.checks` if 1D is used outside the MRS context.

markmikkelsen: 1D does appear to be a [valid value](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.13.4.html#table_C.8-87). And, yes, 1D only makes sense for MRS.

wtclarke: I think 1D does make sense for MRI as well. There's even been some use of linescanning recently: https://pubmed.ncbi.nlm.nih.gov/34415208/

src/appendices/contributors.md
```diff
@@ -202,7 +202,7 @@ If you contributed to the BIDS ecosystem and your name is not listed, please add
 -   Giacomo Mazzamuto 📖🔣
 -   David McAlpine 📖
 -   Manuel Mercier 📖🤔
--   Mark Mikkelsen 📖
+-   Mark Mikkelsen 📖💻
```
Remi-Gau, line 205: TODO: update the tributors.yml instead

src/schema/objects/metadata.yaml
```diff
@@ -1847,6 +1925,16 @@ MRAcquisitionType:
...
     - $ref: objects.enums.TwoD.value
     - $ref: objects.enums.ThreeD.value
+MRAcquisitionType__mrs:
+  name: MRAcquisitionType
+  display_name: MR Acquisition Type
+  description: |
+    Type of sequence readout.
+  type: string
+  enum:
+    - 1D
```
Remi-Gau, line 1935: add an enum for 1D?

src/schema/objects/metadata.yaml
```diff
@@ -2168,6 +2270,19 @@ NonlinearGradientCorrection:
...
+  display_name: Number Of Spectral Points
+  description: |
+    The number of complex data points in each recorded transient of the detected time-domain MR
+    signal, equivalent to the number of points in a single spectrum.
+  type: integer
+NumberOfTransients:
+  name: NumberOfTransients
+  display_name: Number Of Transients
+  description: |
+    The number of single applications of the pulse sequence recorded during an MRS acquisiton.
```
Remi-Gau, line 2284:
```suggestion
    The number of single applications of the pulse sequence recorded during an MRS acquisition.
```

src/schema/objects/metadata.yaml
markmikkelsen: I think we should as 1D MRS and MRI are both possible (albeit very rarely done). And 1D is a [valid value](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.13.4.html#table_C.8-87).

src/schema/objects/entities.yaml
```diff
@@ -408,6 +419,15 @@ tracer:
     Please note that the `<label>` does not need to match the actual value of the field.
   type: string
   format: label
+volume:
+  name: voi
+  display_name: Volume of Interest
+  description: |
+    The `voi-<label>` entity can be used to distinguish acquisitions localized to different regions.
+    The label SHOULD be the name of the body region or part scanned.
+    If used, the entities `"BodyPart"` and `"BodyPartDetails"` MUST also be defined in the JSON file.
```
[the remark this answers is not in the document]
guiomar: Is there any related entity to VOI which could be used instead? I see `part` has a similar meaning, what would be the difference? Also maybe related to the field [ROI](https://bids-specification.readthedocs.io/en/stable/glossary.html#roi-enums)

src/modality-specific-files/magnetic-resonance-spectroscopy.md
```diff
@@ -0,0 +1,191 @@
...
+   "NumberOfSpectralPoints": 2048,
+   "NumberOfTransients": 64,
+   "RepetitionTime": 2,
+   "AcquisitionVoxelSize": [40, 20, 30],
+   "BodyPart": "Brain",
+   "BodyPartDetails": "Anterior cingulate cortex",
+   "ReferenceSignal": "bids::sub-01/mrs/sub-01_acq-press_ref.nii.gz",
+   "AnatomicalImage": "bids::sub-01/anat/sub-01_T1w.nii.gz"
+}
+```
```
guiomar, line 191:
```suggestion
```
### Combining MRS with anatomical MRI data
For combining MRS with anatomical MRI data, see [MRS-MRI correspondence](https://bids-specification.readthedocs.io/en/stable/appendices/cross-modality-correspondence.html#mrs-mri-correspondence) on the cross modality correspondence section of the Appendix.
```

File: src/schema/objects/metadata.yaml [hunk 10]
```diff
@@ -882,6 +921,43 @@ EchoTime__fmap:
   type: number
   unit: s
   exclusiveMinimum: 0
+EditCondition:
```
guiomar, line 924: Are there any other fields where EditCondition, etc may be necessary in the future in a different context? Otherwise this could be named `SpectralEditCondition`, but if you don't forsee any, good to keep it like this.

src/schema/objects/metadata.yaml
```diff
@@ -3563,6 +3799,23 @@ VascularCrushingVENC:
       items:
         type: number
         unit: cm/s
+VolumeAffineMatrix:
+  name: VolumeAffineMatrix
```
guiomar, line 3803: Could be this somehow related with the [ChunkTransformationMatrix](https://bids-specification.readthedocs.io/en/stable/glossary.html#chunktransformationmatrix-metadata)?

guiomar (commented):
Looks quite good to me :) I've left a few comments and suggestions. 
Thanks for all the work!

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,97 @@
...
+      level_addendum: recommended for SVS and unloc
+    RepetitionTime__mrs: recommended
+    MixingTime: recommended
+    InversionTime: recommended for inversion recovery data
+    FlipAngle: recommended
+    AcquisitionVoxelSize: recommended
+    MatrixSize: recommended for MRSI
+    VolumeAffineMatrix: recommended for MRSI
+    BodyPart: recommended
+    BodyPartDetails: recommended
```
guiomar, line 64:
```suggestion
    BodyPartDetails: recommended
    BodyPartDetailsOntology: recommended
```

src/modality-specific-files/magnetic-resonance-spectroscopy.md
```diff
@@ -0,0 +1,191 @@
...
+| ---------------------------------------- | ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Single-voxel spectroscopy                | svs          | MRS acquisitions where the detected MR signal is spatially localized to a single volume.                                                                                              |
+| Magnetic resonance spectroscopic imaging | mrsi         | MRS acquisitions where additional imaging gradients are used to detect the MR signal from 1, 2, or 3 spatial dimensions.                                                              |
+| Unlocalized spectroscopy                 | unloc        | MRS acquisitions run without localization. This includes signals detected using coil sensitivity only.
+|
+| Concentration or calibration reference   | ref          | A separate MRS acquisition acquired to detect a signal to serve as a concentration reference for absolute quantification or for preprocessing (for example, eddy-current correction).
```
effigies, line 70: I think `ref` is too general a term to want it to go bare. Would `refspec` or `specref` for reference spectroscopy or spectroscopy reference work? The comparison I'm thinking of here is "single-band reference" (`sbref`) to accompany multiband BOLD or DWI sequences.

In those cases, `sbref` generally matches `bold` or `dwi` images in all but suffix. Is that expected here, or would one `ref` be good for the entire session?

src/schema/objects/metadata.yaml
markmikkelsen: I've resolved this by simply adding `"1D"` as a possible value for `MR Acquisition Type` in [ff370d7](https://github.com/bids-standard/bids-specification/pull/1377/commits/ff370d7c8df59b2e8a3691b8341069eafc229077).

markmikkelsen:
@guiomar I've amended the PR based on most of your suggestions (some I still need to think about) - thanks so much for reviewing!

src/modality-specific-files/magnetic-resonance-spectroscopy.md
markmikkelsen: How often is a single-band reference image acquired when MRI users perform multi-band MRI? Because acquiring a reference MRS acquisition is pretty much standard in MRS. I'm not sure if we should have to conform to MRI when it's far more niche to MRI than MRS.

markmikkelsen: I've thought about this, and you're right: I suggest we use `specref`.

src/appendices/contributors.md
```diff
@@ -246,7 +246,7 @@ If you contributed to the BIDS ecosystem and your name is not listed, please add
 | Marco Castellaro                                     | 💬🐛💻📖💡⚠️📢🚇                       |
 | Maria de la Iglesia                                  | 📖                                     |
 | Marie-Hélène Bourget                                 | 📖🔣💻🤔                               |
-| Mark Mikkelsen                                       | 📖                                     |
+| Mark Mikkelsen                                       | 📖💻                                     |
```
Remi-Gau, line 249: @markmikkelsen 
for updating your contribution info, you must do it through the wiki: https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors#updating-your-info

the listing of contributors in this file is automatically generated and the maintainers take care of updating it before a new release

markmikkelsen: Noted! Thanks for the heads up.

src/modality-specific-files/magnetic-resonance-spectroscopy.md
```diff
@@ -0,0 +1,191 @@
...
+| Spin-echo spectroscopy             | spinecho    | An MRS experiment whereby the MR signal is detected using a spin-echo acquisition: for example, 90°–180°–acq.                                                                                                                                                                                                                                                                                                                                                                                           |
+
+Each `<label>` in the table MAY be combined with another to better describe the acquisition used.
+For example, `megaspecial`, `jpress`, `dwslaser`, and so on.
+
+The OPTIONAL `nuc-<label>` key/value pair can be used to distinguish acquisitions tuned to detect different nuclei.
```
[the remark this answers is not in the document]
markmikkelsen: I've not added this edit as this is a new extension, and new entities will necessarily be expected, and there's no need to mention that they are new, in my opinion.

src/schema/objects/entities.yaml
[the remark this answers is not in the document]
markmikkelsen: `part` is specific to the component of the complex signal of MR data, (e.g., real, imaginary, phase, magnitude), thus, not relevant here. 
As BIDS has defined it, `ROI` is specific to a (derivative) mask that is used when analyzing fMRI data. So, again, also not appropriate here.

src/modality-specific-files/magnetic-resonance-spectroscopy.md
markmikkelsen: Added this just now ([873fd79](https://github.com/bids-standard/bids-specification/pull/1377/commits/873fd79846278d2218d56ee6e668312f8262b2aa))

src/schema/objects/metadata.yaml
markmikkelsen: "Editing" is equivalent to "spectral editing", which is used for fat suppression in MRI. So this might need to be harmonized with the `WaterSuppression` and `OuterVolumeSuppression` metadata fields we've already proposed.

src/schema/objects/metadata.yaml
markmikkelsen: I don't believe this equivalent to what VolumeAffineMatrix is. @wtclarke could you weigh in here?

src/modality-specific-files/magnetic-resonance-spectroscopy.md
markmikkelsen: Thinking back on this: `specref` could eventually (potentially) be confused with SPECT (assuming some group decides to BIDSify SPECT - which someone probably should...).

I guess `mrsref` would make more sense.

src/modality-specific-files/magnetic-resonance-spectroscopy.md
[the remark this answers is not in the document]
markmikkelsen: This is a BEP. Is it necessary to explicitly state that we have new key/value pairs?

src/schema/objects/metadata.yaml
wtclarke: Fundamentally I think that the two encode the same thing. However, `Chunk` is not a term that anyone uses in MRS, so it is going to confuse users as to its purpose. Instead this gives an explicit description of what it's used for and relates it to a real-world user set parameter `VOI`.

src/modality-specific-files/magnetic-resonance-spectroscopy.md
markmikkelsen: Addressed in [248de47](https://github.com/bids-standard/bids-specification/pull/1377/commits/248de47ec238325a784df3ecfda2a3d053f3b8a1)

effigies:
Assigning myself to remind me to go through with a fine-toothed comb next week. I think this is basically ready for community review, so I want to make sure we've got all our ducks in a row. We may need to set up a last meeting with maintainers/steering before review, but I need to remind myself of that process and any changes that may have been made to it since we started this.

Thanks very much, @markmikkelsen, for your endurance on this!

src/schema/objects/metadata.yaml
markmikkelsen: DICOM doesn't have a specific tag for spectral editing, so I think what we have is suitable.

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+    - modality == "mrs"
+  fields:
+    PulseSequenceType:
+      level: recommended
+      description_addendum: |
+        For example, `"sLASER"`, `"MEGA-PRESS"`, `"EPSI"`, `"Metabolite-cycled MRSI"`.
+    ScanningSequence__mrs: recommended
+    SequenceName: recommended
+    PulseSequenceDetails: recommended
+    MRAcquisitionType: recommended for MRSI
```
effigies, line 48: I think this would be:

```suggestion
    MRAcquisitionType: optional, but recommended for MRSI
```

Then we should add an additional table:

```
MRSIRecommendedFields:
  selectors:
    - modality == "mrs"
    - suffix == "mrsi"
  fields:
    MRAcquisitionType: recommended
```

We can add the other ones from MRSRecommendedFields, as well.

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+    EchoTime: required
+
+MRSRecommendedFields:
+  selectors:
+    - modality == "mrs"
+  fields:
+    NumberOfSpectralPoints: recommended
+    NumberOfTransients:
+      level: optional
+      level_addendum: recommended for SVS and unloc
```
effigies, line 71: This should get an extra table to promote to recommended for these suffixes. You can use `intersects([suffix], ["svs", "unloc"])` as a selector.

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+    RepetitionTime__mrs: recommended
+    MixingTime: recommended
+    InversionTime: recommended for inversion recovery data
+    FlipAngle: recommended
+    AcquisitionVoxelSize: recommended
+    MatrixSize: recommended for MRSI
+    VolumeAffineMatrix: recommended for MRSI
+    EditTarget: recommended for edited data
+    EditPulse: recommended for edited data
+    EditCondition: recommended for edited data
```
effigies, line 81: Is there a way to identify edited data apart from the presence of these fields?

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+    EncodingTechnique: recommended for MRSI
+    ReferenceSignal: recommended if MRS reference data are present
+    AnatomicalImage: recommended if anatomical MRI data are present
```
effigies, line 84: This is a fun one:

```suggestion
    AnatomicalImage: optional, but recommended if anatomical MRI data are present
```

\+

```yaml
MRSAnatomicalImageRec:
  selectors:
    - modality == "mrs"
    - '"anat" in dataset.datatypes'
  fields:
    AnatomicalImage: recommended
```

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+    InversionTime: recommended for inversion recovery data
+    FlipAngle: recommended
+    AcquisitionVoxelSize: recommended
+    MatrixSize: recommended for MRSI
+    VolumeAffineMatrix: recommended for MRSI
+    EditTarget: recommended for edited data
+    EditPulse: recommended for edited data
+    EditCondition: recommended for edited data
+    EncodingTechnique: recommended for MRSI
+    ReferenceSignal: recommended if MRS reference data are present
```
effigies, line 83: Q: Would there be a single reference signal for many `svs/mrsi/unloc` scans? If not, possibly this could be left as an association, where `mrsref` with all the same entities applies?

I'm trying to think how to validate this, and checking for all of the suffixes present in the dataset is a bit more of a pain than checking for all of the datatypes or modalities.

src/schema/rules/sidecars/mrs.yaml
```diff
@@ -0,0 +1,98 @@
...
+  selectors:
+    - modality == "mrs"
+  fields:
+    NumberOfSpectralPoints: recommended
+    NumberOfTransients:
+      level: optional
+      level_addendum: recommended for SVS and unloc
+    RepetitionTime__mrs: recommended
+    MixingTime: recommended
+    InversionTime: recommended for inversion recovery data
```
effigies, line 74: Is this identifiable otherwise?

effigies (commented):
I think this is in great shape!

I have a number of minor suggestions and a few questions for you.

Thinking about validation now:

* Are there any metadata fields that are duplicated from the NIfTI header or the NIfTI-MRS header extension? Any duplication should be explicitly listed so that validators have a clear target. If the duplication is not exact but one is a transform of another or places a constraint, we should know those, too.
* Do any of the suffixes place constraints on the contents of the NIfTI files? For example I think `svs` should have a shape 1x1x1? Which suffixes can have a time dimension? Can any have a dimension >4? For MRSI, must `MatrixSize == nifti_header.dim[1:4]`?

src/schema/objects/metadata.yaml
```diff
@@ -844,6 +870,13 @@ EOGChannelCount:
     Number of EOG channels.
   type: integer
   minimum: 0
+EchoAcquisition:
+  name: EchoAcquisition
+  display_name: Echo Acquisition
+  description: |
+    How the detected echo was acquired when the analog-to-digital converter was turned on.
+    For example, `"Half echo"`, `"Full echo"`.
+  type: string
```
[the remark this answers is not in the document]
markmikkelsen: A free-form string.

src/schema/objects/metadata.yaml
```diff
@@ -896,6 +923,43 @@ EchoTime__fmap:
...
+  display_name: Editing Pulse
+  description: |
+    If spectral editing was applied, this details the editing parameters.
+    For example:
+    `{"ON": {"FrequencyOffset": 1.9, "PulseDuration": 16},
+    "OFF": {"FrequencyOffset": 7.5, "PulseDuration": 16}}`.
+  anyOf:
+    - type: object
+      additionalProperties:
+        type: object
```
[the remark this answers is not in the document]
markmikkelsen: Those two fields are the most relevant for describing an editing pulse's parameters for the sake of analysis.

Note that this example is for a common two-step editing experiment. But a four- or more step experiment is also a used approach. In addition, while the values will be numeric, they could also be an array (say when the editing pulse has a dual-band frequency). So, the following would be valid:

`{"A": {"FrequencyOffset": [1.9, 4.56], "PulseDuration": 20}, "B": {"FrequencyOffset": 1.9, "PulseDuration": 20}, "C": {"FrequencyOffset": 4.56, "PulseDuration": 20}, "D": {"FrequencyOffset": 7.5, "PulseDuration": 20}}`

src/schema/objects/metadata.yaml
```diff
@@ -2591,6 +2711,16 @@ ReceiveCoilName:
...
+ReceiveGain:
+  name: ReceiveGain
+  display_name: Receive Gain
+  description: |
+    The gain of the receive coil.
+  anyOf:
+    - type: number
+    - type: array
+      items:
+        type: number
```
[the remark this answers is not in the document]
markmikkelsen: I'm not very versed in RF coil electronics, but dB seems to be correct.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: I'm slightly unclear what you mean by this? Could you elaborate?

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: In the NIfTI-MRS format, edited data can be identified by dimension 7 of the data block, which encodes subspectra.

CC'ing @wtclarke here.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: Do you mean for a subject that has, in theory, `svs`, `mrsi`, and `unloc` scans, would a single reference apply to them all? If so, no, because the reference has to be specific to the acquisition. So, each of the three would have to have their own references.

If you mean instead if, say, a subject has repeat `svs` (or `mrsi/unloc`) scans (e.g., two scans for test-retest), would a single reference apply to them both? Then, that is more of a study design decision. Personally, I would acquire a separate reference signal for each individual `svs/mrsi/unloc` scan I'd acquire.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: Oh, hold on, I've looked up examples in the other sidecars. I know what you mean.

src/schema/rules/sidecars/mrs.yaml
wtclarke: Not quite. The 5th to 7th dimensions of NIfTI-MRS are flexible in their use. For example, they could be used to store uncombined coils, averages, or editing cases. The use of a dimension is encoded in the `dim_{5,6,7}` keys of the header extension, using a range of pre-defined tags (`DIM_COIL`, `DIM_DYN`, `DIM_EDIT` for the examples of the previous sentence). The order is flexible.

However, this identification process assumes that the data hasn't been processed. For the most common case (MEGA editing), the final output of processing is a single spectrum (constructed from multiple editing cases). This means that the `DIM_EDIT` dimension will have been removed.

effigies: Okay. These are probably hard to make recommended in the sense that their absence should display a warning. I would suggest just making them optional, as RECOMMENDED would have no force.

If it would be helpful, we could check whether the MRS extension fields are set to DIM_EDIT and warn if the `Edit*` metadata is missing, but if that won't cover all cases of edited data, it's still not generally enforceable.

src/schema/rules/sidecars/mrs.yaml
effigies: Okay, so it may frequently be, but is not always, 1-1. I'll mull on this a little.

src/schema/rules/sidecars/mrs.yaml
effigies: Just `InversionTime`. When we have conditional "RECOMMENDED" status, we need to think about how to identify when it should be recommended and when not. For example, here we might use the presence of the `inv-<label>` entity to indicate that `InversionTime` should be in the sidecar?

src/schema/objects/metadata.yaml
```diff
@@ -896,6 +927,43 @@ EchoTime__fmap:
...
+  display_name: Editing Pulse
+  description: |
+    If spectral editing was applied, this details the editing parameters.
+    For example:
+    `{"ON": {"FrequencyOffset": 1.9, "PulseDuration": 16},
+    "OFF": {"FrequencyOffset": 7.5, "PulseDuration": 16}}`.
+  anyOf:
+    - type: object
+      additionalProperties:
+        type: object
```
effigies, line 953: Here's a proposal that I believe should match the examples you have here. Please check whether you would want the fields to be required or recommended, and if I got the units correct.

```suggestion
  type: object
  additionalProperties:
    type: object
    recommended_fields: [FrequencyOffset, PulseDuration]
    properties:
      FrequencyOffset:
        anyOf:
          - type: number
            unit: MHz
          - type: array
            items:
              type: number
              unit: MHz
      PulseDuration:
        type: number
        unit: s
```

@rwblair could you check that my syntax is good?

markmikkelsen: For the units: `FrequencyOffset` should be `ppm` and `PulseDuration` should be `ms`.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: Agreed that these are best set as optional.

markmikkelsen:
Hi @effigies (and cc'ing @wtclarke),

Responding to your two previous queries:
 
> * Are there any metadata fields that are duplicated from the NIfTI header or the NIfTI-MRS header extension? Any duplication should be explicitly listed so that validators have a clear target. If the duplication is not exact but one is a transform of another or places a constraint, we should know those, too.

Metadata fields stored in the NIfTI-MRS header extension will depend on what can be extracted from the source dataset's header. (@wtclarke, could you confirm?). A number of them match the current MRS BEP metadata fields; others are not included in the BEP/BIDS. (Although, it would appear conversion software, like MRIcroGL, may also create non-BIDS metadata fields in their exported BIDSy-fied .json file.)
 
> * Do any of the suffixes place constraints on the contents of the NIfTI files? For example I think `svs` should have a shape 1x1x1?

Yes, `svs` data will have a shape of 1x1x1. A 2D `mrsi` dataset's size could be something like 32x32x1. And a 3D `mrsi` dataset's size could be 64x64x64.

> Which suffixes can have a time dimension?

If you mean the equivalent to, say, the number of volumes in a BOLD-EPI scan, then, yes. This will be stored in one of dimensions 5-7. 

> Can any have a dimension >4?

Yes, dimensions >4 are possible, e.g., for dynamic repeats, uncombined RF channels, or subspectra (such as for edited data). But this will depend on the "rawness" of the source dataset that was converted.

> For MRSI, must `MatrixSize == nifti_header.dim[1:4]`?

For MRSI, if you mean must `MatrixSize == nifti_header.dim[2:4]` (MATLAB indexing syntax), then yes(?).

wtclarke:
Yes, the exact content of the NIfTI-MRS header extension is dependant on input file type. Some file types that vendors output contain lots of information that can be interpreted, others the bare minimum. It also depends on what I can reliably interpret from the headers, as I am much less familiar with some vendor's formats. According to the NIfTI-MRS standard there is very little that is required, only SpectrometerFrequency and ResonantNucleus, and then any tags for higher dimensions (5th - 7th) that are in use.

> Which suffixes can have a time dimension?

@Mark is getting at the fact that all NIfTI-MRS files have a time dimension (in the 4th NIfTI dimension). This (broadly) stores the time points of one FID/echo. A higher dimension (5-7) can then be labelled `DIM_DYN` and be used to store repeated acquisitions.

src/schema/objects/metadata.yaml
```diff
@@ -2702,6 +2844,22 @@ RepetitionTime:
   type: number
   exclusiveMinimum: 0
   unit: s
+RepetitionTime__mrs:
+  name: RepetitionTime
+  display_name: Repetition Time
+  description: |
+    The time between repetitions (TR), specified in seconds.
+    For dynamic acquisitions, such as cardiac-triggered acquisitions where the TR
+    may differ over the scan, an array of numbers can be used.
```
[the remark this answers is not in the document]
markmikkelsen: This would correspond to `RepetitionTimeExcitation`, [DICOM Tag 0018, 0080](https://dicom.innolitics.com/ciods/mr-image/mr-image/00180080). But `RepetitionTime` seems to make more sense. I've done it this way to permit arrays of numbers (like in the case of cardiac triggering). But that said, `VolumeTiming` would also technically fit the case of such dynamic TRs.

src/schema/objects/metadata.yaml
effigies: Updated:

```suggestion
  type: object
  additionalProperties:
    type: object
    recommended_fields: [FrequencyOffset, PulseDuration]
    properties:
      FrequencyOffset:
        anyOf:
          - type: number
            unit: ppm
          - type: array
            items:
              type: number
              unit: ppm
      PulseDuration:
        type: number
        unit: ms
```

src/schema/rules/sidecars/mrs.yaml
effigies: I have not been able to think of a way in the current schema to make this reliably produce a warning when there are `mrsref` present but not when there are none. As written, it will simply always warn, which would be no more than annoying to someone who didn't collect reference scans.

I think this is fine.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: I've set these fields as `optional`.

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: The only way I can think of is if the user adds the `inv-<label>` entity to the filename.

markmikkelsen:
@effigies Just updated `mrs.yaml` and `metadata.yaml`. Could you please check if it's what you had recommended?

src/schema/rules/sidecars/mrs.yaml
markmikkelsen: I'm not sure how that suggested change and that extra code work together with respect to what appears in `src/modality-specific-files/magnetic-resonance-spectroscopy.html`. That is, I've included:

```
MRSRecommendedFieldsAnatomicalImage:
  selectors:
    - modality == "mrs"
    - '"anat" in dataset.datatypes'
  fields:
    AnatomicalImage: recommended
```

and:

```
{{ MACROS___make_sidecar_table([
  "mrs.MRSRecommendedFields",
  "mrs.MRSRecommendedFieldsNumTransients",
  "mrs.MRSRecommendedFieldsAnatomicalImage",
  "mrs.MRSIRecommendedFields"
  ])
}}
```

But now `AnatomicalImage` is shown as `RECOMMENDED` rather than `OPTIONAL, but RECOMMENDED if anatomical MRI data are present`.

effigies:
Hi @markmikkelsen. Been organizing (and now attending thru Friday) a workshop for the last couple weeks, so apologies for being unable to respond in a timely manner. Will get back to this early next week. I would like to get this to the community review stage very soon. Would you like to have a call to sync up next week?

markmikkelsen:
Hi @effigies, sounds good! I'll only be available next week Monday or Tuesday. Feel free to email me to set up a time.

src/schema/objects/metadata.yaml
```diff
@@ -2287,6 +2397,12 @@ OtherAcquisitionParameters:
   description: |
     Description of other relevant image acquisition parameters.
   type: string
+OuterVolumeSuppression:
+  name: OuterVolumeSuppression
+  display_name: Other-Volume Suppression
```
markmikkelsen, line 2402: This needs to be corrected to `Outer-Volume Suppression`

effigies:
Notes from call:

1) We need to increase the requirement level for BodyPart+BodyPartDetails if the `voi` entity is present. You can use `entities.volume` as a selector.
2) We need a check that ResonantNucleus and SpectrometerFrequency (both required fields) are defined in the MRS header extension. I will plan to submit a PR for this on Thursday afternoon.
3) Please review the descriptions for EchoTime and FlipAngle and make sure that the wording makes sense in an MRS context.
4) Set InversionTime to recommended if `inv` entity is present. (`entities.inversion`)
5) Move to RepetitionTime and VolumeTiming, rather than RepetitionTime as an array in the variable-timing case.
6) Add check that `sidecar.MatrixSize == nifti_header.dim[1:4]`, as in
    https://github.com/bids-standard/bids-specification/blob/181c50e08fd143d481e22be9a15c31f612e6f906/src/schema/rules/checks/func.yaml#L32-L43
7) ReferenceSignal should be uniformly recommended.
8) Examples should be updated to use `ReferenceSignal` and `AnatomicalImage` where appropriate.
9) The fmrs example needs sidecar files.

Note for (6) that the expression language doesn't have slices, so that would be `sidecar.MatrixSize[0] == nifti_header.dim[1] && ...`.

src/schema/objects/metadata.yaml
```diff
@@ -2757,6 +2936,19 @@ Resolution:
...
+  description: |
+    The isotope of interest of an MR experiment (for example, `"1H"`, `"13C"`, `"31P"`).
+    For multi-nuclei experiments such as <sup>1</sup>H-[<sup>13</sup>C] MR,
+    an array can be used: `["1H", "13C"]`.
+    Corresponds to DICOM Tag 0018, 9100 `Resonant Nucleus`.
+  anyOf:
+    - type: string
+    - type: array
+      items:
+        type: string
```
effigies, line 2951: Looking through the MRS standard, this field is always an array. I suspect more people would value consistency between the MRS header and the BIDS sidecar than the ability to drop a pair of brackets:

```suggestion
  type: array
  items:
    type: string
```

src/schema/objects/metadata.yaml
```diff
@@ -3287,6 +3489,30 @@ SpecificRadioactivityUnits:
...
+    to the resonant frequency of <sup>1</sup>H.
+    For multi-nuclei experiments such as <sup>1</sup>H-[<sup>13</sup>C] MR at 3T,
+    an array can be used: `[127.731, 32.125]`.
+  anyOf:
+    - type: number
+      unit: MHz
+    - type: array
+      items:
+        type: number
+        unit: MHz
```
effigies, line 3515: Likewise:

```suggestion
  type: array
  items:
    type: number
    unit: MHz
```

markmikkelsen:
@effigies I've done another look over the BEP and, yes, I believe it's ready for review.

effigies:
@markmikkelsen I've implemented all of the above suggestions (except 3, 8 and 9) in https://github.com/bids-standard/bids-specification/pull/1830. Please review that and merge if you're happy with the changes.

src/schema/objects/metadata.yaml
markmikkelsen: Good point!

src/schema/objects/metadata.yaml
markmikkelsen: Corrected in [2c969bb](https://github.com/bids-standard/bids-specification/pull/1377/commits/2c969bb35e448f475d856f8dbf9fff505d67172e)

ctr:
Hi @wtclarke  and others. I hope this is the right place to comment.

1. Would you consider adding a "methodDoi" header to link to the DOI for the main paper describing the acquisition approach used? This may be a helpful historical record to bake into MRS data as there are so many one-off special methods. It could then aid an understanding of how the other parameters are (ab)used for new sequences beyond what is anticipated today.

2. Can the JSON note whether the MRS spatial coordinates are defined with or without distortion correction? I presume normally it's without distortion correction but not sure that is always the case. This will otherwise be a source of mis-alignment.

3. Does the specification offer guidance as to how to complete the Vendor, etc labels in cases of simulated data?

4. Does it provide a space to specify samples that should be dropped before analysis, e.g. initial points before a spin echo or initial points that are known to be corrupted during acquisition?

Thanks

Chris Rodgers.

src/schema/rules/entities.yaml
```diff
@@ -7,6 +7,8 @@
 - task
 - tracksys
 - acquisition
+- nucleus
+- volume
```
effigies, line 11: What order should these be in? I think `nucleus` makes sense to group with `ceagent` and `tracer`, but `volume` doesn't seem to group here, intuitively. Possibly with `reconstruction` or `chunk`?

markmikkelsen: Hi @effigies, in my opinion, `volume ` should be grouped with `sample`. `nucleus` should be grouped with `acq`.

tsalo: Placing `nucleus` between `acquisition` and `ceagent`/`tracer` sounds good to me. It seems like `volume` could fit well next to either `sample` or `chunk` (honestly, it seems like `sample` and `chunk` should be next to each other). I agree that `sample`'s a closer fit than `chunk` though.

src/schema/rules/sidecars/func.yaml
```diff
@@ -41,6 +41,10 @@ MRIFuncVolumeTiming:
     VolumeTiming:
       level: required
       level_addendum: mutually exclusive with `RepetitionTime`
+      description_addendum: |
+        This field is mutually exclusive with `"DelayTime"`.
+        If defined, this requires acquisition time (TA) be defined via either
+        `"SliceTiming"` or `"AcquisitionDuration"` be defined.
```
tsalo, line 47:
```suggestion
        If defined, this requires acquisition time (TA) be defined via either
        `"SliceTiming"` or `"AcquisitionDuration"`.
```

src/schema/rules/entities.yaml
effigies: Sample is considered on par with subject and session (the microscopy BEP would have made sample first and subject optional, if it were feasible). I'm not sure that it holds the same place of priority here, but you may disagree.

Anyway, if we insert `volume` after `sample`, then the ordering for MRS would be:

```
sub-<label>[_ses-<label>][_voi-<label>][_task-<label>][_acq-<label>][_nuc-<label>][_rec-<label>][_run-<index>][_echo-<index>][_inv-<label>]_<suffix>.nii.gz
```

Lexically sorting file names, entities to the left will tend to indicate a bigger change than those to the right. Does this match your expectations when looking at a study?

tsalo: That's a really good point. Maybe it should go near chunk instead.

markmikkelsen: Yeah, `voi` being there doesn't seem right to me either. I agree with @tsalo; I guess it belongs closer to `chunk`.

effigies: Do you think before or after run? Do you think of it as doing multiple runs of multiple VOIs (`voi-<label>_run-<index>`) or multiple VOIs for each run (`run-<index>_voi-<label>`)?

markmikkelsen: I would say the former: `voi-<label>_run-<index>`. In my opinion, the VOI (particularly for single-voxel spectroscopy [`svs`]), specifying `voi` is hugely important and should take precedence in file-naming.

effigies: Okay. Then the current ordering is fine.

Remi-Gau:
- I think this will close https://github.com/bids-standard/bids-specification/issues/680, no?

src/modality-specific-files/magnetic-resonance-spectroscopy.md
```diff
@@ -0,0 +1,260 @@
...
+| Spin-echo spectroscopy                      | spinecho    | An MRS experiment whereby the MR signal is detected using a spin-echo acquisition: 90°–180°–acq.                                                                                                                                                                                                                                                                                                                                                                                                                  |
+
+Each `<label>` in the table above MAY be combined with another to better describe the acquisition used.
+For example, `megaspecial`, `jpress`, `dwslaser`, `mcdwsteam`, and so on.
+
```
bpinsard, line 123: In [examples](https://github.com/bids-standard/bids-examples/pull/425) a number is sometimes suffixed to the label (eg. megapress80/64). If that is part of the spec, I think it should be explained here.

markmikkelsen: Hi @bpinsard. The `acq` label can take any form that a user wishes (see: https://bids-specification.readthedocs.io/en/bep022/appendices/entities.html#acq). The examples are just that: examples. Affixing a number is possible if it distinguishes acquisitions appropriately.

bpinsard: Oh right! I thought it had a meaning like number of spectrum acquired or else.
I am not really versed in MRS and was mainly trying to understand how to BIDS-format some CMRR-megapress data. 
Just found in the README of that dataset:
> - TE/TR = 68 (for standard GABA editing) or 80 (for symmetric macromolecule suppression)/2000 ms

Using `acq-`  to store MR parameters seems a bit of a BIDS anti-pattern, but I'm not an MRS expert.
Anyway, thanks for the explanation!

effigies (approved):

PR #1385, opened 2023-01-19
SCHEMA: Add rules for checking existence of referenced files, non-existence of duplicate files

effigies:
This PR introduces a new function `exists()` that operates on a URI/path or list of URI/paths


src/schema/meta/expression_tests.yaml
```diff
@@ -45,6 +45,10 @@
   result: false
 - expression: '"VolumeTiming" in null'
   result: false
+- expression: exists(null, "bids-uri")
+  result: false
+- expression: exists([], null)
+  result: false
```
effigies, line 51: I assume we want these to be `false` for consistency, but `exists([], *)` seems like it could safely evaluate true, and so `exists(null, *)` could also evaluate true. It's a degenerate `and()`/`product()`.

bendhouseart (dismissed):

rwblair (approved):

src/schema/README.md
```diff
@@ -246,21 +250,30 @@ This value propagates through all of the above operations in a fully-defined,
...
+| `!null`                    | `null` |
+| `null + 1`                 | `null` |
+| `null - 1`                 | `null` |
+| `null * 1`                 | `null` |
+| `null / 1`                 | `null` |
+| `match(null, pattern)`     | `null` |
+| `intersects(list, null)`   | `null` |
+| `substr(null, 0, 1)        | `null` |
+| `substr(str, null, 1)      | `null` |
+| `substr(str, 0, null)      | `null` |
```
tsalo, line 269:
```suggestion
| `substr(null, 0, 1)`        | `null` |
| `substr(str, null, 1)`      | `null` |
| `substr(str, 0, null)`      | `null` |
```

bendhouseart (approved):
Comment, good.

PR #1386, opened 2023-01-20
SCHEMA: Add DuplicateFiles rule that must be implemented by each validator

effigies:
Converts issue 74 (`DUPLICATE_NIFTI_FILES`) to generic `DUPLICATE_FILES` error if the same file exists both compressed and uncompressed.

If we add a slicing operator (or substring function), we could write the check in schema:

```YAML
selectors:
  - match(extension, '\.gz$')
check:
  - path[:length(path) - 3] in dataset.files
```

### DIFF
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -214,30 +214,33 @@ These are to be populated by an *interpreter* of the schema, and provide the
 
 The following operators should be defined by an interpreter:
 
-| Operator  | Definition                                                    | Example                                       |
-| --------- | ------------------------------------------------------------- | --------------------------------------------- |
-| `==`      | equality                                                      | `suffix == "T1w"`                             |
-| `!=`      | inequality                                                    | `entities.task != "rest"`                     |
-| `<`/`>`   | less-than / greater-than                                      | `sidecar.EchoTime < 0.5`                      |
-| `<=`/`>=` | less-than-or-equal / greater-than-or-equal                    | `0 <= 4`                                      |
-| `in`      | object lookup, true if RHS is a subfield of LHS               | `"Units" in sidecar`                          |
-| `!`       | negation, true if the following value is false, or vice versa | `!true == false`                              |
-| `&&`      | conjunction, true if both RHS and LHS are true                | `"Units" in sidecar && sidecar.Units == "mm"` |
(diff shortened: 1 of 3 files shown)


effigies:
Alternately:

```YAML
selectors:
  - ! match(extension, '\.gz$')
check:
  - (path + ".gz") in dataset.files
```

We do have checks that require `+` for int, which needs to be added to the schema README. If we permit it for strings, then we don't need to add anything else.

rwblair:
If #1385 is merged it could read:
```
selectors:
  - match(extension, '\.gz$')
check:
  - exists(path[:length(path) - 3])
 ```

Javascript doesn't have square brackets with colon slice, so making it a function would be preferable. That being said it is possible to override property accessors to add a feature like this. Currently in the js validator paths in contexts are plain strings, so they'd need to be instantiated differently no idea on how that might affect performance.

Triggering the check on '.gz' existing vs it not existing would be more performant. String concatenation is probably more generally useful in the future.

**Comment by rwblair** [comment]
Does not go down so smooth for readme, but this is a pretty low priority rule:
```
selectors:
  - substr(path, 0 , 7) == '/README'
check:
  - count([exists('/README'), exists('/README.md'), exists('/README.rtf'), exists('/README.txt')], true) == 1
```

Are we missing an easier way to handle large numbers of mutually exclusive filenames?

**Comment by effigies** [comment]
What if exists returns a count? Usually we only care about >0, but we could specifically check ==1.

**Comment by rwblair** [comment]
> What if exists returns a count? Usually we only care about >0, but we could specifically check ==1.

I like it, then should it take wildcards/regex as an arg? Would it be terrible to implicitly coerce check statements that resolve to ints to booleans? Semantics between python and js are similar:
```python
>>> "nice" if 4 else "nah"
'nice'
>>> "nice" if 0 else "nah"
'nah'
>>> "nice" if -1 else "nah"
'nice'
```
```js
> 4 ? 'nice' : 'nah'
"nice"
> 0 ? 'nice' : 'nah'
"nah"
> -1 ? 'nice' : 'nah'
"nice"
```

effigies:
I think I would be okay accepting integers as truthy values. We would just need to be explicit and add some test cases, like we do with `null`.

PR #1387, opened 2023-01-23
SCHEMA: Remove inconsistent name field from objects.common_principles

effigies:
In `objects`, we use the `name` field for terms that are the name in name-value pairs, such as column headers, entities or metadata fields. We updated `objects.common_principles` to use `display_name` but did not remove the inapplicable `name` field.


erdalkaraca (approved):
systematic change, removed redundant data

tsalo (approved):
Great catch! This change fits with the README.

It kind of makes me wish I'd figured out meta-testing in https://github.com/bids-standard/bids-specification/pull/1048. Then, we'd be able to automatically catch this kind of thing.

PR #1401, opened 2023-02-06
Add to schemacode_ci.yml a workflow to publish master version of schema

yarikoptic:
The idea is to automate population/updates to  https://github.com/bids-standard/bids-schema/tree/main/versions/master . original issue is https://github.com/bids-standard/bids-schema/issues/7 but likely we want to make workflow here to react immediately on pushes to master and releases (only when typing this thought : "why? couldn't we just make workflow triggered from pushes to master here or releases?")

Unfortunately we cannot use this workflow for publishing on releases since it reacts to schema- tags and not v* tags and anyways not taking care about actual releases of the bids-specification itself.  So we would either need to move this into release workflow (is it the circleci one?) or just duplicate this directly in  bids-schema  making it to run there on cron and checking version tags.

Also probably I need some custom to that repo CHANGE_TOKEN to be able to push to it.

But before anything else, may be @effigies @sappelhoff or @tsalo have better ideas on how to wrap it all up?  May be indeed just keeping it in `bids-schema` but having it  triggered from here within `bids-schema` repo?

### DIFF
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -125,6 +125,43 @@ jobs:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
 
+  publish-schema:
+    runs-on: ${{ matrix.os }}
+    name: Publish to dandi-schema (on master)
+    needs: [ publish ]
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: "Checkout dandi-schema"
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          repository: bids-standard/bids-schema
+
+      - name: Set Up Python environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+
+      - name: Install the schemacode package for bst tool used by inject-schema
+        run: |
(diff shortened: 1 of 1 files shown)


effigies:
I think a daily job on `bids-schema` would be cleaner.

yarikoptic:
ok, I will move there I guess. My  dislikes of such cron job approach 
- is is that they "flood" the list of jobs and then it is harder to locate the one of interest (e.g. passed and did something useful not just skipped all the steps after initial step). E.g. look at https://github.com/datalad/git-annex/actions/workflows/update-mirror.yml and find the last one which actually updated the mirror ;)  
- we might miss some `master` state... but I guess it is ok
- we also would likely not account for changes to jsonification until `bst` release (but that could be taken as feature)

yarikoptic:
might get back to it later

yarikoptic:
FTR: implemented now there as a cron job.  May be we should look into adding triggering of that workflow somehow from here upon merge, but I no urgent need.  It will also keep updating master version of the schema as new ones come in. Details are in

- https://github.com/bids-standard/bids-schema/blob/main/tools/inject-schema-fully-auto
- https://github.com/bids-standard/bids-schema/blob/main/.github/workflows/inject.yml to run it

PR #1404, opened 2023-02-07
[WIP] [INFRA] add workflow for precommit autoupdate

Remi-Gau:


effigies:
Wouldn't it be better just to add https://pre-commit.ci/ than our own equivalent?

Remi-Gau:
I would tend to say yes.

But this comes at the cost of having pre commit ci adding commits into pr which may confuse users:

>why do I have to pull before pushing again eventhough I am the only one working on this at the moment?

Best to discuss at our next maintenance meeting.

Remi-Gau:
It was decided to let pre-commit ci handle this

PR #1405, opened 2023-02-07
SCHEMA: Add sorted() and index() functions, update expression tests

effigies:
To enable some check implementations, we need sorted() and index(). Definitions and tests are provided. Checks will come in a separate PR.


bendhouseart (approved):
Sure.

src/schema/meta/expression_tests.yaml
```diff
@@ -77,15 +84,61 @@
...
+- expression: count([1, 2, 3], 1)
+  result: 1
+- expression: index(["i", "j", "k"], "i")
+  result: 0
+- expression: index(["i", "j", "k"], "j")
+  result: 1
+- expression: index(["i", "j", "k"], "x")
+  result: null
+- expression: sorted([3, 2, 1])
+  result: [1, 2, 3]
```
tsalo, line 136: Are there any non-numeric cases for sorted in the spec?

effigies: The only use case we've found is for checking that certain arrays are monotonically increasing.

tsalo: awesome!

tsalo (approved):

PR #1406, opened 2023-02-07
SCHEMA: Add convenience objects to nifti_header context

effigies:
Some existing checks assumed that the NIfTI header had been interpreted, and not simply exposed as fields. This adds `shape == dim[1:dim[0] + 1]` and `voxel_sizes == pixdim[1:dim[0] + 1]`, which can range in length from 0 to 7.

This permits us to do things like check `min(shape) > 0`, which would be more inconvenient to do directly on `dim`.

### DIFF
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -266,6 +266,22 @@ context:
           maxItems: 8
           items:
             type: number
+        shape:
+          name: 'Data shape'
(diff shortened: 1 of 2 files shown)


bendhouseart (approved):
You close out those validator issues Chris.

tsalo (approved):
LGTM

PR #1412, opened 2023-02-09
ENH: Add expression language parser

effigies:
Adding a Python parser for the schema expression language

### DIFF
--- a/tools/schemacode/bidsschematools/expressions.py
+++ b/tools/schemacode/bidsschematools/expressions.py
@@ -1,0 +1,98 @@
+from pyparsing import (
+    Forward,
+    Group,
+    Literal,
+    Optional,
+    StringEnd,
+    StringStart,
+    Suppress,
+    ZeroOrMore,
+    common,
+    delimited_list,
+    one_of,
+    quoted_string,
+)
+
+# EBNF-ish grammar
+#
+# # In order of binding (loosest to tightest)
+# cmpOp :: '==' | '!=' | '<' | '<=' | '>' | '>=' | 'in'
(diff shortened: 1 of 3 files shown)


tools/schemacode/bidsschematools/expressions.py
```diff
@@ -0,0 +1,243 @@
...
+
+# Arithmetic expressions
+factor <<= atom + (expOp + factor)[...]  # Right-associative
+term = factor + (mulOp + factor)[...]
+expr = term + (addOp + term)[...]
+
+# Logic expressions (tests, to avoid name collision)
+comparison = expr + (compOp + expr)[...]
+notTest <<= notOp + notTest | comparison
+andTest = notTest + (andOp + notTest)[...]
```
effigies, line 99: Should be reverted to right associative to simplify short circuiting.

tools/schemacode/bidsschematools/expressions.py
```diff
@@ -0,0 +1,243 @@
...
+test <<= andTest + (orOp + andTest)[...]
```
effigies, line 100: And here

effigies:
@rwblair AST complete. What do you think? Should I go ahead and implement interpretation before merging, or get this in and make interpretation a new PR? And any other tests you can think of?

rwblair:
Looks great. For a real thorough test could try and parse every selector and check.

Guess go ahead an merge lest it turn into one of my validator PRs...

effigies:
Tests added. Which goes to show that parsing isn't type checking... Either way, let's get moving.

PR #1413, opened 2023-02-10
[ENH] Some missing docstrings for bidsschematools

anibalsolon:
Adding some more docstrings for the bidsschematools.

### DIFF
--- a/tools/schemacode/bidsschematools/validator.py
+++ b/tools/schemacode/bidsschematools/validator.py
@@ -514,7 +514,7 @@ def select_schema_dir(
         according to the respective schema version, e.g. "1.7.0".
         If the path starts with the string "{module_path}" it will be expanded relative to the
         module path.
-    schema_version : str or None
+    schema_version : str | None
         Version of BIDS schema, or path to schema.
         If a path is given, this will be expanded and used directly, not concatenated with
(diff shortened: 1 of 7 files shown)


bendhouseart (approved):
Looks good, approval and merge should push to rtd's.

anibalsolon:
hey @TheChymera tagging you here as well, if you'd like to take a look!

tools/schemacode/bidsschematools/validator.py
```diff
@@ -514,7 +514,7 @@ def select_schema_dir(
         according to the respective schema version, e.g. "1.7.0".
         If the path starts with the string "{module_path}" it will be expanded relative to the
         module path.
-    schema_version : str or None
+    schema_version : str | None
```
effigies, line 517: Above you went in the other direction.

effigies (commented):
Mostly good. We should pick a convention for describing types and stick with it, though.

tools/schemacode/bidsschematools/validator.py
anibalsolon: Ah missed this one- it should be an `or`. From what I saw from numpydoc that's what's used

effigies (approved):

PR #1439, opened 2023-03-15
[ENH] Clarify that BIDS specification is in American English

yarikoptic:
"Explicit is better than implicit." and the question of the spelling and syntax does come up from time to time, and ATM I observe BEPs prepared by European Colleagues are using British English in their documents.

Making it explicit would help to minimize possible tensions and required later work to prepare contributions into acceptable state.

I placed it first in the common principles since IMHO it is the shortest and most "basic" principle to agree upon before going into which words to use etc.


src/common-principles.md
```diff
@@ -1,5 +1,9 @@
 # Common principles
 
+## Language
+
+BIDS specification is written in American English.
```
sappelhoff, line 5:
```suggestion
The BIDS specification is written in American English.
```

sappelhoff (approved):

arokem (approved):

Remi-Gau (approved):

effigies:
Was this agreed by the SC? Could we get one more approval @guiomar @robertoostenveld @CPernet?

Along with #1444, it seems these could go in either the spec or the contributing guide. I'm slightly in favor of having spec-writing "principles" in the contributing guide as they have no effect on the interpretation of the spec or a dataset, but I would not attempt to block this.

CPernet:
''I observe BEPs prepared by European Colleagues are using British English in their documents.'' yeh, if that was up to me I take British anytime ... but since you guys have way more contribution and besides India, the rest of Asia uses American English I guess you win -- SC has not discuss it, but it has my vote, so just need one more

robertoostenveld (approved):
I approve.

Remi-Gau:
I have a crazy suggestion: can we have it in both? (I usually don't like duplication but I may make an exception for this one.)

yarikoptic:
I would vote strongly against `transfer` since that would invalidate the purpose of formalizing the language of the specification (as we have already for some special WORDS), and would vote "ok" for `both`.

PR #1440, opened 2023-03-15
[ENH] Remove explicitly worded requirement of events file association with the task

yarikoptic:
The discussion came up during reviewing BEP032 that association of events file with an active agent performing a task might be of a stretch for many animal experiments.

Other arguments
- we used _events suffix, not _taskevents  so to me it makes total sense to generalize _events to all events
- wording already suggests that events could be stimuli and participant responses thus offering support that events could be related to both influences over participant and behavioral responses thus mixing two domains of data. Metadata fields such as event_type remains there to possibly disentangle multiple domains

As a result we would open possibility to make record of events during non-functional runs, e.g.

- someone interested to investigate effect of various "events" which occured during acquisition of anatomical data (which can span minutes), e.g. "change of screen intensity", "instruction input", etc, which could later be used e.g. for removal of confounds etc.

This is an incomplete change to first spark discussion and further improvements

Attn @bids-standard/bep032 and other teams

Possible TODOs
- [ ] Further text editing
- [ ] Verify that we do not need to make it more explicit about the time point against which onset is relative to
- [x] `{{ MACROS___make_columns_table("task.TaskEvents") }}` suggests we might want to change schema to dissociate events from tasks at the level of the schema
- [ ] Decide on what to recommend in case of conflicts (e.g. `_T1w` and `_T2w` and `_events`) either recommend explicit additional entity (e.g. `_acq`) or implicit (`_mod`) to disambiguate (similarly to how done for `_defacemask`).


VisLab:
I would be interested in participating in the investigation of removing the association of task and events and concur with the changes proposed in this PR.

Might we also consider changing the title of the section `Task-events`  to just `Events`?

effigies:
Overall, I think removing the specificity to task seems fine. I think that for time series data (BOLD, EEG, PET, etc.), there will always be enough information in the entities to ensure a clear mapping between data file and events.

While I understand the desire to make annotations available universally, I think it will be hard to make it work with anatomical acquisitions. For example, if I have:

```
anat/
  sub-01_T1w.nii.gz
  sub-01_T2w.nii.gz
  sub-01_events.tsv
```

Which volume does the events file apply to? I suppose we could add a unique `acq-<label>` to each, but is this something that is desired enough to justify the added complication?

yarikoptic:
> Which volume does the events file apply to? I suppose we could add a unique `acq-<label>` to each, but is this something that is desired enough to justify the added complication?

great point @effigies ! well, I am not going to even exercise this aspect here, but I think it is a great point to keep in mind while seeing how to generalize this concept in other modalities etc.

mkdocs.yml
```diff
@@ -10,7 +10,7 @@ nav:
         - Magnetoencephalography: modality-specific-files/magnetoencephalography.md
         - Electroencephalography: modality-specific-files/electroencephalography.md
         - Intracranial Electroencephalography: modality-specific-files/intracranial-electroencephalography.md
-        - Task events: modality-specific-files/task-events.md
+        - Events: modality-specific-files/events.md
```
effigies, line 13: Should also update the redirect on L86. Since the redirects haven't been in a released version, I don't think we need another redirect from `modality-specific-files/task-events.md` to `modality-specific-files/events.md`.

yarikoptic: yeap, just did that! pushed rewrite

robertoostenveld:
I approve of the proposed changes, although it does not solve everything. Regarding the sentence "Each events file REQUIRES a corresponding data file" (which was before "Each task events file REQUIRES a corresponding task data file") and comment https://github.com/bids-standard/bids-specification/pull/1440#issuecomment-1470369771, I would say that so far we have relied on the `task-<label>` entity in the filename for linking specific acquisitions to their corresponding events.tsv. I would also be fine using any other entity, as long as it results in a unique link. The example from @effigies then becomes. 

```
anat/
  sub-01_acq-T1w_T1w.nii.gz
  sub-01_acq-T2w_T2w.nii.gz
  sub-01_acq-T1w_events.tsv
  sub-01_acq-T2w_events.tsv
```

However, although recording events during a long anatomical scan is possible, I don't think that this link between the T1w and events is sufficiently concrete, as the T1w does not specify timing information, whereas the events.tsv must have a first column with onset times relative to something. Would that also be "start of T1w acquisition sequence"?

yarikoptic:
re which entity to use to add needed disambiguation, I thought of the [`_mod-`](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#mod) which was invented to absorb the original `_suffix` upon adding `_defacemask` and alike.  We might want to look/consider may be that one for explicit (have in both original `_T1w` and `_events`) or implicit (have only for `_events`) disambiguation that way since we already presumably do it for `_defacemask`s.

effigies (approved):
I'm good with this, as-is, if it helps clear the way for anybody. @yarikoptic Are there still things to do?

neuromechanist:
+1,
While we are at it, can we also try to address https://github.com/bids-standard/bids-specification/issues/1751, probably moving **Events** from modality-specific section to modality-agnostic?

WDYT, @effigies, @yarikoptic and @VisLab? Can we make this move as well within this PR?

neuromechanist:
I'll checkout a branch to make the change to `modality-agnostic` and merge into this PR.

mkdocs.yml
```diff
@@ -168,5 +168,7 @@ plugins:
         "99-appendices/11-qmri.md": "appendices/qmri.md"
         "99-appendices/12-arterial-spin-labeling.md": "appendices/arterial-spin-labeling.md"
         "99-appendices/13-cross-modality-correspondence.md": "appendices/cross-modality-correspondence.md"
+        "modality-specific-files/task-events.md": "modality-agnostic-files/events.md"
+        "modality-specific-files/events.md": "modality-agnostic-files/events.md"
```
effigies: This file has not existed except in this PR. No need to create a redirect for it.

```suggestion
```

effigies (approved):
LGTM. Small nitpick.

mkdocs.yml
neuromechanist: Ohhh, yes! 🙈

src/common-principles.md
```diff
@@ -677,7 +677,7 @@ Example of a hypothetical `*_bold.json` file, accompanying a `*_bold.nii` file:
 Example of a hypothetical `*_events.json` file, accompanying an
 `*_events.tsv` file. Note that the JSON file contains a key describing an
 *arbitrary* column `stim_presentation_side` in the TSV file it accompanies.
-See [task events section](modality-specific-files/task-events.md)
+See [task events section](modality-agnostic-files/events.md)
```
VisLab: Remove "task" in [task events section]

src/modality-agnostic-files/events.md
```diff
@@ -1,8 +1,9 @@
-# Task events
+# Events
 
 The purpose of this file is to describe timing and other properties of events
-recorded during a run.
+recorded during data acquisition.
 Events are, for example, stimuli presented to the participant or participant responses
```
VisLab: Reword -- this sounds like they an only be for tasks.  Suggested rewording:

"Events may include stimuli presented to the participant or participant responses that are associated with tasks or markers of other incidents that occurred during the execution of the experiment. See the [definition of Event](../common-principles.md#definitions) for a more detailed explanation."

src/modality-agnostic-files/events.md
```diff
@@ -1,8 +1,9 @@
...
+during acquisition of functional data and thus associated with specific tasks performed
 (see [Definitions](../common-principles.md#definitions)).
 A single event file MAY include any combination of stimulus, response, and other events.
 Events MAY overlap in time.
```
VisLab, line 10: On the next line:  "Please mind ..... "   should be "Please keep in mind ...."

src/modality-agnostic-files/events.md
```diff
@@ -18,7 +19,7 @@ duration).
    )
 }}
 
-Each task events file REQUIRES a corresponding task data file.
+Each events file REQUIRES a corresponding data file.
```
VisLab: Each `events.tsv` file....

VisLab: Also, this is REQUIRES a corresponding data file is confusing, since it can be associated with more than one (see next line).  Can you have an `events.tsv` file with no corresponding data file --- I thought you could?

src/modality-agnostic-files/events.md
```diff
@@ -35,7 +36,7 @@ The definitions of these fields can be found in
 and a guide for using macros can be found at
  https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
 -->
-{{ MACROS___make_columns_table("task.TaskEvents") }}
+{{ MACROS___make_columns_table("events.Events") }}
```
VisLab, line 40: The description for `duration` should be reviewed.  The last sentence:  "Must be a number greater than or equal to 0" repeats previous information and leads to confusion about `n/a`.

VisLab: the description of `HED` column:  "Hierarchical Event Descriptor (HED) tags."

VisLab: The `stim_file` column description -- there shouldn't be a space between / and stimuli so it doesn't break badly.

VisLab: I didn't know that we were requiring sorting by `onset` time --- is this actually being validated?

VisLab: on line 61 in the for fMRI data, there should be a comma after "sequence has started"

VisLab: The contents of the events TSV file has weird channel F,1 and annots F,2,  I think that the trial_type for this should be "artifact"  not n/a.

VisLab: The explanation also mentions CZ, but I didn't see any CZ.

effigies:
It was a fluke. Fixed by rerunning.

src/modality-agnostic-files/events.md
VisLab: The Stimuli section is pretty abrupt.  The /stimuli directory doesn't add information in the events.tsv and events.json file.

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -38,7 +38,7 @@ The definitions of these fields can be found in
 and a guide for using macros can be found at
  https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
 -->
-{{ MACROS___make_columns_table("task.Behavioral") }}
+{{ MACROS___make_columns_table("events.Behavioral") }}
```
VisLab, line 41: On line 31 above, it mentions `_beh.tsv`.  Should it also mention `_beh.json`?

Some comment about having a space between / and stimuli in some places resulting in it being split up on separate lines.

VisLab: Same comment about the HED column in the table.  Should read Hierarchical Event Descriptor (HED) tags.

VisLab: On line 52 below -- "it is RECOMMENDED to add the following metadata to the JSON files of this directory."  should these be specifically `_beh.json` files?

VisLab: In the Example below can we add HED tags for "congruent" and "incongruent" in the `_beh.json` file?

src/pregh-changes.md
```diff
@@ -9,7 +9,7 @@
 
 -   Added support for MEG data (merged BEP008).
 -   Added `SequenceName` field.
--   Added support for describing events with Hierarchical Event Descriptors: [4.3 Task events](modality-specific-files/task-events.md).
+-   Added support for describing events with Hierarchical Event Descriptors: [4.3 Task events](modality-agnostic-files/events.md).
```
VisLab: Line 12 still has Task in it.

VisLab (commented):
Minor comments interspersed.  Overall looks good but various little nits....

src/pregh-changes.md
neuromechanist: This is changelog. Might be ok to keep it as is.

src/modality-specific-files/behavioral-experiments.md
neuromechanist: Most of them are probably true, I wonder if we need to address it under another PR, to review event-related text across the spec.

src/modality-agnostic-files/events.md
neuromechanist: Resolved, but I don't think that you can do that right now

src/modality-agnostic-files/events.md
neuromechanist: I wonder if these commnets require their own PR? They are ok, but not directly related to the PR. Or should we expand the PR?

PR #1441, opened 2023-03-15
ENH: Clarify sample column values to be 0-based and possibly non-integer

effigies:
Based on discussion in https://github.com/bids-standard/bids-examples/pull/356, it seems that defining the `sample` column of `events.tsv` as containing integer values may have been mistaken. This column was introduced in 1.2.0, when EEG and iEEG were added to the specification, and there does not appear to be consensus among contributors to those BEPs that this was intended to be an integer, with @arnodelorme stating:

> Sample latencies can sometimes be fractional because they come from a different machine that has more resolution than the EEG sampling frequency. This is important for reaction time, for example. Even with EEG sampled at 250 Hz, you need to be able to determine reaction time with 1ms (1000 Hz equivalent sampling rate) precision.

If sample is not purely an index, then it is a value where the time unit is not seconds. Therefore it should be possible to say "0 samples from the start of the run", resolving issue #499 in favor of 0-based "indexing".

This PR is necessary in order to correctly validate events.tsv files that contain non-integer sample indices.

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -394,10 +394,13 @@ sample:
   description: |
     Onset of the event according to the sampling scheme of the recorded modality
     (that is, referring to the raw data file that the `events.tsv` file accompanies).
+    A sample MAY contain a fractional part when the precision of the event recording
+    is higher than the sampling rate of the corresponding data file.
+    Sample indices start at 0.
     When there are several sampling schemes present in the raw data file (as can be
     the case for example for `.edf` files), this column is ambiguous and
     SHOULD NOT be used.
-  type: integer
+  type: number
 sample_id:
   name: sample_id
   display_name: Sample ID


arnodelorme:
We need fractional samples for EEG, MEG, and iEEG - and I think this update goes against this. We need fractional samples because events are often acquired with different hardware, which has a resolution higher than the EEG. For example, a typical EEG file can be sampled at 500 Hz, and events can be at 1000 Hz or more (this is useful because, in psychophysics, you need millisecond precision on events).

effigies:
Could you recommend alternative wording? I was aiming to allow fractional samples, but if it can be read as restricting them, we have a problem.

sappelhoff:
This is what I remember from the BEP development process:

- Many (almost all) ephys recording involves sending "event markers" (or "triggers") from the experiment presentation hardware (or some other hardware) to the recording hardware (typically an amplifier) that digitizes the neural signal (we have samples then)
- The incoming event markers are then digitized together with the neural signal
- As such, each event marker is associated with a specific, discrete sample -- it cannot lie in between samples

The [`sample` metadata](https://bids-specification.readthedocs.io/en/latest/glossary.html#objects.columns.sample) for `events.tsv` files was introduced together with the [`value` metadata](https://bids-specification.readthedocs.io/en/latest/glossary.html#objects.columns.value) to cover exactly this situation --> these two metadata entries were supposed to duplicate (and make accessible) information that is present in most (almost all) ephys data formats: Event markers associated with discrete samples in the data files (=indices into the data).

If we make `sample` a zero-based and fractional metadata entry, I don't really see the difference to what we already cover with `onset`.

VisLab:
Back in version 1.2.1 sample was a recommend column:

https://github.com/bids-standard/bids-specification/blob/3d55770c21ac5cc9015c93adb9bc145d0318bfa1/src/04-modality-specific-files/05-task-events.md

Maybe we just want to it take out of the events altogether.

robertoostenveld:
Ah, thanks @sappelhoff. I now see that it is mentioned in the current spec regarding the [task events](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html).

In general I personally prefer the BIDS specification to be as small as possible, but as large as needed. I don't see the need for this being defined in the spec, as it can also be specified in the events.json. Hence I would not object if `sample` were removed rather than being listed as OPTIONAL. As there is no tight definition right now anyway, it would not hurt to make it completely the user's responsibility. The option of "Additional Columns" remains.

sappelhoff:
If we were to remove (effectively "demote") `sample` as a BIDS-defined column in `events.tsv`, that would mean that all BIDS datasets that have `sample` in `events.tsv` but no `sample` in an associated `events.json` will trigger the warning that an arbitrary column `sample` is not described in the JSON file.

Not sure whether that would classify as a backward incompatible change, as some dataset curators may have relied on using the definition of `sample` when preparing their datasets.

I am in favor of making the definition of `sample` more clear/strict for it to serve as a way to duplicate event marker information from ephys raw files (together with the accompanying `value` column, see my comment: https://github.com/bids-standard/bids-specification/pull/1441#issuecomment-1471461680). I am pretty sure that this was the original intent.

The only "wonky" thing about my proposal is that we have to define post-hoc whether `sample` is supposed to be zero or one indexed. Here I would vote for one-indexed, as we are not dealing with a "time-like" variable (like `onset`) but simply an index.

---

In other news: What happens if a dataset curator uses a pre-defined column (like `sample`) in their dataset, but supply a definition for `sample` in their JSON file that completely diverges from the BIDS definition? I don't think we have a section for this case in the specification, but it feels like we should. And I believe this should be NOT RECOMMENDED, however **if** it's being done, we could state that the "user override" is be authoritative.

PR #1445, opened 2023-03-16
FIX: Link to index/label definitions in glossary

### LINKED ISSUE #1307: [BUG] `<label> ` in name templates should link to label definition and not the the label entity definition

Only checked for MRI but I suspect the problem will be the same for other datatype as this is "macro generated".

See for this name template. 

https://bids-specification.readthedocs.io/en/latest/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#anatomy-imaging-data

clicking on label links to:

https://bids-specification.readthedocs.io/en/latest/appendices/entities.html#label

I would expect it to link to this:

https://bids-specification.readthedocs.io/en/latest/glossary.html#label-common_principles

probably all name templates

effigies:
Fixes #1307.

### DIFF
--- a/tools/schemacode/bidsschematools/render/text.py
+++ b/tools/schemacode/bidsschematools/render/text.py
@@ -330,10 +330,11 @@ def make_filename_template(
                         heading=entity["name"],
                         pdf_format=pdf_format,
                     )
+                    fmt = entity.get("format", "label")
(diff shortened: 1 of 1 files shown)


Remi-Gau (approved):
thanks @effigies

PR #1451, opened 2023-03-17
[ENH] Add noRF suffix for MR excitation-free noise scans

### LINKED ISSUE #1024: Key for noise scans in MRI time series
There are a several situations when someone would collect MRI noise scans by using a standard sequence but with no Radio Frequency excitation.
These noise maps are particularly useful for estimating properties of the scanner and sequence that are independent from the object being imaged.

The NORDIC thermal noise reduction method discussed in the following papers for Diffusion & then FMRI are two examples of this. This method has garnered some recent interest and may start to see broader adaption.
- Moeller et at "NOise reduction with DIstribution Corrected (NORDIC) PCA in dMRI with complex-valued parameter-free locally low-rank processing" NeuroImage Feb 2021
https://doi.org/10.1016/j.neuroimage.2020.117539
It's discussed the paragraph beginning with "Following these steps, the noise component of has zero-mean..."

- Vizioli et al "Lowering the thermal noise barrier in functional brain mapping with magnetic resonance imaging" Nature Communication Aug 2021 https://doi.org/10.1038/s41467-021-25431-8

These noise scans are also collected as part of some studies that lean more towards MRI physics.
For example this study collected RF excitation off noise scans as one condition:
Gonzalez-Castillo et al "Physiological noise effects on the flip angle selection in BOLD fMRI" NeuroImage 2011 https://doi.org/10.1016/j.neuroimage.2010.11.020

I looked, but I don't see any key for storing this information in the JSON files in either the main MRI parameter specification or in the fMRI or diffusion parameters.
These noise scans can be collected as a stand-alone run or at the beginning or end of a standard run.
I propose adding the following optional key

| **Key name** | **Description**                                          |
|---------------|----------------------------------------------------------|
| RFoffVols        | Volumes in a data set where there is no radio frequency excitation which will result in noise scans. Either (`"all"`) if every volume in a scan has no RF excitation or a list of volume numbers where RF excitation is turned off like (`"300 301 302 303"`)    |

This would fit either in MRI->Sequence Specifics or MRI->RF & Contrast

tsalo:
Closes #1024.

Changes proposed:

- Add noRF suffix to schema, representing excitation-free noise scans. These scans are typically acquired as additional volumes at the end of an fMRI run, for use in denoising with a tool like NORDIC. This proposes splitting them out into separate files, for better compatibility with tools.


tsalo:
@handwerkerd I'd love your input on the description.

effigies:
This LGTM. It seems like this needs to be done manually, but worth looping in @neurolabusc and @yarikoptic for any possible implications for dcm2niix/heudiconv.

neurolabusc:
@tsalo can you provide exemplar DICOM validation datasets as  [sourcedata](https://bids-extensions.readthedocs.io/en/latest/guide/#not-considering-dicom-fields)? It would be great to see examples from the major manufacturers (GE, Siemens, Philips). It is very hard to support a specification without concrete examples. BIDS fields are often defined by experts that have a clear vision of what is needed, but there can be unintended consequences if the implementation is left to end users that have no idea what this data looks like.

src/schema/objects/suffixes.yaml
```diff
@@ -673,11 +673,25 @@ motion:
...
+    A noise scan acquired alongside a `bold` scan.
+
+    These noise scans are acquired by disabling the radio frequency excitation,
+    while maintaining all other parameters from the associated `bold` scan.
+    In sequences that support `noRF` scans, one or more noise volumes are typically acquired
+    at the beginning or end of the `bold` scan, and must be separated out into separate files
+    outside of the DICOM conversion.
+
+    `noRF` are typically used to estimate properties of the scanner and
+    sequence that are independent from the object being imaged.
```
handwerkerd, line 694: I clean up the wording, but the main changes are that noRF scans can be collected independently and are used for things besides `bold` Given these are going to be distinct files in BIDS, it is better to not specify `bold`
```suggestion
    A noise scan acquired alongside another scan.

    These noise scans are acquired by disabling the radio frequency excitation,
    while maintaining all other parameters from the associated scan.
    These can be aquired using some other sequence with RF excitation removed
    or as part of sequences that turn off RF at the beginning or end of a run.
    For example a `bold` scan where the file few volumes are acquired with RF off.
    There is no consistent DICOM denotation for these scans so separating and
    labeling noRF volumes would need to be done outside of automatic DICOM
    conversion.

    `noRF` are typically used to estimate properties of the scanner and
    sequence that are independent from the object being imaged.
```

handwerkerd (commented):
As a separate file, should there be a key(s) that denotes whether a `noRF` scan is linked to another scan and, if they were collected as part of one acqusition, whether it was at the beginning or end of the scan.

handwerkerd:
> @tsalo can you provide exemplar DICOM validation datasets as [sourcedata](https://bids-extensions.readthedocs.io/en/latest/guide/#not-considering-dicom-fields)? It would be great to see examples from the major manufacturers (GE, Siemens, Philips). It is very hard to support a specification without concrete examples. BIDS fields are often defined by experts that have a clear vision of what is needed, but there can be unintended consequences if the implementation is left to end users that have no idea what this data looks like.

Here's some data shared from when this was an issue: https://github.com/bids-standard/bids-specification/issues/1024#issuecomment-1059453132
That is from a tweaked version of the CMRR Siemens multiband sequence. At that point, this was still a bit hacky from the pulse sequence side of things so it's unclear if there would be a clear and consistent field to use. At least for now, it might be somethings uses might need to manually specify for after automated dicom conversion.

tsalo:
@handwerkerd in that case, maybe we should recommend using an IntendedFor field in the noRF's metadata.

I think events files should be adjusted as part of the BIDSification process, rather than within the BIDS dataset, but I'd like to get @effigies' thoughts on it.

effigies:
I have little context for how these files are used, so my thoughts are more based on experiences with "associated data" files in BIDS, such as fieldmaps.

I think `IntendedFor` is one of my least favorite aspects of BIDS, as it encodes processing decisions within the raw data. I much prefer the `B0FieldIdentifier` approach to indicate which files can be used collectively to estimate a fieldmap, and then the task is to associate a fieldmap to a file. (The `B0FieldSource` metadata still encodes processing decisions in data, which I don't like, but at least they are associated in the BOLD file's metadata. I would prefer it if there were some way to map files )

If the idea here is to use the nearest noRF scan to aid in processing a given file, it seems like the simplest way to encode that is through the `acq_time` column in `scans.tsv`. The selection process is then part of the algorithm of the correcting tool. If you really need to encode intent, then I would suggest following the lead of `B0FieldIdentifier` and only use `IntendedFor` if you are unable.

effigies:
Also: What files might be corrected by this, in addition to BOLD? Should we add it to DWI, ASL, anat, etc?

tsalo:
>I much prefer the B0FieldIdentifier approach to indicate which files can be used collectively to estimate a fieldmap, and then the task is to associate a fieldmap to a file. (The B0FieldSource metadata still encodes processing decisions in data, which I don't like, but at least they are associated in the BOLD file's metadata. I would prefer it if there were some way to map files )

That sounds reasonable... maybe something like `ThermalNoiseIdentifier` and `ThermalNoiseSource`? @handwerkerd WDYT? I also agree that using the scan time can be a good heuristic for defining the field, as folks often do with field maps.

>Also: What files might be corrected by this, in addition to BOLD? Should we add it to DWI, ASL, anat, etc?

I know it has been applied to DWI data at minimum, but I'm not sure about other modalities. It's meant as a measure of the thermal noise level, I believe, so I can imagine it being useful in other modalities. At least ones that typically involve some kind of denoising over time (e.g., ASL)... but I'm very new to this acquisition, so I could be wrong.

handwerkerd:
I think this might be getting overly complex (partially my fault). Any MRI sequence can be collected with noRF and this can be useful to measure the noise floor for any pulse sequence. In practice, you can tell exactly what it is is linked to because all the other parameters should match. 

The only thing that makes this messy is that there is currently a single pulse sequence that collects some noise volumes a the end of an fMRI run so that a dataset will have multiple noise scans with each one collected as part of a specific run. The only reason we would need to know if it is from the beginning or end of a run is because BIDS requires actively cutting up the runs to separate the noRF from the normal volumes, which risks creating timing issues. As long as a script is included that documents what is done, this run order might be something that's not worth specifying in BIDS.

effigies:
> BIDS requires actively cutting up the runs to separate the noRF from the normal volumes

I'm not sure that's true, but I guess that would reopen discussion of https://github.com/bids-standard/bids-specification/issues/1024#issuecomment-1065415197, and this has apparently been open for a year, so I'll try to move forward and not look too much backward.

> As long as a script is included that documents what is done, this run order might be something that's not worth specifying in BIDS.

This might be best until there are are enough such scans that it is worth developing a consensus on how to encode it. I think we could still suggest calculating the two split files' `acq_time`s from the original BOLD file's acq_time and duration, as that does not require introducing any more new concepts to BIDS.

src/schema/objects/suffixes.yaml
```diff
@@ -678,6 +678,23 @@ nirs:
   value: nirs
   display_name: Near Infrared Spectroscopy
   description: Data associated with a Shared Near Infrared Spectroscopy Format file.
+noRF:
+  value: noRF
+  display_name: No Radio Frequency Excitation Scan
+  description: |
+    An MR noise scan acquired alongside another MR scan.
```
handwerkerd, line 685:
```suggestion
    An MR noise scan often acquired alongside another MR scan.
```

src/schema/objects/suffixes.yaml
```diff
@@ -678,6 +678,23 @@ nirs:
...
+
+    These noise scans are acquired by disabling the radio frequency excitation,
+    while maintaining all other parameters from the associated scan.
+    These can be aquired using some other sequence with RF excitation removed
+    or as part of sequences that turn off RF at the beginning or end of a run.
```
handwerkerd, line 690:
```suggestion
    or as part of sequences that turns off RF at the beginning or end of a run.
```

src/schema/objects/suffixes.yaml
```diff
@@ -678,6 +678,23 @@ nirs:
...
+    For example a `bold` scan where the file few volumes are acquired with RF off.
```
handwerkerd, line 691:
```suggestion
    For example a `bold` scan where a few volumes are acquired with RF off.
```

handwerkerd (commented):
Given the discussion with @effigies, the actual content of the PR is fine with me. I just noticed a few typos.

FWIW, part of the challenge issue here is that BIDS is trying to define pulse sequences by application so that a single band fMRI time series is labeled `bold` while the exact same sequence is labeled `sbref` if it's acquired  before a multiband sequence and it has a different label if it's acquired as a b0 scan during a diffusion run, while a multi-echo sequence is `bold` even if the earliest echoes shouldn't be called bold-weighted. That's how we get into requiring a label in a file name for what really should be a parameter in the header. Unfortunately, this PR really is the best solution given the currently overall set-up of BIDS.

handwerkerd (approved):

effigies (approved):

handwerkerd:
I noticed this is still sitting around unmerged since last March. @ericearl or @erdalkaraca Can one of you review?

tsalo:
@yarikoptic it's a bit of a judgment call, but I think `noRF` is more similar to `sbref` scans than a separate `acq-noRF` run. Here's why:

1. The `noRF` volumes are typically (always?) acquired as part of the BOLD run. rather than as a separate scan. The main reason this proposal splits them into a separate file is because adding a new metadata field to index no-RF volumes in the time series would make things much harder for tools like fMRIPrep.
2. The `noRF` volumes are always associated with the run they're acquired with. You wouldn't acquire a separate `noRF` run and apply that to multiple BOLD runs. Given that association, we would need an `IntendedFor` field (or an equivalent to `B0FieldSource`/`B0FieldIdentifier`) if we used `acq-noRF` instead of a suffix. As it stands, we can rely on the implicit links we have due to overlapping entities, much like we have with `sbref` files at the moment.
3. The `noRF` volumes are not useful on their own.

handwerkerd:
@tsalo and @yarikoptic The current sequence I'm using (CMRR) has an option to tag noRF scans to a longer BOLD run, but, to my knowledge, this is the only common sequence where this is an option, and I can definitely point to studies where noRF scans were run on their own (see: https://www.sciencedirect.com/science/article/pii/S1053811910014503 ). 

That said, what makes `noRF` interesting is that it can be used on any pulse sequence. That is, one can run a an EPI BOLD acquisition with RF off or an MPRAGE with RF off. In that context, noRF is always linked to whatever sequence would have been run if RF were on. In that context, it feels like a suffix to me.

As I grumbled above, the real issue here is the brittleness of the BIDS file naming structure compared to the range of things people do with MRI acquisitions. That's a harder problem to solve.

In relationship to this PR, my goal is to get something specified and merged and I'm fine with any of the currently proposed options.

Perhaps one compromise is to to allow `noRF` to either be a suffix if it's directly part of another sequence or an `acq` if it's a stand alone acquisition.

yarikoptic:
> or an MPRAGE with RF off. In that context, noRF is always linked to whatever sequence would have been run if RF were on. In that context, it feels like a suffix to me.

so would not there be a conflict if you have multiple modalities (thus having suffixes e.g. T1w, T2w, ...) with noRF for each one of them -- how to discern what that `_noRF` for?

handwerkerd:
> > or an MPRAGE with RF off. In that context, noRF is always linked to whatever sequence would have been run if RF were on. In that context, it feels like a suffix to me.
> 
> so would not there be a conflict if you have multiple modalities (thus having suffixes e.g. T1w, T2w, ...) with noRF for each one of them -- how to discern what that `_noRF` for?

This is the problem with BIDS focusing on contrasts rather than pulse sequences. A noRF image would be neither T1w nor T2w. It would be an acquisition with no excitation (and no flip angle) thus the echo spacing between the non-existent excitation and readout is also meaningless. An EPI acquisition where the flip angle and echo times were tuned for T1w or T2w should generate similar noise profiles with noRF.

yarikoptic:
I understand that, but my point was - if you have `sub-1_T1w.nii.gz` and `sub-1_T2w.nii.gz` and for each one you get `noRF` -- how those two files are to be named?  

if we are to with a suffix, I guess we could piggy back on [mod](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#mod) and thus have them `sub-1_mod-T1w_noRF.nii.gz` and `sub-1_mod-T2w_noRF.nii.gz`

handwerkerd:
That looks fine to me (and I think it's what this PR would currently do?)

src/schema/objects/suffixes.yaml
```diff
@@ -670,6 +670,23 @@ nirs:
...
   description: Data associated with a Shared Near Infrared Spectroscopy Format file.
+noRF:
+  value: noRF
+  display_name: No Radio Frequency Excitation Scan
+  description: |
+    An MR noise scan often acquired alongside another MR scan.
+
+    These noise scans are acquired by disabling the radio frequency excitation,
+    while maintaining all other parameters from the associated scan.
+    These can be aquired using some other sequence with RF excitation removed
```
yarikoptic, line 681: that is nice -- github actions/codespell annotates for a line, coolio, hence

```suggestion
    These can be acquired using some other sequence with RF excitation removed
```

src/schema/objects/suffixes.yaml
```diff
@@ -670,6 +670,23 @@ nirs:
...
+    or as part of sequences that turns off RF at the beginning or end of a run.
+    For example a `bold` scan where a few volumes are acquired with RF off.
+    There is no consistent DICOM denotation for these scans so separating and
```
yarikoptic, line 684: I think at most what it should say is

```suggestion
    There is no consistent DICOM denotation for these scans.
```

since the rest is up to conversion tool, and should nohow be enforced by standard.

src/schema/objects/suffixes.yaml
```diff
@@ -670,6 +670,23 @@ nirs:
...
+    labeling noRF volumes would need to be done outside of automatic DICOM
+    conversion.
```
yarikoptic, line 686: I think at most what it should say is

```suggestion
    There is no consistent DICOM denotation for these scans.
```

since the rest is up to conversion tool, and should nohow be enforced by standard.

src/schema/objects/suffixes.yaml
```diff
@@ -670,6 +670,23 @@ nirs:
...
+
+    `noRF` are typically used to estimate properties of the scanner and
```
yarikoptic, line 688: Echoing discussion, if decide to go with the suffix, might be valuable to say how to distinguish in case of conflicts, e.g. smth like

```suggestion
    If multiple sequences with different suffixes but otherwise the same entities have `noRF`
    sequences, disambiguate using modality entity, e.g. `mod-T1w_noRF.nii` and `mod-T2w_noRF.nii`.
    
    `noRF` are typically used to estimate properties of the scanner and
```

src/schema/objects/suffixes.yaml
```diff
@@ -670,6 +670,24 @@ nirs:
...
+
+    These noise scans are acquired by disabling the radio frequency excitation,
+    while maintaining all other parameters from the associated scan.
+    These can be acquired using some other sequence with RF excitation removed
+    or as part of sequences that turns off RF at the beginning or end of a run.
+    For example a `bold` scan where a few volumes are acquired with RF off.
+    There is no consistent DICOM denotation for these scans.
+
+    If multiple sequences with different suffixes but otherwise the same entities have `noRF`
+    sequences, disambiguate using modality entity, e.g. `mod-T1w_noRF.nii` and `mod-T2w_noRF.nii`.
```
tsalo, line 687:
```suggestion
    sequences, they SHOULD be disambiguated using the modality entity
    (for example, `mod-T1w_noRF.nii.gz` and `mod-T2w_noRF.nii`).
```

handwerkerd (approved):

ericearl (approved):
I got pinged on this and the changes LGTM. But I don't know the schema language well enough to know if this would break anything. @effigies or @rwblair this looks okay, right? Does the specification have text referring back to this or would that need a separate PR?

effigies:
All this PR will do right now is make `noRF` suffix valid anywhere `bold` is valid. It does not provide for the `mod-<label>` entity that is included in the description, which seems like a problem. There will be no required metadata for this file. Is that okay?

tsalo:
I'll add the mod entity now. I see noRF files are similar to sbrefs in that they should generally rely on an implicit connection to the original file, so I can't think of any metadata they would need.

tsalo:
I changed the `mod-` examples to `cbv` and `bold`, but I can add `noRF` as a valid suffix for other MRI rulesets.

tsalo:
Should I start the 5-day clock now that there are two approvals?

tsalo:
@bids-standard/raw-mri @bids-standard/raw-mri-func This is now open for voting and will be merged on Wednesday, April 24 if no one raises any major concerns.

bpinsard:
Joining the party a bit late and have more technical questions about implementing noRF scan (that I just learned about):
- In the data that @handwerkerd shared here https://github.com/bids-standard/bids-specification/issues/1024#issuecomment-1059453132 it seems that it is exported in the same series as the main acquisition (not like the SBRef). So running it through heudiconv/dcm2niix will not produce a separate nifti file by default, unless these are modified to split it based on the dicoms. However, looking at the 2 dicoms, there seems to be no dicoms/CSA/MRProtocol tags that allow to distinguish the noRF from the regular acquisition. 
- Is this a CMRR-MB feature only, or is it possible with product sequence ? I can't find how that option can be enabled in the docs here https://www.cmrr.umn.edu/multiband/Multi-Band_C2P_Instructions_R017.pdf . Is that a hidden option that needs to be activated through the config file?
- Having more info, we could then get back to CMRR folks to ask if this dicom could be exported as a separate series, similar to SBRef, to allow more straightforward BIDS-conversion. 

Thanks!

bpinsard:
I think I found an answer to my question https://github.com/CMRR-C2P/MB/wiki#extra-shadowy-shadow-header-parameters

tsalo:
>it seems that it is exported in the same series as the main acquisition (not like the SBRef). So running it through heudiconv/dcm2niix will not produce a separate nifti file by default, unless these are modified to split it based on the dicoms. However, looking at the 2 dicoms, there seems to be no dicoms/CSA/MRProtocol tags that allow to distinguish the noRF from the regular acquisition.

Unfortunately, the only solution I've come up with is to do the split after running heudiconv.

>Is this a CMRR-MB feature only, or is it possible with product sequence ?

Sorry, I don't know. I only know of it in the context of the CMRR-MB sequence.

>we could then get back to CMRR folks to ask if this dicom could be exported as a separate series

I hadn't considered asking CMRR to separate the series. That would be awesome.

PR #1459, opened 2023-03-30
[ENH] Allow fractional seconds in StartTime

### LINKED ISSUE #1447: Allow `StartTime` metadata to be specified with sub-second precision (optionally)
[`StartTime`](https://bids-specification.readthedocs.io/en/latest/glossary.html#objects.metadata.StartTime) does not specifically say it can be sub-second precision, which would be important to indicate.

_Originally posted by @JuliusWelzel in https://github.com/bids-standard/bids-specification/issues/1436#issuecomment-1472113881_

sappelhoff:
closes #1447

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -2,5 +2,7 @@
 skip = *.js,*.svg,*.eps,.git
 ignore-words-list = fo,te,als
 builtin = clear,rare
-dictionary = .codespell_dict
+# this overloads default dictionaries and I have not yet figured out
+# how to have multiple https://github.com/codespell-project/codespell/issues/2727
+# dictionary = .codespell_dict
 exclude-file = src/CHANGES.md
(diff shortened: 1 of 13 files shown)


JuliusWelzel:
Thanks for clarifying this, that will be helpful in the future!

src/schema/objects/metadata.yaml
```diff
@@ -3277,6 +3277,9 @@ StartTime:
   description: |
     Start time in seconds in relation to the start of acquisition of the first
     data sample in the corresponding (neural) dataset (negative values are allowed).
+    This data MAY be specified with sub-second precision using the syntax `s[.000000]`,
+    where `s` reflects seconds, and `000000` reflects fractional seconds. The square
+    brackets indicate that supplying fractional seconds is OPTIONAL.
```
effigies, line 3282: This feels a little over-wordy to me. Is the following clear enough?

```suggestion
    where `s` reflects whole seconds, and `.000000` reflects OPTIONAL fractional seconds.
```

effigies (approved):

PR #1464, opened 2023-04-07
[FIX] channel description table in motion and latency channel description

### LINKED ISSUE #1462: motion: clarify "latency" channel type description
> Regarding the comment from @sappelhoff about `latency` channel, yes, the onset is intended to be the `acq_time` of `scans.tsv` and the text could be updated to make that clear. 
> That would also imply that the first sample should always have latency of 0, considering `acq_time` = when the first data point in each run was acquired.
 
> ss[.000000] was not intended to imply integer digit limit. But I see how it is misleading. 
> Should we update the notation to s[.000000] and add that s refers to an integer of arbitrary number of digits?

_Originally posted by @sjeung in https://github.com/bids-standard/bids-specification/issues/1436#issuecomment-1471599790_
            
            
This issue is to explicitly track these points which came up in another thread. Just to prevent this from being buried and forgotten, as it's important.

Related to:

- https://github.com/bids-standard/bids-specification/pull/1459

JuliusWelzel:
closes #1462


src/modality-specific-files/motion.md
```diff
@@ -168,19 +168,19 @@ the axial components that corresponds to the three spatial axes MUST be specifie
...
-| POS         | Position in space, one channel for each spatial axis. Column `component` for the axis MUST be added to the `*_channels.tsv` file (x, y or z).                                                                                | NONE                                                                |
-| VEL         | Velocity, one channel for each spatial axis. Column `component` for the axis MUST be added to the `*_channels.tsv` file (x, y or z).                                                                                         | NONE                                                                |
+| Keyword |                                                                                                        Description                                                                                                       |
+|-------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
sappelhoff, line 172:
```suggestion
|---------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```

src/modality-specific-files/motion.md
```diff
@@ -168,19 +168,19 @@ the axial components that corresponds to the three spatial axes MUST be specifie
...
-| POS         | Position in space, one channel for each spatial axis. Column `component` for the axis MUST be added to the `*_channels.tsv` file (x, y or z).                                                                                | NONE                                                                |
-| VEL         | Velocity, one channel for each spatial axis. Column `component` for the axis MUST be added to the `*_channels.tsv` file (x, y or z).                                                                                         | NONE                                                                |
+| Keyword |                                                                                                        Description                                                                                                       |
+|---------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
sappelhoff, line 172:
```suggestion
| **Keyword** | **Description**                                                                                                                                                                                                              |
|-------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```

PR #1465, opened 2023-04-09
[FIX] Clean up qMRI RB1COR notes

lukeje:
- The documentation of the RB1COR receive sensitivity mapping file collection referred to different "transmit" coils, rather than different "receive" coils. This is fixed by this pull request. Note that because the body coil can typically be used as a receive coil and a transmit coil, I have used the phrasing "coil in receive mode" rather than "receive coil".
-  It was unclear to me whether "upon subsequent scans" referred to the two RB1COR images being acquired sequentially, or to the possibility of acquiring a separate receive sensitivity map for each element of an MPM file collection. I have attempted to clarify this.

### DIFF
--- a/src/appendices/qmri.md
+++ b/src/appendices/qmri.md
@@ -706,9 +706,12 @@ The example above applies to the `TB1RFM` suffix as well.
 
 #### `RB1COR` specific notes
 
-This method generates a sensitivity map by combining two low resolution images
-collected by two transmit coils (the body and the head coil) upon subsequent scans
-with identical acquisition parameters.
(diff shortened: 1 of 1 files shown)


lukeje:
@Gilles86 Thanks. As you've been invited as an "approving reviewer" in the system, it would be good if you could also use the [github pull request review framework](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request) to approve these changes.

effigies (approved):

PR #1470, opened 2023-04-18
Logging info on error-less validation, better warning/error separation

TheChymera:
Ideally we would inform the user that validation was successful. Could also do it with a print call... but logging seems like it would prevent stdout from getting cluttered when this is used as a lib.

Let me know :)


effigies (approved):
LGTM. Small change to wording and a "SUCCESS" indicator to help people interpret what they're reading that much quicker.

PR #1477, opened 2023-04-27
SCHEMA ASL rule check missing qualifier

rwblair:


effigies (commented):
Can you do the same on L84, L133, L158, L213, L266 and L298?

effigies (approved):

PR #1480, opened 2023-05-01
Record the actual expressions we use.

TheChymera:
Minor fix. Currently we add a prefix substitution in the validation function, but that's not recorded anywhere.


TheChymera:
@effigies I was wondering if this prefix could be added further upstream, i.e. in the rules that generate the expressions..

Issue #644, opened 2020-10-08
Harmonize .tsv examples (and their rendering)


yarikoptic:
Inspired by discussion in https://github.com/bids-standard/bids-specification/pull/619 I came to realization that there is no consistency ATM: .tsv "Text" examples come both in

<details>
<summary>proper TSV form</summary> 

```shell
$> git grep -e '        ' | grep '\.md'
src/02-common-principles.md:onset	duration	response_time	correct	stop_trial	go_trial
src/02-common-principles.md:200	200	0	n/a	n/a	n/a
src/03-modality-agnostic-files.md:filename	acq_time
src/03-modality-agnostic-files.md:func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
src/03-modality-agnostic-files.md:func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
src/04-modality-specific-files/08-genetic-descriptor.md:participant_id	age	sex	group	genetic_id	idh_mutation
src/04-modality-specific-files/08-genetic-descriptor.md:sub-control01	34	M	control	124587	yes
src/04-modality-specific-files/08-genetic-descriptor.md:sub-control02	12	F	control	548936	yes
src/04-modality-specific-files/08-genetic-descriptor.md:sub-patient01	33	F	patient	489634	no
src/05-derivatives/03-imaging.md:	    anat/
src/05-derivatives/03-imaging.md:		    sub-001_space-orig_probseg.nii.gz
src/05-derivatives/03-imaging.md:		    sub-001_space-orig_probseg.json
```
</details>
and more often in some shape of "replace tab with some spaces" - some times with nice layout into a tabular form, some times not really (even canonical example had no tabs but varying number of spaces used for delimiting, #643 ).  IMHO ideally

- markdown sources should use TABs where TABs are intended to be used
- rendering into HTML or PDF should do its best to signal that they are tabs while providing visually coherent display of the table.

I.e. ideally it should be a job of remake (some plugin) to do use TABS (ATM they are replaced by spaces by remake I guess), and just use HTML/CSS to provide nice visual alignment into the table (should be possible) while having TAB as the actual character so content could be copy pasted. Even though it is unlikely the actual entries would be copy pasted, I would expect some readers copy-paste headers (which also should be tab-delimited).  Could even use (although not really needed and overkill) smth like https://www.datatables.net/, where copy pasting a row from that fancy table does produce tab delimited entries:
```
[...]

PR #1483, opened 2023-05-02
ENH: Define channel column for events and Delimiter field for column descriptions

### LINKED ISSUE #1446: ENH: Array data in `.tsv` cells
Potentially to be dealt with:

- [x] https://github.com/bids-standard/bids-specification/pull/1605

Discussion originated in this thread:

- https://github.com/bids-standard/bids-examples/pull/324#issuecomment-1423276788

Currently, cells in BIDS `.tsv` files may take on numeric or string values (see [tabluar data](https://bids-specification.readthedocs.io/en/latest/common-principles.html#tabular-files)).

Here I list some examples:

- a string `n/a`, reflecting a missing value or that a value is not applicable for the present cell
- a string `mystr`, that is further specified in the `column_name.Levels` object in a `.json` file that accompanies the `.tsv` file, where `column_name` refers to the name of the column in the `.tsv` file under which the string `mystr` occurs
- floating point numbers, reflecting time since acquisition for the `onset` column in `events.tsv`
    - often associated with a unit, defined in `column_name.Units`, analogously to the `mystr` example above

**However, for some cases, it would be desirable to define an array of values in a particular `.tsv` cell.** For example when a given event (in a row) in an `events.tsv` file is associated with multiple channels from a neural recording (e.g., EEG, iEEG, NIRS, ...):

> This event at `onset` X and with `duration` Y is associated with channels A, B, and C

Currently, there is no formal way to specify such an array in BIDS. We think that it could be helpful for several data modalities (now and in the future) to have such a concept to work with.

In https://github.com/bids-standard/bids-examples/pull/324#issuecomment-1462816253, @effigies suggests that we use [JSON arrays ](https://www.json.org/json-en.html) for this case (see his comment for advantages over more basic structures, like a comma-separated string).

> An array is an ordered collection of values. An array begins with `[` (left bracket) and ends with `]` (right bracket). Values are separated by `,` (comma).

grafik

I think if we were to adopt JSON arrays as a valid input for `.tsv` cells, we would would benefit by (at least initially) restricting the kinds of values that may be listed in a given JSON array. Values are defined as such:

> A value can be a string in double quotes, or a number, or true or false or null, or an object or an array. These structures can be nested.

For our present purposes, I think it would suffice to confine ourselves to:

- string
- number
- boolean (true/false)

---

What are your opinions? Do you see other use cases apart from annotations in ephys data (in `events.tsv` files)? Please comment and advise.

effigies:
* See #1446 for extensive discussion.

Closes #1446.

### DIFF
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -82,9 +82,10 @@ A guide for using macros can be found at
 Example of the content of the TSV file:
 
 ```Text
-onset	duration	trial_type	response_time	stim_file
-1.23	0.65	start	1.435	images/red_square.jpg
-5.65	0.65	stop	1.739	images/blue_square.jpg
+onset	duration	trial_type	response_time	stim_file	channel	artifact
+1.23	0.65	start	1.435	images/red_square.jpg	n/a	n/a
+5.65	0.65	stop	1.739	images/blue_square.jpg	n/a	n/a
+12.1	2.35	n/a	n/a	n/a	F,1|F,2|Cz	sweat
 ```
 
 In the accompanying JSON sidecar, the `trial_type` column might look as follows:
@@ -98,12 +99,42 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
             "start": "A red square is displayed to indicate starting",
             "stop": "A blue square is displayed to indicate stopping"
         }
+    },
+    "channel": {
+        "Description": "Channel(s) associated with the event",
+        "Delimiter": "|"
+    },
+    "artifact": {
+        "Description": "Description of an artifact."
     }
 }
 ```
 
-Note that all other columns SHOULD also be described but are omitted for the
-sake of brevity.
+Note that in the example above:
+
+1.  Only a subset of columns are described for the sake of brevity.
+    In a real dataset, all other columns SHOULD also be described.
+
+1.  The `channel` column contains a list of values that are separated
+    by a non-default delimiter (that is, *not* with a comma `,`).
+    See the `Delimiter` field in the accompanying JSON sidecar.
+    Thus, the channels related to the event in the third row of the example
+    are called `F,1`, `F,2`, and `Cz`.
+
+1.  The example contains a column called `artifact`.
+    This column is not defined in BIDS, and constitutes additional, arbitrary
+    (that is, unofficial) metadata.
+    In the present case, it is used to describe artifacts in the data,
+    in reference to the `channel` column.
+    Note that in practice, such a hypothetical `artifact` column should make
+    use of the powerful HED system for documenting events, see below.
+
+Events MAY also be documented in machine-actionable form
+using HED (Hierarchical Event Descriptor) tags.
+This type of documentation is particularly useful for datasets likely to be used
+in event-related analyses.
+See [Hierarchical Event Descriptors](../appendices/hed.md)
+for additional information and examples.
 
 For multi-echo files, the `events.tsv` file is applicable to all echos of
 a particular run:
@@ -124,13 +155,6 @@ A guide for using macros can be found at
       },
    }
 ) }}
-
-Note: Events can also be documented in machine-actionable form
-using HED (Hierarchical Event Descriptor) tags.
-This type of documentation is particularly useful for datasets likely to be used
-in event-related analyses.
-See [Hierarchical Event Descriptors](../appendices/hed.md)
-for additional information and examples.
 
 ## Stimuli
 
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -82,10 +82,10 @@ A guide for using macros can be found at
 Example of the content of the TSV file:
 
 ```Text
-onset	duration	trial_type	response_time	stim_file	channel	artifact
+onset	duration	trial_type	response_time	stim_file	channel	annots
 1.23	0.65	start	1.435	images/red_square.jpg	n/a	n/a
 5.65	0.65	stop	1.739	images/blue_square.jpg	n/a	n/a
-12.1	2.35	n/a	n/a	n/a	F,1|F,2|Cz	sweat
+12.1	2.35	n/a	n/a	n/a	F,1|F,2|Cz	musc
 ```
 
 In the accompanying JSON sidecar, the `trial_type` column might look as follows:
@@ -104,8 +104,15 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
         "Description": "Channel(s) associated with the event",
         "Delimiter": "|"
     },
-    "artifact": {
-        "Description": "Description of an artifact."
+    "annots": {
+        "LongName": "Annotations",
+        "Description": "Hierarchical Event Descriptors (HED) annotations",
+        "Levels": {
+            "musc": "Muscle artifact. A very common, high frequency, sharp artifact that corresponds with agitation/nervousness in a patient."
+        },
+        "HED": {
+            "musc": "EMG-artifact"
+        }
     }
 }
 ```
@@ -116,17 +123,17 @@ Note that in the example above:
     In a real dataset, all other columns SHOULD also be described.
 
 1.  The `channel` column contains a list of values that are separated
-    by a non-default delimiter (that is, *not* with a comma `,`).
-    See the `Delimiter` field in the accompanying JSON sidecar.
+    by a delimiter (`|`), as is declared in the `Delimiter` metadata
+    field of the `events.json file.
     Thus, the channels related to the event in the third row of the example
     are called `F,1`, `F,2`, and `Cz`.
 
-1.  The example contains a column called `artifact`.
+1.  The example contains a column called `annots`.
     This column is not defined in BIDS, and constitutes additional, arbitrary
     (that is, unofficial) metadata.
     In the present case, it is used to describe artifacts in the data,
     in reference to the `channel` column.
-    Note that in practice, such a hypothetical `artifact` column should make
+    The `annots` column is making
     use of the powerful HED system for documenting events, see below.
 
 Events MAY also be documented in machine-actionable form
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -106,7 +106,7 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
     },
     "annots": {
         "LongName": "Annotations",
-        "Description": "Hierarchical Event Descriptors (HED) annotations",
+        "Description": "Annotations associated with channels indicated in the channel column.",
         "Levels": {
             "musc": "Muscle artifact. A very common, high frequency, sharp artifact that corresponds with agitation/nervousness in a patient."
         },
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -48,6 +48,14 @@ cardiac:
   description: |
     continuous pulse measurement
   type: number
+channel:
+  name: channel
+  display_name: Channel
+  description: |
+    Channel(s) associated with an event.
+    If multiple channels are specified, separate with commas (`,`) or a
+    delimiter specified in the `"Delimiter"` field describing the `channel`
+    column.
 color:
   name: color
   display_name: Color label
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -56,6 +56,7 @@ channel:
     If multiple channels are specified, separate with commas (`,`) or a
     delimiter specified in the `"Delimiter"` field describing the `channel`
     column.
+  type: string
 color:
   name: color
   display_name: Color label
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -53,9 +53,12 @@ channel:
   display_name: Channel
   description: |
     Channel(s) associated with an event.
-    If multiple channels are specified, separate with commas (`,`) or a
-    delimiter specified in the `"Delimiter"` field describing the `channel`
-    column.
+    If multiple channels are specified, they MUST be separated by a delimiter
+    specified in the `"Delimiter"` field describing the `channel` column.
+    For example, channels separated with a comma (`,`) require the `events.json`
+    file to contain `"channel": {"Delimiter": ","}`.
+    In the absence of a delimiter, tools MUST interpret any character as being part
+    of a channel name.
   type: string
 color:
   name: color
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -570,6 +570,13 @@ DelayTime:
     This field is mutually exclusive with `"VolumeTiming"`.
   type: number
   unit: s
+Delimiter:
+  name: Delimiter
+  display_name: Delimiter
+  description: |
+    If a column may be interpreted as a list of values, the character that
+    separates one value from the next.
+  type: string
 Density:
   name: Density
   display_name: Density
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -574,7 +574,7 @@ Delimiter:
   name: Delimiter
   display_name: Delimiter
   description: |
-    If a column may be interpreted as a list of values, the character that
+    If rows in a column may be interpreted as a lists of values, the character that
     separates one value from the next.
   type: string
 Density:
--- a/src/schema/rules/tabular_data/task.yaml
+++ b/src/schema/rules/tabular_data/task.yaml
@@ -10,6 +10,11 @@ TaskEvents:
     response_time: optional
     HED: optional
     stim_file: optional
+    channel:
+      level: optional
+      description_addendum: |
+        Note that this column only applies to data types where
+        channels are specified, such as EEG, iEEG, MEG or NIRS.
   additional_columns: allowed
   initial_columns:
     - onset
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -483,6 +483,7 @@ and a guide for using macros can be found at
         ),
         "Levels": "RECOMMENDED",
         "Units": "RECOMMENDED",
+        "Delimiter": "OPTIONAL",
         "TermURL": "RECOMMENDED",
         "HED": "OPTIONAL",
    }


src/schema/objects/columns.yaml
```diff
@@ -45,6 +45,14 @@ cardiac:
...
     continuous pulse measurement
   type: number
+channel:
+  name: channel
+  display_name: Channel
+  description: |
+    Channel(s) associated with an event.
+    If multiple channels are specified, separate with commas (`,`) or a
+    delimiter specified in the `"Delimiter"` field describing the `channel`
+    column.
```
sappelhoff, line 55: seems like a type needs to be declared for this entry?

> ValueError: Type could not be inferred for channel

https://app.circleci.com/pipelines/github/bids-standard/bids-specification/6246/workflows/b6e08071-be84-4391-b1a8-e3ae6bd28ac0/jobs/18265

sappelhoff (commented):
Thanks! Perhaps we could extend the example here? --> https://bids-specification.readthedocs.io/en/latest/common-principles.html#tabular-files

effigies:
> Thanks! Perhaps we could extend the example here? --> https://bids-specification.readthedocs.io/en/latest/common-principles.html#tabular-files

I would be up for an added example in https://bids-specification.readthedocs.io/en/latest/modality-specific-files/task-events.html, but I think it would be cluttering in common principles. @christinerogers Would you be able to put together a small example of how `channel` would be used in EEG tasks?

christinerogers:
> > Thanks! Perhaps we could extend the example here? --> https://bids-specification.readthedocs.io/en/latest/common-principles.html#tabular-files
> 
> I would be up for an added example in https://bids-specification.readthedocs.io/en/latest/modality-specific-files/task-events.html, but I think it would be cluttering in common principles. @christinerogers Would you be able to put together a small example of how `channel` would be used in EEG tasks?

Sure @effigies - @andesha and i will get back to you on that

Also, point taken about cluttering common principles, though how about a simple "See also" link, @effigies  ?  Just so the pointer is there to help the next person embedding a data structure to find supported formats.  Would help avoid divergence across modalities and major headaches for the bids-validator and other software platforms at the end of the day.  
Or what's a better place contributors will come across it like BEP maintainers guide?

dorahermes:
> Thanks for that. It looks like that has both comma separation and dashes with unclear meaning. Would this be considered valid under the current proposal and everybody agrees `FP1-F7,F7-T3` would parse as `["FP1-F7", "F7-T3"]` but it's up to tools to understand what `"FP1-F7"` means? Or would there be an expectation that the validator would check that all listed channels are in `channels.tsv`, so these would fail?

Thanks for noting. This is actually something that should be corrected in this example. This example is an EEG set with a bipolar montage, so each channel is a difference between 2 channels, hence the channel name is typically `F7-T3`. However, in this case the corresponding channels.tsv file correctly names the channel `F7` and the reference `T3`. I think the channel columns should be at least recommended to match the channel names. In the example I made comments to correct this, such that the channel column in the events file matches the channel name. 

> 
> What tools are generating and consuming these columns at this point?

The artifact and seizure examples were translated to BIDS from the [TUH EEG corpus ](https://isip.piconepress.com/projects/tuh_eeg/index.shtml). This is a large dataset where EEG data were annotated by experts. I use custom code to process the channel column. Perhaps @christinerogers can comment on other tools that are generating and consuming these.

sappelhoff:
> I think the channel columns should be at least recommended to match the channel names.

agreed, I would make this REQUIRED, because otherwise it may lead to confusion (i.e., when documenting that an event is related to channels that are nowhere to be found in the data).

effigies:
That sounds pretty painful to validate, especially trying to write as a declarative rule instead of custom code. We need to take each entry in a channel column, check to see if there's a specified delimiter (falling back to comma) in the `sidecar.channel` object, split by that delimiter, collect all the resulting channels, and verify they exist in the associated `channels.tsv`.

Andesha:
@effigies My new draft PR on bids-examples [here](https://github.com/bids-standard/bids-examples/pull/379) is the first step in contributing a dataset to your point. There should be more commits soon with the channel lists.

I believe @VisLab is planning commits to bids-examples to cover some more cases, too.

cc @christinerogers

sappelhoff:
I will add an example within the next days/weeks and merge this. We have agreed on this at the BIDS derivatives meeting Copenhagen 2023.

christinerogers:
> I finally got around to adding an example. please feel free to review:
> 
> old	new
> image	image

looping @dorahermes  and @Andesha in here --- **Dora** does this work for you as well?

effigies:
This is adding "Delimiter" as a valid field in all TSV columns. What is the obligation of the validator for a column that declares a "Delimiter"?

If I have channels that include `F,1` and I specify a `channel` column with no delimiter because I do not use lists, I think the proposed behavior above would cause us to parse this as two channels (`F` and `1`) which do not exist in `channels.tsv`.

I think at best this should be a warning, because otherwise the complexity of writing a validation rule in schema that covers all edge cases is going to be painful.

sappelhoff:
> If I have channels that include F,1 and I specify a channel column with no delimiter because I do not use lists, I think the proposed behavior above would cause us to parse this as two channels (F and 1) which do not exist in channels.tsv.

Or we make the `Delimiter` metadata REQUIRED. If it's not declared, then `channels` cells are not parsed as arrays.

However, the majority of people in #1146 was in favor of a (in my opinion only "seemingly") simpler solution that is currently proposed in this PR.

---

> I think at best this should be a warning, because otherwise the complexity of writing a validation rule in schema that covers all edge cases is going to be painful.

you mean that IF

1. `channels` is a column in `channels.tsv`
2. cells in the `channels` column contain `,`
3. no `Delimiter` is declared in `channels.json` under the `channels` key

THEN

1. we warn

???

> This is adding "Delimiter" as a valid sidecar field for any column in any TSV.

yes, but defined columns other than `channels` do not declare that their cells are `,`-separated arrays by default ... so how I read it, for all non-`channels`-columns, you would HAVE TO declare `Delimiter` to parse cells as arrays.

VisLab:
I would not be opposed to requiring DELIMITER if column is to be a list.
You are right in that it will be clearer and it is better to avoid coupling
requirements from multiple files.

src/modality-specific-files/task-events.md
```diff
@@ -98,12 +99,42 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
...
+1.  The `channel` column contains a list of values that are separated
+    by a non-default delimiter (that is, *not* with a comma `,`).
+    See the `Delimiter` field in the accompanying JSON sidecar.
+    Thus, the channels related to the event in the third row of the example
+    are called `F,1`, `F,2`, and `Cz`.
+
+1.  The example contains a column called `artifact`.
+    This column is not defined in BIDS, and constitutes additional, arbitrary
+    (that is, unofficial) metadata.
+    In the present case, it is used to describe artifacts in the data,
```
dorahermes, line 127: Just a suggestion, but I would probably change this to have an HED example, with the clear statement that this is optional, and change to something like: This example describes an artifact in the data, with the optional use of a HED annotation for documenting a Sweat-artifact.

src/modality-specific-files/task-events.md
```diff
@@ -98,12 +99,42 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
             "start": "A red square is displayed to indicate starting",
             "stop": "A blue square is displayed to indicate stopping"
         }
+    },
+    "channel": {
+        "Description": "Channel(s) associated with the event",
+        "Delimiter": "|"
+    },
+    "artifact": {
```
dorahermes, line 107: Perhaps change "artifact" to HED-annotations:
"HED_annotations": {
        "LongName": "Hierarchical Event Descriptors annotations",
        "Description": "EEG interpretation Hierarchical Event Descriptors annotations"
    }

dorahermes:
I agree with requiring a `Delimiter` in the channels.json in order for a column to be parsed as an array. Otherwise there may be many tricky edge cases. 

Just a suggestion, but for the `artifact` column in the example, we could perhaps provide a HED example instead? This is to help avoid the potential abundance of terms using in XEG-BIDS data to describe artifacts. In that case the column name would be HED_annotations, and `Sweat-artifact` would be listed in the content. The json sidecar would list the following for the HED_annotations column.

"HED_annotations": {
        "LongName": "Hierarchical Event Descriptors annotations",
        "Description": "EEG interpretation Hierarchical Event Descriptors annotations"
    },

src/schema/objects/columns.yaml
```diff
@@ -48,6 +48,15 @@ cardiac:
...
```
sappelhoff, line 58:
```suggestion
    Channel(s) associated with an event.
    If multiple channels are specified, they MUST be separated with a delimiter
    (for example, a comma `,`) specified in the `"Delimiter"` field describing
    the `channel` column (in the corresponding `channels.json` file).
```

sappelhoff (commented):
I think my code review suggestion here is sufficient to "REQUIRE" a delimiter to be specified.

Currently, a delimiter is described as:

>     If rows in a column may be interpreted as a lists of values, the character that
    separates one value from the next.

With my suggestion, I am effectively removing the "default delimiter" of a comma from the new `channels` column that we want to introduce.

As a result, if people want to use a `channels` column containing lists of values, they *must* specify a `Delimiter` in the corresponding `channels.json` file -- such as a comma: `,`.

> Just a suggestion, but I would probably change this to have an HED example, with the clear statement that this is optional

Thanks for the feedback Dora, I will adjust the example accordingly.

BTW, comments about the style / visual appearance of the TSV examples should be directed here:

- https://github.com/bids-standard/bids-specification/issues/644

I think we should focus on the content exclusively here, and try to get this feature merged soon.

src/schema/objects/columns.yaml
effigies: I'd rather use more sentences than multiple parenthetical clauses.

What do you think about:

```suggestion
    Channel(s) associated with an event.
    If multiple channels are specified, they MUST be separated by a delimiter
    specified in the `"Delimiter"` field describing the `channel` column.
    For example, channels separated with a comma (`,`) require the `events.json`
    file to contain `"channel": {"Delimiter": ","}`.
    In the absence of a delimiter, tools MUST interpret any character as being part
    of a channel name.
```

The final sentence is intended to prevent some tools from adopting an implicit convention that would make them incompatible with other tools.

sappelhoff (approved):

src/modality-specific-files/task-events.md
sappelhoff: done ✅

effigies:
Okay. So it seems we're agreed on language. I think to validate this, we will need to add/modify schema functions. We don't currently have a way of saying "x is a subset of y", or splitting a list of strings into a list of lists and combining into a list.

sappelhoff:
> I think to validate this, we will need to add/modify schema functions. We don't currently have a way of saying "x is a subset of y", or splitting a list of strings into a list of lists and combining into a list.

when talking about validation, what exactly do you want to validate?

1. for each row in the `channel` column
2. splitting a potential array into values
3. for each value
4. checking whether that channel is in the corresponding `channels.tsv`
5. if not, warn

is that it? If you expected an error instead of a warning, we need to tune the language to include something like.

> Channel names listed in the `channel` column MUST refer to entries in the `channels.tsv` file

Was there something else you wanted to validate that I am not thinking of right now?

---

I tagged some folks for reviews.

effigies:
> > Channel names listed in the `channel` column MUST refer to entries in the `channels.tsv` file
> 
> Was there something else you wanted to validate that I am not thinking of right now?

Yes, and the check you describe is the one I'm thinking of. The wording would indicate to me that an error needs to be raised if the check fails.

I should clarify though: I don't *want* to validate this at all because it's a pain that requires aggregating data from a minimum of three files (events.tsv, events.json, channels.tsv), handling `n/a` and performing a check that's not currently possible in our schema expression language.

If XEG users/developers are happy to handle this without the validator producing a warning or error, then that works for me. The size of the audience for this whose use wouldn't be mediated by a tool seems like it's probably small?

sappelhoff:
> The size of the audience for this whose use wouldn't be mediated by a tool seems like it's probably small?

yes.

> Yes, and the check you describe is the one I'm thinking of. The wording would indicate to me that an error needs to be raised if the check fails.

given that we do not have said language in the text currently, the situation is slightly ambiguous. We are neither requiring, nor recommending that each listed channel be in `channels.tsv`.

I am fine not validating it, which probably effectively means that we make it OPTIONAL whether or not listed channels are in `channels.tsv`.

I would, however, be in favor of adding REQUIRED language to the spec, but putting off validating to a future timepoint (potentially never), or non-bids-validator tools.

Please react with a thumbs-up here if you agree, and then I'll add the language. Or comment otherwise.

src/schema/rules/tabular_data/task.yaml
```diff
@@ -10,6 +10,11 @@ TaskEvents:
     response_time: optional
     HED: optional
     stim_file: optional
+    channel:
+      level: optional
+      description_addendum: |
+        Note that this column only applies to data types where
+        channels are specified, such as EEG, iEEG, MEG or NIRS.
```
Remi-Gau, line 17: Technically also motion, no?
So maybe better than listing all the relevant data types maybe just say:
`EEG, iEEG, MEG...` ?

Remi-Gau (approved):
Made a minor comment but should not block merging.

src/schema/rules/tabular_data/task.yaml
sappelhoff: I think the "such as" serves as a "for example"

Remi-Gau: ha yeah it can be read that way too: fine then. :-)

effigies:
Oh wow, I really thought I'd read the language that they have to line up. My brain is failing me...

I will leave it to your judgment whether to add the language, but I would treat a MUST as a commitment to eventually either validate it or relax the rule. If we don't add the MUST then I think it can only ever at most be a warning for backwards compatibility's sake, but the difficulty of the validation might justify limiting to a warning.

sappelhoff:
Thanks, I am fine with a warning then. It seems reasonable to me. Waiting some days with this for potential further reviews -- if none are coming in, I will merge it, as this has been discussed at length.

VisLab:
Could you wait until Sunday to merge. I'd like to look at more carefully
but won't have access to a computer until then. Thx

src/modality-specific-files/task-events.md
```diff
@@ -104,8 +104,15 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
         "Description": "Channel(s) associated with the event",
         "Delimiter": "|"
     },
-    "artifact": {
-        "Description": "Description of an artifact."
+    "annots": {
+        "LongName": "Annotations",
+        "Description": "Hierarchical Event Descriptors (HED) annotations",
```
VisLab, line 109: Change to:
"Description": "Annotations associated with specific channels",

src/schema/objects/columns.yaml
```diff
@@ -48,6 +48,14 @@ cardiac:
   description: |
     continuous pulse measurement
   type: number
+channel:
+  name: channel
+  display_name: Channel
+  description: |
+    Channel(s) associated with an event.
+    If multiple channels are specified, separate with commas (`,`) or a
```
VisLab: This should be revised if the Delimiter column is now required (which I think we agreed it should be)?

src/schema/rules/tabular_data/task.yaml
```diff
@@ -10,6 +10,11 @@ TaskEvents:
     response_time: optional
     HED: optional
     stim_file: optional
+    channel:
+      level: optional
+      description_addendum: |
```
VisLab, line 15: I don't completely understand why this is in the task.yaml file.  Is task.yaml where it determines if tabular data applies to a particular modality?

src/schema/objects/metadata.yaml
```diff
@@ -570,6 +570,13 @@ DelayTime:
     This field is mutually exclusive with `"VolumeTiming"`.
   type: number
   unit: s
+Delimiter:
+  name: Delimiter
+  display_name: Delimiter
+  description: |
+    If a column may be interpreted as a list of values, the character that
```
VisLab: Should something be mentioned here about the default being a comma?

src/schema/objects/metadata.yaml
```diff
@@ -574,7 +574,7 @@ Delimiter:
   name: Delimiter
   display_name: Delimiter
   description: |
-    If a column may be interpreted as a list of values, the character that
+    If rows in a column may be interpreted as a lists of values, the character that
```
VisLab, line 577: This might be better as:
"If a column entry may be interpreted as..."

Also, should you mention here that the default is a comma (since you are not requiring the Delimiter to be given?

src/modality-specific-files/task-events.md
```diff
@@ -82,9 +82,10 @@ A guide for using macros can be found at
 Example of the content of the TSV file:
 
 ```Text
-onset	duration	trial_type	response_time	stim_file
-1.23	0.65	start	1.435	images/red_square.jpg
-5.65	0.65	stop	1.739	images/blue_square.jpg
+onset	duration	trial_type	response_time	stim_file	channel	artifact
+1.23	0.65	start	1.435	images/red_square.jpg	n/a	n/a
+5.65	0.65	stop	1.739	images/blue_square.jpg	n/a	n/a
+12.1	2.35	n/a	n/a	n/a	F,1|F,2|Cz	sweat
```
VisLab: The "sweat"  was "musc" in a later example and the "artifact" column was called "annot".  I think the same example should be continued along.

src/modality-specific-files/task-events.md
```diff
@@ -98,12 +99,42 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
...
 ```
 
-Note that all other columns SHOULD also be described but are omitted for the
-sake of brevity.
+Note that in the example above:
+
+1.  Only a subset of columns are described for the sake of brevity.
+    In a real dataset, all other columns SHOULD also be described.
+
+1.  The `channel` column contains a list of values that are separated
```
VisLab, line 125: This is confusing.  I am liking the idea of the `Delimiter` field being optional less and less. 

I somehow thought from the discussion that we were going with  the `channel` column is ONLY interpreted as a list if the `Delimiter` is given in the JSON sidecar?  Also, the next line sounds like we have more than one default delimiter.  If there is a default delimiter -- it should only be the comma?

src/modality-specific-files/task-events.md
```diff
@@ -98,12 +99,42 @@ In the accompanying JSON sidecar, the `trial_type` column might look as follows:
...
+    by a non-default delimiter (that is, *not* with a comma `,`).
+    See the `Delimiter` field in the accompanying JSON sidecar.
+    Thus, the channels related to the event in the third row of the example
+    are called `F,1`, `F,2`, and `Cz`.
+
+1.  The example contains a column called `artifact`.
```
VisLab: I think that the two examples that relate to this should have this column either called "artifact" or "annot".

src/schema/objects/columns.yaml
```diff
@@ -53,9 +53,12 @@ channel:
   display_name: Channel
   description: |
     Channel(s) associated with an event.
-    If multiple channels are specified, separate with commas (`,`) or a
-    delimiter specified in the `"Delimiter"` field describing the `channel`
-    column.
+    If multiple channels are specified, they MUST be separated by a delimiter
```
VisLab, line 56: As in the task.yaml, I think that this is confusing.  Is the Delimiter required as line 59 seems to indicate it is, but the earlier changes indicate that it isn't and still include the notion of a default delimiter.

effigies:
@vislab Some of your comments are marked "Outdated". I'm not sure if you're using a link to a specific commit, or just need to refresh. Can you check that you're reviewing the latest version?

VisLab:
Yeah, I see the outdated, but I think I am looking at the latest version.
I tried to resolve what I think was fixed.  I think that your updates now
make it clear that the entries are only treated as lists if the Delimiter
field is specified.

My concerns are:
1.  Does Delimiter only apply to the "channel" column.  (Some places not
clear.)
2.  Where is the right place to enforce modality constraints for using
channel column as a special column.
3.  The examples with the annot or artifact columns should use rectified.

Thanks!

VisLab:
BTW --- I think you should go ahead and merge this PR.  If any clarifications are needed, they can be added in an additional PR.

src/modality-specific-files/task-events.md
sappelhoff:
```suggestion
        "Description": "Annotations associated with channels indicated in the channel column.",
```

src/schema/objects/metadata.yaml
sappelhoff: Delimiter will from now on apply to **any** column.

sappelhoff (approved):
Thanks for the review @VisLab, I am merging this now.

src/schema/rules/tabular_data/task.yaml
effigies: `task.yaml` is an arbitrary name. It could have been called `events.yaml`, but the rule `schema.rules.tabular_data.task.TaskEvents` defines what are valid columns in an `_events.tsv` file. The filename construction (and hence what modalities apply) are defined in the filename rules:

https://github.com/bids-standard/bids-specification/blob/426de564d995f2316ccf13ca66999623186f5bac/src/schema/rules/files/raw/task.yaml#L2-L19

https://github.com/bids-standard/bids-specification/blob/426de564d995f2316ccf13ca66999623186f5bac/src/schema/rules/files/raw/task.yaml#L2-L19

Issue #1499, opened 2023-05-22
Roadmap for BIDS Schema Tools docs


effigies:
### Your idea

Thanks to @bendhouseart, we are now pushing documentation for the schema tooling to https://bidsschematools.readthedocs.io/en/latest/. This issue is intended to track tasks for improving this site to make it more useful. Please add/edit freely, or comment below.

### Appearance

* [ ] Update theme. [Furo](https://github.com/pradyunsg/furo) is a good, clean theme with dark/light support.
* [ ] Logo. We should have BIDS logos for dark and light themes.

### Content

* [ ] Description of project on landing page. Single-sourcing README for PyPI (https://pypi.org/project/bidsschematools/) and the docs seems like a good idea.
* [ ] Links to other major sites in BIDS.
* [ ] Flesh out API docs. Aim for 100% docstrings of public functions/methods/classes.
* [ ] Add examples, run doctests (if we aren't already)

Remi-Gau:
Should the documentation for the macros be also moved into the official schema doc?

I mean this document
https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md

At least we could symlink it.

WDYT?

PR #1503, opened 2023-05-25
DOC: Prettify schemacode docs

effigies:
Doing some stuff I mentioned in #1499.

I also want to do full autosummary generation, but I don't want to mix that up with this.

Render: https://bidsschematools--1503.org.readthedocs.build/en/1503/

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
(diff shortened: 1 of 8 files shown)


bendhouseart (approved):
I'm not attached to the previous theme; this newly rendered doc looks good to me. I especially like the inclusion of the readme on the main page.

It's nice to see that you enabled PR renderings on RTD.

PR #1507, opened 2023-05-26
DOC: Schema tools doc improvements

effigies:
Adding module-level docstrings for modules that are missing them. (Less worried about test modules just now.)

Also realized that we were unnecessarily calling `as_file()` for reading version files.

### DIFF
--- a/tools/schemacode/bidsschematools/types/namespace.py
+++ b/tools/schemacode/bidsschematools/types/namespace.py
@@ -1,4 +1,9 @@
-"""Namespace types"""
+"""Namespace types
+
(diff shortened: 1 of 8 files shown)


tools/schemacode/bidsschematools/types/namespace.py
```diff
@@ -1,4 +1,9 @@
-"""Namespace types"""
+"""Namespace types
+
+The purpose of the :class:`~Namespace` type is to make a directory of
```
bendhouseart, line 3: Is `~< whatever >` a mark up thing?

bendhouseart (approved):

PR #1530, opened 2023-06-27
[ENH] Add a Consistency principle to Common principles

oesteban:
Explicitly adds a new Consistency principle which I think is implicit in the motivation and other principles. Still, I believe we need to make it explicit to establish a shared standpoint on the interpretation of the specs.

The reason I think we want to discuss about this proposed "Consistency principle" is that interpretation of modifications, etc. may change dramatically. I informally voiced this here https://github.com/bids-standard/bids-specification/pull/105#issuecomment-1361474135 and at least for that particular PR to go forward, I think we want to make several of these basic agreements/conventions explicit.


effigies:
Do we have examples where this isn't followed and has caused problems? Or is this a guideline that rule crafters should ensure remains the easiest behavior for users?

How does the development of two mechanisms, for example, B0FieldIdentifier and IntendedFor comport with this principle? Does the increased expressiveness justify an exception? Is it an exception? As another example, BIDS-URIs are intended to unify (and expand) existing file links; would that be acceptable under this rule?

yarikoptic:
I think BIDS is not there (yet) in terms of rules for "conversion" (what subject label to choose if there is no subject ID in dicom? what session label to choose?) to establish such somewhat stringent rule.  Having said that, it might be something to inspire and strive for.
But then in current formulation `"virtually"` is not defined, and thus possibly adds just confusion and not clarification really, hence obliterating the value of the principle. If it could be described as a set of rules/algorithmic descriptions ("chosen labels need to have 1-to-1 mapping between conversions", "set of entities present be identical between conversions", etc) then it would be formalized nicely, but (again) I am not yet 100% on board with it. E.g. I might have liked something like `sub-X_rec-res1_desc-filt1_T1w.nii.gz`,  `sub-X_rec-res1_desc-filt2_T2w.nii.gz` to signal that those two were acquired at the same resolution but filtered differently, where neither `rec` nor `desc` would be strictly needed since suffixes differ.  So I still think that there is some value in researcher ability to be "expressive" in pointing to some important distinctions/metadata.

oesteban:
> Do we have examples where this isn't followed and has caused problems? Or is this a guideline that rule crafters should ensure remains the easiest behavior for users?

If this is a principle we all share, it should be stated as such, and then act upon examples and problems. Nonetheless, I mentioned #105 where I felt there are several interpretations of the specs' spirit that signify we may not agree on this simple (or derivation thereof) articulation of "consistency".

In fact, I have several minor edits here and there prepared in case this is accepted as a foundational standpoint. I was planning to propose those on a subsequent PR, so that there's no conflict between discussing this and secondary effects. I believe we should agree on the basics first before we go into the details.

> How does the development of two mechanisms, for example, B0FieldIdentifier and IntendedFor comport with this principle? Does the increased expressiveness justify an exception? Is it an exception? As another example, BIDS-URIs are intended to unify (and expand) existing file links; would that be acceptable under this rule?

While I agree accepting this modification opens the door to new decisions and potential conflicts such as those you bring up, my point here is that this addition is necessary to better support currently unclear and, more worryingly, hard to interpret in the specs.

That said, as it reads right now, I don't think this has any implication in the B0FieldIdentifier vs. IntendedFor as they are optional (though the spirit is openly toward trying to make a more directive stance in this regard). Unfortunately, I am not familiar enough with BIDS-URIs, but I don't see why it would not be acceptable under this principle (i.e., not a rule).

oesteban:
Isn't a fundamental principle of replicability that given the same dataset, the same method should give you the same output no matter the platform you run it in or the operator?

oesteban:
> Would you call those virtually identical?

No - but there are two different concepts conflated. If participants came on 2 different days, then this principle would open the door to stipulate the (currently not existing) rule that two different experiments must be organized with different sessions (even if they are done on the same day but one after another). Precisely to limit that degree of freedom of the encoder.

Example two -  participants did 2 tasks. is this in the same session (i.e., simultaneous fMRI and EEG) or different sessions. Although I agree it is possible, I find that your `ds3` should explicitly be disallowed in the specs founded on this very principle, hence justifying this PR.

What I find "virtually" equivalent is this:

```
├── ds1
│   └── sub-01
│       ├── anat
│       ├── eeg
│       └── func

├── ds2
│   └── sub-AA
│       ├── anat
│       ├── eeg
│       └── func
```

or

```
├── ds1
│   └── sub-01
│       ├── ses-eeg
│       │   └── eeg
│       └── ses-mri
│           ├── anat
│           └── func
├── ds2
│   └── sub-01
│       ├── ses-1
│       │   └── eeg
│       └── ses-2
│           ├── anat
│           └── func
```

PR #1533, opened 2023-06-28
ENH: Add reference volumes to common imaging derivatives

### LINKED ISSUE #1532: Derivative reference volumes for 4D series
### Your idea

In fMRIPrep, we generate `_boldref.nii.gz` files from some combination of the source bold file, an associated sbref (if available) and within-space transforms (such as SDC warp). This is used for performing SDC, coregistration, visualization, and as a resampling target for a full BOLD series.

I was initially going to put this into #519 as `_boldref` and `_cbvref`, but is it more general? Does this make sense to put into [Imaging data types](https://bids-specification.readthedocs.io/en/latest/derivatives/imaging.html)?

Would something like the following ever be generated?

* `_dwiref` (cc @oesteban, @arokem, @Lestropie, @bids-standard/derivatives-mri-dwi)
* `_aslref` (cc @patsycle, @HenkMutsaerts)
* `_petref` (cc @mnoergaard @melanieganz, @bids-standard/derivatives-pet)

effigies:
It was discovered that fmriprep, aslprep and qsiprep all generate `<suffix>ref` volumes for their respective input modalities. Here is an initial proposal. I do not think we need any additional metadata over the universally RECOMMENDED `Description` and OPTIONAL `Sources`.

I kept `space-<label>` and `res-<label>` but not `den-<label>`, since these do make sense to resample into target spaces and possibly with resolutions differing from the input images, but there doesn't seem a clear analog for surface meshes. I suppose you could theoretically sample the `boldref` to a surface as a diagnostic, but I figure we should let the use case come up before specifying it.

Contradicting that, I threw `cbvref` in there, since it's functional data like `bold`. I have no clue how that's processed, but it seems reasonable to say "if you want a reference file, call it `cbvref`". In any case, there's nothing stopping modalities from adding another suffix that in practice they use as a reference volume. It is just useful to have a name for the thing that's not an actual average image but is some attempt to make a good registration target from what we are given.

Closes #1532.

### DIFF
--- a/src/derivatives/imaging.md
+++ b/src/derivatives/imaging.md
@@ -140,6 +140,31 @@ And the corresponding `sub-001_task-rest_run-1_space-fsLR_bold.json` file:
 }
 ```
 
+## Reference volumes
+
+Template:
+
+```Text
+<pipeline_name>/
+    sub-<label>/
+        func|dwi|perf/
+            <source_entities>[_space-<space>][_res-<label>][_desc-<label>]_<suffix>.nii.gz
+```
+
+A reference volume is a 3D image that is used to represent a 4D series.
+Reference volumes are often used for registration, as resampling targets,
+as well as for generating quality control reports.
+
+The suffix of a reference volume is derived by adding `ref` to the suffix
+of the series for which it is a reference.
+
+| Data type | Original suffix | Reference volume suffix |
+|-----------|-----------------|-------------------------|
+| `func`    | `bold`          | `boldref`               |
+| `func`    | `cbv`           | `cbvref`                |
+| `dwi`     | `dwi`           | `dwiref`                |
+| `perf`    | `asl`           | `aslref`                |
+
 ## Masks
 
 Template:
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -510,6 +510,11 @@ asllabeling:
     An anonymized screenshot of the planning of the labeling slab/plane
     with respect to the imaging slab or slices.
     This screenshot is based on DICOM macro C.8.13.5.14.
+aslref:
+  value: aslref
+  display_name: ASL reference image
+  description: |
+    Reference image for arterial spin labeling (ASL) series
 beh:
   value: beh
   display_name: Behavioral recording
@@ -529,11 +534,21 @@ bold:
   display_name: Blood-Oxygen-Level Dependent image
   description: |
     Blood-Oxygen-Level Dependent contrast (specialized T2\* weighting)
+boldref:
+  value: boldref
+  display_name: BOLD reference image
+  description: |
+    Reference image for blood-oxygen-level-dependent (BOLD) series
 cbv:
   value: cbv
   display_name: Cerebral blood volume image
   description: |
     Cerebral Blood Volume contrast (specialized T2\* weighting or difference between T1 weighted images)
+cbvref:
+  value: cbvref
+  display_name: CBV reference image
+  description: |
+    Reference image for cerebral blood volume (CBV) series
 channels:
   value: channels
   display_name: Channels File
@@ -563,6 +578,11 @@ dwi:
   display_name: Diffusion-weighted image
   description: |
     Diffusion-weighted imaging contrast (specialized T2 weighting).
+dwiref:
+  value: dwiref
+  display_name: DWI reference image
+  description: |
+    Reference image for diffusion-weighted imaging (DWI) series
 eeg:
   value: eeg
   display_name: Electroencephalography


Lestropie:
In the current proposal, the filename suffix is the concatenation of two items: some other pre-existing suffix indicating the contrast present in some other image, and "`ref`". I'm going to throw a spanner in the works and question whether this should be the case, despite some limited precedent.

1.  Not all cases of "3D reference images generated from a > 3D dataset for the purpose of registration" are faithful to the suffix from which they may have originated. For instance:

    1.   A mean *b*=0 from DWI explicitly has *no* diffusion-weighting; it's just a T2-weighted image.
    2.   The reference volume used for registration of DWI data could look nothing like raw DWI data. Could be an FA image (or anything else that could have its own more appropriate filename), could be a pseudo-T1w image like [here](http://www.sciencedirect.com/science/article/pii/S1053811915002451) or [here](https://www.researchgate.net/publication/307862882_Generating_a_T1-like_contrast_using_3-tissue_constrained_spherical_deconvolution_results_from_single-shell_or_multi-shell_diffusion_MR_data), could be anything else hypothetically.

2.  This is mixing up *image content* with *processing*. Pre-BIDS, if I were to point within a dataset at a particular file, where that file were generated by taking the mean of an fMRI time series, and I were to describe those data to a third party, I would not say "that's the fMRI reference volume used for registration".

    1.  If describing the data themselves, which is IMO the purview of BIDS, I'd say that the best descriptor of that image is the mean statistic taken along the 4th axis of the BOLD timeseries. See eg. https://github.com/bids-standard/bids-bep016/issues/61 (even though I've not convinced myself that that's the right way to go).
    
    2.  If describing *how the data were used*, ie. specifically as the reference volume for registration in a previously completed processing pipeline, that is IMO provenance.

    3.  If describing *what the data are intended to be used for*, ie. here's an image that this pipeline thinks would serve well as a reference volume for registration in some subsequent pipeline that uses this derivative dataset as an input, then:
        1.  It should be up to that pipeline to decide what of the derivatives available it wants to use for that purpose;
        2.  That falls in the domain of *intended utilisation in processing*. This I spoke about in https://github.com/bids-standard/bids-2-devel/issues/53 as I think it should be separated from the data as much as possible.

3.  In the worst case scenario, this could lead to a doubling of the number of suffices, despite a desire to minimize such.

effigies:
The purpose of these volumes is for performing, applying and evaluating registrations. They do not map any particular physical quantity, they cannot necessarily be described as a voxel-wise summary statistic of the original series, and two tools generating references do not need to generate the same reference. They are useful for provenance and post-pipeline reuse. For example, when attempting to find a minimal set of derivatives needed to regenerate the rest, this is a critical one.

I'm not sure if your position is that tools should not create these reference volumes, should have heterogeneous names that more closely reflect the contents, or should shove these files in `.bidsignore`. I think you may be trying to achieve a purity in BIDS that is not really possible.

> Not all cases of "3D reference images generated from a > 3D dataset for the purpose of registration" are faithful to the suffix from which they may have originated.

I wrote "A reference volume is a 3D image that is used to represent a 4D series," which is specifically not intended to indicate that the file was generated from the original 4D dataset or that the resulting contrast matches that suffix.

> This is mixing up image content with processing.

I think this derivative is both valuable and inextricable from processing.

> I'd say that the best descriptor of that image is the mean statistic taken along the 4th axis of the BOLD timeseries

That's not what `boldref` is. I don't know if it is what `aslref` or `dwiref` is. I would say you are welcome to use `stat-mean` or images where that applies.

> In the worst case scenario, this could lead to a doubling of the number of suffices, despite a desire to minimize such.

This is hyperbolic. There are currently 5 explicitly 4D suffixes (`asl`, `bold`, `cbv`, `dwi`, `pet`) where this would clearly apply, and three of them have tools that *already do this*. For sensor-based modalities, I don't know if there's an equivalent. If so, I would honestly be willing to move this to common derivatives as a principle, because it is so useful.

I also don't see minimizing suffixes as an explicit goal so much as a heuristic for finding cases where a few suffixes and one or two entities might replace many suffixes. I have very little concern about `boldref` taking a suffix that might be used for something else in another context, while something like `mean` could apply in many cases and it would be good not to claim it for a very narrow case.

tsalo (approved):
The addition of reference volume suffixes makes sense to me.

Should `Sources` be added as a RECOMMENDED metadata field? I know it's OPTIONAL for derivatives, but, as with masks, it seems particularly relevant for reference volumes.

In regards to concerns about adding suffixes, I do think that we generally want to avoid adding unnecessary suffixes when combinations of existing suffixes and entities would be sufficient, but in this case I think these reference volume suffixes are both useful and already widely in use. The only alternative I can think of would be a general `reference` suffix, but that would just complicate things for a lot of tools.

effigies:
> Should Sources be added as a RECOMMENDED metadata field? I know it's OPTIONAL for derivatives, but, as with masks, it seems particularly relevant for reference volumes.

I would be fine with promoting it to RECOMMENDED.

> the precise terms for registration purposes would be "static" and "moving" images, regardless of the modality 

Not sure that that's very useful here, as one process's static image is another process's moving. In the case of a boldref, it is static for motion correction and moving for coregistration. I really think "reference" is a useful term, as it is an image with a definite affine and grid that stands in for any other images that are aligned with it. Its contents need have no meaning apart from their usefulness in registration. If there's a more precise term for this type of image, I'm happy to use it, but it should not be dependent on the direction of registration.

For what it's worth, the original inspiration for `boldref` was the single-band reference (SBRef) from the Human Connectome Project, which has the property of being a useful stand-in for the BOLD series while not being derived from the series.

Lestropie:
> I think you may be trying to achieve a purity in BIDS that is not really possible.

I do that. You may have noticed more generally. :-P  I've had plenty of experience of blurring of logical concepts leading to problems in software design / communication, so will advocate for the cleanest separation even if the consequences of failing to do so aren't clear and won't manifest for years. But I have no expectation of always getting my way. Just trying to provide insights based on that experience and seeing which of them those of authority agree with.

> I wrote "A reference volume is a 3D image that is used to represent a 4D series," which is specifically not intended to indicate that the file was generated from the original 4D dataset or that the resulting contrast matches that suffix.

I think refining this might provide some guidance / consensus. "Used to represent a series" is very vague. 

1.  Which of the following is the case?
    1.  "The reference image" is "the image that is used for registration".
    2.  "The reference image" is something like "a dataset possessing high contrast, potentially of reduced dimensionality with respect to the original dataset, that best localises the spatial position and internal structure of the data content", for which potential applications include registration but also things like visualisation.
    
    This might influence choice of suffix / description.

2.  With respect to the choice of suffices, I had mentioned the (unofficial?) policy of minimising new suffices. The current proposal I suppose I would place in the intermediate range in this respect: it's greater than one, and less than the total possible number of different image contrasts. But the generation of those new proposed suffices introduces two potential problems:
    1.  The image content may look nothing like what the suffix describes (eg. `dwiref`; as mentioned earlier above)
    2.  In most of these cases, the imaging modality (eg. "`dwi`") will already be indicated by the directory in which the file resides, and so being a part of the suffix also will be redundant.

    Reason I re-raise i. and add ii. is that there's an alternative option: introduce *only one* new suffix. The modality of such would be inferred from its directory location. Looks like @tsalo mentioned and discarded it, but I think it's worth considering.
    The data from which it was generated would ideally be encoded via provenance, but it might be appropriate to also define a metadata field that is mandatory for data files with this suffix that lists those data files for which it provides such a spatial reference.
    The stumbling block I see for this approach is where there are two such references generated for a single imaging modality. On first consideration, these could be disambiguated at the file level using `_desc-<suffix>` and more precisely from the metadata field above. But this is beyond my personal experience so curious to know if someone thinks this would be more fundamentally broken.

    And of course it would not be compatible with some BIDS Apps that have already made their own decisions on how to export such images; but as you say, I'm a purist, so I'd prefer to contemplate the decision rather than relying on precedent that may itself have come from just satisfying an immediate requirement. The relationship to SBRef makes sense, and I'd kind of guessed as such already. There "single-band" conveys a lot more about the expected image content, relating to 2.i. above.

> the precise terms for registration purposes would be "static" and "moving" images, regardless of the modality

"Static" and "moving" are best avoided here, since they imply an *asymmetric* registration, which is not always the case. That's also tying even more strongly to actual processing, as opposed to data content or even intent of processing, which I've expressed my objection to above.

PR #1547, opened 2023-07-11
[ENH] Allow UNKNOWN and NONE ContrastBolusIngredients

### LINKED ISSUE #1542: Unknown Contrast Agent in BIDS
Hello,
We (@adhusch, @MeryemABBAD) are trying to convert several datasets from DICOM to the BIDS format. There are some datasets which use a contrast agent but it's not clear which one was used.
There is the ContrastBolusIngredient field in the meta-data but the documentation states that "Allowed values: IODINE, GADOLINIUM, CARBON DIOXIDE, BARIUM, XENON". Secondly there is the ceagend entity but the documentation states that it represents the ContrastBolusIngredient field. We assume that by using "ce: True" and "ContrastBolusIngredient: UNKNOWN" the dataset would not comply with BIDS anymore. How does BIDS handle these cases?
Thanks!

Remi-Gau:
fixes #1542

- allows `unkwown` and `none` contrast agent

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -500,8 +500,8 @@ ContrastBolusIngredient:
     - CARBON DIOXIDE
     - BARIUM
     - XENON
-    - unknown
-    - none
+    - UNKNOWN
+    - NONE
 DCOffsetCorrection:
   name: DCOffsetCorrection
   display_name: DC Offset Correction
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -500,6 +500,8 @@ ContrastBolusIngredient:
     - CARBON DIOXIDE
     - BARIUM
     - XENON
+    - unknown
+    - none
 DCOffsetCorrection:
   name: DCOffsetCorrection
   display_name: DC Offset Correction


src/schema/objects/metadata.yaml
```diff
@@ -500,6 +500,8 @@ ContrastBolusIngredient:
     - CARBON DIOXIDE
     - BARIUM
     - XENON
+    - unknown
+    - none
```
effigies, line 504:
```suggestion
    - UNKNOWN
    - NONE
```

effigies (approved):

effigies:
@yarikoptic Does HeuDiConv need to know about this change?

tsalo (approved):
The change looks good to me. We might want to keep an eye out for this issue with similar entity-metadata field combos.

Remi-Gau:
Task name comes to mind but this is already checked in the validator.
Doubt contrast agent is actually checked.

effigies:
From older specs:

> The key ContrastBolusIngredient MAY be also be added in the JSON file, with the same label.

It's never been even RECOMMENDED. Reading it, it almost seems like it should be disallowed except when `ce-` is present, but MRI does not disallow unknown metadata, so that doesn't seem sensible.

I think it would make sense to make it `RECOMMENDED` if `ce-` is present, which we can easily do in schema.

Remi-Gau:
> MRI does not disallow unknown metadata
Is that new?
From memory extra metadata was fine

tsalo:
Oh actually NONE should probably just be `n/a`, right?

effigies:
> Oh actually NONE should probably just be `n/a`, right?

`n/a` is mainly a TSV convention. For JSON, I believe PET used it for REQUIRED fields that could sometimes be undefined for legitimate reasons, but that seems unnecessary for OPTIONAL/RECOMMENDED fields.

> > MRI does not disallow unknown metadata
> 
> Is that new?
> From memory extra metadata was fine

Yes, extra metadata is fine, and not disallowed.

PR #1548, opened 2023-07-11
SCHEMA: Define associated data with selectors, path components and inheritance

effigies:
One difficulty with validation is knowing when it is worth making the effort to load an associated file. This meta-schema is intended to hint to validators and other implementations when associations may be relevant.

It could make sense to add filename rules here, as well as indicators of whether a file can be found through inheritance (making some path components optional) or not. For example `magnitude` images will never be found through inheritance, but must match all entities with their corresponding `fieldmap` images. `bval` and `bvec`, however, can be inherited.

### DIFF
--- a/src/schema/meta/associations.yaml
+++ b/src/schema/meta/associations.yaml
@@ -1,0 +1,43 @@
+# These rules indicate whether an association (defined in meta.context.context.associations)
+# applies to a given file. These are hints to allow implementations to avoid unnecessary
+# directory and file reads.
+#
+# If the selectors match a given file, a conforming tool MUST attempt to load the association.
+# Tools MAY attempt to load associations when selectors do not match.
+---
+events:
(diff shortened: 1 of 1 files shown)


rwblair (commented):
Looks good! Bval and bvec are pains because they are extensions, so I still need to know implicitly that thats the part of the file name I'm replacing on. But they're the only ones and a legacy rule.

In addition to selectors we could have a `target` key that lists the desired parts of the filename matched by the selectors to replace. So for events:
```
target:
  extension: '.tsv'
  suffix: 'events'
```

The extension along with a boolean for inheritance would let us know how to inherit properly.

effigies:
> Bval and bvec are pains because they are extensions

I think if you implement inheritance logic based off of `target`, where extensions are included, they shouldn't need separate treatment.

PR #1551, opened 2023-07-12
BEP: DICOM "file-format" support

yarikoptic:
I think GitHub issues do not provide adequately powerful medium to discuss complex topics due to single threading etc. In DataLad, DANDI and other projects we adopted an idea of introducing "design documents" which take advantage from "code review" functionality of the GitHub by allowing for context-specific comments or recommended changes.
Hence I am submitting this PR to seek further discussion and improvements to the *idea* of such a BEP first.

### DIFF
--- a/src/extensions/dicom-file-format.md
+++ b/src/extensions/dicom-file-format.md
@@ -2,60 +2,59 @@
 
 ### High-level Description
 
-Make `.dcm` (DICOM) file/folder to become acceptable file format for the imaging, and possibly other modalities.
+Propose the adoption of the `.dcm` (DICOM) file/folder as an acceptable file format for imaging and potentially other modalities.
 
 ### Motivation
 
-BIDS already supports a **wide** range of file formats across different data modalities.
-Choices for any particular format were typically made based on the presence of dominant in research field file format (e.g., NIfTI for neuroimaging data) or absence of such, which lead to allowing for multiple formats (e.g., [`ctf`](#ctf), [`fif`](#neuromagelektamegin), [`4d`](#bti4d-neuroimaging) + 3 more for [MEG](../appendices/meg-file-formats.md)).
-A BRAIN Initiative® supported [NWB](#nwb) file format (although supporting many modalities) was added to support intracranial electroencephalography.
+BIDS already supports a **broad** range of file formats spanning various data modalities.
+The choice of any particular format was typically made based on the dominance of that format in a research field (e.g., NIfTI for neuroimaging data), or the absence of a dominant format, which led to allowing for multiple formats (e.g., [`ctf`](#ctf), [`fif`](#neuromagelektamegin), [`4d`](#bti4d-neuroimaging) + 3 more for [MEG](../appendices/meg-file-formats.md)).
+The BRAIN Initiative®-supported [NWB](#nwb) file format was added for intracranial electroencephalography, even though it supports many modalities.
 
-From a paragraph within "The Future of BIDS" of WiP draft of the PPFofBIDS paper:
+The following extract from "The Future of BIDS" section of the WiP draft of the PPFofBIDS paper discusses this further:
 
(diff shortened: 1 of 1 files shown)


src/extensions/dicom-file-format.md
```diff
@@ -0,0 +1,60 @@
...
+### Summary of Proposed Developments
+
+Develop a BEP to allow `.dcm` to be an accepted file format for MRI and other data types (the BEP work would decide which specifically).
+
+#### Cons: Potential Issues and Solutions
+
+- Duplication of metadata between `.dcm` and BIDS-sidecar `.json`.
+  - This issue is not new or specific to DICOM: it needs to be addressed conceptually in BIDS (https://github.com/bids-standard/bids-specification/pull/761).
+  - In fact, it offers a **benefit** of metadata harmonization, particularly when DICOM or manufacturers do not yet provide information in non-manufacturer-specific sections of DICOM.
+- Libraries needed to support DICOM I/O to support BIDS: experiences with `dcm2niix` suggest that it remains a challenge to consistently access imaging data from DICOMs.
```
yarikoptic, line 49: @neurolabusc please have a look at this BEP idea and please express yourself -- your experience and knowledge would be of the most value here!

arokem:
This was mentioned as related in the BIDS OHBM town hall: https://github.com/ismrmrd/ismrmrd

satra:
@arokem - i believe the scope of that file format was specifically for k-space data.

arokem:
A clear case of "one person's raw data is another person's product"

marcelzwiers:
The idea of dropping nifti and moving over sounds appealing to me, yes. But... that would require all(/most) common neuroimaging packages that even do not support BIDS out of the box (e.g. FSL) + the existing bidsapps to switch over to DICOM. I think hell will freeze over before that happens, while in the meantime we would create more chaos/incompatibilities...

src/extensions/dicom-file-format.md
neurolabusc: I am agnostic about this. The issue with DICOM is that it is like the English language, which is used differently by different individuals (e.g. `pants` means something different in the USA than the UK). Each manufacturer has its own definitions, can use terms like `RepetitionTime` differently, and uses private tags for a lot of details. This seems like an aspirational goal, but it seems like the initial work must come from harmonization efforts by the DICOM work groups.

Assuming a user wants to make a DICOM-based BIDS dataset, how does the BIDS validator work? Consider a Philips fMRI time series, where that manufacturer has no method for defining slice timing. What tool will be used to inject required details not in the raw DICOM? Do we need to define a generic set of tags to define these (presumably with the DICOM work group assigning a public tag)?

I think that dcm2niix (and dicm2nii, SPM) face the red queen problem: we must always update our code as the manufacturers change their DICOM implementation and our understanding of their usage evolves. Adopting DICOM simply shifts the burden of perpetual updating to other tools.

However, I also respect the [view](https://peerj.com/articles/2057/) of @fedorov and his [dcmqi](https://github.com/qiicr/dcmqi) does deliver on the promise of using DICOM directly for science. He started with the DICOM format, while many of the BIDS pipelines are wrappers for tools that do not support DICOM. Therefore, this would require a lot of investment.

yarikoptic: > Do we need to define a generic set of tags to define these (presumably with the DICOM work group assigning a public tag)?

I think so -- we would need to completely avoid relying on manufacture private/specific tags and validate that DICOMs valid and contain all necessary metadata where we expect them.

yarikoptic: Thanks @fedorov . Let's see if I would eliminate the confusion in

> I don't understand the logic of reasoning. "Vendors put things into non-standard places in DICOM" - agreed! "DICOM standard is complex" - agreed! Conclusion: "because of this we will make a new standard" - confused.

I think your statement is *broken* here by the fact that actually nobody is making a new (file) standard ATM -- standards were already made, and not in one day. It just happened historically that DICOM evolved from not being used widely by anyone, to becoming the standard adopted by (nearly) all manufacturers; and in parallel to that, researchers used a "crippled" but "efficient for their needs" NIfTI, which some manufacturers supported before they even supported DICOMs; and through time we just complimented NIfTIs with "extra agreed upon metadata in .json", and "agreed file-naming scheme" since we do not work with/on PACS -- we typically work with files.  A little more on that worded in [The past, present, and future of the brain imaging data structure (BIDS): 3 The Establishment of BIDS](https://direct.mit.edu/imag/article/doi/10.1162/imag_a_00103/119672)

So, I just see this PR as the next step forward to potentially bring two, largely developed independently standards closer together. And thanks to your work on dcmqi, and following the logic and procedures you stated above, do keep doing those baby steps forward to make both "fully compatible" and eventually indeed converging on using dicoms for becoming the basic file format for storing various data types in BIDS, providing more human accessible formatting of datasets (DICOM standard, similarly to NIfTI and other file formats) AFAIK does not care about file names ;-)

PR #1556, opened 2023-07-18
[MISC] Update teonbrooks' info for .tributors

teonbrooks:


sappelhoff:
Thanks Teon!

@Remi-Gau technically updating the `.tributors` file **only** would be the way to go, right? and then letting the `CITATION.cff` file be updated automatically upon next release?

Remi-Gau:
> Thanks Teon!
> 
> @Remi-Gau technically updating the `.tributors` file **only** would be the way to go, right? and then letting the `CITATION.cff` file be updated automatically upon next release?

yes that is correct: but I did open an issue to check that if only the citation.cff were to be updated we would know (some check in CI, or the update script would notify...)

sappelhoff (approved):

PR #1561, opened 2023-07-21
Update contributors.md

francopestilli:
I updated my contribution, in preparation for the BIDS town hall

### DIFF
--- a/src/appendices/contributors.md
+++ b/src/appendices/contributors.md
@@ -142,7 +142,7 @@ If you contributed to the BIDS ecosystem and your name is not listed, please add
 | Fidel Alfaro Almagro                                 | 💬📖💡🔌                               |
 | Filip Szczepankiewicz                                | 📖                                     |
 | Filippo Maria Castelli                               | 📖🔣                                   |
-| Franco Pestilli                                      | 📖💻🎨💡🤔👀🔧📋🔍🚇                   |
+| Franco Pestilli                                      | 📖💻🎨💡🤔👀🔧📋🔍🚇📢                   |
 | Franklin W. Feingold                                 | 📋📝✅💬🤔🎨📢👀🚇🖋️📆💻               |
 | François Tadel                                       | 📖🔌💡                                 |
 | Gaia Rizzo                                           | 📖                                     |


Remi-Gau:
@francopestilli 

This file should not be edited directly: see https://github.com/bids-standard/bids-specification/blob/7fccfc371221dc6f395b596e83ad21209c8b01da/src/appendices/contributors.md?plain=1#L39

We should probably find a way to make this more unmissable or have CI notify people when only this file is edited.

sappelhoff:
@francopestilli I have added your contribution here: https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors

This follows our workflow to update contributions to BIDS. Could you please also tell us an email we can use in your contributor profile?

@Remi-Gau, see how I added Franco (screenshot below):

image

I only put the name and added the single emoji that was updated (added). The docs currently don't say whether for an **update** all previous information must be reproduced, or whether adding the **new** items is sufficient. What do you think?

The docs on how to sync contributors currently say:

> - add new contributors info to the `tools/new_contributors.tsv` file.

However, it's not entirely clear to me with which workflow (copy paste / manual editing?) one should go from this (screenshot below):

image


To this (screenshot below):

image

see also how in the screenshot of the TSV file, it says "idea" instead of the idea emoji :bulb: --> shouldn't we always paste the exact emojis into that file?

---

Another issue I see relates to the workflow in the makefile:

```
update_contributors:
	python tools/add_contributors.py
	python tools/print_contributors.py
	yarn all-contributors generate
```

I don't think it says anywhere in the docs that `yarn` is needed

Remi-Gau:
I think I am probably going to add a github action workflow that posts in a PR if it modifies the contributors file in the appendix.

The message should link to the contributors section in the wiki and tell people to add their info there instead.

Remi-Gau:
> which workflow (copy paste / manual editing?) one should go from this (screenshot below)

currently this must be done manually (copy pasting)

Remi-Gau:
> shouldn't we always paste the exact emojis into that file?

Thinking again about this: we should probably only allow just the words a not the emoji to be added in the tsv (easier for the code and less cognitive effort when adding contributions)

Remi-Gau:
> I only put the name and added the single emoji that was updated (added). The docs currently don't say whether for an **update** all previous information must be reproduced, or whether adding the **new** items is sufficient. What do you think?

I think that the sensible behavior is:
- for contributions are just added (we just check there is no duplicates): I don't think we will want to remove contributions
- for all other fields:
  - if the value in the tsv is empty we do not change it
  - if there is a non empty value in the tsv then it replaces the previous one we have in store

Remi-Gau:
@francopestilli your contribution will be updated as part of #1651
I will close this PR after #1651 is merged but will keep track of what was discussed here to improve our "contributors workflow".

sappelhoff:
- closed in https://github.com/bids-standard/bids-specification/pull/1651

PR #1570, opened 2023-08-02
Allow for "echo" entity to be used with T1w (and other "anat" nonparametric) sequences

### LINKED ISSUE #654: Add support for multi-echo T1w scans
BIDS currently does not seem to support multi-echo anatomical T1w scans such as:

    sub-A00000909_ses-20110101_acq-mprage_run-01_echo-01_T1w.json
    sub-A00000909_ses-20110101_acq-mprage_run-01_echo-01_T1w.nii.gz

It would be nice if this could be added in the future. See [this post on neurostars.org](https://neurostars.org/t/how-to-query-multi-echo-files-using-pybids/17212), where this issue has already been discussed more detailed.

yarikoptic:
Closes #654

https://github.com/bids-standard/bids-specification/issues/654 provides more of discussion with a number of people bringing up the  use cases which keep coming about.

As EchoTime is a recommended sidecar file metadata field for any file of 'mri' modality, I decided that it would only be logical for at least existing use cases (T1w) and possibly other cases (to not overspecify T1w) to simply allow for "echo" entity to be used if so necessary to distignuish two files while they do remain of that weighting (e.g. T1w) given the EchoTime values.

### DIFF
--- a/src/schema/rules/files/raw/anat.yaml
+++ b/src/schema/rules/files/raw/anat.yaml
@@ -27,6 +27,7 @@ nonparametric:
     ceagent: optional
     reconstruction: optional
     run: optional
+    echo: optional
     part: optional
 
 parametric:


effigies (approved):
Based on @yarikoptic and @neurolabusc's latest comments[[0]][[1]] on #654, I think this is a reasonable change.

[0]: https://github.com/bids-standard/bids-specification/issues/654#issuecomment-1679095002
[1]: https://github.com/bids-standard/bids-specification/issues/654#issuecomment-1679490284

tsalo (approved):
The discussion in #654 does seem to have converged around T1w instead of MEGRE, so this change looks good to me.

pwighton:
I'm a little late chiming in, but I also support this change.

If there is confusion what the weighting (t1, t2, pd) means when dealing with multiple echos, I think we can say the weighting refers to the first (and highest SNR) echo which, to my knowledge, is the common convention.

Issue #1382, opened 2023-01-13
Add `chunk` entity to MRI datatype


jcohenadad:
### Your idea

The `chunk` entity was introduced for the 'Microscopy' datatype in BEP031 https://github.com/bids-standard/bids-specification/pull/881. Given that the acquisition of multiple 'chunks' is not uncommon in MRI (eg: different slabs of the brain or spinal cord that are then stitched together), I would welcome having the `chunk` entity available for the MRI community as well.

Related issues:
- https://github.com/bids-standard/bids-validator/issues/1582
- https://github.com/bids-standard/bids-specification/issues/935

CPernet:
could be, should the `ChunkTransformationMatrix` be defined in the nii header and json ; how is this usually done?

jcohenadad:
> could be, should the ChunkTransformationMatrix be defined in the nii header and json ; how is this usually done?

no need, this is what the qform is for. This is the beauty of the world coordinate system 😊

CPernet:
sorry, that's what I meant - ie chunks are already aligned , but we will need to explicitly state it

effigies:
I'm +1 on this, but we need to be clear on what this means for curation. There is no plausible way for a validator to actually check that two MRI images share a world space that may or may not overlap.

How do these images come off the scanner? Are they scanner-aligned and require manual curation to bring into a common space? Or are there scanner protocols that will annotate the images with their relative alignment?

We have these options (via [nifti1.h](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h)) for qform/sform codes:

| Code | Name | Description | 
|--|--|--|
| 0 | NIFTI_XFORM_UNKNOWN | Scanner-based anatomical coordinates |
| 1 | NIFTI_XFORM_SCANNER_ANAT | Coordinates aligned to another file's, or to anatomical "truth". |
| 2 | NIFTI_XFORM_ALIGNED_ANAT | Coordinates aligned to Talairach-Tournoux Atlas; (0,0,0)=AC, etc. |
| 3 | NIFTI_XFORM_TALAIRACH | MNI 152 normalized coordinates. |
| 4 | NIFTI_XFORM_MNI_152 | Normalized coordinates (for any general standard template space) |
| 5 | NIFTI_XFORM_TEMPLATE_OTHER ||

[...]

PR #881, opened 2021-09-16
[ENH] BEP031: Microscopy

mariehbourget:
This PR aims to incorporate the Microscopy BEP031 into the BIDS specification.

It follows PR #812 that integrated the new `sample` entity and `samples.tsv` file, and PR #816 still open for addition of animal metadata to the `participants.tsv` columns.

### Link to the rendered draft: https://bids-specification--881.org.readthedocs.build/en/881/04-modality-specific-files/10-microscopy.html 

Thank you to everyone who contributed to this BEP!

Moderators:  @mariehbourget, @jcohenadad

-------
Note to the maintainers:
- As discussed, I minimized the number of files for community review by not using yaml files and macro. The only 2 yaml files I edited/added are `src/schema/entities.yaml` and `src/schema/datatypes/microscopy.yaml` for the "Entities" and "Entity table" appendix.
- We have 2 metadata field already existing in BIDS for which our definition is slightly different (`StationName` and `BodyPart`). From my understanding of the schema and macros, this is something that we will be able to fix after community review with the yaml files. Let me know if that is correct.
- ~There is a "Recommended participant data" section that overlaps with PR #816 regarding `strain` and `strain_rrid` columns in participants.tsv. Its content depends on whether or not those columns will be included in the "general" Participants file section.~
EDIT: The "Recommended participant data" section is now up-to-date following the merge of PR #816.

---
**Post community-review checklist**
- [x] implement macros
  - [x] for filename templates
  - [x] for examples
  - [x] for metadata table
  - [x] for modality suffix table
- [x] add contributors to the [wiki](https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors) (so they can be added to the contributors page)
- [x] update link to bids-examples
- [x] remove remaining commented sections

Bonus:
- Add the [FAQ from the google doc](https://docs.google.com/document/d/1Nlu6QVQrbOQtdtcRarsONbX5SrOubXWBvkV37LRRqrc/edit#bookmark=id.39jsd4wp2g3d) to the bids starter kit

PR #1586, opened 2023-08-12
[ENH] Add "chunk" entity to MRI datatype

valosekj:
The `chunk` entity was originally added for the 'Microscopy' datatype in BEP031 https://github.com/bids-standard/bids-specification/pull/881. This PR adds the `chunk` entity also to the MRI datatype. In the context of MRI, `chunks` refer to individual field-of-views (FOVs), for example, one FOV for the brain and the second FOV for the spinal cord.
Since two different FOVs might be acquired not only for anatomical MRI, I added the `chunk` entity also for DWI and fMRI.

Resolves: #1382

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -51,9 +51,12 @@ chunk:
   name: chunk
   display_name: Chunk
   description: |
-    The `chunk-<index>` key/value pair is used to distinguish between different regions,
-    2D images or 3D volumes files,
-    of the same physical sample with different fields of view acquired in the same imaging experiment.
+    The `chunk-<index>` entity is used to distinguish between different regions, 2D images (for example Microscopy),
+    3D volumes files, or 4D volumes files (for example DWI) of the same physical sample with different fields of
+    view acquired in the same imaging experiment.
+
+    In case of MRI data, this entity is used to distinguish between different anatomical regions, for example brain
+    and spinal cord.
   type: string
   format: index
 density:
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -404,6 +404,8 @@ ChunkTransformationMatrix:
     scaling along the second and third axis respectively).
     Note that non-spatial dimensions like time and channel are not included in the
     transformation matrix.
+    In case of MRI data, the ChunkTransformationMatrix corresponds to the NIFTI [qform matrix]
+    (https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html)
   anyOf:
     - type: array
       minItems: 3
--- a/src/schema/rules/files/raw/anat.yaml
+++ b/src/schema/rules/files/raw/anat.yaml
@@ -28,6 +28,7 @@ nonparametric:
     reconstruction: optional
     run: optional
     part: optional
+    chunk: optional
 
 parametric:
   suffixes:
@@ -61,6 +62,7 @@ parametric:
     ceagent: optional
     reconstruction: optional
     run: optional
+    chunk: optional
 
 defacemask:
   suffixes:
@@ -80,6 +82,7 @@ defacemask:
     reconstruction: optional
     run: optional
     modality: optional
+    chunk: optional
 
 multiecho:
   suffixes:
@@ -122,6 +125,7 @@ multiflip:
     echo: optional
     flip: required
     part: optional
+    chunk: optional
 
 multiinversion:
   suffixes:
@@ -142,6 +146,7 @@ multiinversion:
     run: optional
     inversion: required
     part: optional
+    chunk: optional
 
 mp2rage:
   suffixes:
@@ -164,6 +169,7 @@ mp2rage:
     flip: optional
     inversion: required
     part: optional
+    chunk: optional
 
 vfamt:
   suffixes:
@@ -187,6 +193,7 @@ vfamt:
     flip: required
     mtransfer: required
     part: optional
+    chunk: optional
 
 mtr:
   suffixes:
@@ -207,3 +214,4 @@ mtr:
     run: optional
     mtransfer: required
     part: optional
+    chunk: optional
--- a/src/schema/rules/files/raw/dwi.yaml
+++ b/src/schema/rules/files/raw/dwi.yaml
@@ -18,6 +18,7 @@ dwi:
     direction: optional
     run: optional
     part: optional
+    chunk: optional
 
 sbref:
   suffixes:
@@ -36,3 +37,4 @@ sbref:
     direction: optional
     run: optional
     part: optional
+    chunk: optional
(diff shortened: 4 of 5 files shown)


Remi-Gau:
@valosekj would it be possible to work on an example dataset to add to bids-example like it was done when the task entity was added for anat: https://github.com/bids-standard/bids-examples/issues/341?

src/schema/objects/metadata.yaml
```diff
@@ -404,6 +404,8 @@ ChunkTransformationMatrix:
     scaling along the second and third axis respectively).
     Note that non-spatial dimensions like time and channel are not included in the
     transformation matrix.
+    In case of MRI data, the ChunkTransformationMatrix corresponds to the NIFTI [qform matrix]
+    (https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html)
```
effigies, line 408: Currently, this is only going to show up in the microscopy section, which does not seem very helfpul. If we want to this to be applicable to MRI data, then we need to add it to their sidecars.

I'm also unclear on what this means. Does this mean that ChunkTransformationMatrix must be equal to the the qform? Or that the qform is used and therefore `ChunkTransformationMatrix` does not apply to NIfTI images? (If so, then we should just remove this addition.)

Does this need a rule, such that `qform_code != 0` (or `== 1`)? What about `sform`? Can that be anything, or must that be identical to the `qform`, or have `sform_code == 0`?

src/schema/objects/entities.yaml
```diff
@@ -51,9 +51,12 @@ chunk:
...
   description: |
-    The `chunk-<index>` key/value pair is used to distinguish between different regions,
-    2D images or 3D volumes files,
-    of the same physical sample with different fields of view acquired in the same imaging experiment.
+    The `chunk-<index>` entity is used to distinguish between different regions, 2D images (for example Microscopy),
+    3D volumes files, or 4D volumes files (for example DWI) of the same physical sample with different fields of
+    view acquired in the same imaging experiment.
+
+    In case of MRI data, this entity is used to distinguish between different anatomical regions, for example brain
+    and spinal cord.
```
effigies, line 59: What do you think of this wording?

```suggestion
    The `chunk-<index>` key/value pair is used to distinguish between images of
    the same physical sample with different fields of view acquired in the same
    imaging experiment.
    This entity applies to collections of 2D images, 3D volumes or 4D volume series
    (for example, diffusion weighted images), and may be used to indicate different
    anatomical structures or regions of the same structure.
```

effigies (commented):
Broadly makes sense. Do we want to bring over any of the sample metadata (such as `BodyPart`) from micr? If the main purpose in MRI is to distinguish spinal cord from brain, annotating that consistently would presumably be helpful.

https://github.com/bids-standard/bids-specification/blob/e1f853873036e2079d39a4fcb1357a04c8afbb80/src/schema/rules/sidecars/micr.yaml#L46-L69

It would go into https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/sidecars/mri.yaml.

Also, would this apply to fieldmap files as well?

Remi-Gau:
Note. Bringing some of the metadata like bodypart may be easier to do on its own and done via a separate PR.

effigies:
See #1593.

valosekj:
> @valosekj would it be possible to work on an example dataset to add to bids-example like it was done when the task entity was added for anat: [bids-standard/bids-examples#341](https://github.com/bids-standard/bids-examples/issues/341)?

Sure! Should I add the example dataset to this PR or create a separate PR?

effigies:
Examples should be submitted as PRs to https://github.com/bids-standard/bids-examples/.

Please see https://github.com/bids-standard/bids-examples/blob/master/CONTRIBUTING.md for guidance.

src/schema/objects/metadata.yaml
jcohenadad: Agreed with @effigies, there is no need to duplicate the already existing information from the qform. So in the case of MRI, I would suggest to have this value set to `n/a` maybe?

effigies: ChunkTransformationMatrix already isn't defined for MRI, so no need to even call it `n/a`. It can just be absent.

valosekj: Okay! I removed the MRI paragraph from the ChunkTransformationMatrix description in https://github.com/bids-standard/bids-specification/commit/d1e4f7c8a2b5d91c56805384a7e9e5f805257438.

valosekj:
> @valosekj would it be possible to work on an example dataset to add to bids-example like it was done when the task entity was added for anat: [bids-standard/bids-examples#341](https://github.com/bids-standard/bids-examples/issues/341)?

Example dataset added in https://github.com/bids-standard/bids-examples/pull/405

valosekj:
@effigies and @Remi-Gau, is anything else required, or can this PR be merged? 
Note that I will bring some of the metadata like bodypart in a separate PR, as proposed in https://github.com/bids-standard/bids-specification/pull/1586#issuecomment-1688316196.

Remi-Gau:
I think that @effigies [comment](https://github.com/bids-standard/bids-specification/pull/1586#pullrequestreview-1589549811) is worth considering:

> would this apply to fieldmap files as well?

Don't know how much sense it would make for ASL but for fieldmap it may make sense to support chunck: even if in theory we could just rely on the `IntendedFor` field to say that a given fieldmap is for this or that chunk it may make it more explicit if we allow this in the filename.

jcohenadad:
> Don't know how much sense it would make for ASL but for fieldmap it may make sense to support chunck: even if in theory we could just rely on the IntendedFor field to say that a given fieldmap is for this or that chunk it may make it more explicit if we allow this in the filename.

Yes, good point. It is possible to acquire different chunks of EPIs, with their associated B0 fieldmap.

valosekj:
> I think that @effigies [comment](https://github.com/bids-standard/bids-specification/pull/1586#pullrequestreview-1589549811) is worth considering:
> 
> > would this apply to fieldmap files as well?
> 
> Don't know how much sense it would make for ASL but for fieldmap it may make sense to support chunck: even if in theory we could just rely on the `IntendedFor` field to say that a given fieldmap is for this or that chunk it may make it more explicit if we allow this in the filename.

Good point! I added the `chunk` entity to fieldmaps in https://github.com/bids-standard/bids-specification/pull/1586/commits/901fe7fa621b992e17e07c397562cfe4fe6bd79d.
Also, I added `chunk` to multiecho in https://github.com/bids-standard/bids-specification/pull/1586/commits/f6282cc9dee0026d4055ba3bd68da22f5f1e1c19 (I originally missed it).

Remi-Gau (approved):

Remi-Gau:
> Should this be extended to ASL images as well?

I am struggling to imagine how ASL on several chunks would be done in a way that "make sense", but one could also say the same for func. Also I have been known to lack imagination... 
So I am fine either way.

> We should not hold up on that.

Whether it should be added or not should not stop from merging this PR now.

effigies:
For BOLD and ASL, I think you would stitch together statistical results, not individual volumes.

Issue #1573, opened 2023-08-07
allow Levels values (in sidecar for TSV files) to be objects


Remi-Gau:
### Your idea

was raised during a meeting at OHBM that some project invovled with phenotypic annotation ([neurobagel](https://neurobagel.org/)) may greatly benefit if the data type for [TermURL](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.metadata.TermURL) (in the sidecar json of TSV files) could be a json object instead of just string. 

Opening an issue here to to start the conversation.

@ericearl @jbpoline @nikhil153 @surchs @michellewang @bcmcpher

Remi-Gau:
@surchs 

Could you provide an example of what the object could look like?

ericearl:
@Remi-Gau Would a JSON or a HED object do alright?

surchs:
Hey, sorry for the delay here - and thanks @Remi-Gau for creating the issue! 

As far as I understand, the `"Levels"` key in e.g. a `participants.json` sidecar can only have 'string' values (https://bids-specification.readthedocs.io/en/stable/glossary.html#levels-metadata) [like described in the docs](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#phenotypic-and-assessment-data):

```json
{
  "MeasurementToolMetadata": {
    "Description": "Adult ADHD Clinical Diagnostic Scale V1.2",
    "TermURL": "https://www.cognitiveatlas.org/task/id/trm_5586ff878155d"
  }
  "adhd_c_dx": {
    "Description": "As child met A, B, C, D, E and F diagnostic criteria",
    "Levels": {
      "1": "YES",
      "2": "NO"
    }
  }
}
```

The problem is that this only allows us to put a human readable description (e.g. `"1": "YES"` **OR** a machine readable unique identifier there (e.g. `"1": "https://myhappyontology.org/vocab/controlled_term_yes"`). But both are important.

So what we would like to propose is to turn the allowed values for `"Levels"` keys from only `"string"` to either `"string"` or `"object"`. Then we could do something like this:

```json
{
  "MeasurementToolMetadata": {
    "Description": "Adult ADHD Clinical Diagnostic Scale V1.2",
    "TermURL": "https://www.cognitiveatlas.org/task/id/trm_5586ff878155d"
  }
  "adhd_c_dx": {
    "Description": "As child met A, B, C, D, E and F diagnostic criteria",
    "Levels": {
      "1": {
[...]

PR #1524, opened 2023-06-23
[ENH] Propose to use `*_channels.json` for the description of coordinate systems

JuliusWelzel:
Hello, 

as discussed [before](https://github.com/bids-standard/bids-specification/issues/1488), we would like to move the description of coordinate system information for motion data from the `*_motion.json` to the `*_channels.json` file.

@sjeung already implemented this in her [example datasets](https://github.com/sjeung/bids-examples/commit/acc012bea750c5b638fe67f8e2f6bb0327f1f2eb), I will not get around to do this before my holiday :(
We updated the schema accordingly.

Please check if the PR seems ok.

Best,
Julius

PR #1591, opened 2023-08-22
[ENH] Add Level objects to channels.json for motion

effigies:
- Replaces #1524.

This was initially comments on #1524, but it became a lot and easier to just make a proposal. Making pull requests against other people's branches often fails, so might as well just make an alternative PR.

The things worth noting:

1) This is essentially taking #1573 as accepted. I'm fine with it, but this is a change from the current spec, where these fields MUST be strings. This probably needs a PR on its own. 
image
2) I think adding `"n/a"` is strictly worse than omitting a field. In one case I can test for presence.
3) "ReferenceFrameDescription" is redundant when it is in `sidecar.reference_frame.Levels.<rf_name>` object. I just changed it to Description.


sjeung:
Thank you, I agree with all of the proposed changes. 

Example data sets in [this PR](https://github.com/bids-standard/bids-examples/pull/418) reflects them already (renamed ReferenceFrameDescription -> Description). 
Once Julius approves my change to spec [this PR](https://github.com/bids-standard/bids-specification/pull/1524) will be up to date with it too. (Removed unnecessary n/a options and renamed ReferenceFrameDescription -> Description)


> 1. This is essentially taking [allow Levels values (in sidecar for TSV files) to be objects #1573](https://github.com/bids-standard/bids-specification/issues/1573) as accepted. I'm fine with it, but this is a change from the current spec, where these fields MUST be strings. This probably needs a PR on its own.

Do you plan to open this one or should I have a look into implementing this?

effigies:
> Do you plan to open this one or should I have a look into implementing this?

That was done in #1603.

src/schema/objects/columns.yaml
```diff
@@ -377,6 +377,18 @@ reference__ieeg:
     - type: string
       enum:
         - n/a
+reference_frame:
+  name: reference_frame
+  display_name: Reference frame
+  description: |
+    Specification of a reference frame in which the motion data are to be interpreted.
+    The description of the levels in `*_channels.json` SHOULD use `RotationRule`,
+    `RotationOrder` and `SpatialAxis` for the specification.
```
sappelhoff, line 386:
```suggestion
    `RotationOrder`, and `SpatialAxis`, and MAY use `Description` for the specification.
```

sappelhoff (approved):
I think mentioning `Description` is a good idea, but feel free to ignore if you disagree. Anyhow, +1 to merge

Remi-Gau (approved):

Issue #1569, opened 2023-08-01
BIDS-BodyPart extracted from BEP025


jomasator2:
### Your idea

To whom it may concern,
I am Jose Manuel Saborit. Responsible for BEP025. I wanted to inform you that I am currently working on the simplification of BEP025 to include it in other more simplified BEPs. Currently, I have a document ready to upload that would correspond to add the tag body parts to BIDS. This document not only determines that a tag has to be added but also specifies various aspects of when it is possible to use it, what rules have to be followed and whether this change affects the associated metadata. 

I would therefore like to include a new BEP for this feature to be extended to other parts of the body.

I remain at your disposal for any consultation and confirmation in order to upload this BEP. Best regards.

sappelhoff:
is this issue perhaps also a duplicate of:

- #1275 

@jomasator2?

jomasator2:
Jomasator2 is my new account, I lost the previous one and I couldn't recover my old account. That's why I'm doing a new issue.

jomasator2:
Hello @VisLab , regarding which tags we define this "BodyPart" where we advise using some standard like UMLS snomed-ct, the document dicom PS3.16 - L DICOM Content Mapping Resource, Correspondence of Anatomic Region Codes and Body Part Examined Defined Terms, OMOP,... The idea is to use the coding defined in the above standards or use the descriptive name of the standards. For example, use the code in snomed-tc "39607008" or use the keyword "lung".

If you want to add this information in participants.tsv I would add it as a list of anatomical parts in keyword format and give it the meaning of what anatomical parts are present in each subject. Always with an informative character to have an initial discretisation. The reason why I put it this way is because a study can have different anatomical parts. In the new BEP, I support the storage of this data in the "_scans.tsv" since it is data that is integrally related to the image.

I don't know if this information can help in accommodating the "BodyPart" tag in "participants.tsv".

Any suggestions are welcome.

VisLab:
To clarify, we aren't trying to accommodate `BodyPart` in `participants.tsv` in particular.  It is up to the user in which files `.tsv` to place HED columns.  

From the validation point of view, it is a matter of feeding the `.tsv` and their associated `.json` to the HED validator... they are all validated in the same way.  We do need to figure out what HED tags should be used to encode whatever is being standardly being placed in the columns of these `.tsv` files.

We would appreciate links to example datasets of how the information you mentioned is being incorporated in a BIDS dataset.
[...]

PR #1593, opened 2023-08-24
ENH: Add sample metadata to MRI and PET

effigies:
Adds `BodyPart`, `BodyPartDetails` and `BodyPartDetailsOntology` to MRI and PET imaging, taken from additions in BEP 031 - Microscopy. `BodyPart` corresponds to a DICOM tag (0018, 0015 `Body Part Examined`), while the others are for additional annotation when this is insufficient detail.

The more straightforward part of #1396. This complements #1586, which proposes to use the `chunk-<index>` entity to identify different fields of view in MRI, which we expect will most commonly be used for spinal cord imaging.

This also lays some groundwork for BEP22 (MRS), which is waiting on a release with motion before starting their final review process.

### DIFF
--- a/src/schema/rules/sidecars/mri.yaml
+++ b/src/schema/rules/sidecars/mri.yaml
@@ -35,6 +35,16 @@ MRIHardware:
     MatrixCoilMode: recommended
     CoilCombinationMethod: recommended
 
+MRISample:
+  selectors:
+    - modality == "mri"
+  fields:
+    BodyPart:
+      level: optional
+      description_addendum: Corresponds to DICOM Tag 0018, 0015 `Body Part Examined`.
+    BodyPartDetails: optional
+    BodyPartDetailsOntology: optional
+
 MRIScannerHardwareASL:
   selectors:
     - datatype == "perf"
--- a/src/schema/rules/sidecars/pet.yaml
+++ b/src/schema/rules/sidecars/pet.yaml
@@ -35,6 +35,17 @@ PETInstitutionInformation:
       level: recommended
       description_addendum: Corresponds to DICOM Tag 0008, 1040 `Institutional Department Name`.
 
+PETSample:
+  selectors:
+    - modality == "pet"
+    - suffix == "pet"
+  fields:
+    BodyPart:
+      level: optional
+      description_addendum: Corresponds to DICOM Tag 0018, 0015 `Body Part Examined`.
+    BodyPartDetails: optional
+    BodyPartDetailsOntology: optional
+
 PETRadioChemistry:
   selectors:
     - datatype == "pet"
--- a/src/modality-specific-files/magnetic-resonance-imaging-data.md
+++ b/src/modality-specific-files/magnetic-resonance-imaging-data.md
@@ -162,6 +162,18 @@ A guide for using macros can be found at
 -->
 {{ MACROS___make_sidecar_table("mri.MRIEchoPlanarImagingAndB0Mapping") }}
 
+#### Tissue description
+
+<!-- This block generates a metadata table.
+These tables are defined in
+  src/schema/rules/sidecars
+The definitions of the fields specified in these tables may be found in
+  src/schema/objects/metadata.yaml
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_sidecar_table("mri.MRISample") }}
+
 ## Anatomy imaging data
 
 Anatomy MRI sequences measure static, structural features of the brain.
--- a/src/modality-specific-files/positron-emission-tomography.md
+++ b/src/modality-specific-files/positron-emission-tomography.md
@@ -216,6 +216,18 @@ A guide for using macros can be found at
 
 {{ MACROS___make_sidecar_table("pet.PETInstitutionInformation") }}
 
+#### Tissue description
+
+<!-- This block generates a metadata table.
+These tables are defined in
+  src/schema/rules/sidecars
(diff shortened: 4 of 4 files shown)


src/schema/rules/sidecars/mri.yaml
```diff
@@ -35,6 +35,14 @@ MRIHardware:
     MatrixCoilMode: recommended
     CoilCombinationMethod: recommended
 
+MRISample:
+  selectors:
+    - modality == "mri"
+  fields:
+    BodyPart: optional
```
tsalo, line 42: Why make BodyPart optional for MRI, but recommended for PET?

effigies: Unclear. @Remi-Gau?

Remi-Gau: can't remember
would say let's make them optional for both

src/schema/rules/sidecars/pet.yaml
```diff
@@ -35,6 +35,17 @@ PETInstitutionInformation:
       level: recommended
       description_addendum: Corresponds to DICOM Tag 0008, 1040 `Institutional Department Name`.
 
+PETSample:
+  selectors:
+    - modality == "pet"
+    - suffix == "pet"
+  fields:
+    BodyPart:
+      level: recommended
```
effigies, line 44:
```suggestion
      level: optional
```

src/schema/rules/sidecars/mri.yaml
effigies: Should the DICOM tag be mentioned in MRI as well?

Remi-Gau, line 42:
```suggestion
   BodyPart:
      level: optional
      description_addendum: Corresponds to DICOM Tag 0018, 0015 `Body Part Examined`.
```

effigies:
@neurolabusc I would appreciate your review on this one. I understood from #1396 that this was the easy bit, but I don't want to presume too far on a couple posts from 7 months ago.

neurolabusc:
@effigies I only have minor comments.
 

1. For `BodyPart` I do think that any field that corresponds to a DICOM tag should have a validation dataset that includes both the source DICOM as well as the expected BIDS. Examples of source DICOMs could include [UIH](https://github.com/neurolabusc/dcm_qa_uih) data where the body part is `HEAD` or [Siemens](https://github.com/neurolabusc/dcm_qa_xa30) data where the body part is `BRAIN`. I assume that this field is expected to be all capitals and correspond to the values in [DICOM part 16 Table L1](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_L.html#chapter_L). My sense is that providing both a source DICOM and the desired BIDS provides a Rosetta Stone, providing a concrete mapping from one structure to the other.
2. `BodyPartDetails` and `BodyPartDetailsOntology` are loosely defined - are values completely up to user, are values expected to be all capitals. Should these generally follow the style of the DICOM `BodyPart`?

VisLab:
I am having trouble unwinding what exactly is being proposed in this and #1569.  Would it be possible for someone to recap what is being proposed in the two PRs and what is already in BIDS?

I generally support adding `BodyPart` and some sort of `BodyPartID` with a corresponding vocabulary identification. DICOM terminology is widely used but it is flat and isn't a formal ontology. Also terms overlap but are not identical to those found in some actual anatomy ontologies such as Uberon. Maybe a better word than `BodyPartDetailsOntology` could be found? Could a list of potential vocabularies be suggested?

A second question is that `BodyPart` has utility not only as a designator for where imaging is being applied but in other contexts. In HED, for example, a body part tag might be used to indicate what part of the body is engaging in an participant action during the experiment. `BodyPart` might also be used to indicate locations affected by a diagnosis. I don't think that `BodyPart` should necessarily be made equivalent to DICOM `Body Part Examined`.

effigies:
@neurolabusc:

>     1. For `BodyPart` I do think that any field that corresponds to a DICOM tag should have a validation dataset that includes both the source DICOM as well as the expected BIDS. Examples of source DICOMs could include [UIH](https://github.com/neurolabusc/dcm_qa_uih) data where the body part is `HEAD` or [Siemens](https://github.com/neurolabusc/dcm_qa_xa30) data where the body part is `BRAIN`.

Sounds good. Would you like a PR against the QA or dcm2niix repositories, or just an agreement that merging this means you could move in that direction?

> I assume that this field is expected to be all capitals and correspond to the values in [DICOM part 16 Table L1](https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_L.html#chapter_L). My sense is that providing both a source DICOM and the desired BIDS provides a Rosetta Stone, providing a concrete mapping from one structure to the other.

Yes, I think directly using the values of `(0018,0015) Body Part Examined` makes sense, and that appears to be all capitals, corresponding to the values of part 16 table L1, column "Body Part Examined".

>     2. `BodyPartDetails` and `BodyPartDetailsOntology` are loosely defined - are values completely up to user, are values expected to be all capitals. Should these generally follow the style of the DICOM `BodyPart`?

My interpretation is that this would be up to an annotator (person or tool). Given that Table L1 lists SNOMED terms, it would be reasonable for `dcm2niix` to select https://bioportal.bioontology.org/ontologies/SNOMEDCT as its ontology and use "Code meaning" for `BodyPartDetails`.

@VisLab Have your questions been adequately addressed?

VisLab:
Yes this was helpful. Thx

sappelhoff (approved):

Remi-Gau (approved):

PR #1595, opened 2023-08-24
[SCHEMA] Update pet.yaml PlasmaAvail check.

rwblair:
BloodPlasma requirement in tabular data didn't check if the corresponding json was true. All jsons are required to have the PlasmaAvail field, so any dataset with PlasmaAvail false would fail this check.

### DIFF
--- a/src/schema/rules/tabular_data/pet.yaml
+++ b/src/schema/rules/tabular_data/pet.yaml
@@ -28,6 +28,7 @@ BloodPlasma:
     - suffix == "blood"
     - extension == ".tsv"
     - '"PlasmaAvail" in sidecar'
+    - sidecar.PlasmaAvail == true
   columns:
     plasma_radioactivity: required
 


src/schema/rules/tabular_data/pet.yaml
```diff
@@ -28,6 +28,7 @@ BloodPlasma:
     - suffix == "blood"
     - extension == ".tsv"
     - '"PlasmaAvail" in sidecar'
+    - sidecar.PlasmaAvail == true
```
effigies, line 31: Would this be any more efficient, since an absent `PlasmaAvail` will return `null`?

```suggestion
    - sidecar.PlasmaAvail == true
```

tsalo (approved):
The change makes sense to me. Since `PlasmaAvail` is a required field, though, do we need `'"PlasmaAvail" in sidecar'`?

effigies (approved):
We also need to change L17 from `required` to `recommended`, which is inconsistent with L44.

I checked back on https://bids-specification.readthedocs.io/en/v1.6.0/04-modality-specific-files/09-positron-emission-tomography.html#blood-recording-data, and the rule is correct, the level_addendum is wrong.

This and my suggestion can be done in this PR or another one, if you just want this in.

PR #1603, opened 2023-08-29
[ENH] allow Levels values to be objects

### LINKED ISSUE #1573: allow Levels values (in sidecar for TSV files) to be objects

Remi-Gau:
- closes #1573


src/common-principles.md
```diff
@@ -513,6 +513,28 @@ Example:
...
+like in the example below.
+
+```JSON
+{
+    "sex": {
+        "Description": "sex of the participant as reported by the participant",
+        "Levels": {
+            "M": {
+                "Description":"male",
+                "TermURL": "TODO"
```
Remi-Gau, line 527: update

src/common-principles.md
```diff
@@ -513,6 +513,28 @@ Example:
...
+            },
+            "F": {
+                "Description":"female",
+                "TermURL": "TODO"
```
Remi-Gau, line 531: update

Remi-Gau:
@surchs @ericearl @alyssadai

Remi-Gau:
@effigies am I correct that this is one aspect of the spec that has not been schemaified yet?

Remi-Gau:
Should the examples of the json sidecar [for participants.tsv](https://bids-specification.readthedocs.io/en/latest/modality-agnostic-files.html#participants-file) and for [phenotypic assessments](https://bids-specification.readthedocs.io/en/latest/modality-agnostic-files.html#phenotypic-and-assessment-data) be updated to also include this way of describing levels?

src/common-principles.md
```diff
@@ -513,6 +513,28 @@ Example:
 }
 ```
 
+Each level can be described with a string as in the example above,
+or with an object containing the fields `Description` and `TermURL`
```
surchs, line 517: Do we expect BIDS users to know what `TermURL` expects? How about:

```suggestion
or with an object containing the fields `Description` and `TermURL`.
If you use the object format, the value of the `TermURL` key should
be a URL to a term from a controlled vocabulary that describes the
level. For example: https://www.ncbi.nlm.nih.gov/mesh/68008297 for 
"male". The value of the `Description` key should then correspond
to the human readable name of this term (or to the preferred name
if several options exist in the controlled vocabulary). For example: "Male".
```

src/common-principles.md
surchs, line 527:
```suggestion
                "Description": "Male",
                "TermURL": "https://www.ncbi.nlm.nih.gov/mesh/68008297"
```

src/common-principles.md
surchs, line 531:
```suggestion
                "Description": "Female",
                "TermURL": "https://www.ncbi.nlm.nih.gov/mesh/68005260"
```

effigies:
> @effigies am I correct that this is one aspect of the spec that has not been schemaified yet?

Correct. It's a weird state because what we really need to say is that, if the suffix is TSV, then additional properties of sidecars have the form of an object with certain allowed terms. It would have been much more convenient if we had made the column descriptions an object in a single field. (Further, there generally aren't rules against extra metadata, so it's actually that we need to create a typed metadata field dynamically based on the headers in the TSV.)

src/common-principles.md
effigies: If we're going to improve the definition of TermURL, we should do it in https://bids-specification.readthedocs.io/en/stable/glossary.html#termurl-metadata so it appears in the glossary and any tables it's defined in.

 https://bids-specification.readthedocs.io/en/stable/glossary.html#termurl-metadata

Remi-Gau: I could only transfer one sentence to the TermURL generic definition, the rest of @surchs suggestion feels too specific to this example I think.

effigies (approved):

surchs (approved):
:tada: thanks a lot @Remi-Gau. One free bagel on us!

sappelhoff (approved):
LGTM! Thanks a lot @Remi-Gau

PR #1605, opened 2023-08-31
FIX: Do not require notch frequencies to be parsed as numbers, accommodating multiples

effigies:
While schematizing, the `notch` column was defined as a number. In the case of iEEG at least, the BEP leads had submitted examples where `notch` took the form `[F1, F2, F3]`. This PR recognizes this as existing practice and suggests that curators and tools adopt `[F1, F2, ...]` for interoperability, but does not require it.

cc @bids-standard/raw-electrophys for comment.

Closes https://github.com/bids-standard/bids-examples/issues/395.

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -302,13 +302,11 @@ notch:
   display_name: Notch frequencies
   description: |
     Frequencies used for the notch filter applied to the channel, in Hz.
+    If the low and high frequencies are supplied in addition to the center
+    frequency, it is RECOMMENDED that they take the form `[low, center, high]`,
+    for example `[60, 120, 180]`.
     If no notch filter applied, use `n/a`.
-  anyOf:
-    - type: number
-      unit: Hz
-    - type: string
-      enum:
-        - n/a
+  type: string
 onset:
   name: onset
   display_name: Event onset
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -302,9 +302,8 @@ notch:
   display_name: Notch frequencies
   description: |
     Frequencies used for the notch filter applied to the channel, in Hz.
-    If the low and high frequencies are supplied in addition to the center
-    frequency, it is RECOMMENDED that they take the form `[low, center, high]`,
-    for example `[60, 120, 180]`.
+    If the low and high frequencies are supplied in addition to the center frequency,
+    they MAY take the form `[low, center, high]`, for example `[60, 120, 180]`.
     If no notch filter applied, use `n/a`.
   type: string
 onset:
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -302,9 +302,10 @@ notch:
   display_name: Notch frequencies
   description: |
     Frequencies used for the notch filter applied to the channel, in Hz.
-    If the low and high frequencies are supplied in addition to the center frequency,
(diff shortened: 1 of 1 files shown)


src/schema/objects/columns.yaml
```diff
@@ -302,13 +302,11 @@ notch:
   display_name: Notch frequencies
   description: |
     Frequencies used for the notch filter applied to the channel, in Hz.
+    If the low and high frequencies are supplied in addition to the center
+    frequency, it is RECOMMENDED that they take the form `[low, center, high]`,
```
adam2392, line 306: Why not required? 

Would another format be possible if the low and high frequencies are provided?

src/schema/objects/columns.yaml
```diff
@@ -302,13 +302,11 @@ notch:
...
+    for example `[60, 120, 180]`.
     If no notch filter applied, use `n/a`.
-  anyOf:
-    - type: number
-      unit: Hz
-    - type: string
-      enum:
-        - n/a
+  type: string
```
adam2392, line 309: Is there anyway to include a regex pattern like it does below? I feel like that could be useful for the validator too

adam2392 (commented):
I left some comments cuz this has been an issue for me in the past. Hopefully it helps!

src/schema/objects/columns.yaml
effigies: Because it hasn't previously been required. We have another PR proposing to use comma-separated values for multiple channels in a single TSV column. My goal here is to accommodate existing behavior, not impose a new rule.

If the ephys community wants to adopt that, that's fine. I would suggest then that I simply relax the type here and leave the text unchanged. You or someone else who actually uses these data could lead an effort to get buy-in to have more specific requirements for three-value notch specifications.

src/schema/objects/columns.yaml
effigies: Yes, but it would be ugly. You need to match `SQUARE_LEFT SPACE NUMBER COMMA SPACE NUMBER COMMA SPACE NUMBER SQUARE_RIGHT`

And `NUMBER` is 

https://github.com/bids-standard/bids-specification/blob/eff64847adf07638ab4eb39cacb31e193c13bc43/src/schema/objects/formats.yaml#L26-L30

And you probably want to be forgiving and allow extra spaces around the numbers, so you really want to replace `SPACE NUMBER` with `SPACE* NUMBER SPACE*`.

effigies:
@adam2392 Are you comfortable with the current wording leaving open the possibility of a more structured recommendation/requirement in the future?

effigies:
Thank you for the clarification. Is the current wording better?

dorahermes:
Yes, thank you! Just a minor suggestion is to say: "If notch filters are applied at multiple frequencies, these frequencies MAY ..."

sappelhoff (commented):
I am aware that this PR is mostly to get an already existing practice finally incorporated into the spec, but it irks me that this is not very in line with:

- https://github.com/bids-standard/bids-specification/pull/1483

Do you think we should RECOMMEND specifying the frequencies as a comma separated list like `60, 120, 240`, and clarify that an old method of `[60, 120, 240]` is DEPRECATED?

Either way, I am approving this, because I am fine with my suggestion being dealt with in a separate PR. Thanks everyone.

sappelhoff (approved):

christinerogers:
@Andesha is this something you'd feel comfortable taking a moment to review?

rwblair (approved):

PR #1629, opened 2023-10-12
SCHEMA: Make expression tests more manageable

effigies:
We started with a pretty ambitious expression language, thinking it would not be hard to implement in any language. It turns out that it would require a proper parser, which would probably be an intolerable slowdown to the javascript implementation. Making `null` "falsey" is enough, and we don't use much of the fall-through behavior we originally designed.

A Python implementation will be able to do some type-checking on the expressions we actually have to make sure they stay in the bounds the JS implementation can handle.

### DIFF
--- a/src/schema/meta/expression_tests.yaml
+++ b/src/schema/meta/expression_tests.yaml
@@ -61,9 +61,9 @@
 - expression: '"VolumeTiming" in null'
   result: false
 - expression: exists(null, "bids-uri")
-  result: false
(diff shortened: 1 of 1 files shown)


effigies:
cc @rwblair

rwblair:
Do we know if anyone else uses these expression checks?

rwblair (approved):

effigies:
No, I don't think anybody does.

PR #1632, opened 2023-10-18
Allow for .nwb standard/file format  to be used for "micr"

yarikoptic:
See https://pynwb.readthedocs.io/en/stable/tutorials/domain/ophys.html for support of various microscopy modalities within NWB.  As NWB is already supported for other modalities, there is some support for microscopy in NWB in various tools, and as there are datasets with .nwb -- it makes sense to allow for .nwb to be used for microscopy data as well where appropriate.

### DIFF
--- a/src/schema/rules/files/raw/micr.yaml
+++ b/src/schema/rules/files/raw/micr.yaml
@@ -25,6 +25,7 @@ microscopy:
     - .png
     - .tif
     - .json
+    - .nwb
   datatypes:
     - micr
   entities:


satra:
this may be more of a question of physiology versus microscopy. in bids micr is focused on anatomy-based info not physiology. and i don't think nwb quite fits into that space. ophys doesn't fit into micr as far as i can tell. in bids ophys would go into func.

satra:
or `phys` if that was to become a new modality.

yarikoptic:
## re "functional vs anatomical"

related discussions to tease apart those two notions of modality (MRI, micr, ...) and data type (anat vs func):
- https://github.com/INCF/neuroscience-data-structure/issues/18
- https://github.com/bids-standard/bids-2-devel/issues/55

Note that nothing in https://bids-specificabtrcitetcjtion.readthedocs.io/en/stable/04-modality-specific-files/10-microscopy.html explicitly restricts to anatomy (thcbhbnieljus allows for extra), although indeed lacks any annotation for temporal component and events (requires "extending").
Overall the fact that we have `func/` ATM is an artifact of early days orientation toward MRI, and peculiarity that some modalities are purely functional (e.g. EEG, MEG) thus not even raising the question.  But then if we look at [PET](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/09-positron-emission-tomography.html) -- we can have a single volume PET, or "functional PET" for which `_events.tsv` is then required.

So I could argue about similarity to this case here

## re phys vs micr

ATM microscopy is for the modality, and it already supports 2PE which I guess also falls into "phys" domain.
As all other "modality/" folders are really instrumentation modality, I would argue that we better not place microscopy data under electrophysiology folder.

sappelhoff:
has there been a request by a user / a particular group of users to make this happen? 

If not, I would wait with introducing such a change, and remain "use case driven".

Especially when an argument of being "consistent" isn't entirely clear (as I get the impression, reading the discussion above)

yarikoptic:
> has there been a request by a user / a particular group of users to make this happen?

it was brought up as a side-topic in discussion of BEP032 (@bep032) .

> remain "use case driven".

LGTM. @bendichter, are you aware of any DANDI set with nwb microscopy data which we would run to while approaching conversion of DANDI dandisets to BIDS?

bendichter:
Yes, there are lots of ophys dandisets, including 1p, 2p, and more. Here are some good examples:

* [000037 - Allen Institute Openscope - Responses to inconsistent stimuli in somata and distal apical dendrites in primary visual cortex](https://dandiarchive.org/dandiset/000037)
* [000054 - Plitt & Giocomo (2021) Experience Dependent Contextual Codes in the Hippocampus. Nat Neuro](https://dandiarchive.org/dandiset/000054)
* [000711 - Allen Institute - Visual Behavior - Optical Physiology](https://dandiarchive.org/dandiset/000711)

[Among others](https://dandiarchive.org/dandiset/search?search=ImagingPlane)...

yarikoptic:
There is an ongoing effort of https://github.com/con/nwb2bids . Not yet sure how we would deal with multimodal nwb files like in https://dandiarchive.org/dandiset/000037/0.240209.1623/files?location=sub-408021&page=1 but I remain of opinion that NWB is a viable and used format for `micr`... I will rebase and take it out of the draft.

effigies:
What benefits does `micr/*.nwb` have over `micr/*.ome.zarr`? There is a cost to tool writers to proliferating formats within a modality, hence why we do not support MINC.

That said, I know nothing of this space, and IMO this needs input from microscopy contributors. @jcohenadad do you have thoughts here?

yarikoptic:
tagging @just-meng who is working with 2photon imaging data and might be interested to help us out!

yarikoptic:
ok, celebrating 2 years for this PR.  What should we do? I see us 
- C1: promoting 2-photon calcium imaging into `func/`
  - found require HEAVY adjustment since we have `func/` under [Magnetic Resonance Imaging](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html) which 2P is not
- C2: keeping under `micr/`
  - also adding `_events.{tsv,gz}` to cover collection of events/annotations acquired during acquisition (in time for func or not) 
- C3: establishing entire new `ophys/` modality alike `func/` but for "optical physiology"  

any preference or other choices you see @effigies @satra ?

PR #1641, opened 2023-10-26
[SCHEMA] - Adding validator to CLI

suleesan:
In main.py, working to add CLI hook for validator function.

### DIFF
--- a/tools/schemacode/bidsschematools/__main__.py
+++ b/tools/schemacode/bidsschematools/__main__.py
@@ -4,7 +4,7 @@ import os
 import click
 
 from .schema import export_schema, load_schema
-
+from .validator import validate_bids
 
 @click.group()
 @click.option("-v", "--verbose", count=True)
@@ -21,7 +21,7 @@ def export(ctx, schema, output):
     """Export BIDS schema to JSON document"""
     logger = logging.getLogger("bidsschematools")
     schema = load_schema(schema)
-    text = export_schema(schema)
+    text = export_schema(schema) 
     if output == "-":
         logger.debug("Writing to stdout")
         print(text)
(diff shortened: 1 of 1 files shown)


tools/schemacode/bidsschematools/__main__.py
```diff
@@ -32,5 +33,50 @@ def export(ctx, schema, output):
...
+    schema,
+    in_paths,
+    dummy_paths,
+    bids_reference_root,
+    schema_path,
+    bids_version,
+    report_path,
+    suppress_errors,
+    accept_non_bids_dir,
+):
```
effigies, line 58: Okay, looking at most of these, I think we don't really want them. They seem to be aimed at future use cases that don't really exist right now. What I can see is:

```suggestion
@click.option("--schema")
@click.option("--report", is_flag=False, flag_value=True, default=False)
@click.pass_context
def validate(
    ctx,
    schema,
    in_paths,
    report,
):
```

tools/schemacode/bidsschematools/__main__.py
```diff
@@ -32,5 +33,50 @@ def export(ctx, schema, output):
...
+    validation_result = validate_bids(
+        in_paths,
+        dummy_paths,
+        bids_reference_root,
+        schema_path,
+        bids_version,
+        report_path,
+        suppress_errors,
+        accept_non_bids_dir,
+    )
```
effigies, line 72:
```suggestion
    validation_result = validate_bids(
        in_paths,
        schema_path=schema,
        report_path=report,
    )
```

tools/schemacode/bidsschematools/__main__.py
```diff
@@ -32,5 +33,50 @@ def export(ctx, schema, output):
...
+    logger.debug("Printing out path_tracking in validation_result dictionary")
+    if validation_result["path_tracking"]:
+        print(validation_result["path_tracking"])
+
+    # print validation_result["path_tracking"] if not empty
+    # return 0 = success, non-zero = fail use Click exception handling
```
effigies, line 78: At the very top, we'll want to `import sys`, but then we can exit with:

```suggestion
    sys.exit(bool(validation_result["path_tracking"]))
```

effigies (commented):
Great start! I've had a quick look through, and here are some suggestions. Overall, I think there's a lot in this function we don't want to expose to a CLI user.

PR #1672, opened 2023-12-17
SCHEMA: Add file rule for phenotype tables

effigies:
This PR slightly amends how `stem` rules work, by making the stem a [glob](https://www.man7.org/linux/man-pages/man7/glob.7.html) pattern and allowing it to coexist with `datatype`. This hacks 'phenotype' into the `datatype` structure; in the absence of entities, it works correctly for now.

I don't know that this is a long-term solution, but it seems good enough for now with minimal changes required to the schematools or the JS schema validator.

This is intended to be compatible with, but not actually dependent on #1693.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -751,6 +751,10 @@
       "type": "object",
       "properties": {
         "level": { "enum": ["optional", "recommended", "required"] },
+        "datatypes": {
+          "type": "array",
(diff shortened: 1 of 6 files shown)


effigies:
Closed in favor of #1693.

effigies:
@rwblair @tsalo Let me know what you think.

rwblair (approved):

Issue #1646, opened 2023-10-30
Add support for HiP-CT


balbasty:
Dear BIDS community

[Hierarchical Phase-Contrast Tomography (HiP-CT)](https://mecheng.ucl.ac.uk/hip-ct/) is an _ex vivo_ X-ray technique that leverages high-energy X-ray sources -- specifically, the ESRF's [Extremely Brilliant Source (EBS)](https://www.esrf.fr/home/UsersAndScience/Accelerators/ebs---extremely-brilliant-source.html).  HiP-CT is non-destructive, reaches resolutions of 1 to 10 um, and allows specific ROIs to be imaged within the sample without physical sectioning. Although the modality is obviously not widely available, the UCL group has a relatively streamlined acquisition process and we will be collecting data on multiple brains in the coming years.

We would like to add support for HiP-CT in BIDS, and we think it most naturally fits as a new [modality suffix](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/10-microscopy.html#modality-suffixes) within the microscopy extension. This makes sense because micro-CT is already supported as a microscopy suffix. Also, the microscopy extension already includes tags that are required by HiP-CT data (for example, the existence of multiple images of the same samples, potentially with different resolutions and different field-of-views, labeled as `chunk-`). We are therefore proposing to add the `hipCT` suffix to the microscopy extension.

We have uploaded an exampled dataset that follows the proposed convention to DANDI:
https://dandiarchive.org/dandiset/000026/draft/files?location=sub-I58%2Fses-Hip-CT%2Fmicr&page=1
We also provide three neuroglancer links that allow each volume to be visualized online, directly streaming the dandi data:
[...]

PR #1686, opened 2024-01-24
[ENH] Add support for HiP-CT

balbasty:
Resolves: #1646

This PR introduces a new `hipCT` microscopy suffix.

It does **not** make any changes to the required or optional metadata. 
To me, HiP-CT is not a quantitative modality, so I do not see any parameters that would be _required_ for downstream analysis.
Please give feedback if you feel that this is not correct and that hip-ct-specific metadata should be added. 

I am leaving it as a draft PR for now.

cc @satra @ucbpclw


effigies:
This seems fine to me, and there were no objections on the discussion #1646.

@jcohenadad Do you see any problems with this?

balbasty:
Thanks! I removed the draft mode

effigies (approved):

src/modality-specific-files/microscopy.md
```diff
@@ -89,6 +89,7 @@ and a guide for using macros can be found at
          "NLO",
          "OCT",
          "SPIM",
+	 "hipCT",
```
effigies, line 92: Small style cleanup.

```suggestion
         "SPIM",
         "hipCT",
```

effigies:
Note that [by the rules](https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules) we do need a second approving review by a community member. (It does not need to be a maintainer.)

Remi-Gau (approved):

PR #1690, opened 2024-01-29
[ENH] Add TablePosition tag to MRI

### LINKED ISSUE #1643: Consider adding the Relative Table Position information
### Background
The relative table position is essential to convert from the "patient" coordinate system to the "scanner" coordinate system. For our application in MRI, it would allow to know the location of the isocenter in every acquisition, not just in acquisitions where the table was not moved, which would be the origin when converted to RAS+ using the qform affine matrix. This has been a requested feature in an [issue](https://github.com/rordenlab/dcm2niix/issues/726) opened in Dcm2niix for another application; to perform offline gradwarp correction.

The different vendors have different ways of encoding this information in the DICOMs. So there would need to be a discussion about how to unify it.

I have gathered some relevant information here:
From @neurolabusc ([comment](https://github.com/rordenlab/dcm2niix/issues/726#issuecomment-1682473518))
```
(0043,10b2) LO [956.1\320.1\-17.3125\0\75.3604\58.0479] # GE MR Table Position Information
(0019,1014) IS [0\0\0]                                  # Siemens V* TablePosition
(0021,1005) IS [0\0\0]                                  # Siemens XA TablePosition
(2005,143c) FL 0                                        # Philips MRStackTablePosLong
(2005,143d) FL 0                                        # Philips MRStackTablePosLat
(0065,1023) LO [0]                                      # UIH TablePosition
```

From @mr-jaemin ([comment](https://github.com/rordenlab/dcm2niix/issues/726#issuecomment-1683053413)):
GE Private tag (0043,10b2) MR Table Position Information consists of
- isoVectorZ (Distance from ISO to ALIGN)
- landmark
- tablePosition
- dzPETMR
- topOfHead
- distanceFromTopOfHead
Ex) (0043,10b2) LO [956.1\320.1\-17.3125\0\75.3604\58.0479]

### Proposition
All manufacturers include information in the Inferior/Superior direction but others also allow some Posterior/Anterior and Left/Right information. I therefore suggest to have a tag, possibly "RelativeTablePosition" that would be a 3 value vector containing the relative table displacement in RAS+ coordinates from when it was first set to isocenter.

"TablePosition": [x, y, z]

Let me know what you think :)

po09i:
### Description

This PR adds the TablePosition tag to the MRI specification. The tag specifies the location of the table when the acquisition was taken relative to an implementation specific reference point. The table position is defined by an array of 3 numbers corresponding to x, y and z coordinates respectively.

Fixes #1643


po09i:
@tsalo, @erdalkaraca, any feedback on this? I'm happy to make changes if necessary

src/schema/objects/metadata.yaml
```diff
@@ -3366,6 +3366,20 @@ SubjectArtefactDescription:
...
     source of artifacts except cardiac and blinks.
   type: string
+TablePosition:
+  name: TablePosition
+  display_name: Table Position
+  description: |
+    The x, y and z coordinates of the table relative to an implementation
+    specific reference point. The array MUST contain three numeric values
+    corresponding to x, y, and z axis of the coordinate system in that exact
+    order.
```
effigies, line 3376: What coordinate system is this position in? Is it DICOM LPS, or is it the responsibility of the converter software to use RAS for consistency with NIfTI?

I don't fully understand "an implementation specific reference point". I suppose it means that the value cannot be meaningfully compared across datasets unless `ManufacturersModelName` and `SoftwareVersions` are the same. Is it correct or incorrect to say that this is the offset from the origin of the image affine to the scanner isocenter?

@neurolabusc Is this enough information for `dcm2niix` to have an unambiguous directive on what to do?

po09i: > What coordinate system is this position in? Is it DICOM LPS, or is it the responsibility of the converter software to use RAS for consistency with NIfTI?

Since NIfTI is using RAS+ coordinates, I was thinking that to be consistent, it should be using RAS+ as well. I can make it more explicit.

> Is it correct or incorrect to say that this is the offset from the origin of the image affine to the scanner isocenter?

This depends where the origin of the table position is since only the offset is reported. From my point of view, I would make it so that it is the case to be able to know where isocenter is. This implies that we know the origin of the table position (it seems to be at isocenter, at least from my experience on Siemens and Philips), and that the origin of the affine is isocenter if the table position is at its origin. This seems to be my experience.

If we can say with certainty that this is the offset from isocenter, I would go with that. We could even change the name tag to be more explicit (e.g.: affineIsoOffset, ...). Maybe @neurolabusc knows a bit more.

effigies: > Since NIfTI is using RAS+ coordinates, I was thinking that to be consistent, it should be using RAS+ as well. I can make it more explicit.

Agreed. Given that DICOM is LPS, making it explicit that converters will need to adjust would be good.

> This depends where the origin of the table position is since only the offset is reported. From my point of view, I would make it so that it is the case to be able to know where isocenter is. This implies that we know the origin of the table position (it seems to be at isocenter, at least from my experience on Siemens and Philips), and that the origin of the affine is isocenter if the table position is at its origin. This seems to be my experience.

Got it. What about the following text:

> The table position, relative to an implementation-specific reference point, often the isocenter.
> Values must be an array of three distances in millimeters, respectively
> right, anterior and superior (RAS+) of the image affine.

There's some additional text in [(0018,9327)](https://dicom.innolitics.com/ciods/enhanced-ct-image/enhanced-ct-image-multi-frame-functional-groups/52009229/00189326/00189327):

> Shall be relative to the same reference point for all frames in this SOP Instance, but may be different from the reference point in other SOP Instances. Positions as the table moves into the gantry viewed from the front are more negative.

I'm not sure if/how best to include this concept. Possibly "Shall be relative to the same reference point for all images with the same acquisition parameters."? Or maybe best to leave out.

po09i: Great ideas, I will update the PR to reflect these comments. 

> I'm not sure if/how best to include this concept. Possibly "Shall be relative to the same reference point for all images with the same acquisition parameters."? Or maybe best to leave out.

I originally opted to leave it out since the goal is to make it relative to the isocenter.

po09i: I updated the code to reflect the changes. @neurolabusc, do you have any thoughts on this?

neurolabusc: The definition looks clear to me.

src/schema/objects/metadata.yaml
```diff
@@ -3366,6 +3366,20 @@ SubjectArtefactDescription:
...
     source of artifacts except cardiac and blinks.
   type: string
+TablePosition:
+  name: TablePosition
+  display_name: Table Position
+  description: |
+    The table position, relative to an implementation-specific reference point,
+    often the isocenter. Values must be an array of three distances in
+    millimeters, respectively right, anterior and superior (RAS+) of the image
+    affine.
```
effigies, line 3376: Apologies, I think I gave you some slightly unclear text. I think this is clearer, but please feel free to use an alternative that makes sense to you.

```suggestion
    millimeters, respectively right, anterior and superior (RAS+) of the origin
    defined by the image affine.
```

effigies (dismissed):
Sounds good. Let's do this. One small suggestion...

src/schema/objects/metadata.yaml
po09i: Good point! I reworded it a little more

po09i:
@effigies I'm wondering if the coordinate system of the `TablePosition` should be relative to the "scanner" (e.g.: positive/negative if moving into the scanner) rather than relative to the patient coordinate system. The reasoning is that for patients inserted feet first, rather than head first, then the `tablePosition` would have opposite signs when the table moves into the scanner. 

- Using the patient coordinate system makes it easier to use since the `tablePosition` is already in the same space as the image coordinates but feels less intuitive and harder to grasp.
- However, using the scanner coordinate requires to know the `PatientPosition` (e.g.: HFS) to bring it in the same coordinate system as the patient but feels more intuitive.

I'm okay with both definitions but I'm wondering if you have any thoughts on this?

po09i:
Do you have feedback @jcohenadad?

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,23 @@ SubjectArtefactDescription:
...
     source of artifacts except cardiac and blinks.
   type: string
+TablePosition:
+  name: TablePosition
+  display_name: Table Position
+  description: |
+    The table position, relative to an implementation-specific reference point,
+    often the isocenter. Values must be an array of three distances in
+    millimeters. The coordinates are respectively right, anterior and superior
+    (RAS+) as if it were acquired in head first supine (HFS) orientation. In
```
jcohenadad, line 3402: should we maybe say something about the first two values, which, in most cases I guess will be zero

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,24 @@ SubjectArtefactDescription:
...
     source of artifacts except cardiac and blinks.
   type: string
+TablePosition:
+  name: TablePosition
+  display_name: Table Position
+  description: |
+    The table position, relative to an implementation-specific reference point,
+    often the isocenter. Values must be an array (1x3) of three distances in
+    millimeters in absolute coordinates (world coordinates). The
+    absolute coordinates are respectively right, anterior and superior (RAS+)
```
jcohenadad, line 3402: you said it in the previous sentence, no need to repeat I think
```suggestion
    coordinates are respectively right, anterior and superior (RAS+)
```

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,24 @@ SubjectArtefactDescription:
...
+    as if it was acquired in head first supine (HFS) orientation. In
+    other words, moving the table into the scanner will always increase the 3rd
+    value in the array, regardless of the patient positioning (feet
```
jcohenadad, line 3405: I'm wondering if we should just avoid mentioning examples of patient positioning, because in the end TablePosition is independent of patient positioning. The way you defined it is good enough I think (ie: table goes in the scanner means S increases). Adding refs to patient positioning adds confusion I find.

src/schema/objects/metadata.yaml
jcohenadad, line 3402: In general (eg [here](https://slicer.readthedocs.io/en/latest/user_guide/coordinate_systems.html)), we use XYZ for absolute coordinate and RLAPSI for patient coordinate. I'm wondering if we shouldn't use XYZ in the context of table position, since it is defined in the absolute coordinate.

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,24 @@ SubjectArtefactDescription:
...
+    first for example). The origin is defined by the image affine.
```
jcohenadad, line 3406: Really? I thought that the image affine origin was in reference to the scanner isocenter (https://nipy.org/nibabel/coordinate_systems.html). Let's imagine two scenarios: 
1. Patient is installed, isocenter is set around the eyes, table goes in, image is acquired with FOV centered around the eyes, and, let's say, `TablePosition(3)=x`. 
2. Patient is installed, isocenter is set around the chin (which is, let's say -100mm from the eyes), image is acquired with FOV centered around the chin, I would expect `TablePosition(3)=x-100`, so not the same as in the previous scenario, even though both images are acquired with the FOV at isocenter (ie: `affine(3,4)=0`). 

I have to admit it is easy to get confused with this, so it is possible that my demonstration has a glitch somewhere... but in any case, I would find it useful to give similar example scenario to explain how TablePosition is calculated in the BIDS definition.

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,23 @@ SubjectArtefactDescription:
     If this field is set to `"n/a"`, it will be interpreted as absence of major
     source of artifacts except cardiac and blinks.
   type: string
+TablePosition:
+  name: TablePosition
+  display_name: Table Position
+  description: |
+    The table position, relative to an implementation-specific reference point,
+    often the isocenter. Values must be an array of three distances in
```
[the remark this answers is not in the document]
po09i: Fixed here: [2142da3](https://github.com/bids-standard/bids-specification/pull/1690/commits/2142da34091dbe4e7f44e14c2f07c1eedd64c7b4)

src/schema/objects/metadata.yaml
```diff
@@ -3392,6 +3392,23 @@ SubjectArtefactDescription:
...
+    millimeters. The coordinates are respectively right, anterior and superior
+    (RAS+) as if it were acquired in head first supine (HFS) orientation. In
+    other words, moving the table into the scanner will increase the 3rd value
+    in the array (even in the feet first orientation). The origin is defined by
```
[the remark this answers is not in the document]
po09i: In this definition, the table would be dependant on the "absolute" coordinates. I tried to clarify here [2142da3](https://github.com/bids-standard/bids-specification/pull/1690/commits/2142da34091dbe4e7f44e14c2f07c1eedd64c7b4)

src/schema/objects/metadata.yaml
po09i: Great scenario. I took a look at previous acquisitions (brain and spinal cord scans), I almost always see the `TablePosition` DICOM tag (at least on Siemens and Philips) at <|1| on the first scan (localizer). This suggests to me that the tag is an offset once it has been set to isocenter and does not depend on where it was set on the patient.

> Really? I thought that the image affine origin was in reference to the scanner isocenter (https://nipy.org/nibabel/coordinate_systems.html). Let's imagine two scenarios:

I believe the affine origin is isocenter if the TablePosition = [0,0,0], otherwise it is not. Maybe `IsocenterOffset`, `TablePositionOffset` is a better name tag?

po09i:
@effigies I updated the definition to include Julien's comments, what do you think?

po09i:
@tsalo, @erdalkaraca Tagging others so I can merge asap.

neurolabusc:
I really think there should be concrete examples that allow us to validate that different tools (dcm2niix, dicm2nii, etc) populate this information consistently. I think we need a repository like [dcm_qa_enh](https://github.com/neurolabusc/dcm_qa_enh) that provides DICOM inputs and validated BIDS results for each manufacturer.

po09i:
> I really think there should be concrete examples that allow us to validate that different tools (dcm2niix, dicm2nii, etc) populate this information consistently. I think we need a repository like [dcm_qa_enh](https://github.com/neurolabusc/dcm_qa_enh) that provides DICOM inputs and validated BIDS results for each manufacturer.

@neurolabusc Yes I agree it is probably wise to do so. 

I think we can still move forward with the spec and explore the validation side in dcm2niix/dcm2nii.

po09i:
Tagging maintainers since I'm trying to merge this. @Remi-Gau @tsalo @effigies @ericearl

ericearl:
I was just trying to request Ross review, sorry others!

rwblair (approved):

effigies:
I'd like to get on the same page as consumers, producers, and standards maintainers to make sure that the order of operations is clear. IMO consumers and producers should agree, and then once they're happy the standard can put a stamp on it.

If people are happy with this definition (@jcohenadad, are your concerns addressed?) and feel it's unlikely to change substantively, then they can go ahead and start using it before a final merge.

I am inclined to defer to @neurolabusc on the need for QA data, as there's the potential for QA to reveal subtle changes in definition.

The thing I don't want to happen is for this to end up deadlocked, where QA efforts are on hold waiting for a final BIDS merge, while the merge is on hold pending QA efforts.

jcohenadad:
> @jcohenadad, are your concerns addressed?

Yes, 100%. I don't think we should wait for QA data for merging this PR. This could be added subsequently.

mr-jaemin:
> I really think there should be concrete examples that allow us to validate that different tools (dcm2niix, dicm2nii, etc) populate this information consistently. I think we need a repository like [dcm_qa_enh](https://github.com/neurolabusc/dcm_qa_enh) that provides DICOM inputs and validated BIDS results for each manufacturer.

As noted here https://github.com/rordenlab/dcm2niix/issues/726#issuecomment-2253236598, I will collect & provide validation GEHC samples for both Head-first and Feet-first scans.

po09i:
The table position was implemented in [dcm2niix](https://github.com/rordenlab/dcm2niix/issues/726) according to this specification. Datasets for GE, Siemens and Philips were provided. We did not run into issues with the current specification.

@effigies I believe this was the last piece of the puzzle before a final review.

mr-jaemin:
For future reference,

- TablePosition validataion datasets (GE, Siemens and Philips) here [dcm_qa_table](https://github.com/neurolabusc/dcm_qa_table)
   - GE validation dataset documentation [GEHC_BIDS_TablePosition_validation.pdf](https://github.com/mr-jaemin/ge-mri/blob/main/doc/GEHC_BIDS_TablePosition_validation.pdf)
- dcm2niix  v1.0.20240812+

src/schema/rules/sidecars/mri.yaml
```diff
@@ -37,6 +37,7 @@ MRIHardware:
     MatrixCoilMode: recommended
     CoilCombinationMethod: recommended
     NumberTransmitCoilActiveElements: optional
+    TablePosition: recommended
```
effigies, line 40: Please note that `recommended` means that the validator will warn if the metadata are not present. I think it should probably be `optional`.

```suggestion
    TablePosition: optional
```

If we want to make it `recommended` if the `chunk` entity is present:

```suggestion
    TablePosition:
      level: optional
      level_addendum: recommended if `chunk` entity is present

MRIChunkPosition:
  selectors:
    - modality == "mri"
    - entities.chunk
    - match(extension, '\.nii(\.gz)?$')
  fields:
    TablePosition:
      level: recommended
      issue:
        code: TABLE_POSITION_RECOMMENDED
        message: |
          TablePosition is RECOMMENDED if the chunk entity is present.
```

po09i: The use case I have for the TablePosition tag is to find the coordinates of the isocenter in RAS coordinates. The isocenter can then be used for shimming. The TablePosition is therefore not only useful when there are different chunks. If it is optional, then I worry that the information will not be available and we won't have access to the isocenter location. However, I use dcm2niix so since it is implemented there, I believe I won't have issues for the data that I acquire but could run into issues if converted from somewhere else. @effigies you are more aware of the impacts that it might have so let me know what you think.

mr-jaemin: Current definition, "The table position, relative to an implementation-specific reference point,
    often the isocenter." include some vagueness about the reference point, not specific to the isocenter. Is this intentional?

po09i: I could not 100% percent validate that a table position of 0 means that the patient position affine origin is at isocenter. From my experience, it has always been the case (on Siemens and Philips), but I do not have access to all manufacturers private header information to assume this hypothesis. "relative to an implementation-specific reference point" is the wording used for the [DICOM Table position CT tag](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.15.3.5.html) which is why I chose to include it in the definition, however I would love to be able to drop the vagueness if we can.

effigies: I am not trying to say it is only useful with multiple chunks, but that it goes from being nice-to-have to potentially-difficult-to-process-without in a multi-chunk context. It is certainly valid to advertise that you require the metadata in order to use your tool.

I'm coming at this from the perspective of validation. We have accumulated many RECOMMENDED fields over the years, and a validator that warns on all of them produces quite a lot of text. I believe we should be judicious about adding new recommendations, and target them to contexts where they are most likely to do good. I do think it's a good thing for tools like dcm2niix to aim to be as complete as possible, to make the dataset as broadly usable as possible, but a validator needs to raise *actionable* warnings and errors. An error should indicate that a dataset is wrong, and a warning should indicate that curators should make a considered decision about whether to address it.

(As an aside, I also think we should bump many existing recommendations down to optional, for the reasons above.)

po09i: That sounds reasonable :)

effigies: I'm marking this conversation resolved. @mr-jaemin If you want to further discuss wording, please respond on the main thread.

effigies (approved):

mr-jaemin:
As @po09i noted [here](https://github.com/bids-standard/bids-specification/pull/1690#discussion_r1715907618) I personally would love to be able to drop the vagueness in definition if we can, especially considering the motivation; to know the location of the isocenter and an application; to performa offline gradwarp correction' From @po09i's experience, it has always been the isocenter on Philips and Siemens although he could not 100% validate this. 

I confirm this is the case for GE and validation datasets, implementation (dcm2niix), documentation [here](https://github.com/bids-standard/bids-specification/pull/1690#issuecomment-2286931009).

I am okay with current definition but wanted to hear any feedback on removing the vagueness in definition (@neurolabusc @mharms)?

> The table position, relative to ~~an implementation-specific reference point, often~~ the isocenter. Values must be an array (1x3) of three distances in millimeters in absolute coordinates (world coordinates). If an observer stands in front of the scanner looking at it, a table moving to the left, up or into the scanner (from the observer's point of view) will increase the 1st, 2nd and 3rd value in the array respectively. The origin is defined by the image affine.

mr-jaemin:
> @mr-jaemin I believe the origin coordinate is isocenter for all MRI manufacturers, but for CT the origin is the table cetner. Therefore, the `specific reference point` seems worth keeping as it makes this definition flexible for both modalities (I am not sure about PET, perhaps @CPernet knows).

@neurolabusc That makes sense! I was thinking of `TablePosition` as a MRI specific field. Thanks!

effigies:
Thanks for your time and effort!

PR #1693, opened 2024-02-01
SCHEMA: Add schema.rules.directories to specify directory layouts

effigies:
WIP. Let's talk on the files tab.

### DIFF
--- a/src/schema/rules/directories.yaml
+++ b/src/schema/rules/directories.yaml
@@ -1,53 +1,22 @@
 ---
 # This file defines layouts of directories.
-# A directory may have a name, entity or value.
+#
+# A layout defines a collection of directory specifiers.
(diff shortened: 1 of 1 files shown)


src/schema/rules/directories.yaml
```diff
@@ -0,0 +1,96 @@
...
+          datatype:
+            value: datatype
+            level: required
+            opaque: false
+      datatype:
+        value: datatype
+        level: optional
+        opaque: false
+
+alt-raw:
```
tsalo, line 50: `alt-raw` makes more sense to me, though maybe `validate` instead of `opaque`?

effigies: I'm not committed to opaque, but validate seems to be a bit more of a directive than a description of the directory.

PR #1698, opened 2024-02-07
Fix up examples in src/schema/README.md to not use outdated schema paths

yarikoptic:
I also developed a script to catch those no longer valid paths used. I think it is worth adding it on CI

- [x] find where in CI we produce that json dump of the schema, and add there running this script to ensure that all paths in README.md are valid
- (maybe elsewhere) ~~decide if we should not choose simpler example or may be make it fuller (give example of secondary dereference)~~

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -2,12 +2,13 @@ version: 2
 
 build:
   os: ubuntu-22.04
+  apt_packages:
+    - jq
   tools:
     python: "3.11"
   jobs:
     pre_build:
       - bst -v export --output src/schema.json
-      - sudo apt install jq
       - tools/no-bad-schema-paths.sh src/schema.json  # README.md might need fixing
 
 mkdocs:


yarikoptic:
I found where schema.org is produced -- in RTD, but there in env there is no `jq` so RTD succeeds but log says
```
tools/no-bad-schema-paths.sh: line 19: jq: command not found
meta.context.context: not reachable
tools/no-bad-schema-paths.sh: line 19: jq: command not found
objects.columns: not reachable
tools/no-bad-schema-paths.sh: line 19: jq: command not found
objects.common_principles: not reachable
tools/no-bad-schema-paths.sh: line 19: jq: command not found
```

moreover -- RTD succeeds! so that would not be appropriate place.  we need to do it in another CI

yarikoptic:
yeay -- it worked, removing TEMP commit:

image

Issue #197, opened 2019-04-05
[ENH] Proposal for multidimensional array file format


tyarkoni:
At the BIDS-ComputationalModels meeting, it became pretty clear that a wide range of applications require (or would at benefit considerably from) the ability to read in generic n-dimensional arrays from a binary file. There are at least two major questions that should be discussed here, and then we should move to draft a PR modifying the BIDS-Raw specification:

1. What file format should we use? This should be something generic enough that it can be easily read on all common platforms and languages. The main proposals that came up at the meeting were for numpy (`.npy`) or HDF5 containers (.h5). While numpy is technically a Python format, it's sufficiently simple and well-supported that there appear to be available libraries for the major languages. Please suggest other potential solutions.

2. How and where should we represent associated metadata? The generic file format (and naming conventions, etc.) will eventually described in the BIDS-Raw spec, alongside all of the other valid formats (.tsv, nifti, etc.). But some applications are likely to require fairly specific interpretations of the data contained in the file. There appears to be some convergence on the notion of representing the relevant metadata in relevant sections of the BIDS-Derivatives spec (or current BEPs)—i.e., every major use case would describe how data in the binary array format should be interpreted when loaded. We could also associate suffixes with use cases, so that a tool like PyBIDS can automatically detect which rules/interpretations to apply at load time. But if there are other proposals (e.g., a single document describing all uses cases), we can discuss that here.

I'm probably forgetting/overlooking other relevant aspects of the discussion; feel free to add to this. Tagging everyone who expressed interest, or who I think might be interested: @johngriffiths @maedoc @effigies @yarikoptic @satra.

effigies:
Note that there are [two versions](https://docs.scipy.org/doc/numpy/reference/generated/numpy.lib.format.html) of npy, so compatibility levels of 1 and 2 should be assessed.

My primary concern with `npy` is that `npy` is not compressed, `npz` is just a `zip` of a directory of `npy` files, which almost certainly won't handle random read access as well as HDF5.

My primary concern with HDF5 is that it's just a container, and we will find ourselves defining formats. Perhaps just saying it contains only a dataset with name `/data` or similar will resolve that.

satra:
in a different world, but probably related to computational models BRAIN has funded development of the [NWB standard](https://www.nwb.org/how-to-use/). to the extent that needs may become similar, it may be worthwhile thinking about supporting NWB in BIDS.

[...]

PR #1699, opened 2024-02-08
ENH: Clarify the relation of motion.tsv columns to channels.tsv rows

effigies:
The motion spec does not say whether there is a header column or not, only that the columns must match the rows of channels.tsv. From unofficial communications, I believe the intent is that motion.tsv files do not have column headers, and channels.tsv defines the column names.

This changes it to state explicitly that there are no column headers, and that the ordering is defined by the rows in channels.tsv.

cc @sjeung @JuliusWelzel

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -430,7 +430,8 @@ Tabular data MUST be saved as tab delimited values (`.tsv`) files, that is, CSV
 files where commas are replaced by tabs. Tabs MUST be true tab characters and
 MUST NOT be a series of space characters. Each TSV file MUST start with a header
 line listing the names of all columns (with the exception of
-[physiological and other continuous recordings](modality-specific-files/physiological-and-other-continuous-recordings.md)).
+[physiological and other continuous recordings](modality-specific-files/physiological-and-other-continuous-recordings.md)
+as well as [motion recording data](modality-specific-files/motion.md)).
 It is RECOMMENDED that the column names in the header of the TSV file are
 written in [`snake_case`](https://en.wikipedia.org/wiki/Snake_case) with the
 first letter in lower case (for example, `variable_name`, not `Variable_name`).
--- a/src/modality-specific-files/motion.md
+++ b/src/modality-specific-files/motion.md
@@ -42,7 +42,10 @@ each of which is accompanied by `*_tracksys-<label>_motion.json` and `*_tracksys
 Between `tracksys-<label>` entity and `*_motion.tsv`, `*_motion.json`, or `*_channels.tsv` suffixes, optional [`acq-<label>`](../appendices/entities.md#acq) or [`run-<index>`](../appendices/entities.md#run) entity MAY be inserted.
 
 One column in the `*_tracksys-<label>_motion.tsv` file represents one data channel.
-The ordering of columns MUST match the order of rows in the `*_channels.tsv` file for unambiguous assignment.
+Motion files MUST NOT have a header row;
+the ordering of columns is given by the order of rows in the associated `*_channels.tsv` file.
+The number of columns in `_motion.tsv` files MUST equal the number of rows
+in the associated `_channels.tsv` file.
 All relevant metadata about a tracking systems is stored in accompanying sidecar `*_tracksys-<label>_motion.json` file.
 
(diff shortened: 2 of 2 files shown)


effigies:
This relates to https://github.com/bids-standard/bids-validator/issues/1889.

Remi-Gau:
Hum... I am sorry I missed that but I had assumed that those TSV had column headers given our common principles about tabular data:

https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files

> Each TSV file MUST start with a header line listing the names of all columns (with the exception of [physiological and other continuous recordings](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/physiological-and-other-continuous-recordings.html)).

sappelhoff (approved):

effigies:
> Hum... I am sorry I missed that but I had assumed that those TSV had column headers given our common principles about tabular data:
> 
> https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files
> 
> > Each TSV file MUST start with a header line listing the names of all columns (with the exception of [physiological and other continuous recordings](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/physiological-and-other-continuous-recordings.html)).

Yeah, in retrospect I wish we'd pushed a bit to make them either standard TSV files or stim/physio-like `.tsv.gz` files. As it is, they're not either, which means we now have a third way of specifying column headers, but so be it. I do not want to relitigate a merged BEP.

Remi-Gau:
Fair. Should we update common principles though?

effigies:
Possibly. I don't know what I'd want them to say at this point.

sappelhoff:
> in retrospect I wish we'd pushed a bit to make them either standard TSV files or stim/physio-like .tsv.gz files. As it is, they're not either, which means we now have a third way of specifying column headers, but so be it. I do not want to relitigate a merged BEP.

I seem to recall that this was being discussed at some point. I just fail to find the thread (e.g., in https://github.com/bids-standard/bids-examples/pull/425). @sjeung @JuliusWelzel do you remember discussions about this and could potentially link to them?

> I don't know what I'd want them to say at this point.

Not pretty, but could do this:

https://github.com/bids-standard/bids-specification/blob/365f3219aec261654220c4bb88e02498b5b39ad8/src/common-principles.md?plain=1#L431-L433

```diff
Each TSV file MUST start with a header line listing the names of all columns
(with the exception of physiological and other continuous recordings
+, and motion.tsv files). 
```

JuliusWelzel (approved):

JuliusWelzel:
Hi, thanks for picking this up @effigies. I found a [protocol](https://github.com/JuliusWelzel/BIDS-BEP029/blob/master/protocols_tc/tc_2021_03_30.md#_channelstsv-files) from March 21, which said we would use headers in the motion.tsv file. For me, having headers in the `motion.tsv` file makes sense for a future release.

effigies:
IMO we should either migrate to headers right now or not at all in BIDS 1.x. There hasn't been much data uploaded at this point and the text of the spec is ambiguous, so we can clarify in the opposite direction as well.

sjeung:
I think the protocol from 2021 was before we agreed on separating channel files for multiple tracking systems. 
We dropped headers after it became obsolete with the introduction of tracksys entity. 
Having no headers is consistent with BIDS validator and example data set as well.

Unfortunately I could not find any log of .gz discussion which I recall was around community review phase, and the main reason against .gz was time pressure before release of the upcoming BIDS version back then. So I would be happy to make motion spec coherent with other continuous time series data .tsv files in BIDS like stim or physio in the future.  (no headers, and tsv.gz)

effigies:
Ah, yes. I read that as "March 21, 2023", not "March 2021". If that is that out-of-date, then I do not think that qualifies as a decision that was missed during merge, but an outdated decision.

I am +1 for allowing `motion` to be `.tsv` or `.tsv.gz`, but there was just a suggestion in #197 to permit Parquet files, which I think would be a better fit than either. If that goes places, I would say we don't add `.tsv.gz`, but simply add `.parquet`.

sappelhoff (approved):

PR #1700, opened 2024-02-09
ENH: Add CLI tool for filename validation for use in pre-receive hooks

effigies:
This is a basic validator that accepts a `.bidsignore` list and a list of files. The use case is to quickly determine if a git push should be accepted by a server.

Todo:

* [x] Accept `dataset_description.json` from stdin, and enable derivatives
* [ ] More efficient `fnmatch` filtering
* [x] Consider streaming stdin instead of slurping the entire input


effigies:
@nellh What would be a good format for dataset_description? Right now it's:

```
ignore-entry1
ignore-entry2
...
ignore-entryN
0001
file1
file2
...
fileN
```

We could do:

```
{
  ...
}
0000
ignore-entry1
ignore-entry2
...
ignore-entryN
0001
file1
file2
...
fileN
```

This would be kinda backwards compatible, as the `0001` will continue to separate ignore and file listing, and then we could look for `0000\n` in the ignore list. But I'm not sure if we care about backwards compatibility, and if there might be something better.

tools/schemacode/bidsschematools/__main__.py
```diff
@@ -32,5 +40,128 @@ def export(ctx, schema, output):
...
+        stream = chain([first_line], stream)
+        description = {}
+
+    dataset_type = description.get("DatasetType", "raw")
+    logger.info("Dataset type: %s", dataset_type)
+
+    ignore = [line.strip() for line in stream if line != "0001\n"]
+    logger.info("Ignore patterns found: %d", len(ignore))
+
+    schema = load_schema(schema)
```
nellh, line 123:
```suggestion
```
Looks like this isn't needed.

tools/schemacode/bidsschematools/__main__.py
```diff
@@ -32,5 +40,128 @@ def export(ctx, schema, output):
...
+            sys.exit(2)
+    else:
+        # Legacy: ignore patterns, followed by "0001", followed by filenames
+        stream = chain([first_line], stream)
+        description = {}
+
+    dataset_type = description.get("DatasetType", "raw")
+    logger.info("Dataset type: %s", dataset_type)
+
+    ignore = [line.strip() for line in stream if line != "0001\n"]
```
nellh, line 120:
```suggestion
    ignore = []
    for line in stream:
        if line == "0001\n":
            break
        else:
            ignore.append(line.strip())
```
This would consume the entire input as ignore patterns.

nellh (changes requested):
@effigies I merged this too soon on the OpenNeuro side but looks fine with a couple fixes (see suggested changes). Anything else we should address to merge this?

tools/schemacode/bidsschematools/__main__.py
effigies: Slightly adjusted your suggestion to skip the redundant else.

tools/schemacode/bidsschematools/__main__.py
```diff
@@ -1,6 +1,8 @@
+import json
 import logging
 import os
```
effigies, line 3:
```suggestion
import os
import re
```

nellh (approved):

PR #1703, opened 2024-02-13
[FIX] Remove task entity from DWI and PERF time series file templates

### LINKED ISSUE #1701: [BUG] timeseries files support task entity for ASL and DWI

physio / stim files support task entity for ASL and DWI when the data files for ASL and DWI do not.

mentioned in https://neurostars.org/t/bids-validator-asl-filename-error/28116

no file under ASL and DWI should support the task entity

original support for physio / stim was introduced in #894 

see here: https://github.com/bids-standard/bids-specification/pull/894/files#r1486489726

but the current schema for those files for MRI is based on a file pattern that includes task:

https://github.com/bids-standard/bids-specification/blob/365f3219aec261654220c4bb88e02498b5b39ad8/src/schema/rules/files/raw/task.yaml#L75

https://bids-specification.readthedocs.io/en/latest/modality-specific-files/magnetic-resonance-imaging-data.html#arterial-spin-labeling-perfusion-data

https://bids-specification.readthedocs.io/en/latest/modality-specific-files/magnetic-resonance-imaging-data.html#diffusion-imaging-data

Remi-Gau:
- fix #1701

### DIFF
--- a/src/schema/rules/files/raw/task.yaml
+++ b/src/schema/rules/files/raw/task.yaml
@@ -41,7 +41,7 @@ timeseries:
 timeseries__mri_no_task:
   suffixes:
     - physio
-    # - stim # no task --> no stimuli
+    - stim
   extensions:
     - .tsv.gz
     - .json
--- a/src/schema/rules/files/raw/task.yaml
+++ b/src/schema/rules/files/raw/task.yaml
@@ -38,6 +38,25 @@ timeseries:
     run: optional
     recording: optional
 
+timeseries__mri_no_task:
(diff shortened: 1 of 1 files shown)


tsalo (approved):
The schema changes make sense to me. There's probably a way to define a shared base physio/stim rule that the MRI and MEG/EEG rules inherit from, but I don't know if it's worth it. I also looked over the filename templates for fMRI, DWI, ASL, and MEG, and they looked right as well.

effigies (approved):
LGTM

PR #1613, opened 2023-09-13
[ENH] Add `descriptions.tsv` file relating to the `desc-<label>` entity

CPernet:
add descriptions.tsv file as optional based on the derivatives guidelines https://docs.google.com/document/d/1JtTu5u7XTkWxxnCIH6sxGajGn1qG_syJ-p14aejpk3E/edit?usp=sharing

Issue #1706, opened 2024-02-15
Support entity definition TSV files


tsalo:
### Your idea

In #1613, we added a `descriptions.tsv` file to the Derivatives spec. The file describes different `desc-<label>` definitions used in a dataset. I believe this general structure can be applied to other entities, especially `acq-<label>` (`acquisitions.tsv`), `rec-<label>` (`reconstructions.tsv`), `proc-<label>` (`processings.tsv`?), and `recording-<label>` (`recordings.tsv`).

sappelhoff:
> I believe this general structure can be applied to other entities

This is in fact, how the idea for `descriptions.tsv` originated when we discussed in in Copenhagen 2023. There are some minor inconsistencies, like `participants.tsv`, which should rather have been `subjects.tsv` (corresponding to the sub entity).

I am generally in favor of adding support like this wherever it is practically meaningful (that is, if we see a concrete use case).

ericearl:
Just an FYI since @tsalo linked me this to read: Our BEP036 is going to propose a change to allow sessions.tsv at the top-level so long as it has the participant_id, the session_id, and a sessions.json describing the "Levels" of session_id. This will allow imaging data sessions and phenotype-only sessions to live together peacefully/in-harmony.  :smile:

Pinging my fellow BEP-ers here too if they want to read it and/or mention anything else about sessions.tsv.

P.S. I love this Issue's idea.

yarikoptic:
Related!
- https://github.com/bids-standard/bids-specification/issues/2283

PR #1707, opened 2024-02-15
Add `descriptions.tsv` to the schema

tsalo:
Closes none, but continues from #1613.

I didn't figure out how to render the filenames in a macro. I think allowing `descriptions.tsv` files at the subject or session level introduced more complexity than it was worth, unfortunately.

I also haven't figured out how to add meaningful rules about the file to the schema. Specifically, we probably want to raise a warning if any desc values described in the file aren't in the dataset, or vice versa.

Changes proposed:

- Add the `descriptions.tsv` file to the schema.


src/schema/objects/columns.yaml
```diff
@@ -122,6 +129,12 @@ description__optode:
   description: |
     Free-form text description of the optode, or other information of interest.
   type: string
+description__descriptions:
```
tsalo, line 132: If we choose to implement other entity definition TSVs (see #1706), I wonder if we'd want to use the same column name ("description") for the value definitions or if we'd want to use the full entity names (e.g., `acq_id` and `acquisition`). I think `acq_id` and `description` makes more sense TBH, in which case maybe the identifier here should be `description__entity`.

tsalo:
Switching to a draft until I figure out how to render the filename template.

tsalo:
I've given up on rendering the template with a macro, but I'll bring up adding checks for the file in the schema meeting today.

Presumably we would want rules to check if all `desc-<label>` entities are reflected in the file and that no definitions are included that _aren't_ present in the dataset. Both should probably be warnings, rather than errors.

src/derivatives/common-data-types.md
```diff
@@ -256,24 +256,33 @@ static volume, a `RepetitionTime` property would no longer be relevant).
 
 ## descriptions.tsv
 
+Template:
+
+```Text
+sub-<label>/
+    [ses-<label>/]
+        sub-<label>[_ses-<label>]_descriptions.tsv
+        sub-<label>[_ses-<label>]_descriptions.json
```
effigies, line 265:
```suggestion
[sub-<label>/]
    [ses-<label>/]
        [sub-<label>][_ses-<label>]_descriptions.tsv
        [sub-<label>][_ses-<label>]_descriptions.json
```

src/schema/objects/columns.yaml
tsalo: Per chat with @effigies, `description__entity` makes more sense. We can customize the definitions as they appear in the specification text with the macros.

src/derivatives/common-data-types.md
tsalo: Oh yeah good point!

effigies (approved):
LGTM.

src/schema/objects/files.yaml
```diff
@@ -126,3 +126,8 @@ stimuli:
     A directory to store any stimulus files used during an experiment.
     See the [relevant section](SPEC_ROOT/modality-specific-files/task-events.md#stimuli-directory)
     for more information.
+descriptions:
+  display_name: Descriptions Definitions
+  file_type: regular
+  description: |
+    A TSV file describing labels found for the `desc` entity in a Derivatives dataset.
```
tsalo, line 133: This actually might conflict with the suffix definition later. Should I remove it?

tsalo: Removing it.

tsalo:
I will open an issue about my two pain points:

- Filename template building isn't working.
- Rules/checks not implemented in schema.

sappelhoff (approved):
approving for merge as soon as "good enough" and to then cover the pain points in a follow-up (or several).

Remi-Gau:
@tsalo 
Can't remember if there is a reason this was not merged?

tsalo:
No reason- I didn't realize I had two approvals. I'll update from main and then merge once CI passes.

PR #1708, opened 2024-02-15
Fix links

tsalo:
Closes none. Fixes the links that are causing the link-checker CI step to fail.


tsalo:
The INCF website error seems to stem from calling the URL too many times, rather than a broken URL.

sappelhoff (approved):

effigies:
The INCF issue is that they seem to be signed with a certificate that isn't in OS certificate stores yet. Try running `curl https://www.incf.org`.

effigies (approved):

effigies:
Opened a thread at https://neurostars.org/t/incf-website-not-sending-a-complete-certificate-chain/28163. I don't know if either of you have a better notion where to ping INCF folks.

Issue #1481, opened 2023-05-01
rules behind the limited number of electrodes.tsv per session should be made more explicit


Remi-Gau:
Unless I missed it, it seems that the rules limiting the number of eletrodes.tsv per session are not explicitly stated.

For example

> electrodes (positions) and coordinates (in which they are expressed) are required to remain constant in a session. If not, then the data must be represented over multiple sessions (which can be on the same day). Also, if multiple EEG systems are being used in succession on the same subject (e.g. with dry electrodes), it should be represented in multiple sessions. 

Linking to the original thread at the end of this post.

I may be wrong but this feels more then a simple "how to use bids" issue as it adds an extra constraints on the definition of session for EEG and iEEG (as mentioned in the original thread).

Another reason why this may needed is that currently the validator does not enforce this rule and allows many more entities than the spec permits (https://github.com/bids-standard/bids-validator/issues/1198).

The lack of clarity on this point can otherwise lead to user confusion (see https://github.com/bids-standard/bids-matlab/issues/542#issuecomment-1529921282)

--- 

electrodes (positions) and coordinates (in which they are expressed) are required to remain constant in a session. If not, then the data must be represented over multiple sessions (which can be on the same day). Also, if multiple EEG systems are being used in succession on the same subject (e.g. with dry electrodes), it should be represented in multiple sessions. 

Different sampling rates and/or filter settings of the amplifier can happen within a session, as (of course) multiple runs of the same task, or different tasks in different (EEG) files.

Electrode positions are (potentially noisy) measurements that can be done multiple times, hence the `acq` label. The coordinate system is a definition,  which only needs to be specified once. 

I also do not recall why there is an optional `run` label for electrodes. I do not see the use case for it. Looking in the draft [google doc](https://docs.google.com/document/d/1ArMZ9Y_quTKXC-jNXZksnedK2VHHoKP3HCeO5HPcgLE/edit#), I also don't see it there.

_Originally posted by @robertoostenveld in https://github.com/bids-standard/bids-specification/issues/416#issuecomment-587459530_

Remi-Gau:
Just to be clear, I am not trying to argue against the rationale, but more arguing for the rationale to be made more explicit, a bit like in the spirit of some other recent PRs that either clarify and make things more explicit:
- https://github.com/bids-standard/bids-specification/pull/1392
- https://github.com/bids-standard/bids-specification/pull/1444/files

robertoostenveld:
I value clarifications!

[...]

PR #1722, opened 2024-03-07
FIX: Re-add run entity to electrodes.tsv

effigies:
In BIDS 1.3, EEG had

```
sub-<label>[_ses-<label>][_acq-<label>][_run-<index>]_electrodes.tsv
```

and iEEG had 

```
sub-<label>[_ses-<label>][_space-<label>]_electrodes.tsv
```

As of 1.5, presumably due to schema stuff, they both became

```
sub-<label>[_ses-<label>][_acq-<label>][_space-<label>]_electrodes.tsv
```

So EEG lost `run` and gained `space`, and iEEG gained `acq`. Looking in OpenNeuro, there are iEEG electrodes files with `acq` and `run` and EEG with `space`. So we should just permit all three for both, as trying to bottle this genie isn't worth the pain.

### DIFF
--- a/src/schema/rules/checks/electrodes.yaml
+++ b/src/schema/rules/checks/electrodes.yaml
@@ -1,0 +1,15 @@
+---
+ElectrodeSpecificity:
+  issue:
+    code: EXCESSIVE_ELECTRODE_SPECIFICITY
+    message: |
+      Run or acquisition entities detected in electrodes.tsv.
+      Electrode definitions should generally not vary within a session.
+      Consider creating a new session each time electrodes are reconfigured.
+    level: warning
+  selectors:
+    - suffix == 'electrodes'
+    - extension == '.tsv'
+  checks:
+    - '!("run" in entities)'
+    - '!("acquisition" in entities)'
--- a/src/modality-specific-files/electroencephalography.md
+++ b/src/modality-specific-files/electroencephalography.md
@@ -310,6 +310,12 @@ and a guide for using macros can be found at
 -->
 {{ MACROS___make_columns_table("eeg.EEGElectrodes") }}
 
+`*_electrodes.tsv` files SHOULD NOT be duplicated for each data file,
+for example, during multiple runs of a task.
+The [inheritance principle](../common-principles.md#the-inheritance-principle) MUST
+be used to find the appropriate electrode positions for a given data file.
+If electrodes are repositioned, it is RECOMMENDED to use multiple sessions to indicate this.
+
 ### Example `*_electrodes.tsv`
 
 See also the corresponding [`electrodes.tsv` example](#example-channelstsv).
--- a/src/modality-specific-files/intracranial-electroencephalography.md
+++ b/src/modality-specific-files/intracranial-electroencephalography.md
@@ -351,6 +351,12 @@ and a guide for using macros can be found at
 -->
 {{ MACROS___make_columns_table("ieeg.iEEGElectrodes") }}
 
+`*_electrodes.tsv` files SHOULD NOT be duplicated for each data file,
+for example, during multiple runs of a task.
+The [inheritance principle](../common-principles.md#the-inheritance-principle) MUST
+be used to find the appropriate electrode positions for a given data file.
+If electrodes are repositioned, it is RECOMMENDED to use multiple sessions to indicate this.
+
 ### Example `*_electrodes.tsv`
 
 ```Text
--- a/src/schema/rules/files/raw/channels.yaml
+++ b/src/schema/rules/files/raw/channels.yaml
@@ -70,6 +70,7 @@ electrodes:
     subject: required
     session: optional
     acquisition: optional
+    run: optional
     space: optional
 
 # MEG has an additional entity available


sappelhoff (commented):
for EEG at least, changing "electrodes" within a run would typically mean a new session -- so, a "run" entity for electrodes in EEG doesn't make sense. I don't remember what happened between the versions unfortunately, and it's unfortunate that there are datasets that use the run entity on EEG electrodes 🤔 

cc @robertoostenveld @dorahermes

dorahermes:
We typically assumed when we wrote the spec that changing electrode positions would result in a different session. 

Is there perhaps an example for acquisition in iEEG electrodes? I can imagine adding acquisition on the channels as that can pertain to the sampling frequency etc, but not sure what the use case would be for electrodes.

effigies:
I can dig up the data next time I'm at my machine, but the point of this is not that there's a good use case but that the validator has permitted it since the BEPs were merged and this flexibility has led to people using it.

I suspect in many cases they are just copied to match the data files, like sidecars with no inheritance, not actually varying.

dorahermes:
Thinking about it a bit more acquisition for iEEG could potentially indicate CT/MRI/Photo, moving forward I think this is ok. 

I am not sure if run makes sense without an optional task label etc. Not sure how packages deal with this @robertoostenveld

Remi-Gau:
> for EEG at least, changing "electrodes" within a run would typically mean a new session -- so, a "run" entity for electrodes in EEG doesn't make sense. I don't remember what happened between the versions unfortunately, and it's unfortunate that there are datasets that use the run entity on EEG electrodes 🤔 
> 
> cc @robertoostenveld @dorahermes 

Is this related to this?

https://github.com/bids-standard/bids-specification/issues/1481

robertoostenveld:
I can imagine that people copy the full filename of the EEG file, and only replace the `_eeg` suffix with `_electrode`, consequently the electrode filename would also contain the task entity. I could even imagine that happening when people use automatic converters, like EEGLAB or a FieldTrip script.

There is no real point in the task entity (nor in the run entity) for the electrodes, but I also don't see much harm as long as there is an unambiguous mapping between the EEG file and the electrode positions. Here I am thinking of the perspective of a dataset downloader/user: having superfluous entities in the electrode file name does not interfere with that mapping.

I would in general recommend not to use entities that are not needed. For example, with a single EEG recording, I would not use `run-1`. And with digitized EEG electrode positions, the run and task entity are not needed and hence I would not use them either. My rule would be: keep filenames as short as possible, and as long as needed. 

Note that for iEEG I believe that we did agree on the acq entity to be used, and also for EEG I believe that makes sense in some cases (like with a pre and post digitization of EEG electrode positions). We are now working on [OPM MEG coregistration](https://www.fieldtriptoolbox.org/tutorial/coregistration_opm/), and also there I can imagine that versions of the OPM sensor locations can be specified. Note however that OPM locations stored separately from the data file at the moment falls outside of the spec, but there is some work to be done on BIDS to accomodate OPMs (as indicated by the suggested OPM BEP).

effigies:
I suggest we add the following text to each `electrodes.tsv` section:

> `*_electrodes.tsv` files SHOULD NOT be duplicated for each data file,
> for example, during multiple runs of a task.
> The [inheritance principle](../common-principles.md#the-inheritance-principle) MUST
> be used to find the appropriate electrode positions for a given data file.
> If electrodes are repositioned, it is RECOMMENDED to use multiple sessions to indicate this.

This would allow the existing duplication without invalidating datasets, direct the validator to raise a warning when detecting this case (I'll write a schema rule), and indicate clearly to tooling to search for the relevant file rather than hard-code assumptions about what entities should be there.

Would this PR be acceptable if I do the above?

src/schema/rules/checks/electrodes.yaml
```diff
@@ -0,0 +1,15 @@
...
+      Run or acquisition entities detected in electrodes.tsv.
+      Electrode definitions should generally not vary within a session.
+      Consider creating a new session each time electrodes are reconfigured.
+    level: warning
+  selectors:
+    - suffix == 'electrodes'
+    - extension == '.tsv'
+  checks:
+    - '!("run" in entities)'
+    - '!("acquisition" in entities)'
```
sappelhoff, line 15: I would take the "acq" entity out of this check, as "pre" and "post" measurements (using `acq-pre` and `acq-post`) for example would be an understandable use-case.

sappelhoff (approved):

src/schema/rules/checks/electrodes.yaml
effigies: This is a warning, not an error, so people are free to decide that their reasons for violating the recommendation are good.

I'm okay removing `acq` if you think it will generally be a legitimate use.

FWIW, here are the uses in OpenNeuro (removing redundant examples):

```
./ds003688/sub-30/ses-iemu/ieeg/sub-30_ses-iemu_acq-HDgrid_electrodes.tsv
./ds003688/sub-30/ses-iemu/ieeg/sub-30_ses-iemu_acq-clinical_electrodes.tsv
./ds004100/sub-HUP060/ses-presurgery/ieeg/sub-HUP060_ses-presurgery_acq-seeg_space-fsaverage_electrodes.tsv
./ds004100/sub-HUP064/ses-presurgery/ieeg/sub-HUP064_ses-presurgery_acq-ecog_space-fsaverage_electrodes.tsv
./ds004194/sub-p02/ses-umcuiemu01/ieeg/sub-p02_ses-umcuiemu01_acq-clinical_electrodes.tsv
./ds004194/sub-p03/ses-nyuecog01/ieeg/sub-p03_ses-nyuecog01_acq-clinical_electrodes.tsv
```

robertoostenveld: These two are electrodes of ECoG and SEEG, apparently recorded together, but stored in two separate files. That is not what I would have expected, they should be in the same tsv file and be documented with a single coordsys.json.  

```
./ds004100/sub-HUP060/ses-presurgery/ieeg/sub-HUP060_ses-presurgery_acq-seeg_space-fsaverage_electrodes.tsv
./ds004100/sub-HUP064/ses-presurgery/ieeg/sub-HUP064_ses-presurgery_acq-ecog_space-fsaverage_electrodes.tsv
```

effigies: Those are two different subjects. There aren't different acquisitions within a subject in that dataset.

robertoostenveld: oh sorry, my bad!

effigies: @sappelhoff Just checking in about whether you still think the `acq` check should be removed.

sappelhoff: I agree with your reasoning in https://github.com/bids-standard/bids-specification/pull/1722#discussion_r1520070303 and I am fine to keep the acq check. Thanks

effigies:
@dorahermes @robertoostenveld Could I bother either of you for a review?

effigies:
Thanks!

PR #1725, opened 2024-03-11
[ENH] Specify the naming of scanner-generated TRACE and ADC volumes

### LINKED ISSUE #1723: [DWI] How to encode ADC/TRACE volumes?
### Your idea

Scanners are increasingly generating ADC/TRACE volumes, and BIDS now considers scanner-generated derivatives to be raw enough to not contort ourselves into moving portions of the output into a `derivatives/` subdirectory.

These volumes right now are best considered DWI volumes, with no applicable bval/bvecs, but at least the validator expects DWI to be 4D and have bval/bvec sidecars.

I would like to push the issue and just have a vote. Here are our options (use the emoji responses to indicate your vote):

* :tada: Carve out `acq-ADC_dwi.nii[.gz]` and `acq-TRACE_dwi.nii[.gz]` as special DWI images that can be 3D and not require bval/bvec. (Vote for this if you would prefer an entity, even if not `acq`.)
* :rocket: Create TRACE and ADC suffixes that go in the `dwi/` datatype directory to refer to single-volume images with no bval/bvec requirements.
* :+1: Either is fine, but please just fix this.
* :-1: Leave things as is.

#### Rules of engagement

In the interest of keeping this discussion short, I would ask commenters to not address one another, but to limit themselves to a single post that argues for or against one or more of these options. If all you want to do is show support for one option, do it by voting.

#### References

* https://github.com/bids-standard/bids-validator/issues/1858
* https://neurostars.org/t/multiple-lesion-masks-for-bids/27184

effigies:
Closes #1723.

### DIFF
--- a/src/schema/rules/files/raw/dwi.yaml
+++ b/src/schema/rules/files/raw/dwi.yaml
@@ -38,3 +38,22 @@ sbref:
     run: optional
     part: optional
     chunk: optional
+
+isotropic:
+  suffixes:
+    - ADC
+    - TRACE
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  datatypes:
+    - dwi
+  entities:
+    subject: required
+    session: optional
+    acquisition: optional
+    reconstruction: optional
+    direction: optional
+    run: optional
+    chunk: optional
--- a/src/schema/rules/files/raw/dwi.yaml
+++ b/src/schema/rules/files/raw/dwi.yaml
@@ -39,6 +39,7 @@ sbref:
     part: optional
     chunk: optional
 
+# Common scanner-generated derivatives need raw names
 isotropic:
   suffixes:
     - ADC
--- a/src/modality-specific-files/magnetic-resonance-imaging-data.md
+++ b/src/modality-specific-files/magnetic-resonance-imaging-data.md
@@ -656,13 +656,7 @@ The definitions of these fields can be found in
 and a guide for using macros can be found at
  https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
 -->
-{{ MACROS___make_suffix_table(
-      [
-         "dwi",
-         "sbref",
-      ]
-   )
-}}
+{{ MACROS___make_suffix_table(["dwi", "sbref", "ADC", "TRACE"]) }}
 
 <!--
 This block generates a filename templates.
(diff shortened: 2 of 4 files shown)


Remi-Gau (approved):

Remi-Gau:
See HTML template in this section

https://bids-specification--1725.org.readthedocs.build/en/1725/modality-specific-files/magnetic-resonance-imaging-data.html#diffusion-imaging-data

src/schema/rules/files/raw/dwi.yaml
```diff
@@ -38,3 +38,22 @@ sbref:
     run: optional
     part: optional
     chunk: optional
+
+isotropic:
```
tsalo, line 42: Can you add a comment here summarizing what this group is?

CPernet: yeah looks good ! thx @Remi pointing to the rendered version :-)

arokem (approved):

effigies:
Not really a fan of how this looks, prioritizing the derivatives over the raw, due to sorting:

image

Split into a separate table. Rerequested reviews, since this is a significant change from before.

Remi-Gau (approved):
Still good with me.

CPernet:
I was updating my local code and realized that `_FA.nii[.gz]` is not in the list. Are there any reasons for that? (also what the scanner generates)

CPernet:
One small suggestion in the phrasing -- maybe it is worth avoiding using 'derivatives' to avoid confusion? 
  
`scanner-generated derivatives` -->` scanner-generated images`
`such volumes SHOULD be derived from that series `--> `such volumes SHOULD be computed from that series`

effigies:
> I was updating my local code and realized that `_FA.nii[.gz]` is not in the list. Are there any reasons for that? (also what the scanner generates)

It wasn't part of the prior discussion. If you don't mind, I'd prefer to reserve that for a new PR than change this one last-minute.

src/schema/rules/files/raw/dwi.yaml
Lestropie: Scanner-derived images won't be isotropic if the DWI acquisition wasn't isotropic. Elsewhere in this file I see "parametric" vs. "nonparametric"; I wonder if that's the more suitable disambiguation here? Otherwise may need to adopt "scanner-derived" ([comment](https://github.com/bids-standard/bids-specification/pull/1831#issuecomment-2132483613) in #1831).

PR #1730, opened 2024-03-19
[FIX] Add `rawdata` to glossary

kabilar:
## Changes

- [x] Add the `rawdata` description to the glossary by combining text from other parts of the BIDS specification.


effigies:
Rawdata is not a BIDS term. It's a name used in an example, but BIDS does not specify the name of the directory containing a BIDS dataset.

PR #1732, opened 2024-03-20
[ENH] events.tsv should be sorted by onset values

### LINKED ISSUE #1691: Event onset times can be out of order
Some datasets on OpenNeuro do not have their events ordered in ascending order by the `onset` time. I couldn't find anything in the spec about this.... but I think downstream analysts have a reasonable expectation that event times should be non-decreasing and will be in for a nasty surprise (if they notice it) when using these datasets.

Should there be a requirement that time series times be non-decreasing?

Remi-Gau:
- closes #1691

### DIFF
--- a/src/schema/rules/checks/events.yaml
+++ b/src/schema/rules/checks/events.yaml
@@ -38,4 +38,5 @@ SortedOnsets:
     - suffix == "events"
     - extension == ".tsv"
   checks:
+    # n/a values will likely cause false alarms if encountered. Consider alternatives.
     - sorted(columns.onset) == columns.onset
--- a/src/schema/rules/checks/events.yaml
+++ b/src/schema/rules/checks/events.yaml
@@ -27,3 +27,15 @@ StimulusFileMissing:
     - columns.stim_file != null
   checks:
     - exists(columns.stim_file, "stimuli") == length(columns.stim_file) - count(columns.stim_file, "n/a")
+
+SortedOnsets:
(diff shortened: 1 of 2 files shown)


effigies:
Will propose schema rule.

sappelhoff (approved):

src/schema/rules/checks/events.yaml
```diff
@@ -27,3 +27,15 @@ StimulusFileMissing:
...
+  issue:
+    code: EVENT_ONSET_ORDER
+    message: |
+      The onset column in events.tsv files should be sorted.
+    level: warning
+  selectors:
+    - suffix == "events"
+    - extension == ".tsv"
+  checks:
+    - sorted(columns.onset) == columns.onset
```
effigies, line 41: @rwblair Do you agree?

rwblair: n/a would be the only thing to mess it up if allowed. Otherwise looks good.

effigies: Good point. I think it being a warning should make that an annoyance, not a show-stopper. Will add a comment, though.

effigies, line 41:
```suggestion
    # n/a values will likely cause false alarms if encountered. Consider alternatives.
    - sorted(columns.onset) == columns.onset
```

effigies (approved):

PR #1741, opened 2024-03-21
[FIX] Move `rawdata/` into `sourcedata/raw` in alternative structure example, clarify on naming of datasets themselves

### LINKED ISSUE #1519: Clarify that a `rawdata` directory is not a prescription

### LINKED ISSUE #1736: `rawdata/` is described in text but not part of the schema
Inspired by 
- https://github.com/bids-standard/bids-specification/pull/1734

with some relevant prior discussion found in

- https://github.com/bids-standard/bids-specification/issues/1009

apparently only the text talks about `rawdata/` while `src/schema` does not have references beyond mentioning in the tests (@TheChymera could you clarify on either it is specific to `rawdata/` there or could be another path):

```shell
❯ git grep rawdata
src/CHANGES.md:-   \[FIX] typo in "rawdata" example [#1045](https://github.com/bids-standard/bids-specification/pull/1045) ([sappelhoff](https://github.com/sappelhoff))
src/common-principles.md:            "rawdata": {
src/common-principles.md:`rawdata`, **only the `rawdata` subdirectory** needs to be a BIDS-compliant
src/common-principles.md:`derivatives`, or `rawdata` directory names.
src/common-principles.md:  "SourceDatasets": [ {"URL": "../../rawdata/"} ]
tools/schemacode/bidsschematools/tests/data/expected_bids_validator_xs_write.log:   * `/home/chymera/.data2/datalad/000026/rawdata/sub-EXC022/anat/sub-EXC022_ses-MRI_flip-1_VFA.nii.gz
tools/schemacode/bidsschematools/tests/test_validator.py:        "rawdata/sub-EXC022/anat/sub-EXC022_ses-MRI_flip-1_VFA.nii.gz"
tools/schemacode/bidsschematools/tests/test_validator.py:        "rawdata/sub-EXC022/anat/sub-EXC022_ses-MRI_flip-1_VFA.nii.gz"
tools/schemacode/bidsschematools/validator.py:        bids_paths = '~/.data2/datalad/000026/rawdata'
```

IMHO we should fix `src/schema` to "specify/allow" for `rawdata/` in BIDS derivative datasets (not in raw)

yarikoptic:
Individual commits have more rationale.
I can split into multiple PRs but let's see - may be this could provide us a closure.
Reflecting my thoughts on 
- https://github.com/bids-standard/bids-specification/pull/1734/files#r1534475631

---

- Closes #1736 
- Supersedes #1734
- Closes #1519

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -274,7 +274,7 @@ A guide for using macros can be found at
 -->
 {{ MACROS___make_filetree_example(
     {
-    "my_dataset-1": {
+    "my_project-1": {
         "sourcedata": {
             "dicoms": {},
             "raw": {
@@ -290,9 +290,7 @@ A guide for using macros can be found at
             "pipeline_1": {},
             "pipeline_2": {},
             "...": "",
-        },
-        "dataset_description.json": "",
-        "...": "",
+        }
     }
    }
 ) }}
@@ -301,7 +299,7 @@ In this example, `sourcedata/dicoms` is not nested inside
 `sourcedata/raw`, **and only the `sourcedata/raw` subdirectory** is a BIDS-compliant dataset among `sourcedata/` subfolders.
 The subdirectories of `derivatives` MAY be BIDS-compliant derivatives datasets
 (see [Non-compliant derivatives](#non-compliant-derivatives) for further discussion).
-The above example is a fully compliant BIDS dataset, providing a convention useful for organizing source, raw BIDS, and derived BIDS data while maintaining overall BIDS compliance.
+The above example is just a convention useful for organizing source, raw BIDS, and derived BIDS data while maintaining BIDS compliance of the raw data directory.
 When using this convention it is RECOMMENDED to set the `SourceDatasets`
 field in `dataset_description.json` of each subdirectory of `derivatives` to:
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -255,8 +255,8 @@ source data (such as the raw event logs or parameter files, before conversion to
 However, in the case that these data are to be included:
 
 1.  These data MUST be kept in separate `sourcedata` directory.
-    BIDS does not prescribe a particular naming scheme for source data.
-    But it is recommended for it to follow BIDS naming convention where possible.
+    BIDS does not prescribe a particular naming scheme for source data,
+    but it is recommended for it to follow BIDS naming convention where possible.
     For example: `sourcedata/sub-01/ses-pre/func/sub-01_ses-pre_task-rest_bold.dicom.tgz` or
     `sourcedata/sub-01/ses-pre/func/MyEvent.sce`.
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -379,6 +379,7 @@ Derivatives can be stored/distributed in two ways:
             "sub-01": {},
             "sub-02": {},
             "...": "",
+            "dataset_description.json": "",
             }
         }
     ) }}
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -101,7 +101,7 @@ instead of specified with an `n/a` value, or included as an empty file
 
 BIDS does not prescribe a particular naming scheme for directories containing individual BIDS datasets.
 However, it is RECOMMENDED to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens to separate words.
-BIDS datasets embedded within a larger BIDS dataset MAY follow some convention (see e.g. [Storage of derived datasets](#storage-of-derived-datasets)).
+BIDS datasets embedded within a larger BIDS dataset MAY follow some convention (see for example [Storage of derived datasets](#storage-of-derived-datasets)).
 
 ## Filesystem structure
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -268,40 +268,38 @@ A guide for using macros can be found at
 {{ MACROS___make_filetree_example(
     {
     "my_dataset-1": {
-            "sourcedata": "",
-            "...": "",
-            "rawdata": {
+        "sourcedata": {
+            "dicoms": {},
+            "raw": {
                 "dataset_description.json": "",
-                "participants.tsv": "",
                 "sub-01": {},
                 "sub-02": {},
                 "...": "",
             },
-            "derivatives": {
-                "pipeline_1": {},
-                "pipeline_2": {},
-                "...": "",
-            },
-        }
+            "..." : "",
+        },
+        "derivatives": {
+            "pipeline_1": {},
+            "pipeline_2": {},
+            "...": "",
+        },
+        "dataset_description.json": "",
+        "...": "",
     }
+   }
 ) }}
 
-In this example, where `sourcedata` and `derivatives` are not nested inside
-`rawdata`, **only the `rawdata` subdirectory** needs to be a BIDS-compliant
-dataset.
+In this example, `sourcedata/dicoms` is not nested inside
+`sourcedata/raw`, **and only the `sourcedata/raw` subdirectory** is a BIDS-compliant dataset among `sourcedata/` subfolders.
 The subdirectories of `derivatives` MAY be BIDS-compliant derivatives datasets
 (see [Non-compliant derivatives](#non-compliant-derivatives) for further discussion).
-This specification does not prescribe anything about the contents of `sourcedata`
-directories in the above example - nor does it prescribe the `sourcedata`,
-`derivatives`, or `rawdata` directory names.
-The above example is just a convention that can be useful for organizing raw,
-source, and derived data while maintaining BIDS compliance of the raw data
-directory. When using this convention it is RECOMMENDED to set the `SourceDatasets`
+The above example is a fully compliant BIDS dataset, providing a convention useful for organizing source, raw BIDS, and derived BIDS data while maintaining overall BIDS compliance.
+When using this convention it is RECOMMENDED to set the `SourceDatasets`
 field in `dataset_description.json` of each subdirectory of `derivatives` to:
 
 ```JSON
 {
-  "SourceDatasets": [ {"URL": "../../rawdata/"} ]
+  "SourceDatasets": [ {"URL": "../../sourcedata/raw/"} ]
 }
 ```
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -278,10 +278,11 @@ A guide for using macros can be found at
         "sourcedata": {
             "dicoms": {},
             "raw": {
-                "dataset_description.json": "",
                 "sub-01": {},
                 "sub-02": {},
                 "...": "",
+                "dataset_description.json": "",
+				"...": "",
             },
             "..." : "",
         },
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -100,7 +100,7 @@ instead of specified with an `n/a` value, or included as an empty file
 ## Dataset naming
 
 BIDS does not prescribe a particular naming scheme for directories containing individual BIDS datasets.
-However, it is RECOMMENDED to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens to separate words.
+However, it is recommended to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens or underscores to separate words.
 BIDS datasets embedded within a larger BIDS dataset MAY follow some convention (see for example [Storage of derived datasets](#storage-of-derived-datasets)).
 
 ## Filesystem structure
@@ -256,7 +256,7 @@ However, in the case that these data are to be included:
 
 1.  These data MUST be kept in separate `sourcedata` directory.
     BIDS does not prescribe a particular naming scheme for source data.
-    But it is RECOMMENDED for it to follow BIDS naming convention where possible.
+    But it is recommended for it to follow BIDS naming convention where possible.
     For example: `sourcedata/sub-01/ses-pre/func/sub-01_ses-pre_task-rest_bold.dicom.tgz` or
     `sourcedata/sub-01/ses-pre/func/MyEvent.sce`.
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -97,6 +97,12 @@ and/or files (like `events.tsv`) are fully omitted *when they are unavailable or
 instead of specified with an `n/a` value, or included as an empty file
 (for example an empty `events.tsv` file with only the headers included).
 
+## Dataset naming
+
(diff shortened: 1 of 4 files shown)


Remi-Gau (approved):
I am good with this approach. Hopefully this should prevent confusion.

yarikoptic:
> What the example looks like

anal me notices now that `dataset_description.json` in that example is "inconsistent" it its ordering with directories... 

<details>
<summary>apparently we do not have full consistency but mostly list it AFTER the folders... but then logically it might need ... before if there are sub- folders, and after to signal that more files are possible...</summary> 

```shell
src/appendices/qmri.md-   {
src/appendices/qmri.md-    "ds-example": {
src/appendices/qmri.md-        "derivatives": {
src/appendices/qmri.md-            "qMRLab": {
src/appendices/qmri.md:                "dataset_description.json": "",
src/appendices/qmri.md-                "sub-01": {
src/appendices/qmri.md-                    "anat": {
src/appendices/qmri.md-                        "sub-01_T1map.nii.gz": "",
src/appendices/qmri.md-                        "sub-01_T1map.json": "",
--
src/common-principles.md-    "my_dataset-1": {
src/common-principles.md-        "sourcedata": {
src/common-principles.md-            "dicoms": {},
src/common-principles.md-            "raw": {
src/common-principles.md:                "dataset_description.json": "",
src/common-principles.md-                "sub-01": {},
src/common-principles.md-                "sub-02": {},
src/common-principles.md-                "...": "",
src/common-principles.md-            },
--
src/common-principles.md-            "pipeline_1": {},
src/common-principles.md-            "pipeline_2": {},
src/common-principles.md-            "...": "",
src/common-principles.md-        },
src/common-principles.md:        "dataset_description.json": "",
src/common-principles.md-        "...": "",
src/common-principles.md-    }
src/common-principles.md-   }
src/common-principles.md-) }}
--
src/common-principles.md-            },
src/common-principles.md-            "sub-01": {},
src/common-principles.md-            "sub-02": {},
src/common-principles.md-            "...": "",
src/common-principles.md:            "dataset_description.json": "",
src/common-principles.md-            }
src/common-principles.md-        }
src/common-principles.md-    ) }}
src/common-principles.md-
--
src/common-principles.md-        },
src/common-principles.md-    "derivatives": {},
src/common-principles.md-    "README": "",
src/common-principles.md-    "participants.tsv": "",
src/common-principles.md:    "dataset_description.json": "",
src/common-principles.md-    "CHANGES": "",
src/common-principles.md-    }
src/common-principles.md-) }}
src/common-principles.md-
--
src/derivatives/common-data-types.md-    "raw/": {
src/derivatives/common-data-types.md-        "CHANGES": "",
src/derivatives/common-data-types.md-        "README": "",
src/derivatives/common-data-types.md-        "channels.tsv": "",
src/derivatives/common-data-types.md:        "dataset_description.json": "",
src/derivatives/common-data-types.md-        "participants.tsv": "",
src/derivatives/common-data-types.md-            "sub-001": {
src/derivatives/common-data-types.md-                "eeg": {
src/derivatives/common-data-types.md-                    "sub-001_task-listening_events.tsv": "",
--
src/longitudinal-and-multi-site-studies.md-            },
src/longitudinal-and-multi-site-studies.md-        "sub-control01_sessions.tsv": "",
src/longitudinal-and-multi-site-studies.md-        },
src/longitudinal-and-multi-site-studies.md-    "participants.tsv": "",
src/longitudinal-and-multi-site-studies.md:    "dataset_description.json": "",
src/longitudinal-and-multi-site-studies.md-    "README": "",
src/longitudinal-and-multi-site-studies.md-    "CHANGES": "",
src/longitudinal-and-multi-site-studies.md-    }
src/longitudinal-and-multi-site-studies.md-) }}

```
</details>

I will at least unify for that now in a separate commit (easy to drop if we decide not worth it)

yarikoptic:
note: failed to upload coverage, I restarted that pipeline to get it hopefully green.

yarikoptic:
yes, it is all green. Waiting for more blessings.  @effigies @tsalo -- WDYT?

src/common-principles.md
```diff
@@ -97,6 +97,12 @@ and/or files (like `events.tsv`) are fully omitted *when they are unavailable or
 instead of specified with an `n/a` value, or included as an empty file
 (for example an empty `events.tsv` file with only the headers included).
 
+## Dataset naming
+
+BIDS does not prescribe a particular naming scheme for directories containing individual BIDS datasets.
+However, it is RECOMMENDED to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens to separate words.
```
effigies, line 103: RECOMMENDED implies a warning. I would remove this line, unless you want to add a validator rule.

```suggestion
```

effigies:
This feels more confusing to me.

```
└─ my_dataset-1/
   ├─ sourcedata/
   │  ├─ dicoms/
   │  ├─ raw/
   │  │  ├─ sub-01/
   │  │  ├─ sub-02/
   │  │  ├─ ... 
   │  │  └─ dataset_description.json 
   │  └─ ... 
   ├─ derivatives/
   │  ├─ pipeline_1/
   │  ├─ pipeline_2/
   │  └─ ... 
   ├─ dataset_description.json 
   └─ ... 
```

Why is there now a `dataset_description.json`? What are the contents of this outer dataset apart from `sourcedata/` and `derivatives/`?

I still don't understand why what is present is confusing, but maybe it would be better just to remove it. I would like to move to a [Diataxis](https://diataxis.fr/) approach, where we clearly separate out the helpful hints and how-tos from the reference material. Maybe it could be revived in a how-to guide in a way that causes fewer problems, if there's any value to it.

src/common-principles.md
```diff
@@ -248,9 +254,10 @@ recommending a particular naming scheme for including different types of
 source data (such as the raw event logs or parameter files, before conversion to BIDS).
 However, in the case that these data are to be included:
 
-1.  These data MUST be kept in separate `sourcedata` directory with a similar
-    directory structure as presented below for the BIDS-managed data. For example:
-    `sourcedata/sub-01/ses-pre/func/sub-01_ses-pre_task-rest_bold.dicom.tgz` or
+1.  These data MUST be kept in separate `sourcedata` directory.
+    BIDS does not prescribe a particular naming scheme for source data.
+    But it is RECOMMENDED for it to follow BIDS naming convention where possible.
```
effigies, line 259:
```suggestion
    BIDS does not prescribe a particular naming scheme for source data,
    but we suggest following BIDS naming conventions to the extent possible.
```

effigies (commented):
Some nit-picks, if the consensus is that this is an improvement and gets cleared to merge. I don't think we want to imply validation of these new bits.

yarikoptic:
> └─ my_dataset-1/
> ...
> Why is there now a `dataset_description.json`? What are the contents of this outer dataset apart from `sourcedata/` and `derivatives/`?

IMHO in BIDS specification we SHOULD only talk about BIDS datasets. We SHOULD NOT talk about arbitrary conventions which could exist outside of BIDS specification.  Hence IMHO this example SHOULD be a BIDS dataset as well -- AFAIK nothing in BIDS specification would state that it is an illegitimate BIDS dataset, or am I wrong?

> I still don't understand why what is present is confusing, but maybe it would be better just to remove it. I would like to move to a [Diataxis](https://diataxis.fr/) approach, where we clearly separate out the helpful hints and how-tos from the reference material. Maybe it could be revived in a how-to guide in a way that causes fewer problems, if there's any value to it.

IMHO it is useful to give users guidance. My problem with current formulation is that it is misleading. This PR IMHO clarifies it, unless it is indeed "not BIDS compliant".

src/common-principles.md
effigies: It is our goal. It is more true in the schema validator than in the legacy validator.

yarikoptic: So -- I would not mind having  validator support for that.
Also I could use the "trick" you suggested below instead for now
```suggestion
However, we suggest to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens to separate words.
```

;-)  but I do not think we actually want to introduce such a new non-standardized notion as "we suggest" (here or below), do we ?

effigies: Right, I don't really like these unenforceable recommendations. It makes it unclear what's spec and what's not.

yarikoptic:
> there are no validatable files

wrong. There is `dataset_description.json`. Might be a bug in bids-validator then ;-)

> You can argue that's wrong, but from an OpenNeuro perspective, I would want to reject such a dataset as it's effectively a blanket `.bidsignore`.

as we discussed -- this sounds like an archive specific desire/behavior.

> But we would also reject what you're replacing, since it wasn't a BIDS dataset, but a layout demonstrating one way to include BIDS datasets alongside their sources and derivatives.

ok, let's meet half-way: what if I remove `dataset_specification.json` for now , and mostly revert that added [statement](https://github.com/bids-standard/bids-specification/pull/1741/files#diff-38cc3c29d8cceeb219373964b67b2286b204e420aedb3029bb9344e5dad02956R304) that this is a valid bids dataset?  and then propose separate follow up PR on that particular aspect since I think we would want also some explicit `DatasetType = "study"` or alike (e.g. "project" to possibly avoid conflict with BEP035) for such cases.

Then we stay with "convention"

effigies:
Sure, if we drop `dataset_description.json` and `...` and do not say or imply that the outer dataset is a BIDS dataset, then this is basically the same as before. I suppose that's fine.

src/common-principles.md
yarikoptic: to me it is more of using an established in common principles terminology (REQUIRED, RECOMMENDED, OPTIONAL). In there `REQUIRED` and `OPTIONAL` are really two opposite ends pertinent to what spec mandates or allows for. RECOMMENDED also semantically means to me advisory similar to "we suggest" and does not necessarily imply that it is part of the specification "mandate".  But I am ok also to make it e.g. lowercase to somewhat emphasize that it is really not the RECOMMENDED as part of the specification. WDYT?

  ```suggestion
However, it is recommended to use a short descriptive name that reflects the content of the dataset, avoid spaces in the name, and use hyphens to separate words.
```

File: src/common-principles.md [hunk 3]
```diff
@@ -267,41 +274,38 @@ A guide for using macros can be found at
...
+`sourcedata/raw`, **and only the `sourcedata/raw` subdirectory** is a BIDS-compliant dataset among `sourcedata/` subfolders.
 The subdirectories of `derivatives` MAY be BIDS-compliant derivatives datasets
 (see [Non-compliant derivatives](#non-compliant-derivatives) for further discussion).
-This specification does not prescribe anything about the contents of `sourcedata`
-directories in the above example - nor does it prescribe the `sourcedata`,
-`derivatives`, or `rawdata` directory names.
-The above example is just a convention that can be useful for organizing raw,
-source, and derived data while maintaining BIDS compliance of the raw data
-directory. When using this convention it is RECOMMENDED to set the `SourceDatasets`
+The above example is just a convention useful for organizing source, raw BIDS, and derived BIDS data while maintaining BIDS compliance of the raw data directory.
```
effigies, line 302: Couldn't this line could equally apply to the original structure? If this is enough to relieve the confusion, why not just do this and 

```diff
-In this example, where `sourcedata` and `derivatives` are not nested inside
-`rawdata`, **only the `rawdata` subdirectory** needs to be a BIDS-compliant
-dataset.
+In this example, `sourcedata` is not nested inside `rawdata`,
+**and only the `rawdata` subdirectory** is a BIDS-compliant dataset.
```

above?

Sorry to be really belaboring this, but I feel like a lot of effort is going into a minor clarification, and someone who notices this change is going to have to spend effort to understand that it's purely superficial.

src/common-principles.md
effigies: That would be fine with me. Do we care about "hyphens" being overly restrictive, or are we okay leaving it to people's judgment? I often use underscores as well when naming directories.

src/common-principles.md
yarikoptic: per  discussion [above](https://github.com/bids-standard/bids-specification/pull/1741/files#r1568910761)

```suggestion
    BIDS does not prescribe a particular naming scheme for source data.
    But it is recommended for it to follow BIDS naming convention where possible.
```
to be committed along when we converge in suggestions

src/common-principles.md
yarikoptic: Yes, correct **but** my main problem with the original formulation is that we were suggesting something  which is **not** BIDS-compliant if adopted within a BIDS dataset, and context is not entirely clear here that we are not talking about a BIDS dataset. And if it is such a great idea to have `rawdata/` then why BIDS itself doesn't have it?

I agreed to remove `dataset_description.json` (which IMHO would have immediately made it "legit" according to spec, not validator, BIDS dataset), but I would like to preserve this adjusted recommended structure since if adopted within a BIDS dataset, it would remain legit, thus we are not somewhat misguiding a reader suggesting that `rawdata/` subdirectory is a legit way to describe BIDS-compliant dataset *within* a BIDS dataset.

yarikoptic:
To expedite and ease re-reviewing -- I committed that final form of suggestions .

src/common-principles.md
```diff
@@ -248,9 +254,10 @@ recommending a particular naming scheme for including different types of
 source data (such as the raw event logs or parameter files, before conversion to BIDS).
 However, in the case that these data are to be included:
 
-1.  These data MUST be kept in separate `sourcedata` directory with a similar
-    directory structure as presented below for the BIDS-managed data. For example:
-    `sourcedata/sub-01/ses-pre/func/sub-01_ses-pre_task-rest_bold.dicom.tgz` or
+1.  These data MUST be kept in separate `sourcedata` directory.
+    BIDS does not prescribe a particular naming scheme for source data.
+    But it is recommended for it to follow BIDS naming convention where possible.
```
effigies, line 259: Grammar nitpick.

```suggestion
    BIDS does not prescribe a particular naming scheme for source data,
    but it is recommended for it to follow BIDS naming convention where possible.
```

effigies (approved):

yarikoptic:
there is no 'ideal' setup per se, but if datalad used, then 2. (`sourcedata/raw`) could also have (possibly subset of only relevant ones) of `sourcedata/` (like `sourcedata/dicoms`) be installed into its own `sourcedata/raw/sourcedata` .  The idea here is just to make it all "flat" and separated only to `sourcedata/` (bids or not) and `derivatives/` (what computed from those).

yarikoptic:
@effigies @sappelhoff so are we go on this one or more tune ups?

PR #1742, opened 2024-03-21
SCHEMA: Allow surrounding whitespace in integers

effigies:
Discovered thanks to @ericearl's magical database. Some people generated TSVs that space-pad their numerical columns. Actual tools don't have a problem interpreting these numbers (pandas, deno, Python), just our regular expressions that we use to validate the columns as strings.

### DIFF
--- a/src/schema/objects/formats.yaml
+++ b/src/schema/objects/formats.yaml
@@ -22,12 +22,12 @@ integer:
   display_name: Integer
   description: |
     An integer which may be positive or negative.
-  pattern: '[+-]?\d+'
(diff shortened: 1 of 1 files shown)


effigies:
@sappelhoff Would appreciate your thoughts here.

sappelhoff (approved):
> Some people generated TSVs that space-pad their numerical columns

I am interested in why people would do that.

However **if** as you say the most commonly used tools do not have a problem parsing this, I am +1 for the changes here.

effigies:
The example I saw was so that onsets <10 would line up with the onsets >=10.

PR #1743, opened 2024-03-21
[MAINT] add BIDS papers references in citation.cff

### LINKED ISSUE #1660: [INFRA] Update `CITATION.cff` with all BIDS paper references
Within this file: https://github.com/bids-standard/bids-specification/blob/f7d12988107ff056e049cb6e54458ca57775e190/CITATION.cff#L3

we should add a section "References", as described here: https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#references

And in that section, we should add the data for all the papers that we list here: https://bids-specification.readthedocs.io/en/latest/introduction.html#citing-bids

Remi-Gau:
- fixes #1660

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -28,6 +28,46 @@ identifiers:
   - description: PDFs of the BIDS specification for versions 1.0.0 to 1.8.0
     type: doi
     value: 10.5281/zenodo.3686061
+  # papers to inject in the how to cite section
+  # ### Original publication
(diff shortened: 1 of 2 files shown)


Remi-Gau:
OK this is using the mkdocs extension to rely on bibtex

image

but this requires switching to conda for installing some python dependencies (I think)...

So that is starting to feel a bit more invovled than I would like...

Issue #472, opened 2020-05-15
Allow for .tsv (in addition to .tsv.gz) for _physio files


yarikoptic:
- Have you checked our contributing guide? **yes (awhile back ;-))**

- Is your idea backwards compatible? **yes**

- Is there already a group working on your idea? **didn't find any issue***

- Will your idea potentially require a large effort? **not at all, I will submit a PR if there is some approval**

ATM we do allow for both `.nii` and `.nii.gz`; typically we demand `.tsv` (without `.gz`) for tabular data and then for `_physio` we demand it to be `.gz`-ed.  No option is given to have them not `.gz'ed`.

Use case: HCP dataset provides physio recordings a tab separated file (with an extension `.txt`), e.g. as in `MNINonLinear/Results/rfMRI_REST1_LR/rfMRI_REST1_LR_Physio_log.txt`. There is [an ongoing effort](https://github.com/datalad-datasets/hcp-functional-connectivity/pull/1) to provide ready to use datalad datasets with HCP data layed out to BIDS.  As files are not publicly available, we cannot just gzip them and place under datalad control [*] - they must come "raw" ;)

So what would be your opinion on allowing `.tsv` in addition to `.tsv.gz` for `_physio`? 

[*] theoretically speaking someone could provide a git-annex special remote opposite of `datalad-archive` which would compress them from uncompressed version "on the fly".  Any takers?

effigies:
Would this be more like our usual tabular data, with column names as headers in the TSV file, or like `physio.tsv.gz`, with column names as in JSON sidecars?

yarikoptic:
there is no header in those files, so I guess ideally we should provide those in sidecar .json indeed (which we could just commit to git, no secrets there I hope ;))

sappelhoff:
I am not aware why physio TSVs are required to be gzipped.

Rather from the opposite angle, I remember a discussion where we stopped supporting `fif.gz` files (now we only allow `.fif` files) because:

- the amount of saved space is negligible (so was the argument)
- the increased burden on the user outweighs the benefits

so from a naive viewpoint I don't see what's wrong with allowing TSV for physio files. But please enlighten me about the drawbacks.

effigies:
[...]

PR #1749, opened 2024-03-25
ENH: Add compressed TSV files to the common principles

oesteban:
Their description is hedged within the physiological recordings so upcasting them to the common principles seems important.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -434,9 +434,13 @@ Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
 (with [extension `.tsv`](glossary.md#tsv-extensions)),
 that is, CSV files where commas are replaced by tab characters.
 Tabs MUST be true tab characters and MUST NOT be a series of space characters.
-Tabular data containing large numbers of rows MAY be saved as
-[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
-as prescribed below.
+Tabular data such as continuous physiology recordings typically containing
+large numbers of rows MAY be saved as
+[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files),
+which are introduced below.
+Plain-text TSV and compressed TSV are not interchangeable, that is, each section
+of the specification prescribes which one SHOULD be used for the data type at
+hand.
 Each TSV file MUST start with a header line listing the names of all columns
 with two exceptions:
 
@@ -558,8 +562,9 @@ like in the example below.
 
 ### Compressed tabular files
 
-Large tabular information such as physiological recordings MAY be stored with
-[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions).
+Large tabular information, such as physiological recordings, SHOULD be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
+so established by the specifications.
 Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
 
 1.  The contents of TSVGZ files MUST be compressed with
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -463,8 +463,8 @@ TSV files MUST be in UTF-8 encoding.
 Example:
 
 ```Text
-onset	duration	response_time	correct	stop_trial	go_trial
-200     200         0               n/a     n/a         n/a
+onset   duration    response_time   trial_type        trial_extra
+200     20.0        127.34e-1       ⌂ (UTF-8 char)    n/a
 ```
 
 !!! warning "The TSV examples in this document (like the one above this note) are occasionally formatted using space characters instead of tabs to improve human readability."
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -430,16 +430,22 @@ NIfTI header.
 
 ### Tabular files
 
-Tabular data MUST be saved as plain-text, tab-delimited values (`.tsv`) files,
+Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
+(with [extension `.tsv`](glossary.md#tsv-extensions)),
 that is, CSV files where commas are replaced by tab characters.
 Tabs MUST be true tab characters and MUST NOT be a series of space characters.
 Tabular data containing large numbers of rows MAY be saved as
 [compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
 as prescribed below.
 Each TSV file MUST start with a header line listing the names of all columns
-with the exception of [compressed tabular files](#compressed-tabular-files)
-where column names are defined in a sidecar metadata
-[JSON object](https://www.json.org/json-en.html) described below.
+with two exceptions:
+
+1.  [compressed tabular files](#compressed-tabular-files),
+    for which column names are defined in a sidecar metadata
+    [JSON object](https://www.json.org/json-en.html) described below; and
+1.  [motion recording data](modality-specific-files/motion.md),
+    which use plain-text TSV and columns are defined as described
+    in its corresponding of the specifications.
 
 It is RECOMMENDED that the column names in the header of the TSV file are
 written in [`snake_case`](https://en.wikipedia.org/wiki/Snake_case) with the
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -571,8 +571,7 @@ Rules for formatting plain-text tabular files apply to TSVGZ files with three ex
     [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
 1.  Compressed tabular files MUST NOT contain a header in the first row
     indicating the column names.
-1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
-    corresponding tabular file but with a `.json` extension.
+1.  TSVGZ files MUST have an associated JSON file that defines the columns in the tabular file.
 
 
 !!! warning "Attention"
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -430,36 +430,41 @@ NIfTI header.
 
 ### Tabular files
 
-Tabular data MUST be saved as tab delimited values (`.tsv`) files, that is, CSV
-files where commas are replaced by tabs. Tabs MUST be true tab characters and
-MUST NOT be a series of space characters. Each TSV file MUST start with a header
-line listing the names of all columns (with the exception of
-[physiological and other continuous recordings](modality-specific-files/physiological-and-other-continuous-recordings.md)
-as well as [motion recording data](modality-specific-files/motion.md)).
+Tabular data MUST be saved as plain-text, tab-delimited values (`.tsv`) files,
+that is, CSV files where commas are replaced by tab characters.
+Tabs MUST be true tab characters and MUST NOT be a series of space characters.
+Tabular data containing large numbers of rows MAY be saved as
+[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
+as prescribed below.
+Each TSV file MUST start with a header line listing the names of all columns
+with the exception of [compressed tabular files](#compressed-tabular-files)
+where column names are defined in a sidecar metadata
+[JSON object](https://www.json.org/json-en.html) described below.
+
 It is RECOMMENDED that the column names in the header of the TSV file are
 written in [`snake_case`](https://en.wikipedia.org/wiki/Snake_case) with the
 first letter in lower case (for example, `variable_name`, not `Variable_name`).
-As for all other data in the TSV files, column names MUST be separated with tabs.
+Column names defined in the header MUST be separated with tabs as for the data contents.
 Furthermore, column names MUST NOT be blank (that is, an empty string) and MUST NOT
 be duplicated within a single TSV file.
-String values containing tabs MUST be escaped using double
-quotes. Missing and non-applicable values MUST be coded as `n/a`. Numerical
-values MUST employ the dot (`.`) as decimal separator and MAY be specified
+String values containing tabs MUST be escaped using double quotes.
+Missing and non-applicable values MUST be coded as `n/a`.
+Numerical values MUST employ the dot (`.`) as decimal separator and MAY be specified
 in scientific notation, using `e` or `E` to separate the significand from the
-exponent. TSV files MUST be in UTF-8 encoding.
+exponent.
+TSV files MUST be in UTF-8 encoding.
 
 Example:
 
 ```Text
 onset	duration	response_time	correct	stop_trial	go_trial
-200	200	0	n/a	n/a	n/a
+200     200         0               n/a     n/a         n/a
 ```
 
-**Note**: The TSV examples in this document (like the one above this note)
-are occasionally formatted using space characters instead of tabs to improve
-human readability.
-Directly copying and then pasting these examples from the specification
-for use in new BIDS datasets can lead to errors and is discouraged.
+!!! warning "The TSV examples in this document (like the one above this note) are occasionally formatted using space characters instead of tabs to improve human readability."
+
+    Directly copying and then pasting these examples from the specification
+    for use in new BIDS datasets can lead to errors and is discouraged.
 
 Tabular files MAY be optionally accompanied by a simple data dictionary
 in the form of a JSON [object](https://www.json.org/json-en.html)
@@ -536,7 +541,7 @@ like in the example below.
             "F": {
                 "Description": "Female",
                 "TermURL": "https://www.ncbi.nlm.nih.gov/mesh/68005260"
-            },
+            }
         }
     }
 }
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -571,9 +571,11 @@ Rules for formatting plain-text tabular files apply to TSVGZ files with three ex
     In contrast to plain-text TSV files, compressed tabular files files
     MUST NOT include a header line.
     Column names MUST be specified in the JSON file following the
-    [`Columns` metadata](glossary.md#columns-metadata) specifications.
-    As plain-text tabular data, column names MUST NOT be blank (that is, an empty string),
-    and MUST NOT be duplicated within a single JSON file describing a TSVGZ file.
+    [`Columns` metadata](glossary.md#columns-metadata) specifications
+    provided with plain-text tabular data.
+    Similarly, specific column metadata stored with dictionaries such as
+    the `sex` column in the example above MUST be specified following the
+    plain-text tabular metadata prescriptions.
 
     TSVGZ are header-less to improve compatibility with existing software
     (for example, FSL, or PNM), and to facilitate the support for other file formats
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -468,7 +468,7 @@ onset   duration    response_time   trial_type        trial_extra
 240     5.0         17.34e-1        visual            n/a
 ```
 
-!!! warning "TSV example formatting"
+!!! warning "Attention"
 
     The TSV examples in this document (like the one above this note) are occasionally
     formatted using space characters instead of tabs to improve human readability.
@@ -570,7 +570,7 @@ Rules for formatting plain-text tabular files apply to TSVGZ files with three ex
     corresponding tabular file but with a `.json` extension.
 
 
-!!! note "Column definitions in compressed tabular files"
+!!! warning "Attention"
 
     In contrast to plain-text TSV files,
     compressed tabular files files MUST NOT include a header line.
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -467,8 +467,10 @@ onset   duration    response_time   trial_type        trial_extra
 200     20.0        127.34e-1       ⌂ (UTF-8 char)    n/a
 ```
 
-!!! warning "The TSV examples in this document (like the one above this note) are occasionally formatted using space characters instead of tabs to improve human readability."
+!!! warning "TSV example formatting"
 
+    The TSV examples in this document (like the one above this note) are occasionally
+    formatted using space characters instead of tabs to improve human readability.
     Directly copying and then pasting these examples from the specification
     for use in new BIDS datasets can lead to errors and is discouraged.
 
@@ -566,16 +568,15 @@ Rules for formatting plain-text tabular files apply to TSVGZ files with three ex
 1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
     corresponding tabular file but with a `.json` extension.
 
-???+ warning "Columns of TSVGZ files MUST be defined in the corresponding JSON sidecar and the tabular content MUST NOT include a header line."
 
-    In contrast to plain-text TSV files, compressed tabular files files
-    MUST NOT include a header line.
-    Column names MUST be specified in the JSON file following the
-    [`Columns` metadata](glossary.md#columns-metadata) specifications
-    provided with plain-text tabular data.
-    Similarly, specific column metadata stored with dictionaries such as
-    the `sex` column in the example above MUST be specified following the
-    plain-text tabular metadata prescriptions.
+!!! note "Column definitions in compressed tabular files"
+
+    In contrast to plain-text TSV files,
+    compressed tabular files files MUST NOT include a header line.
+    Column names MUST be provided in the JSON file with the
+    [`Columns`](glossary.md#columns-metadata) field.
+    Each column MAY additionally be described with a column description,
+    as described in [Tabular files](#tabular-files).
 
     TSVGZ are header-less to improve compatibility with existing software
     (for example, FSL, or PNM), and to facilitate the support for other file formats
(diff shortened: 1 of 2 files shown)


src/common-principles.md
```diff
@@ -430,36 +430,41 @@ NIfTI header.
...
+Tabular data MUST be saved as plain-text, tab-delimited values (`.tsv`) files,
+that is, CSV files where commas are replaced by tab characters.
+Tabs MUST be true tab characters and MUST NOT be a series of space characters.
+Tabular data containing large numbers of rows MAY be saved as
+[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
+as prescribed below.
+Each TSV file MUST start with a header line listing the names of all columns
+with the exception of [compressed tabular files](#compressed-tabular-files)
+where column names are defined in a sidecar metadata
+[JSON object](https://www.json.org/json-en.html) described below.
```
oesteban, line 442: Watch out - there are small consistency changes that are less apparent because of the use of semantic line breaks - hope that's fine for such a short excerpt.

src/common-principles.md
```diff
@@ -430,36 +430,41 @@ NIfTI header.
...
-exponent. TSV files MUST be in UTF-8 encoding.
+exponent.
+TSV files MUST be in UTF-8 encoding.
 
 Example:
 
 ```Text
 onset	duration	response_time	correct	stop_trial	go_trial
-200	200	0	n/a	n/a	n/a
+200     200         0               n/a     n/a         n/a
```
oesteban, line 461: the use of actual tabs voided the purpose of the note below.

Remi-Gau:
Slightly related to #644

Remi-Gau:
I think that this reverts some of #1699, no?
Specifically the exception to the rule for tsv for motion data as motion.tsv are header less.

oesteban:
It feels lines 46-48 of the diff should stay. The first parenthesis is
addressed, but can be made more explicit either here or (better, IMHO)
within motion.

I can take care of harmonizing this.

oesteban:
> I think that this reverts some of #1699, no? Specifically the exception to the rule for tsv for motion data as motion.tsv are header less.

Okay, I'm realizing a problem -- motion's tsv files ARE NOT compressed, is that right?

I'll push a commit improving some of my changes not to undo the motion exception.

Remi-Gau:
> > I think that this reverts some of #1699, no? Specifically the exception to the rule for tsv for motion data as motion.tsv are header less.
> 
> Okay, I'm realizing a problem -- motion's tsv files ARE NOT compressed, is that right?
> 
> I'll push a commit improving some of my changes not to undo the motion exception.

Yup. Sorry. Pre coffee me was not clear. 
This is an "inconsistency" we realized only recently .

oesteban:
Okay, compatibilization with #1699 is done (https://github.com/bids-standard/bids-specification/pull/1749/commits/c4ee8d68cdff253055f86ed82c3992de8562bd35).

> Slightly related to #644

I doubled down with https://github.com/bids-standard/bids-specification/pull/1749/commits/4eadad536ca962c9a113b4f4aeaa6f195938f20f. I am happy to see that changed if #644 leads the table somewhere else later on.

src/common-principles.md
```diff
@@ -536,12 +547,38 @@ like in the example below.
...
 }
 ```
 
+### Compressed tabular files
+
+Large tabular information such as physiological recordings MAY be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions).
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files SHOULD be compressed with
```
effigies, line 562:
```suggestion
1.  The contents of TSVGZ files MUST be compressed with
```

src/common-principles.md
```diff
@@ -536,12 +547,38 @@ like in the example below.
...
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files SHOULD NOT contain a header in the first row
```
effigies, line 564:
```suggestion
1.  Compressed tabular files MUST NOT contain a header in the first row
```

src/common-principles.md
```diff
@@ -536,12 +547,38 @@ like in the example below.
...
+    indicating the column names.
+1.  TSVGZ files SHOULD be accompanied by a JSON file with the same name as their
```
effigies, line 566:
```suggestion
1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
```

src/common-principles.md
```diff
@@ -536,12 +547,38 @@ like in the example below.
...
+    corresponding tabular file but with a `.json` extension.
+
+???+ warning "Columns of TSVGZ files MUST be defined in the corresponding JSON sidecar and the tabular content MUST NOT include a header line."
+
+    In contrast to plain-text TSV files, compressed tabular files files
+    MUST NOT include a header line.
+    Column names MUST be specified in the JSON file following the
+    [`Columns` metadata](glossary.md#columns-metadata) specifications.
+    As plain-text tabular data, column names MUST NOT be blank (that is, an empty string),
+    and MUST NOT be duplicated within a single JSON file describing a TSVGZ file.
```
effigies, line 576: I'm confused. This seems to contradict the previous SHOULD lines. What am I missing?

src/modality-specific-files/physiological-and-other-continuous-recordings.md
```diff
@@ -38,8 +35,10 @@ before the suffix.
 For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 `<matches>` would correspond to `sub-control01_task-nback_run-1`.
 
-Note that when supplying a `*_<physio|stim>.tsv.gz` file, an accompanying
-`*_<physio|stim>.json` MUST be supplied as well.
+!!! warning "TSVGZ files SHOULD NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))"
```
effigies, line 38:
```suggestion
!!! warning "TSVGZ files MUST NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))"
```

src/common-principles.md
oesteban: I believe this was in the physio section meaning that you should not define the same column twice in a JSON object (which is a bit redundant because you can't have the same key twice). Happy to rewrite or remove -- what do you prefer?

effigies: Ah. I would say it just once unless there is a specific exception or caveat here. Otherwise it's unclear why you're saying it twice with slightly different words.

Apologies, I'm not feeling my best, so if what I say doesn't make sense, feel free to ignore me.

oesteban: Sad to hear you are not feeling great :( However, this is definitely confusing in the first place and is confusing to me, too. The current docs read:

https://github.com/bids-standard/bids-specification/blob/bd08602760d2fa0fd61e8f579a2677ea9bda8588/src/modality-specific-files/physiological-and-other-continuous-recordings.md#L75-L77

I just moved this here. I'll try to check if this is duplicated from the standard tabular files and remove if so.

oesteban: Edited to make it clear that this comes from the plain text specs: https://github.com/bids-standard/bids-specification/pull/1749/commits/3811a8c4262a3ef32d7f37e5675ffdc5f41351c6

oesteban: Marking resolved -- please suggest again if the direction I took is not satisfactory.

src/common-principles.md
```diff
@@ -430,36 +430,47 @@ NIfTI header.
...
+as prescribed below.
+Each TSV file MUST start with a header line listing the names of all columns
+with two exceptions:
+
+1.  [compressed tabular files](#compressed-tabular-files),
+    for which column names are defined in a sidecar metadata
+    [JSON object](https://www.json.org/json-en.html) described below; and
+1.  [motion recording data](modality-specific-files/motion.md),
+    which use plain-text TSV and columns are defined as described
+    in its corresponding of the specifications.
```
Remi-Gau, line 448:
```suggestion
    in its corresponding section of the specifications.
```

src/common-principles.md
```diff
@@ -536,12 +547,40 @@ like in the example below.
...
+    Column names MUST be specified in the JSON file following the
+    [`Columns` metadata](glossary.md#columns-metadata) specifications
+    provided with plain-text tabular data.
+    Similarly, specific column metadata stored with dictionaries such as
+    the `sex` column in the example above MUST be specified following the
+    plain-text tabular metadata prescriptions.
+
+    TSVGZ are header-less to improve compatibility with existing software
+    (for example, FSL, or PNM), and to facilitate the support for other file formats
+    in the future.
```
effigies, line 582: Again, too much content in the admonition title. Reading this through again, I would change this to info and write:

```suggestion

!!! note "Column definitions in compressed tabular files"

    In contrast to plain-text TSV files,
    compressed tabular files files MUST NOT include a header line.
    Column names MUST be provided in the JSON file with the
    [`Columns`](glossary.md#columns-metadata) field.
    Each column MAY additionally be described with a column description,
    as described in [Tabular files](#tabular-files).

    TSVGZ are header-less to improve compatibility with existing software
    (for example, FSL, or PNM), and to facilitate the support for other file formats
    in the future.
```

effigies (commented):
Okay, reading through, I think this is good overall. Some specific proposals.

src/common-principles.md
```diff
@@ -536,12 +547,40 @@ like in the example below.
...
+Large tabular information such as physiological recordings MAY be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions).
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files MUST be compressed with
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files MUST NOT contain a header in the first row
+    indicating the column names.
+1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
+    corresponding tabular file but with a `.json` extension.
```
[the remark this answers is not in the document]
oesteban: I would leave it for later, it is not mentioned either for the plain-text tsv files.

src/common-principles.md
```diff
@@ -536,12 +550,39 @@ like in the example below.
...
+
+
+!!! note "Column definitions in compressed tabular files"
```
oesteban, line 573: @effigies - if the title is not going to be long, I'd simplify even more:
```suggestion
!!! warning "Attention"
```
and for this one, I'd re-up it to warning because this is a common friction point for newcomers.

src/common-principles.md
```diff
@@ -430,36 +430,47 @@ NIfTI header.
...
 
-**Note**: The TSV examples in this document (like the one above this note)
-are occasionally formatted using space characters instead of tabs to improve
-human readability.
-Directly copying and then pasting these examples from the specification
-for use in new BIDS datasets can lead to errors and is discouraged.
+!!! warning "The TSV examples in this document (like the one above this note) are occasionally formatted using space characters instead of tabs to improve human readability."
+
+    Directly copying and then pasting these examples from the specification
+    for use in new BIDS datasets can lead to errors and is discouraged.
```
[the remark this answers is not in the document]
oesteban: I have changed the character to a Chinese word ("Chinese", as per google translator is "中国人" so someone who knows please update if necessary).

I believe we do want pandoc to render Chinese characters. Icons are probably unimportant but non-ascii characters are relevant.

Remi-Gau (approved):
the html looks good and the pdf good enough

oesteban:
> the html looks good and the pdf good enough

I just simplified the admonitions' titles so the pdf should be even less far from the HTML now.

effigies (approved):

effigies:
<details><summary>By the regulations, we should give this until Monday (Apr 1) for community review.</summary>

regulations

</details>

src/common-principles.md
```diff
@@ -536,12 +550,39 @@ like in the example below.
...
                 "Description": "Female",
                 "TermURL": "https://www.ncbi.nlm.nih.gov/mesh/68005260"
-            },
+            }
         }
     }
 }
 ```
 
+### Compressed tabular files
```
sappelhoff, line 559: I would make this a sub-heading under tabular files.

```suggestion
#### Compressed tabular files
```

src/modality-specific-files/physiological-and-other-continuous-recordings.md
```diff
@@ -38,8 +35,12 @@ before the suffix.
 For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 `<matches>` would correspond to `sub-control01_task-nback_run-1`.
 
-Note that when supplying a `*_<physio|stim>.tsv.gz` file, an accompanying
-`*_<physio|stim>.json` MUST be supplied as well.
+!!! note "TSVGZ headers are specified in metadata files."
+
+    TSVGZ files MUST NOT include a header line,
+    as established by the [common-principles](../common-principles.md#compressed-tabular-files)).
```
sappelhoff, line 41:
```suggestion
    as established by the [common-principles](../common-principles.md#compressed-tabular-files).
```

sappelhoff (commented):
Thanks for the proposal!

Do I understand it correctly that the current proposal makes gzipped TSV files valid for **all** TSV files (not just physio and stim, as it was before)?

If yes, wouldn't that also need some additions to the BIDS schema?

I read the PR and left two comments. In general, I am in favor, but I feel like this is a big change: With this, users may supply any TSV file they want as TSVGZ. This might break a lot of tools initially. Perhaps we need to clarify this bit:

> Tabular data containing large numbers of rows MAY be saved as
[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
as prescribed below.

What's a "large number of rows"?

---

Somewhat related:

- https://github.com/bids-standard/bids-specification/issues/472

oesteban:
No, the intent is not to allow TSVGZ beyond what it is used for already.
I'll try to think how to make it more clear.

sappelhoff:
> I'll try to think how to make it more clear.

Thanks, I'll mark the parts that I find a bit ambiguous.

src/common-principles.md
```diff
@@ -430,36 +430,50 @@ NIfTI header.
...
-line listing the names of all columns (with the exception of
-[physiological and other continuous recordings](modality-specific-files/physiological-and-other-continuous-recordings.md)
-as well as [motion recording data](modality-specific-files/motion.md)).
+Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
+(with [extension `.tsv`](glossary.md#tsv-extensions)),
+that is, CSV files where commas are replaced by tab characters.
+Tabs MUST be true tab characters and MUST NOT be a series of space characters.
+Tabular data containing large numbers of rows MAY be saved as
+[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files)
+as prescribed below.
```
sappelhoff, line 439: IMHO ambiguous whether this allows TSVGZ for ALL TSV files (xref: https://github.com/bids-standard/bids-specification/pull/1749#issuecomment-2021000568)

src/common-principles.md
```diff
@@ -536,12 +550,39 @@ like in the example below.
...
+            }
         }
     }
 }
 ```
 
+### Compressed tabular files
+
+Large tabular information such as physiological recordings MAY be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions).
```
sappelhoff, line 562: IMHO ambiguous whether this allows TSVGZ for ALL TSV files (xref: https://github.com/bids-standard/bids-specification/pull/1749#issuecomment-2021000568)

src/common-principles.md
oesteban: I'm not a fan of this; creating one more level of depth without siblings feels artificial (\*), and it also makes the section less findable. Curious about what others think.

(\*) At least, it would seem logical to do:

```
## Tabular files
### Plain-text tabular files
### Compressed tabular files
```

src/common-principles.md
oesteban:
```suggestion
Large tabular information, such as physiological recordings, SHOULD be stored with
[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
so established by the specifications.
```

oesteban (commented):
@sappelhoff please have a look and maybe iterate over my suggestions to make the reading of the specs unequivocal regarding TSV vs TSVGZ

oesteban:
@effigies @Remi-Gau @sappelhoff please check the changes to make unambiguous that .tsv and .tsv.gz are not interchangeable https://github.com/bids-standard/bids-specification/pull/1749/commits/48ffe9d1a920b4fe9bde3be69063b703e8ebd3ed

It feels too verbose to me, but I agree with @sappelhoff we need to avert any diverting readings of the specs. TBH, understanding that tsv.gz can be used instead of tsv at every instance felt a bit forced to me, but it's better to make the interpretation absolutely unambiguous.

sappelhoff (approved):
Thanks for the adjustments and the general proposal! I am happy to see this merged either with or without my proposed subheading change.

src/common-principles.md
```diff
@@ -430,36 +430,54 @@ NIfTI header.
...
+Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
+(with [extension `.tsv`](glossary.md#tsv-extensions)),
+that is, CSV files where commas are replaced by tab characters.
+Tabs MUST be true tab characters and MUST NOT be a series of space characters.
+Tabular data such as continuous physiology recordings typically containing
+large numbers of rows MAY be saved as
+[compressed tabular files (with extension `.tsv.gz`)](#compressed-tabular-files),
+which are introduced below.
+Plain-text TSV and compressed TSV are not interchangeable, that is, each section
+of the specification prescribes which one SHOULD be used for the data type at
```
oesteban, line 442: for consistency with the RFC-controlled terms . cc/ @effigies 
```suggestion
of the specification prescribes which one MUST be used for the data type at
```

src/common-principles.md
```diff
@@ -536,12 +554,40 @@ like in the example below.
...
-            },
+            }
         }
     }
 }
 ```
 
+### Compressed tabular files
+
+Large tabular information, such as physiological recordings, SHOULD be stored with
```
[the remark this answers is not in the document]
effigies: As phrased here, this would be a kind of meta-requirement on spec authors to use TSVGZ if they're large, not a requirement on curators/converters, who will receive direction from the individual data type specifications. The problem here is that motion has violated this, using yet another TSV + `channels.tsv` configuration.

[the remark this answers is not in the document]
oesteban: You mean, it would be better to roll back to SHOULD? Happy to change as needed if the problem is not the particular term.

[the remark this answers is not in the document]
effigies: Actually, reading more context, it's ` MUST be stored [...] when so established by the specifications.`

That's fine then.

src/common-principles.md
```diff
@@ -536,12 +554,40 @@ like in the example below.
...
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
+so established by the specifications.
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files MUST be compressed with
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files MUST NOT contain a header in the first row
+    indicating the column names.
+1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
+    corresponding tabular file but with a `.json` extension.
```
oesteban, line 575: From https://github.com/bids-standard/bids-specification/pull/1750#discussion_r1543499316
```suggestion
1.  TSVGZ files MUST have an associated JSON file that defines the columns in the tabular file.
```

src/common-principles.md
```diff
@@ -536,12 +554,39 @@ like in the example below.
...
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
+so established by the specifications.
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files MUST be compressed with
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files MUST NOT contain a header in the first row
+    indicating the column names.
+1.  TSVGZ files MUST have an associated JSON file that defines the columns in the tabular file.
+
```
oesteban, line 575:
```suggestion
1.  TSVGZ files MUST have an associated JSON file that defines the columns in the tabular file.
```

src/common-principles.md
```diff
@@ -430,36 +430,54 @@ NIfTI header.
...
 
-Tabular data MUST be saved as tab delimited values (`.tsv`) files, that is, CSV
-files where commas are replaced by tabs. Tabs MUST be true tab characters and
-MUST NOT be a series of space characters. Each TSV file MUST start with a header
-line listing the names of all columns (with the exception of
-[physiological and other continuous recordings](modality-specific-files/physiological-and-other-continuous-recordings.md)
-as well as [motion recording data](modality-specific-files/motion.md)).
+Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
+(with [extension `.tsv`](glossary.md#tsv-extensions)),
+that is, CSV files where commas are replaced by tab characters.
```
yarikoptic, line 435: this reminded me a (possibly an "incorrect") anecdote about two color blind folks.  To add to the insult -- I do not remember it truly ;)  But here it goes for the same meaning:

One color blind person asks another :
- have you seen my new green car?
- what kind of color is that "green"?
- have you seen the red sunset?
- yes
- so it is like that color but just "green".

We never talk about CSV files in the specification and there is an assumption that users know what those are, and we do not even have a URL.  Let's may be at least add a link if to keep it at all?

```suggestion
(with [extension `.tsv`](glossary.md#tsv-extensions)),
that is, [CSV files](https://en.wikipedia.org/wiki/Comma-separated_values) where commas are replaced by tab characters.
```

src/common-principles.md
```diff
@@ -430,36 +430,54 @@ NIfTI header.
...
-are occasionally formatted using space characters instead of tabs to improve
-human readability.
-Directly copying and then pasting these examples from the specification
-for use in new BIDS datasets can lead to errors and is discouraged.
+!!! warning "Attention"
+
+    The TSV examples in this document (like the one above this note) are occasionally
+    formatted using space characters instead of tabs to improve human readability.
+    Directly copying and then pasting these examples from the specification
+    for use in new BIDS datasets can lead to errors and is discouraged.
```
yarikoptic, line 480: such note would be missed for most users who end up at any given TSV example. Can we take advantage of our MACROs to make it so hover over each such decorated TSV brings such a warning?

src/common-principles.md
```diff
@@ -536,12 +554,38 @@ like in the example below.
...
+            }
         }
     }
 }
 ```
 
+### Compressed tabular files
+
+Large tabular information, such as physiological recordings, MUST be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
```
yarikoptic, line 566: I ack that TSVGZ is already in our spec. But it is odd to me to not be TSV.GZ . FWIW it was 517e2ffb2851a342930a0a866c1813aabdeb1dbf which was a part of the huge monumental effort of 
- https://github.com/bids-standard/bids-specification/pull/1107
but I wonder may be @effigies you remember why you did not go for  `tsv.gz` ?

src/common-principles.md
oesteban: 👍 (I assume such a macro would have its own separate PR, where the PDF alternative is also covered)

src/common-principles.md
oesteban: Not sure what you mean, can you suggest how this would be addressed in your opinion?

yarikoptic:
```suggestion
[compressed tab-delineated (TSV.GZ) files](glossary.md#tsvgz-extensions) when
```
and
- https://github.com/bids-standard/bids-specification/pull/1755

src/common-principles.md
yarikoptic: not sure if there is any "tooltip" functionality in PDF though. Anyways -- this PR just changes already existing text. So I think we better make a separate issue for this, or better two ;) (I like 2nd one better).

- https://github.com/bids-standard/bids-specification/issues/1756
- https://github.com/bids-standard/bids-specification/issues/1757

yarikoptic: but I will resolve this conversation here since not to be addressed in this PR

oesteban:
Considering that the minimum period for discussion has been safely satisfied, the very incremental nature of the proposal, the absence of major side effects on the specs, and the possibility of ironing around yet-to-be-identified wrinkles through the PR mechanism anytime in the future, I think I can go ahead and merge.

There are a few aspects to follow up from here:
- #1750 following upon this PR improving `_stim.tsv.gz` specs
- #1755 normalizing extensions as terms in the glossary
- #1128 BEP020, where we are looking at physio specs with fresh eyes and trying to solidify BIDS foundations and consistency.

Issue #1748, opened 2024-03-24
[BUG] ``<matches>`` is sometimes used in templates but description in the glossary is missing


oesteban:
### Describe your problem in detail.

Some parts of the spec employ the placeholder `<matches>` to refer to the entities that characterize a run or relevant subset of the dataset.

It feels like a reasonable way of making the spec clearer. I think the spec just misses that it was defined in the glossary.

### Describe what you expected.

The definition of `<matches>` in the glossary, instead of every time it is used in a subsection of the spec.

### BIDS specification section

Example of use:
https://bids-specification.readthedocs.io/en/stable/modality-specific-files/task-events.html

Remi-Gau:
Along the same line, in you opinion, do you think it would make sense to also have the "source_entities" mentioned in the derivatives in the glossary?

https://bids-specification.readthedocs.io/en/latest/derivatives/introduction.html#file-naming-conventions

oesteban:
Totally

PR #1750, opened 2024-03-25
FIX: Move ``_stim.<ext>`` specification within the Task Events module

oesteban:
This fix builds on #1749 to separate ``_stim`` and ``_physio`` in the specs.

Currently ``_stim`` files are specified sideways as an addon or extra branch of ``_physio``, which can be confusing and misrepresent their intent.

Moving them with the other stimuli definitions increases the consistency of the spec and the findability of ``_stim`` specifications.

This fix requires #1749 for a more consistent prescription of `tsv.gz` files.

### DIFF
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -46,9 +46,9 @@ the eyetracking data in a certain sampling frequency, and
 `sub-01_task-bart_recording-breathing_physio.tsv.gz` to contain respiratory
 measurements in a different sampling frequency.
 
-Physiological recordings (including eyetracking) SHOULD use the `_physio` suffix.
+Physiological recordings (including eyetracking) MUST use the `_physio` suffix.
 
-The following table specifies metadata fields for the `*_physio.json` file.
+The following tables specifiy metadata fields for the `*_physio.json` file.
 
 <!-- This block generates a metadata table.
 These tables are defined in
@@ -169,7 +169,7 @@ stored in separate files
 entity MAY be used to distinguish these files).
 
 For motion parameters acquired from MRI scanner side motion correction, the
-`_physio` suffix SHOULD be used.
+`_physio` suffix MUST be used.
 
 For multi-echo data, a given `physio.tsv` file is applicable to all echos of
 a particular run.
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -46,7 +46,7 @@ the eyetracking data in a certain sampling frequency, and
 `sub-01_task-bart_recording-breathing_physio.tsv.gz` to contain respiratory
 measurements in a different sampling frequency.
 
-Physiological recordings (including eyetracking) SHOULD use the `_physio`.
+Physiological recordings (including eyetracking) SHOULD use the `_physio` suffix.
 
 The following table specifies metadata fields for the `*_physio.json` file.
 
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -48,7 +48,7 @@ measurements in a different sampling frequency.
 
 Physiological recordings (including eyetracking) MUST use the `_physio` suffix.
 
-The following tables specifiy metadata fields for the `*_physio.json` file.
+The following tables specify metadata fields for the `*_physio.json` file.
 
 <!-- This block generates a metadata table.
 These tables are defined in
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -2,7 +2,7 @@
 
 Physiological recordings such as cardiac and respiratory signals MAY be
 specified using a [compressed tabular file](../common-principles.md#compressed-tabular-files)
-([TSVGZ file](../glossary.md#tsvgz-extensions)) and a corresponding
+([TSV.GZ file](../glossary.md#tsvgz-extensions)) and a corresponding
 JSON file for storing metadata fields (see below).
 
 !!! example "Example datasets"
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -34,11 +34,10 @@ For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 
 !!! warning "Caution"
 
-    TSVGZ files MUST NOT include a header line, as established by the
+    `<matches>_physio.tsv.gz` files MUST NOT include a header line, as established by the
     [common-principles](../common-principles.md#compressed-tabular-files).
-
     As a result, when supplying a `<matches>_physio.tsv.gz` file, an accompanying
-    `<matches>_physio.json` MUST be supplied as well.
+    `<matches>_physio.json` MUST be present to indicate the column names.
 
 The [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish between several recording files.
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -32,10 +32,11 @@ before the suffix.
 For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 `<matches>` would correspond to `sub-control01_task-nback_run-1`.
 
-!!! note "TSVGZ headers are specified in metadata files."
+!!! warning "Caution"
 
-    TSVGZ files MUST NOT include a header line,
-    as established by the [common-principles](../common-principles.md#compressed-tabular-files).
+    TSVGZ files MUST NOT include a header line, as established by the
+    [common-principles](../common-principles.md#compressed-tabular-files).
+
     As a result, when supplying a `<matches>_physio.tsv.gz` file, an accompanying
     `<matches>_physio.json` MUST be supplied as well.
 
--- a/src/modality-specific-files/physiological-recordings.md
+++ b/src/modality-specific-files/physiological-recordings.md
@@ -1,9 +1,6 @@
-# Physiological and other continuous recordings
+# Physiological recordings
 
-## Physiological recordings
-
-Physiological recordings such as cardiac and respiratory signals and other
-continuous measures (such as parameters of a film or audio stimuli) MAY be
+Physiological recordings such as cardiac and respiratory signals MAY be
 specified using a [compressed tabular file](../common-principles.md#compressed-tabular-files)
 ([TSVGZ file](../glossary.md#tsvgz-extensions)) and a corresponding
 JSON file for storing metadata fields (see below).
@@ -24,8 +21,6 @@ sub-<label>/[ses-<label>/]
     <datatype>/
         <matches>[_recording-<label>]_physio.tsv.gz
         <matches>[_recording-<label>]_physio.json
-        <matches>[_recording-<label>]_stim.tsv.gz
-        <matches>[_recording-<label>]_stim.json
 ```
 
 For the template directory name, `<datatype>` can correspond to any data
@@ -41,8 +36,8 @@ For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 
     TSVGZ files MUST NOT include a header line,
     as established by the [common-principles](../common-principles.md#compressed-tabular-files).
-    As a result, when supplying a `*_<physio|stim>.tsv.gz` file, an accompanying
-    `*_<physio|stim>.json` MUST be supplied as well.
+    As a result, when supplying a `<matches>_physio.tsv.gz` file, an accompanying
+    `<matches>_physio.json` MUST be supplied as well.
 
 The [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish between several recording files.
@@ -51,10 +46,9 @@ the eyetracking data in a certain sampling frequency, and
 `sub-01_task-bart_recording-breathing_physio.tsv.gz` to contain respiratory
 measurements in a different sampling frequency.
 
-Physiological recordings (including eyetracking) SHOULD use the `_physio`
-suffix, and signals related to the stimulus SHOULD use `_stim` suffix.
+Physiological recordings (including eyetracking) SHOULD use the `_physio`.
 
-The following tables specify metadata fields for the `*_<physio|stim>.json` file.
+The following table specifies metadata fields for the `*_physio.json` file.
 
 <!-- This block generates a metadata table.
 These tables are defined in
@@ -173,11 +167,6 @@ Recordings with different sampling frequencies or starting times should be
 stored in separate files
 (and the [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish these files).
-
-If the same continuous recording has been used for all subjects (for example in
-the case where they all watched the same movie), one file MAY be used and
-placed in the root directory.
-For example, `task-movie_stim.tsv.gz`
 
 For motion parameters acquired from MRI scanner side motion correction, the
 `_physio` suffix SHOULD be used.
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -337,8 +337,7 @@ sub-<label>/[ses-<label>/]
 ```
 
 For the template directory name, `<datatype>` can correspond to any data
-recording modality, for example `func`, `anat`, `dwi`, `meg`, `eeg`, `ieeg`,
-or `beh`.
+recording modality.
 
 In the template filenames, the `<matches>` part corresponds to task filename
 before the suffix.
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -346,10 +346,11 @@ For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 
 !!! warning "Caution"
 
-    TSVGZ files SHOULD NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))
-
+    `<matches>[_recording-<label>]_stim.tsv.gz` files MUST NOT include a header line,
+    as established by the [common-principles](../common-principles.md#compressed-tabular-files).
     As a result, when supplying a `<matches>[_recording-<label>]_stim.tsv.gz` file,
-    an accompanying `<matches>[_recording-<label>]_stim.json` MUST be supplied as well.
+    an accompanying `<matches>[_recording-<label>]_stim.json` MUST be present to indicate
+    the column names.
 
 The [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish between several recording files,
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -124,7 +124,7 @@ Note that in the example above:
 
 1.  The `channel` column contains a list of values that are separated
     by a delimiter (`|`), as is declared in the `Delimiter` metadata
-    field of the `events.json file.
+    field of the `events.json` file.
     Thus, the channels related to the event in the third row of the example
     are called `F,1`, `F,2`, and `Cz`.
 
--- a/src/modality-specific-files/task-events.md
+++ b/src/modality-specific-files/task-events.md
@@ -309,3 +309,93 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
     "VisionCorrection": "lenses"
 }
 ```
+
+### Continuously-sampled, stimulus-related signals
+
+!!! example "Example datasets"
+
+      At the time of writing, there are no
+      [example datasets](https://bids-standard.github.io/bids-examples/#dataset-index)
+      showcasing stimulus-related signals.
+      However, the following datasets shared on [OpenNeuro](https://openneuro.org)
+      include these type of signals and may be used (with caution, because they are not
+      "official" examples) as a reference when curating a new dataset:
+
+      -   [*Trial timing for multivariate pattern
+          analysis*](https://openneuro.org/datasets/ds000238/versions/00002)
+
+Signals related to stimuli (such as parameters of a film or audio stimuli) that are
+evenly recorded at a constant sampling frequency SHOULD be stored as follows.
+
+Template:
+
+```Text
+sub-<label>/[ses-<label>/]
+    <datatype>/
+        <matches>[_recording-<label>]_stim.tsv.gz
+        <matches>[_recording-<label>]_stim.json
+```
+
+For the template directory name, `<datatype>` can correspond to any data
+recording modality, for example `func`, `anat`, `dwi`, `meg`, `eeg`, `ieeg`,
+or `beh`.
+
+In the template filenames, the `<matches>` part corresponds to task filename
+before the suffix.
+For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
+`<matches>` would correspond to `sub-control01_task-nback_run-1`.
+
+!!! warning "TSVGZ files SHOULD NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))"
+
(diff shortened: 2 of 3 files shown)


src/common-principles.md
```diff
@@ -536,12 +547,40 @@ like in the example below.
...
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files MUST be compressed with
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files MUST NOT contain a header in the first row
+    indicating the column names.
+1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
+    corresponding tabular file but with a `.json` extension.
+
+???+ warning "Columns of TSVGZ files MUST be defined in the corresponding JSON sidecar and the tabular content MUST NOT include a header line."
```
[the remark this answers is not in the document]
oesteban: Perhaps parsing the html would be easier?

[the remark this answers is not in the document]
Remi-Gau: nope because we need to be able to turn this into a pdf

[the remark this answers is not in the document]
oesteban: Are we using pandoc?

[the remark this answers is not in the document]
Remi-Gau: yeah but also for other contributors I feel that maybe not mixing markdwn + macros + html may be a good thing.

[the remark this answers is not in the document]
oesteban: https://github.com/chdemko/pandoc-latex-admonition

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,93 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+```Text
+sub-<label>/[ses-<label>/]
+    <datatype>/
+        <matches>[_recording-<label>]_stim.tsv.gz
+        <matches>[_recording-<label>]_stim.json
+```
+
+For the template directory name, `<datatype>` can correspond to any data
+recording modality, for example `func`, `anat`, `dwi`, `meg`, `eeg`, `ieeg`,
+or `beh`.
```
Remi-Gau, line 341:
```suggestion
For the template directory name, `<datatype>` can correspond to any data
recording modality.
```

Tempted to remove those because it is out of data (motion, fnirs are missing) and I don't think we want to keep updating that sentence for every new raw datatype.

sappelhoff:
> This fix builds on https://github.com/bids-standard/bids-specification/issues/1748 to separate _stim and _physio in the specs.

did you mean to reference:

- https://github.com/bids-standard/bids-specification/pull/1749 instead?

sappelhoff (commented):
> Currently _stim files are specified sideways as an addon or extra branch of _physio, which can be confusing and misrepresent their intent.
>
> Moving them with the other stimuli definitions increases the consistency of the spec and the findability of _stim specifications.

agreed!

Let's fix up #1749 first and then get this one ready!

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,92 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+
+In the template filenames, the `<matches>` part corresponds to task filename
+before the suffix.
+For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
+`<matches>` would correspond to `sub-control01_task-nback_run-1`.
+
+!!! warning "TSVGZ files SHOULD NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))"
+
+    As a result, when supplying a `<matches>[_recording-<label>]_stim.tsv.gz` file,
+    an accompanying `<matches>[_recording-<label>]_stim.json` MUST be supplied as well.
```
oesteban, line 350:
```suggestion
!!! warning "Caution"

    TSVGZ files SHOULD NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))
    
    As a result, when supplying a `<matches>[_recording-<label>]_stim.tsv.gz` file,
    an accompanying `<matches>[_recording-<label>]_stim.json` MUST be supplied as well.
```

src/modality-specific-files/physiological-recordings.md
```diff
@@ -38,8 +32,10 @@ before the suffix.
 For example for the file `sub-control01_task-nback_run-1_bold.nii.gz`,
 `<matches>` would correspond to `sub-control01_task-nback_run-1`.
 
-Note that when supplying a `*_<physio|stim>.tsv.gz` file, an accompanying
-`*_<physio|stim>.json` MUST be supplied as well.
+!!! warning "TSVGZ files MUST NOT include a header line (as established by the [common-principles](../common-principles.md#compressed-tabular-files))"
+
+    As a result, when supplying a `*_physio.tsv.gz` file, an accompanying
+    `*_physio.json` MUST be supplied as well.
```
oesteban, line 38:
```suggestion
!!! warning "Caution"

    TSVGZ files MUST NOT include a header line (as established by the
    [common-principles](../common-principles.md#compressed-tabular-files))

    As a result, when supplying a `*_physio.tsv.gz` file, an accompanying
    `*_physio.json` MUST be supplied as well.
```

oesteban:
Since #1749 is looking very close to the finish line, the following link https://github.com/oesteban/bids-specification/compare/fix/large-tabular-files...oesteban:bids-specification:fix/stim-files-with-stimuli can be used to look at the changes specific of this PR without distractions.

src/modality-specific-files/physiological-recordings.md
```diff
@@ -48,10 +46,9 @@ the eyetracking data in a certain sampling frequency, and
 `sub-01_task-bart_recording-breathing_physio.tsv.gz` to contain respiratory
 measurements in a different sampling frequency.
 
-Physiological recordings (including eyetracking) SHOULD use the `_physio`
-suffix, and signals related to the stimulus SHOULD use `_stim` suffix.
+Physiological recordings (including eyetracking) SHOULD use the `_physio`.
```
effigies, line 49:
```suggestion
Physiological recordings (including eyetracking) SHOULD use the `_physio` suffix.
```

src/modality-specific-files/physiological-recordings.md
```diff
@@ -168,11 +156,6 @@ stored in separate files
 (and the [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish these files).
 
-If the same continuous recording has been used for all subjects (for example in
-the case where they all watched the same movie), one file MAY be used and
-placed in the root directory.
-For example, `task-movie_stim.tsv.gz`
-
```
effigies, line 175: Should we note something like

> For simultaneous recordings, such as fMRI and EEG,
> physiological recordings MAY be placed as a single file in the subject or session directory.

Or should we leave well enough alone?

effigies (approved):
LGTM. Small comments.

src/modality-specific-files/physiological-recordings.md
oesteban: A note on that makes sense, but I think it deserves opening a new issue.

While it is not at all weird to set on your physiological recording gear at the beginning of the session and switch it off only at the end, cropping and splitting these files is usually a difficult task. I'm afraid that stuffing a single file could turn the recordings useless if the end-user does not have sufficient information to cut them at the right places (e.g., they come without triggers or flags). This is made worse by the fact that BIDS is quite sparse in explicit metadata recommendations to ensure the splitting of these files is feasible.

cc/ @smoia who will be extremely interested in this.

oesteban: That said, if the recordings would  be labeled with a column "experiment time" given by the scanner and in sync with the `scans.tsv` and start of each run, this option would definitely be the best option (with some caveats such as dummy scans if they are not properly marked).

smoia: I have the feeling that @effigies and @oesteban you're raising awareness on two different points - I'll comment both what I interpreted @effigies to say and what @oesteban I think you're referring to.

@effigies if by simultaneous recordings you mean cases like concurrent EEG/fMRI setups or EEG/motion cameras, that would normally create two different set of data having their own matching physio files, then I guess your issue is not storing the physio files twice when it's clearly all coming from one run.
IMHO this comes from (not being able to) defining "dependencies" or hierarchy between different files/modalities, and it inversely relates to another issue we're facing in physiopy, that is where to store "redundant" physio data that are collected before/after a "matching neuro run" (currently phys2bids approaches this as leaving it in the first run, but we're thinking about splitting them and moving them to a `beh` folder without a clear "matching dependency" for lack of currently better options).
Going back to your point, how would you declare that those physio files can be used in either modality? Would it be better to have a subject or session folder with 3 to 10 extra files, or to consider instead what equipment "created" those files (e.g. MRI vs external physio equipment vs EEG equipment)? Alternatively, could we consider having a `physio` folder for these types of cases?

@oesteban I think you're referring more to the multiple-neuro-runs-in-a-physio-recording issue, where we normally leave our physio equipment recording for the longest time and we then have to find a solution to split it to match it with neuro data. In that case, besides the fact that it's not compatible with the current specification, I agree with you that dumping a the single longest timeseries ever might confuse users (although that is the best option to allow extra recording for lag-related analyses).

oesteban: I think you are right; I misinterpreted @effigies' message.

oesteban: And I made another "mistake" in that I thought about physio, when here we are talking about `_stim.tsv.gz`. That also pushed your comment away.

- For `_physio` this is worth opening an issue (cc @Remi-Gau) because whatever we do (`beh/`, or `physio/`, or single file at session level) requires some thought.
- For `_stim` this is relatively easier because it feels this should go into `beh/` if a `stim` file responds to fMRI and EEG recorded simultaneously.

WDYT?

smoia: (@isesteves @me-pic @mmiedema if you want to follow the convo)

oesteban: Please redirect the physio part of this problem to #1753. I'll fix this in the mean time for `stim` and point at `beh`.

In the case of `stim`, you can definitely run the same movie (please read stimulus) with fMRI and EEG separately, so this is actually more likely to happen in standard experiments.

src/common-principles.md
```diff
@@ -536,12 +554,40 @@ like in the example below.
...
         }
     }
 }
 ```
 
+### Compressed tabular files
+
+Large tabular information, such as physiological recordings, MUST be stored with
+[compressed tab-delineated (TSVGZ) files](glossary.md#tsvgz-extensions) when
+so established by the specifications.
```
tsalo, line 567:
```suggestion
so established by the specification.
```

src/common-principles.md
```diff
@@ -536,12 +554,40 @@ like in the example below.
...
+Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
+
+1.  The contents of TSVGZ files MUST be compressed with
+    [gzip](https://datatracker.ietf.org/doc/html/rfc1952).
+1.  Compressed tabular files MUST NOT contain a header in the first row
+    indicating the column names.
+1.  TSVGZ files MUST be accompanied by a JSON file with the same name as their
+    corresponding tabular file but with a `.json` extension.
```
tsalo, line 575: This seems to ignore the inheritance principle. Are you saying that the inheritance principle doesn't apply to `.tsv.gz` files' metadata?

tsalo: It seems like the admonition's text about column names in the JSON file should go here.

tsalo (commented):
I just have one small concern/question about inheritance with `.tsv.gz` files.

src/common-principles.md
oesteban: > This seems to ignore the inheritance principle. Are you saying that the inheritance principle doesn't apply to `.tsv.gz` files' metadata?

Can you suggest an alternative phrasing that makes it clear that these files also follow the inheritance principle?

tsalo: Perhaps

```suggestion
1.  TSVGZ files MUST have an associated JSON file that defines the columns in the tabular file.
```

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,95 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+entity MAY be used to distinguish between several recording files,
+as prescribed by the specifications for
+[physiological recordings](physiological-recordings.md).
+
+If the same continuous recording has been used for all subjects (for example in
+the case where they all watched the same movie), one file placed in the
+root directory (for example, `<root>/task-movie_stim.<tsv.gz|json>`) MAY be used
+and will apply to all `<matches>_task-movie_<matches>_<suffix>.<ext>` files.
+In the following example, the two `task-nback_stim.<json|tsv.gz>` apply
+to all the `task-nback` runs across the two available subjects:
```
oesteban, line 365: @effigies - I think your comment (https://github.com/bids-standard/bids-specification/pull/1750/files#r1541231904) can be transferred here.

The question would be whether this (already existing before this PR) text for stim files covers what you are indicating above (https://github.com/bids-standard/bids-specification/pull/1750/files#r1541231904).

The other facet of this problem (involving physio files) can be addressed then separately within #1753).

I'm resolving the old conversation to discuss anything relating stim under this comment or relating to physio with #1753.

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,95 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+    `<matches>[_recording-<label>]_stim.tsv.gz` files MUST NOT include a header line,
+    as established by the [common-principles](../common-principles.md#compressed-tabular-files).
+    As a result, when supplying a `<matches>[_recording-<label>]_stim.tsv.gz` file,
+    an accompanying `<matches>[_recording-<label>]_stim.json` MUST be present to indicate
+    the column names.
+
+The [`recording-<label>`](../appendices/entities.md#recording)
+entity MAY be used to distinguish between several recording files,
+as prescribed by the specifications for
+[physiological recordings](physiological-recordings.md).
```
oesteban, line 358: This is transferred directly from the original location in physio, but I'm not sure this applies anymore. I'm in favor of removing specific mentions to `[_recording-<label>]`.

```suggestion
The [`recording-<label>`](../appendices/entities.md#recording)
entity MAY be used to distinguish between several recording files,
as prescribed by the specifications for
[physiological recordings](physiological-recordings.md).
```

Remi-Gau:
I introduced a merged conflict. Will fix.

oesteban:
> I introduced a merged conflict. Will fix.

I can take care, will be faster

src/modality-specific-files/physiological-recordings.md
```diff
@@ -51,10 +46,9 @@ the eyetracking data in a certain sampling frequency, and
 `sub-01_task-bart_recording-breathing_physio.tsv.gz` to contain respiratory
 measurements in a different sampling frequency.
 
-Physiological recordings (including eyetracking) SHOULD use the `_physio`
-suffix, and signals related to the stimulus SHOULD use `_stim` suffix.
+Physiological recordings (including eyetracking) SHOULD use the `_physio` suffix.
```
Remi-Gau, line 49: Shouldn't that be a MUST and not a SHOULD?

src/modality-specific-files/physiological-recordings.md
```diff
@@ -51,10 +46,9 @@ the eyetracking data in a certain sampling frequency, and
...
 
-The following tables specify metadata fields for the `*_<physio|stim>.json` file.
+The following table specifies metadata fields for the `*_physio.json` file.
```
Remi-Gau, line 51: Actually because we've just split things into several tables, I think we should keep the `tables specify`, no?

Sorry about this.

src/modality-specific-files/physiological-recordings.md
```diff
@@ -174,11 +168,6 @@ stored in separate files
 (and the [`recording-<label>`](../appendices/entities.md#recording)
 entity MAY be used to distinguish these files).
 
-If the same continuous recording has been used for all subjects (for example in
-the case where they all watched the same movie), one file MAY be used and
-placed in the root directory.
-For example, `task-movie_stim.tsv.gz`
-
 For motion parameters acquired from MRI scanner side motion correction, the
 `_physio` suffix SHOULD be used.
```
Remi-Gau, line 172: unrelated but this `SHOULD` should be a `MUST` too, no?

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,97 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+
+!!! example "Example datasets"
+
+      At the time of writing, there are no [example datasets][bids-examples] showcasing
+      stimulus-related signals.
+      However, the following datasets shared on [OpenNeuro](https://openneuro.org)
+      include these type of signals and may be used (with caution, because they are not
+      "official" examples) as a reference when curating a new dataset:
+
+      -   [*Trial timing for multivariate pattern analysis*][ds000238]
```
Remi-Gau, line 323: Thanks for looking this up: I can try to open a PR to the bids example by adapting this dataset to make sure we got an official example.

oesteban: That'd be lovely

src/modality-specific-files/task-events.md
Remi-Gau: I am reading this as a "soft" removing of the recording label for stim files. Let me know if this is not the correct interpretation of your comment.

I agree that this recording label now reads strange out of the physio context.

However we've always had it and it's been part of the schema for stim files for a long time now, so I think it would be breaking backward compatibility to remove it.

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,97 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+
+Signals related to stimuli (such as parameters of a film or audio stimuli) that are
+evenly recorded at a constant sampling frequency SHOULD be stored as follows.
```
Remi-Gau, line 326:
```suggestion
evenly recorded at a constant sampling frequency MUST be stored as follows.
```

Another one?

src/modality-specific-files/task-events.md
Remi-Gau: There is remark failure for the markdown formatting: seems that the admonition are also messing up putting "markdown links" at the bottom of the page when using an admonition.

Not 100% sure about this though.

src/modality-specific-files/physiological-recordings.md
Remi-Gau, line 49:
```suggestion
Physiological recordings (including eyetracking) MUST use the `_physio` suffix.
```

src/modality-specific-files/physiological-recordings.md
Remi-Gau, line 51:
```suggestion
The following tables specifiy metadata fields for the `*_physio.json` file.
```

src/modality-specific-files/physiological-recordings.md
Remi-Gau, line 172:
```suggestion
`_physio` suffix MUST be used.
```

src/modality-specific-files/task-events.md
Remi-Gau: OK there is technically a bids example with stim files: the synthetic example.

https://github.com/bids-standard/bids-examples/blob/master/synthetic/sub-01/ses-01/func/sub-01_ses-01_task-nback_run-01_stim.tsv.gz

I still think it'd be nice for users to have an example that links also to a real dataset.

Thoughts? @effigies @oesteban @sappelhoff

src/modality-specific-files/task-events.md
oesteban: Yes, a soft removal of the recording label.

src/modality-specific-files/task-events.md
effigies: Sure.

Remi-Gau (approved):
Good with me

oesteban:
Perhaps moving them inside the admonition?

oesteban:
> Perhaps moving them inside the admonition?

That actually did the trick!

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,97 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+
+      The following [BIDS-Examples][] showcase stimulus-related signals:
+
+      -   ["synthetic" example dataset][synthetic-example].
+
+      The following examples include these type of signals and may be used
+      (with caution, because they are not "official" examples)
+      as a reference when curating a new dataset:
+
+      -   [*Trial timing for multivariate pattern analysis*][ds000238]
```
Remi-Gau, line 325: OK I just checked and it seems like the stim.tsv.gz of this dataset should be a beh.tsv

Also the error thrown by openneuro validation seems wrong: 
https://openneuro.org/datasets/ds000238/versions/00002

```
Location:

/sub-01/beh/sub-01_task-test_stim.tsv.gz
Reason:

You should define 'ArterialSpinLabelingType' for this file. 'ArterialSpinLabelingType' can be CASL, PCASL, PASL.
```

Maybe another dataset would be better.

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,97 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
+
+      <!-- Link Definitions -->
+      [ds000238]: https://openneuro.org/datasets/ds000238/versions/00002
+      [bids-examples]: https://bids-standard.github.io/bids-examples/#dataset-index
+      [synthetic-example]: https://github.com/bids-standard/bids-examples/tree/master/synthetic
```
Remi-Gau, line 330: At this stage, doesn't it make sense to just have the "typical" markdown link?

src/modality-specific-files/task-events.md
oesteban: Good catch. It doesn't make any sense to point at an example we know is a bad example.

src/modality-specific-files/task-events.md
```diff
@@ -309,3 +309,97 @@ in the accompanying JSON sidecar as follows (based on the example of the previou
...
 }
 ```
+
+### Continuously-sampled, stimulus-related signals
+
+!!! example "Example datasets"
+
+      The following [BIDS-Examples][] showcase stimulus-related signals:
+
+      -   ["synthetic" example dataset][synthetic-example].
```
oesteban, line 319:
```suggestion
      The following [BIDS-Examples](https://bids-standard.github.io/bids-examples/#dataset-index)
      showcase stimulus-related signals and may be used as a reference
      when curating a new dataset:

      -   ["synthetic" example dataset](https://github.com/bids-standard/bids-examples/tree/master/synthetic).
```

src/modality-specific-files/task-events.md
oesteban, line 325:
```suggestion
```

src/modality-specific-files/task-events.md
oesteban, line 330:
```suggestion
```

Remi-Gau:
I think this should be ready to merge after the 5 days period.

src/modality-specific-files/task-events.md
oesteban: Since this is only a "transference" PR, I'll let it be like it was and open a sibling issue to #1753. I've realized that stim files present an additional challenge - you definitely can run the same stimuli (hence single stim file) in two different experiments (e.g., fMRI and EEG) performed separately in different moments.

Remi-Gau:
> I agree with the changes here in principle, pending this review item:
> 
>     * https://github.com/bids-standard/bids-specification/pull/1750/files#r1548105139
> 
> 
> it'd be important that the reference to an example points at a valid example

BTW: did a quick check on openneuro via the datalad superdataset and I could not find a dataset with stim files except the one that @oesteban had mentioned.

sappelhoff (approved):

oesteban:
@Remi-Gau @sappelhoff @effigies - may I merge this? I'd like to rebase BEP020, which will benefit substantially from getting stim out of physio.

PR #1754, opened 2024-03-29
[ENH][SCHEMA] Allow .bval and .bvec files for pepolar fmaps

### LINKED ISSUE #1724: [DWI] Allow bval and bvec for PEPOLAR fieldmaps
### Your idea

Siemens scanners typically don't let you acquire just a b=0 image and will produce some b>0 images as well. For PEPOLAR epi fieldmaps this means that there is a relevant bval and bvec that are useful for determining which volumes should be used for distortion correction. 

Since there is some activity on DWI in #1723, I was hoping to get some input on this idea too. 

In our group, we keep the epi bval and bvec files so they can be used during processing, but list them in .bidsignore so the data is still bids compatible. This is not a great workaround.

Pinging @cookpa @pierre-nedelec @tsalo

mattcieslak:
Closes #1724


src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -1253,7 +1253,8 @@ The phase-encoding polarity (PEpolar) technique combines two or more Spin Echo
 EPI scans with different phase encoding directions to estimate the distortion
 map corresponding to the nonuniformities of the *B<sub>0</sub>* field.
 These `*_epi.nii[.gz]` - or `*_m0scan.nii[.gz]` for arterial spin labeling perfusion data - files can be 3D or 4D --
-in the latter case, all timepoints share the same scanning parameters.
+in the latter case, all timepoints share the same scanning parameters. Some 4D scans intended
+for correcting DWIs may have accompanying `*_epi.bval` and `*_epi.bvec` files.
```
tsalo, line 1257:
```suggestion
in the latter case, all timepoints share the same scanning parameters. 
Some 4D scans intended for correcting DWIs may have accompanying `*_epi.bval` and `*_epi.bvec` files.
```
Splitting lines by punctuation will minimize the diff.

tsalo:
I think the other approach I mentioned (adding `.bval` and `.bvec` to `pepolar` and override the extensions in `pepolar_m0scan`) might be easier to integrate in the schema.

tsalo:
Thanks! I think you just need to address the issues in the `remark` job (misaligned bars in the contributors table and trailing space from my suggestion).

src/schema/rules/files/deriv/preprocessed_data.yaml
```diff
@@ -104,6 +104,13 @@ fmap_pepolar_common:
     space: optional
     description: optional
 
+fmap_pepolar_m0scan:
+  $ref: rules.files.raw.fmap.pepolar_m0scan
+  entities:
+    $ref: rules.files.raw.fmap.pepolar_m0scan.entities
```
tsalo, line 110: I think schemacode might not be dereferencing `rules.files.raw.fmap.pepolar_m0scan` before `rules.files.deriv.preprocessed_data.fmap_pepolar_m0scan`, unfortunately.

```suggestion
    $ref: rules.files.raw.fmap.pepolar.entities
```

src/schema/rules/files/deriv/preprocessed_data.yaml
```diff
@@ -104,6 +104,13 @@ fmap_pepolar_common:
     space: optional
     description: optional
 
+fmap_pepolar_m0scan:
+  $ref: rules.files.raw.fmap.pepolar_m0scan
```
tsalo, line 108: @mattcieslak it looks like the problem is that the `dereference` function can't handle multiple levels of references, so since `rules.files.raw.fmap.pepolar_m0scan` has a `$ref` in it as well, it's not being expanded.

tsalo: @effigies that's a bug in `dereference`, right?

mattcieslak: looks like copy/paste solved it for now

tsalo: That's definitely good, but if it _is_ a bug I'll want to open an issue about it.

tsalo (approved):
The change looks good to me!

tsalo:
I'm seeing `.bval` and `.bvec` in the filename templates for field map case 4: https://bids-specification--1754.org.readthedocs.build/en/1754/modality-specific-files/magnetic-resonance-imaging-data.html#case-4-multiple-phase-encoded-directions-pepolar

Should it be showing up somewhere else?

effigies:
Ah, no. Sorry, it turned out I opened the wrong PR's render.

src/appendices/contributors.md
```diff
@@ -257,6 +257,7 @@ If you contributed to the BIDS ecosystem and your name is not listed, please add
 | Mateusz Pawlik                                       | 📖🐛🤔🚧👀                             |
 | Mathias Goncalves                                    | 💻🔧📢                                 |
 | Mathieu Boudreau                                     | 💬🤔📢                                 |
+| Matt Cieslak                                         | 📖                                     |
```
effigies, line 260: Please update https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors instead of directly modifying this.

src/schema/rules/files/deriv/preprocessed_data.yaml
effigies: Likely what happened was that references of objects with references in them do not get dereferenced enough times. So that probably is a bug.

effigies (commented):
Overall this LGTM. It would be good to update/duplicate these checks:

https://github.com/bids-standard/bids-specification/blob/bd08602760d2fa0fd61e8f579a2677ea9bda8588/src/schema/rules/checks/dwi.yaml#L4-L44

For example, you can update the suffix check to match multiple suffixes with:

```diff
-    - suffix == "dwi"
+    - intersects([suffix], ["dwi", "epi"])
```

And do we want to add a check for `epi.bval` there must be at least one `0` value?

tsalo:
>And do we want to add a check for epi.bval there must be at least one 0 value?

I think datasets can have non-zero, but low, b-values in these images. In QSIPrep there's a b0-threshold parameter that determines which volumes to treat as b=0.

effigies:
> I think datasets can have non-zero, but low, b-values in these images. In QSIPrep there's a b0-threshold parameter that determines which volumes to treat as b=0.

What are typical thresholds? 1, 1e-2, 1e2?

effigies:
Just looked, the default is 100. Maybe let's make the check that there is at least one volume with `b<100`. cc @oesteban @arokem for a sanity check.

effigies:
Yeah, this is just EPI that may have b-values, not DWI, so that doesn't seem like a problem here. 

@mattcieslak @tsalo Do you want to make a schema check? Or need guidance?

arokem (approved):
Overall LGTM! Regarding threshold, I am guessing the b=100 default implemented in qsiprep is based on some use-case that @mattcieslak encountered where the low b-value was larger than 50? [This](https://github.com/dipy/dipy/pull/1667#issuecomment-441060602) is how DIPY chose that number 😄

src/schema/rules/checks/fmap.yaml
```diff
@@ -49,3 +49,16 @@ MagnitudeFileWithTooManyDimensions:
     - nifti_header != null
   checks:
     - nifti_header.dim[0] == 3
+
+EPIWithBValsHasHighBVals:
+  issue:
+    code: EPI_WITH_BVALS_HAS_HIGH_BVALS
+    message: |
+      EPI field maps with b-values should not include high b-values (in essence, b-values > 100).
```
arokem, line 57:
```suggestion
      EPI field maps with b-values should not include high b-values (defined as b-values > 100).
```

arokem: Just a suggestion here. Also "high b-values" is sometimes understood to imply b > 1,000, where DTI fails. I wonder whether there is another phrase we could use here (sorry, can't come up with something good right now and short on time atm).

src/schema/rules/checks/fmap.yaml
```diff
@@ -49,3 +49,16 @@ MagnitudeFileWithTooManyDimensions:
...
+    level: error
+  selectors:
+    - datatype == "fmap"
+    - suffix == "epi"
+    - '"bval" in associations'
+  checks:
+    - max(associations.bval.rows[0]) > 100
```
tsalo, line 64: @effigies not sure if this works with the syntax.

effigies: Do we care about requiring `max() < 100` or `min() < 100`? I thought the idea was that we should be able to identify b0 volumes, not that they all must be.

src/schema/rules/checks/fmap.yaml
tsalo: I agree. @mattcieslak do you have ideas for wording?

src/schema/rules/checks/fmap.yaml
tsalo: @mattcieslak can correct me, but I thought that epi scans used for distortion correction would _all_ effectively be b0 volumes, unlike the DWI scans.

src/schema/rules/checks/fmap.yaml
effigies, line 57: If we prefer to just require that there be some small b-values than all small:

```suggestion
EPISmallBVals:
  issue:
    code: EPI_WITH_BVALS_NEEDS_SMALL_BVALS
    message: |
      EPI field maps with b-values should include small (b<100) values.
```

effigies: What about "should include only small b-values (defined as b-values <= 100)"?

src/schema/rules/checks/fmap.yaml
mattcieslak: @tsalo the issue is that some vendors require that a non b=0 image be collected so there are usually a couple b=0s and one or move b>0s. The reason the bvals are so useful here is so we can identify which volumes to use for sdc

tsalo: In that case @effigies has it right. I'll make the changes to look for at least one low bval instead.

src/schema/rules/checks/fmap.yaml
```diff
@@ -49,3 +49,16 @@ MagnitudeFileWithTooManyDimensions:
...
+    code: EPI_WITH_BVALS_NEEDS_SMALL_BVALS
+    message: |
+      EPI field maps with b-values should include small (b<100) values.
+    level: error
+  selectors:
+    - datatype == "fmap"
+    - suffix == "epi"
+    - '"bval" in associations'
+  checks:
+    - min(associations.bval.rows[0]) > 100
```
effigies, line 64: Note that checks fail if `false`, not `true`.

```suggestion
    - min(associations.bval.values) < 100
```

We also need to add some way to access the bvals to the context:

https://github.com/bids-standard/bids-specification/blob/01025da63297468a3801f8fe43f97a140f2f9551/src/schema/meta/context.yaml#L173-L185

I would recommend

```yaml
            values:
              description: 'B-values contained in .bval file'
              type: array
              items:
                type: number

```

effigies: And would you mind bumping the SCHEMA_VERSION minor? NBD if not.

tsalo: I totally forgot about that. Thanks!

effigies (approved):
LGTM! Since the spec text has not changed in well over a week, and we've been fiddling around in the schema, I think this should be okay to merge without additional delay. I will allow someone else to do the honors, to ensure I am not alone in this opinion.

tsalo:
I agree. The actual content hasn't changed since @arokem and I approved, so I think we can say that there are still three approvals.

PR #293, opened 2019-08-02
[INFRA+FIX] Use linkchecker (from a dedicated docker image) to check all URLs

yarikoptic:
Supersedes #79 

TODOs
- [x] make sure it works
- [ ] possibly change tag and transfer docker image under `bids-standard` organization on hub.docker.org
- [x] fix broken internal (anchored) URLs
- [x] add `--check-extern` option
- [x] fix broken external URLs
- [x] fixup what broke (thanks travis)

PR #1755, opened 2024-03-30
Make extensions identifiers contain . to separate extensions where multiple used

yarikoptic:
origin: 
- https://github.com/bids-standard/bids-specification/pull/1749#discussion_r1545497672

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -563,7 +563,7 @@ like in the example below.
 ### Compressed tabular files
 
 Large tabular information, such as physiological recordings, MUST be stored with
-[compressed tab-delineated (TSV.GZ) files](glossary.md#tsvgz-extensions) when
+[compressed tab-delineated (TSV.GZ) files](glossary.md#tsv_gz-extensions) when
 so established by the specifications.
 Rules for formatting plain-text tabular files apply to TSVGZ files with three exceptions:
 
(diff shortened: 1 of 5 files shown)


effigies:
Dots are separators in the schema. This seems like a bad idea.

effigies:
We could use an underscore instead.

yarikoptic:
good idea, done, taking out of draft.

effigies:
After https://github.com/bids-standard/bids-specification/pull/1749, we need to update the anchors in these links:

```console
❯ grep -rI 'tsvgz\|niigz' src/**/*.md
src/common-principles.md:[compressed tab-delineated (TSV.GZ) files](glossary.md#tsvgz-extensions) when
src/modality-specific-files/physiological-recordings.md:([TSV.GZ file](../glossary.md#tsvgz-extensions)) and a corresponding
src/modality-specific-files/task-events.md:([TSV.GZ file](../glossary.md#tsvgz-extensions)) and a corresponding
```

yarikoptic:
sorry , I am missing why this PR is "relevant" or why they were not spotted/adjusted in #1749 - don't linkchecker tests those as well?

effigies:
I don't think anchors fail to resolve.

yarikoptic:
> I don't think anchors fail to resolve.

[at least in the past I made sure they did](https://github.com/bids-standard/bids-specification/pull/293)... may be when we reverted to use upstream version of linkchecker we lost that... might want to make sure some time so we do not breed them

effigies (approved):

effigies:
Impact on rendered doc is essentially nil. Invoking admin privileges.

PR #1450, opened 2023-03-17
[ENH] Add explicit wording on DICOM terms correspondence

yarikoptic:
I was under impression that we had such wording somewhere but I failed to find. Such aspect came up during recent presentation by @rordenlab to SC,

<details>
<summary>and we already have lots of metadata fields formalized that way</summary> 

```shell
❯ git grep DICOM | grep -i -e based -e correspond
src/modality-specific-files/magnetic-resonance-imaging-data.md:Volume types are defined in the following table, based on DICOM Tag 0018, 9257 `ASL Context`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 9073 `Acquisition Duration`.
src/schema/objects/metadata.yaml:    Based on DICOM Tag 0018, 925F `ASL Bolus Cut-off Delay Time`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 925C `ASL Bolus Cut-off Flag`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 925E `ASL Bolus Cut-off Technique`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 1048 `Contrast/Bolus Ingredient`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0054, 1322 `Dose Calibration Factor`.
src/schema/objects/metadata.yaml:    For Siemens, this corresponds to DICOM field 0019, 1018 (in ns).
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 0081 `Echo Time`
src/schema/objects/metadata.yaml:    Corresponds to: DICOM Tag 0018, 1314 `Flip Angle`.
src/schema/objects/metadata.yaml:    This corresponds to DICOM Tag 0018, 1242 `Actual Frame Duration` converted
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 101A `Hardcopy Device Software Version`.
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 1074 `Radionuclide Total Dose`.
src/schema/objects/metadata.yaml:    This corresponds to DICOM Tag 0018, 1072 `Contrast/Bolus Start Time`
src/schema/objects/metadata.yaml:    Corresponds to DICOM Tag 0018, 0082 `Inversion Time`
src/schema/objects/metadata.yaml:    Based on DICOM macro C.8.13.5.14.
[...]

PR #1758, opened 2024-03-31
ENH: Use "Tag XXXX, YYYY" not "Tag (XXXX,YYYY)" for DICOM Tags

yarikoptic:
In general I like more `(XXXX,YYYY)` form but the other form more prevalent and thus what I also suggested in https://github.com/bids-standard/bids-specification/pull/1450/files

Attention to this was triggered by @CPernet 's comment in
- https://github.com/bids-standard/bids-specification/pull/1450#issuecomment-2028576595

src/schema/objects/metadata.yaml and src/schema/rules/checks/asl.yaml src/schema/rules/sidecars/mri.yaml all use form without `()` and pet sometimes only used with ()

### DIFF
--- a/src/schema/rules/sidecars/pet.yaml
+++ b/src/schema/rules/sidecars/pet.yaml
@@ -114,10 +114,10 @@ PETPharmaceuticals:
   fields:
     PharmaceuticalName:
       level: recommended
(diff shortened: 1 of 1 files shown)


Remi-Gau (approved):

Remi-Gau:
@tsalo @effigies forgot to mention this light PR to reveiw yesterday

sappelhoff (approved):

PR #1760, opened 2024-04-02
FIX: Markdown linter throws false positives if definitions are only used inside admonition blocks

oesteban:
What we see in the context of #1750 seems to respond to remarkjs/remark-lint#307.

This bugfix adds the dependency (as suggested in the issue) and adds one canary to ensure it is working.

cc/ @Remi-Gau


oesteban:
Converting to draft because this doesn't seem to work. And it seems the common-principles.md file is not linted?

Remi-Gau:
not sure if it helps but I had to tweak some remark linting rules when I introduced admonitions:
https://github.com/bids-standard/bids-specification/blob/dbcb237d14970bc75c4837a4c1b317aa5a267cc8/.remarkrc#L13

oesteban:
> not sure if it helps but I had to tweak some remark linting rules when I introduced admonitions:
> 
> https://github.com/bids-standard/bids-specification/blob/dbcb237d14970bc75c4837a4c1b317aa5a267cc8/.remarkrc#L13

Thanks for the suggestion. A naive addition of the plugin did not work (https://github.com/bids-standard/bids-specification/pull/1750/commits/cd9c926e6c52bd96c758e739d1044ee5912d966f)

oesteban:
Defining the links inside the admonition at hand does the trick. Closing as not needed.

PR #1769, opened 2024-04-11
[ENH] Add warning about deidentification when sharing sourcedata

### LINKED ISSUE #1744: Suggestion: add some admonition text warning that sharing under sourcedata/ should be done carefully
we want to include here some language that the specification is completely naive about whether those data can be shared. Along these lines, it would be a good place to mention that source files requiring anonymization can be stored here and still be considered "source data" (although there's been one processing step in between).

The point is, many students and young researchers may be unaware of the circumstances under which the data may or may not shared and not adding some sort of warning or language here to say: hey, discuss with your PI or your university or whoever this needs be discussed before dropping anything and using this as a means of releasing data public.

_Originally posted by @oesteban in https://github.com/bids-standard/bids-specification/pull/1737#discussion_r1534513241_

Remi-Gau:
- closes #1744 

Just to get the conversation started.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -309,10 +309,26 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 }
 ```
 
-!!! warning "non-anonymised data in `sourcedata`"
+!!! danger "Caution"
 
-    Source files requiring anonymization MAY be stored in this subfolder.
-    However, be aware that this could make it harder to share the dataset.
+    Sharing source data may help amend errors and missing data discovered
+    only with the reuse of the raw dataset in practice.
+    Therefore, from an Open Science perspective, it is RECOMMENDED to share
+    the source data whenever it is possible.
+    
+    However, more stringent sharing limitations may apply to
+    the source data than those applicable to the raw data.
+    For example, human data almost always requires anonymization before they
+    can be redistributed, or the subjects' consent form did not explicitly
+    stated that the source files would be shared after anonymization.
+    Further examples in which sharing source data may not be possible
+    include original data formats that are not redistributable as per the
+    acquisition device's license.
+    
+    As for raw data, all regulatory, ethical, and legal aspects SHOULD
+    be carefully considered before sharing data through the `sourcedata/` folder
+    mechanism.
+    In the case of source data, these aspects are likely more stringent.
 
 ### Storage of derived datasets
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -309,6 +309,11 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 }
 ```
 
+!!! warning "non-anonymised data in `sourcedata`"
+
+    Source files requiring anonymization MAY be stored in this subfolder.
+    However, be aware that this could make it harder to share the dataset.
+
 ### Storage of derived datasets
 
 Derivatives can be stored/distributed in two ways:
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -318,10 +318,10 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 
     However, more stringent sharing limitations may apply to the source data
     than those applicable to the raw data.
-    For example, human data almost always requires anonymization
+    For example, human data almost always requires deidentification
     before they can be redistributed,
     or the subjects' consent form did not explicitly state that the source files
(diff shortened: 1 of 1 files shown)


src/common-principles.md
```diff
@@ -309,6 +309,11 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 }
 ```
 
+!!! warning "non-anonymised data in `sourcedata`"
+
+    Source files requiring anonymization MAY be stored in this subfolder.
+    However, be aware that this could make it harder to share the dataset.
```
Remi-Gau, line 315: @oesteban realizing that I am not a 100% sure what you meant in #1744

oesteban, line 315: I mean something like this:

```suggestion
    Sharing source data may help amend errors and missing data discovered
    only with the reuse of the raw dataset in practice.
    Therefore, from an Open Science perspective, it is RECOMMENDED to share
    the source data whenever it is possible.
    
    However, more stringent sharing limitations may apply to
    the source data than those applicable to the raw data.
    For example, human data almost always requires anonymization before they
    can be redistributed, or the subjects' consent form did not explicitly
    stated that the source files would be shared after anonymization.
    Further examples in which sharing source data may not be possible
    include original data formats that are not redistributable as per the
    acquisition device's license.
    
    As for raw data, all regulatory, ethical, and legal aspects SHOULD
    be carefully considered before sharing data through the `sourcedata/` folder
    mechanism.
    In the case of source data, these aspects are likely more stringent.
```

oesteban: @mszinte - this explains my resistance in BEP 020 to say anything about `sourcedata/`. I hope it makes sense to you :)

oesteban (approved):

src/common-principles.md
```diff
@@ -309,6 +309,28 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
...
+    Sharing source data may help amend errors and missing data discovered
+    only with the reuse of the raw dataset in practice.
+    Therefore, from an Open Science perspective, it is RECOMMENDED to share
+    the source data whenever it is possible.
+
+    However, more stringent sharing limitations may apply to the source data
+    than those applicable to the raw data.
+    For example, human data almost always requires anonymization
+    before they can be redistributed,
+    or the subjects' consent form did not explicitly state that the source files
```
DimitriPapadopoulos, line 323: I would use `pseudonymization` or `deidentification` instead of `anonymization`, as `anonymization` in the sense used in the GDPR is not really possible, unless data lose most of their scientific interest.

DimitriPapadopoulos: See #1782 to apply to the whole repo.

Remi-Gau: good point

What is the technical difference between `pseudonymization` or `deidentification`?

oesteban: No objections to change. I think I prefer deidentification in this case.

DimitriPapadopoulos: I don't have access to a normative document such as the above ISO standards, but I see **deidentification** as the more general term which includes **anonymisation** and **pseudonymisation**.

**Pseudonymisation** merely substitutes directly identifying information. For example, names might be replaced by hashes and MRI images might be defaced. However, it remains possible to identify the person behind the personal data, often using additional information.

A quick [Google search](pseudonymisation vs anonymisation ISO standard) comes up  with[Anonymisation vs Pseudonymisation](https://hoggo.io/anonymisation-vs-pseudonymisation/) and other articles. It would be great if someone with access to ISO standards could show us the original wording and recommendations in these standards.

ericearl: I also favor "deidentification" as the term used.

sappelhoff: +1 for deidentification, and applying it to the whole spec, as suggested in 

- https://github.com/bids-standard/bids-specification/issues/1782

src/common-principles.md
```diff
@@ -309,6 +309,28 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
...
+    would be shared after anonymization.
```
effigies, line 324:
```suggestion
    For example, human data almost always requires deidentification
    before they can be redistributed,
    or the subjects' consent form did not explicitly state that the source files
    would be shared after deidentification.
```

effigies (approved):
Seems everyone is in agreement. Green check.

PR #1778, opened 2024-04-13
Make wording more explicit for _mod and remove MRI specificity of _acq

yarikoptic:
`_acq` is also used in other modalities so description should be generic.

`_mod-` change relates to a discussion of _mod destiny in

https://github.com/bids-standard/bids-2-devel/issues/70

and to accent that it is not really "modality" term (e.g. `mri`, `eeg`, ...) we use generally through out BIDS specification.

Some might argue about use of "contrast" to describe `T1w` etc but we do use "contrast" to that extent in other places already AFAIK.

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -8,16 +8,16 @@ acquisition:
   display_name: Acquisition
   description: |
     The `acq-<label>` entity corresponds to a custom label the user MAY use to distinguish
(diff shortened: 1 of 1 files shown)


yarikoptic:
I was wide off -- we did clarify modality definition in https://bids-specification.readthedocs.io/en/stable/common-principles.html#definitions and for better or for worse it was Ok and no need to bring in "contrast" here... I will adjust now

PR #1781, opened 2024-04-15
[ENH] Add `matches` and `source_entity` to glossary and file templates

### LINKED ISSUE #1748: [BUG] ``<matches>`` is sometimes used in templates but description in the glossary is missing

Remi-Gau:
- Examples of templates using `matches`:
  - [Events files](https://bids-specification--1781.org.readthedocs.build/en/1781/modality-specific-files/task-events.html#task-events)
  - [Physio files](https://bids-specification--1781.org.readthedocs.build/en/1781/modality-specific-files/physiological-recordings.html)
- New terms in glossary:
  - [`matches`](https://bids-specification--1781.org.readthedocs.build/en/1781/glossary.html#matches-metaentities)
  - [`source_entities`](https://bids-specification--1781.org.readthedocs.build/en/1781/glossary.html#source_entities-metaentities)

---

- [x] improve filenaming templates with those placeholders to link to glossary.
- [x] link to glossary when they occur in text
- [x] ensure that entities pages are not affected
- [x] replace existing hardcoded templates using `matches` with the macro
- [ ] ~~replace existing hardcoded templates using `matches` with the macro~~ -> deferred to #2070.
- [ ] ~~ensure glossary properly builds in PDF format~~ -> deferred to #2069

---
Closes #1748


src/schema/objects/entities.yaml
Remi-Gau, line 1: adding those in the entities.yml for now

not entirely convinced this is the best place

oesteban: perhaps just the glossary? in entities it will interact with the schema, right?

oesteban: okay, there's no "glossary.yml" hehehehe sorry for shooting before thinking.

oesteban:
bumping this - @Remi-Gau what's the status here?

oesteban:
@Remi-Gau indeed, entities was a bad place to include them because the validator would try an interpret them, right?

We have been using the common principles to integrate this kind of meta-metadata (e.g., suffix as a placeholder for a number of extensions).

However, matches and source_entities didn't feel right under common principles.

I still neet to address the schema validation issue, but what people feel about a new "metaentities" file where these placeholders, wildcards (I'm thinking of an `any` keyword that would match any valid label/data_type), etc?

cc/ @effigies

oesteban:
BTW, this is a great safety check for BEPs. I've updated the task events specs to use matches (instead of hard-coding the template) and https://bids-specification--1781.org.readthedocs.build/en/1781/modality-specific-files/task-events.html

You can see what datatypes allow events files and which ones do not.

tools/schemacode/src/bidsschematools/render/text.py
```diff
@@ -183,7 +178,9 @@ def make_glossary(schema, src_path=None):
             levels = [level["name"] if isinstance(level, dict) else level for level in levels]
             text += f"**Allowed values**: `{'`, `'.join(levels)}`\n\n"
 
-        text += f"**Description**:\n{obj_desc}\n\n"
+        # Convert description into markdown and append to text
+        obj_desc = MarkdownIt().render(f"**Description**:\n{obj_desc}")
```
oesteban, line 182: @Remi-Gau the pdf build doesn't like this

effigies:
image

Was:

image

I think you may want to look at how extensions are collapsed. We could do `<func|eeg|...>/` if it's few enough, or just `<datatype>/` when there are more than, say, 3 or 4.

src/schema/README.md
```diff
@@ -369,6 +369,7 @@ The namespaces are:
 | --------------------------- | ----------------------------------------------------------------------------------- | ---------------- |
 | `objects.common_principles` | Terms that are used throughout BIDS                                                 | General terms    |
 | `objects.modalities`        | Broad categories of data represented in BIDS, roughly matching recording instrument | General terms    |
+| `objects.metaentities`      | Placeholders and wildcards to reduce verbosity of some templates in BIDS            | General terms    |
```
effigies, line 372: I agree that these concepts do not fit elsewhere, but I'm not sure that `metaentity` is right. I'm +1 on meta, -1 on entity. Not sure what would be better just now.

oesteban: Happy to take any suggestions :)

oesteban:
> I think you may want to look at how extensions are collapsed. We could do `<func|eeg|...>/` if it's few enough, or just `<datatype>/` when there are more than, say, 3 or 4.

I'd go with this now, and leave this improvement for the future:

- I'm not very enthusiastic about just `<datatype>/` unless that means absolutely all the possible datatypes. That said, maybe we want to have one more of this "meta" entities that is called `any` as I suggest in the new yaml file:
  ```
  sub-<label>/
      [ses-<label>/]
          <any>/
              <matches>_...
  ```
- Otherwise, I'd stick with explicit rendering of the actual datatypes accepted for two reasons: (i) it is explicit and less prone to errors; and (ii) brings transparently potential issues such as events not being supported everywhere (which this would quickly show if finally they are accepted).

I actually considered the `<beh|eeg|...>` option, but the current implementation of the renderer is a bit prickly to do that and didn't have the time to write proper tests.

I'd support collapsing them in a future PR, but for now, I think this resolves the problem and comes with testing.

src/schema/README.md
oesteban: @effigies - what do you think? Some new name comes up?

oesteban:
I'm adding @tsalo as a reviewer. This new feature is not very heavy lifting but improves the readability a lot.

@Remi-Gau can you give it a glimpse again?

effigies:
@oesteban Will get back to this by the end of the week. Currently traveling. Please re-ping me on Friday if I haven't found time by then.

Remi-Gau:
> I'd support collapsing them in a future PR, but for now, I think this resolves the problem and comes with testing.

agree to have this in now and improve rendering in a separate PR

effigies:
I get these are entities, but I was thinking there might be other things that we want to summarize, and then adding them to `metaentities` would not be appropriate and we'd end up needing a new category that's basically the same. An example is if we want to formalize `.<ext>` to indicate a collection of extensions that's too large to render. Right now, that's just a rule in the rendering code.

We could do something like `objects.rendering` for concepts that are only useful in simplifying spec rendering.

That said, not a hill I'll die on. `metaentities` is good enough. Let me review the content now.

tools/schemacode/src/bidsschematools/render/text.py
oesteban: Description is still gone for the PDF. The problem is that other renderers are aware of whether the PDF or the Markdown versions are to be rendered. However, here we do not know that information.

If this is not Markdowned, then the HTML version doesn't render well the ` ``` ` blocks. If Markdowned, it's missing from the PDF.

Remi-Gau: fine with me: in any case the schema rendering code needs a big refactor IMO and with probably some jinja-ification of the whole thing

oesteban: It wasn't clear at all, I would've saved you time with just a few more words ;)

src/schema/README.md
effigies, line 1: Please add metaentities to the ordering rules section (~L1006).

effigies (commented):
I think we should now be updating these?

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/modality-specific-files/physiological-recordings.md?plain=1#L19-L24

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/imaging.md?plain=1#L10-L15

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/imaging.md?plain=1#L147-L152

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/imaging.md?plain=1#L264-L269

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/imaging.md?plain=1#L328-L333

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/imaging.md?plain=1#L402-L407

https://github.com/bids-standard/bids-specification/blob/0142056a79a292292697cf2bf086618ce12a9a51/src/derivatives/common-data-types.md?plain=1#L180-L185

oesteban:
@effigies I'm trying to use the macro for derivatives, but I'm afraid it is broken for derivatives (in general):

```
WARNING -  [macros] - ERROR # _Macro Rendering Error_

           _File_: `derivatives/imaging.md`

           _AttributeError_: 'Namespace' object has no attribute 'datatypes'

           ```
           Traceback (most recent call last):
             File "/data/home/oesteban/workspace/bids-specification/tools/schemacode/src/bidsschematools/types/namespace.py", line 220, in
           __getattribute__
               return self[key]
                      ~~~~^^^^^
             File "/data/home/oesteban/workspace/bids-specification/tools/schemacode/src/bidsschematools/types/namespace.py", line 236, in
           __getitem__
               val = mapping[subkey]
                     ~~~~~~~^^^^^^^^
           KeyError: 'datatypes'

           During handling of the above exception, another exception occurred:

           Traceback (most recent call last):
             File "/home/oesteban/.miniconda/envs/bids-spec/lib/python3.11/site-packages/mkdocs_macros/plugin.py", line 688, in render
               return md_template.render(**page_variables)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             File "/home/oesteban/.miniconda/envs/bids-spec/lib/python3.11/site-packages/jinja2/environment.py", line 1301, in render
               self.environment.handle_exception()
             File "/home/oesteban/.miniconda/envs/bids-spec/lib/python3.11/site-packages/jinja2/environment.py", line 936, in handle_exception
               raise rewrite_traceback_stack(source=source)
             File "<template>", line 8, in top-level template code
             File "/data/home/oesteban/workspace/bids-specification/tools/mkdocs_macros_bids/macros.py", line 94, in make_filename_template
               codeblock = render.make_filename_template(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             File "/data/home/oesteban/workspace/bids-specification/tools/schemacode/src/bidsschematools/render/text.py", line 323, in
           make_filename_template
               for datatype in rule.datatypes:
                               ^^^^^^^^^^^^^^
             File "/data/home/oesteban/workspace/bids-specification/tools/schemacode/src/bidsschematools/types/namespace.py", line 222, in
           __getattribute__
               raise err
             File "/data/home/oesteban/workspace/bids-specification/tools/schemacode/src/bidsschematools/types/namespace.py", line 214, in
           __getattribute__
               return super().__getattribute__(key)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           AttributeError: 'Namespace' object has no attribute 'datatypes'

           ```
```

(this is with placeholders=False)

Remi-Gau:
> @effigies I'm trying to use the macro for derivatives, but I'm afraid it is broken for derivatives (in general):

my hunch is that some things may not be defined for derivatives in the schema

oesteban:
@effigies @Remi-Gau I don't think this PR should fix everything, so I've opened #2070 for this second issue.

Side-effect is I cannot port into using the macro any of the derivatives templates Chris suggested.

Remi-Gau:
am ok with leaving things as they are in this PR and I can have a look at dealing with the issues it raised in follow up PRs

oesteban:
> am ok with leaving things as they are in this PR and I can have a look at dealing with the issues it raised in follow up PRs

Fantastic. In that case, this is ready for final review and merge.

effigies (approved):

Remi-Gau:
@oesteban 
I can't really approve my own PR.
Maybe we should ping someone else for another review.

oesteban:
@tsalo @yarikoptic @bendichter @erdalkaraca would one of you be able to have a look?

bendichter:
The update to the metadata looks good to me syntactically (but I don't know enough about the BIDS standard to have an opinion about whether to expand it in this way)

oesteban:
> The update to the metaschema looks good to me syntactically (but I don't know enough about the BIDS standard to have an opinion about whether to expand it in this way)

Thanks for having a look. I understand you don't feel ready to give a green tick to this PR.

@effigies @Remi-Gau I think I'll push this branch as mine and request another PR, WDYT?

Remi-Gau:
I'd say that with @bendichter +1 this is good enough to merge here: @effigies WDYT?

oesteban:
> I'd say that with @bendichter +1 this is good enough to merge here: @effigies WDYT?

I've opened #2082 just in case, happy with whatever is more convenient to everyone :)

PR #1784, opened 2024-04-15
[ENH] use full width entity page

Remi-Gau:
closes none

Tiny tweak that helps for all tables but that of MRI.

### Before

image

https://bids-specification.readthedocs.io/en/latest/appendices/entity-table.html#biopotential-amplification-eeg-and-ieeg

### After

image

https://bids-specification--1784.org.readthedocs.build/en/1784/appendices/entity-table.html

### DIFF
--- a/src/appendices/entity-table.md
+++ b/src/appendices/entity-table.md
@@ -1,7 +1,7 @@
 ---
 hide:
-  - navigation
-  - toc
+- navigation
+- toc
 ---
 
 # Entity table
--- a/src/appendices/entity-table.md
+++ b/src/appendices/entity-table.md
@@ -1,6 +1,5 @@
 ---
 hide:
--   navigation
 -   toc
 ---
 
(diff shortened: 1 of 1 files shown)


effigies:
Definitely an improvement, but losing navigation doesn't seem great. Maybe just hide the TOC?

Once https://github.com/squidfunk/mkdocs-material/issues/3428 exists, we could see about starting navigation in collapsed mode.

Remi-Gau:
> Definitely an improvement, but losing navigation doesn't seem great. Maybe just hide the TOC?

Let me try.

src/appendices/entity-table.md
```diff
@@ -1,3 +1,9 @@
+---
+hide:
+-   navigation
```
Remi-Gau, line 3:
```suggestion
```

effigies (approved):
I think this is still an improvement over the old, and it doesn't leave the user stranded, only able to get back to the index.

tsalo (approved):
I like it!

PR #1787, opened 2024-04-16
[INFRA] Introduce metaschema

bendichter:
This PR contains:

1. `metaschema.json`: A json schema document that specifies the entire schema language
2. `validate_schema.py`: code for loading the schema, dereferencing it, and validating against the metaschema
3. very minor adjustment to the actual schema, changing "deprecated" to "DEPRECATED", which appears to be the more dominant.
4. tests ensuring that this correctly identifies erroneous additions to the schema

When I started this at the meeting in Seattle I did not expect the metaschema to be >700 lines, but there it is. I used definitions a bit to make it more concise. It's possible this could benefit from more definitions.


src/metaschema.json
Remi-Gau, line 1: Maybe better to place it in https://github.com/bids-standard/bids-specification/tree/master/src/schema ?

src/validate_schema.py
Remi-Gau, line 1: Probably better to put into the rest of the code for the bids-schema, no?

https://github.com/bids-standard/bids-specification/tree/master/tools/schemacode/bidsschematools

src/metaschema.json
bendichter: this sounds good

src/validate_schema.py
bendichter: this sounds good too

src/metaschema.json
bendichter: I tried moving it there and then I realized that since the metaschema is validating the schema folder, that means if it is in that folder it is validating itself. That's possible, but it seems more natural to me to put it outside the very schema that it is meant to validate.

tools/schemacode/bidsschematools/validate_schema.py
```diff
@@ -0,0 +1,73 @@
...
+        str(schema_path).endswith(".yaml") or str(schema_path).endswith(".yml")
+    ):
+        with open(schema_path, "r") as f:
+            return yaml.safe_load(f)
+    else:
+        with open(schema_path, "r") as f:
+            return f.read()
+
+
+def derefence_schema(schema):
```
tsalo, line 26:
```suggestion
def dereference_schema(schema):
```

src/metaschema.json
Remi-Gau: ha yeah indeed!

bendichter:
depends on #1790

Remi-Gau (changes requested):
OK so if I read this correclty the schema gets validated in CI wiht the metaschema, so the latter won't get out of synch.

Do you think you could add a couple of line of doc about all this?

effigies:
Made a few small changes that were easier to do than suggest. Please have a look and let me know if you disagree.

tools/schemacode/bidsschematools/data/metaschema.json
```diff
@@ -0,0 +1 @@
+../../../../src/metaschema.json
```
effigies, line 1: This gets resolved by setuptools during `sdist` build, so it's safe to use a symlink here.

tools/schemacode/bidsschematools/validate_schema.py
```diff
@@ -0,0 +1,17 @@
+import json
+from importlib.resources import files
+
+from jsonschema import ValidationError, validate
+
+
+def validate_schema(schema):
```
effigies, line 7: I think it makes sense to include this in `bidsschematools.schema` rather than its own module.

src/schema/objects/common_principles.yaml
```diff
@@ -47,7 +47,7 @@ dataset:
     A set of neuroimaging and behavioral data acquired for a purpose of a particular study.
     A dataset consists of data acquired from one or more subjects, possibly from multiple sessions.
 deprecated:
-  display_name: DEPRECATED
+  display_name: deprecated
```
bendichter, line 50:
```suggestion
  display_name: DEPRECATED
```

tools/schemacode/bidsschematools/schema.py
```diff
@@ -274,3 +284,21 @@ def filter_schema(schema, **kwargs):
...
+    # print in the terminal
+    try:
+        validate(instance=schema.to_dict(), schema=metaschema)
+    except ValidationError as e:
+        with tempfile.NamedTemporaryFile(
+            prefix="schema_error_", suffix=".txt", delete=False, mode="w+"
+        ) as file:
+            file.write(str(e))
+            # ValidationError does not have an add_note method yet
+            # e.add_note(f"See {file.name} for full error log.")
```
effigies, line 303: Apparently this is a 3.11+ builtin: 

https://docs.python.org/3/library/exceptions.html#BaseException.add_note

It would be nice to display. Not sure if there's a way to hack this until 3.11 is our minimum.

effigies (approved):
LGTM.

Remi-Gau (approved):
Good with me. thanks @bendichter

effigies:
@bendichter Please add yourself to https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors if you aren't already on the contributors list.

Issue #1633, opened 2023-10-18
Formalize participants' age to actually allow for other than year units/durations 


yarikoptic:
ATM it must be in years (according to schema and hardcoded there assumptions), although [text](https://github.com/bids-standard/bids-specification/blob/0dcb141fccec5bbd92e482a071fefba0d0bbfdb1/src/modality-agnostic-files.md?plain=1#L263) suggests differently 

```
If no `units` is provided for age, it will be assumed to be in years relative to date of birth.
```

which was suggested/accepted in code review of #459 but 1. no alternative units are specified (thus unknown/useless) ; 2. such relaxation was not **later** reflected in the schema which now just [hardcode](https://github.com/bids-standard/bids-specification/blob/HEAD/src/schema/objects/columns.yaml#L35) via `unit: years` without alternative unit values provided. (not even sure if we formalize units to be machine readable ATM, e.g. see #1631 to fix some values)

@bids-standard/bep032  is working on  animal ecephys. In "animal world" restriction to age in years is often too restrictive, ambiguous in conversion (# of days in a year changes, # of days in month variable, etc) or not even expressive enough (pre-gestational age etc; let's keep that for a separate issue).  So we need to improve on specification of age by allowing other "units" and also allow for metadata field following some dictionary which would allow specification of the "age reference".  Here is what is done in other standards/uses:

-  NWB uses "age" of "text" type . For DANDI then it allows for treating  [ISO 8601 Durations](https://en.wikipedia.org/wiki/ISO_8601#Durations) values if those are specified, enforced during metadata validation using [nwbinspector](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/checks/nwbfile_metadata.py#L133) . Use of ISO 8601 makes any duration inambigous.

Since we are "unit" oriented ATM, I think the easiest resolution for this aspect would be to just do implement in the schema on what is promised in the text:

- Consider "Units" sidecar file value
- Define explicit "unit" values which would be accepted for age: year, month, day, hour, minute, s, ms

Question to schema folks (@effigies @rwblair @nellh): how difficult would be to generalize current check

```yaml
CheckAge89:
  issue:
    code: AGE_89
    message: .... trimmed...
    level: warning
  selectors:
    - path == 'participants.tsv'
    - columns.age != null
  checks:
    - max(columns.age) < 89
```

to support use of the "Units" in the side car?

Remi-Gau:
Sort of related: 
[...]

PR #1791, opened 2024-04-17
Clarify that age Units could be overriden + that type is number OR "89+" string

yarikoptic:
Ultimately closes https://github.com/bids-standard/bids-specification/issues/1633

FWIW here is what I see among openneuro datasets ATM

```
$> for j in ds*/participants.json; do jq '.age.Units' $j; done | sort | uniq -c
jq: error (at ds002873/participants.json:577): Cannot index array with string "age"
      2 "Measurement units. [<prefix symbol>]<unit symbol> format following the SI standard is RECOMMENDED"
    227 null
      1 "weeks"
    206 "years"
     21 "Years"
      2 "years old"
      1 "years (rounded down)"
      3 "Years, with one quantile precision"
```

and among examples

```
❯ for j in ds*/participants.json; do jq '.age.Units' $j; done | sort | uniq -c
      2 null
      7 "year"
      9 "years"
```

so nobody uses it really besides 1 "weeks".  If we decide to go for it we need
- [ ] extend list of units with "month", "week", "day", ...
- [ ] see if / how we could adjust validation rules in the schema (attn @effigies as guided me on #1633)

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -36,11 +36,16 @@ age:
   name: age
   display_name: Subject age
   description: |
-    Numeric value in years (float or integer value).
+    Numeric value (float or integer value), by default in Years.
 
     It is recommended to tag participant ages that are 89 or higher as 89+,
     for privacy purposes.
-  type: number
+    If age is in not in years, Units field should specify the unit.
+  anyOf:
+    - type: number
+    - type: string
+      enum:
+        - 89+
   unit: year
 cardiac:
   name: cardiac


sappelhoff (commented):
I think it's a good idea to support other units than "Years" and have it be specified under "units"

yarikoptic:
conflict since since then JSON schema `definition` was added thus making it not overridable, ref: describing the difference between two types of columns:
- https://github.com/bids-standard/bids-specification/issues/2048 

so @effigies, to get back on this one and to allow to overload units for the age (needed or desired for animal and babies research and #1839 ) -- I should replace current `definition:` for age with YAML style one, correct?

effigies:
The JSON-sidecar-style column definitions are overridable by JSON sidecars, so age is overridable.

What we have here is something that's not *describable* by the JSON sidecar definitions. There's no notion of "anyOf" in column definitions in BIDS.

effigies:
I suppose our options are:

1) Revert to JSON schema in YAML and
   1) Lose the ability to override the `age` column OR
   2) Allow any column to be overridden, losing the ability to enforce columns whose types shouldn't change OR
   3) Come up with some way to indicate that a column may be overridden at the cost of expressiveness.
2) Figure out a way to at least partially unify the BIDS column description and JSON schema to cover this case.
3) Hack this particular case and see how far it gets us.

For option (3), we could allow `"Units"` and `"Levels"` to turn into (pardon the mix of YAML and Python):

```yaml
anyOf:
  - type: number
  - type: string
    enum: {Levels.keys()}
```

Then we would update this description to be:

```diff
   definition: {
     "LongName": "Subject age",
     "Description": "Subject age in postnatal years",
     "Units": "year",
+    "Levels": {
+      "89+": "Subject at least 89 years old"
+    }
   }
```

cc @rwblair for thoughts.

effigies:
Maybe it would be good to have a call? We've gone back and forth several times, and I'm not really sure what else to write.

yarikoptic:
Worst comes to worst, let's meet and discuss late July I am back to East coast, may be in some convenient BBQ atmosphere.

PR #1798, opened 2024-04-19
Bids 2.0: work out testing against bids-examples

yarikoptic:
To minimize "traffic" on bids-2.0 PR


yarikoptic:
woohoo -- we are green on validate! I thought to add detection and do `git mv` instead of `mv` for a good measure... will do that although not sure if it would actually be needed in my next idea: push those into `bids-2.0` on `bids-examples`...

PR #1803, opened 2024-04-24
[ENH] Add use of session entity as example of encoding multi-site studies in a single dataset

yarikoptic:
- add missing trailing `)`
- point to specific issue within BIDS 2.0
- mention that different session indicators could be used

relevant discussions: https://github.com/bids-standard/bids-2-devel/issues/11


effigies (approved):

tsalo (approved):
This looks good to me.

PR #1805, opened 2024-04-24
[MAINT] Update contributor table generator style

effigies:
pre-commit is unhappy. Looks like flake8 has started linting the contents of f-strings (more).


src/appendices/contributors.md
```diff
@@ -37,6 +37,7 @@ ecosystem (in alphabetical order).
 If you contributed to the BIDS ecosystem and your name is not listed, please add it.
 
 <!-- THE SECTION BELOW IS AUTOMATICALLY GENERATED -->
+<!--lint ignore -->
```
effigies, line 40: Spent way too long trying to figure out why what I wrote is giving a slightly different result. Who cares about fences on an auto-generated table?

Remi-Gau: ha yeah I recognize those lines with wrong fencing issues... 

I had not managed to figure out either what was going on... So resorted to manual fixes the few times I ran it.

I prefer you very pragmatic approach.

Issue #1339, opened 2022-10-24
Proposed changes to the inheritance principle


arokem:
The steering group has now had the opportunity to take up the proposal to make changes to the inheritance principle (e.g., in https://github.com/bids-standard/bids-specification/pull/1003). 

We discussed these proposed changes in a recent meeting together with the maintainers. Our discussion was intended to assess the feasibility, impact and importance of such changes in moving the standard forward within the 1.0 series, especially as it pertains to the evolution of BIDS derivatives. 

Though we realize that these changes could bring clarity to some of the issues brought up in the discussions on the development of some of the derivatives (e.g., https://github.com/bids-standard/bids-bep016/issues/50), we also wanted to assess the impact that these changes would have on already-existing datasets and on the way that existing tools would operate with datasets that implement changes to the principle. 

As noted in our meeting and in some of the discussions on GitHub (e.g., https://github.com/Lestropie/bids-specification/pull/5#issuecomment-1248319397), it is our understanding that changes to the inheritance principle can be introduced without breaking backwards compatibility. However, the main risk we foresee in introducing these changes within BIDS 1.x is that existing tools (mainly those reading BIDS-formatted data) would operate with new datasets under faulty assumptions, which would be hard to account for, potentially introducing subtle errors (as described in the aforementioned comment). We recognize the potential value of the proposal in reducing duplication of metadata, but do not consider the metadata duplication a show-stopper for continued development of derivatives under BIDS 1.x.  

Therefore after careful consideration of the implications of proposed changes to the inheritance principle we have decided to push these back to a future BIDS 2.0 and to keep using the inheritance principle as-is within BIDS 1.x. We suggest that developments proceed apace with the options that are not in violation of the inheritance principle for now.

PR #1807, opened 2024-04-26
[ENH] Describe Inheritance Principle in Common Principles

Lestropie:
Currently attempting to revive BEP016, after having had #1003 rejected via #1339 and alternative solution #1280 rejected also.

In #946 I rewrote the Inheritance Principle. Previous text sort of tried to both explain what it was and how it worked and why all together; with that change I focused on defining a set of rules that were both human-readable and actionable for software developers.

However it jumps out at me now that the specification goes straight into those systematized rules with zero introduction of what the Inheritance Principle actually is or why it exists. So here I've tried to give a very succinct explanation. It only needs to be a user-friendly introduction, so doesn't need to be hardcore optimized, happy to take on modifications.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -663,10 +663,8 @@ but *not* be erroneously associated with other data files
 to which the metadata contained within are not applicable.
 The Inheritance Principle defines a systematized set of rules
 to determine which metadata files to associate with which data files.
-Further, due to the prospect of an individual data file having
-multiple metadata files applicable to it,
-the Principle defines the *order of precedence* with which such
-metadata file contents should be considered.
+Further, because multiple metadata files may apply to an individual data file,
+the Principle defines the *order of precedence* of such metadata files contents.
 
 ### Rules
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -661,7 +661,7 @@ and be placed on the filesystem in such a way that those files
 are deemed to be *applicable* to each relevant data file individually,
 but *not* be erroneously associated with other data files
 to which the metadata contained within are not applicable.
-The Inheritance Principle defines a systematized set of rules by which
+The Inheritance Principle defines a systematized set of rules
 to determine which metadata files to associate with which data files.
 Further, due to the prospect of an individual data file having
 multiple metadata files applicable to it,
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -653,6 +653,23 @@ for more information.
 
 ## The Inheritance Principle
 
+In some circumstances, there can be multiple data files for which
+all or a subset of the relevant metadata is precisely equivalent.
+Where this occurs,
+it may be preferable to define those metadata *only once*,
+and be placed on the filesystem in such a way that those files
+are deemed to be *applicable* to each relevant data file individually,
+but *not* be erroneously associated with other data files
+to which the metadata contained within are not applicable.
+The Inheritance Principle defines a systematized set of rules by which
+to determine which metadata files to associate with which data files.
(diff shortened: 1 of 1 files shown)


effigies (approved):
This reads very cleanly to me.

arokem (approved):

arokem:
+1 thanks for writing this!

src/common-principles.md
```diff
@@ -653,6 +653,23 @@ for more information.
...
 
+In some circumstances, there can be multiple data files for which
+all or a subset of the relevant metadata is precisely equivalent.
+Where this occurs,
+it may be preferable to define those metadata *only once*,
+and be placed on the filesystem in such a way that those files
+are deemed to be *applicable* to each relevant data file individually,
+but *not* be erroneously associated with other data files
+to which the metadata contained within are not applicable.
+The Inheritance Principle defines a systematized set of rules by which
```
yarikoptic, line 664:
```suggestion
The Inheritance Principle defines a systematized set of rules
```
to make it only 2 `which`, 3 of which cause my mental reset in parsing the sentence.

yarikoptic (approved):
Overall great, just one minor suggestion

Lestropie:
Good catch @yarikoptic.

Re-running CI has overcome the prior failure due to a dead link.

sappelhoff (approved):
This is a great addition, thanks a lot!

src/common-principles.md
```diff
@@ -653,6 +653,23 @@ for more information.
...
+all or a subset of the relevant metadata is precisely equivalent.
+Where this occurs,
+it may be preferable to define those metadata *only once*,
+and be placed on the filesystem in such a way that those files
+are deemed to be *applicable* to each relevant data file individually,
+but *not* be erroneously associated with other data files
+to which the metadata contained within are not applicable.
+The Inheritance Principle defines a systematized set of rules
+to determine which metadata files to associate with which data files.
+Further, due to the prospect of an individual data file having
```
DimitriPapadopoulos, line 666: I'm not a native speaker, but I would prefer a shorter sentence:

_Further, because multiple metadata files may apply to an individual data, the Principle defines the *order of precedence* of such metadata files contents._

src/common-principles.md
```diff
@@ -653,6 +653,23 @@ for more information.
...
+multiple metadata files applicable to it,
+the Principle defines the *order of precedence* with which such
+metadata file contents should be considered.
```
effigies, line 669: Adding @DimitriPapadopoulos's suggestion.

```suggestion
Further, because multiple metadata files may apply to an individual data file,
the Principle defines the *order of precedence* of such metadata files contents.
```

effigies:
@Lestropie Are you happy with the changes? I think both the current text and the proposed fix have been up long enough that there's been plenty opportunity to object.

src/common-principles.md
Lestropie: Adopted in 8154a6b648878634b6581e25bdd98336aca70c61.

Lestropie:
Suggest a minor extension to contextualise why such an order is important, partly because it gives the opportunity to recommend against it. Happy for you to merge with or without b7a3ab625a7c1994d3984f3eb79c4811449e7e2b.

effigies:
I would not add a recommendation here. That should be paired with a validator warning, and I'm not positive people would agree that it should be a warning. Nor am I sure how much work it would be to validate.

effigies:
@Lestropie I did drop b7a3ab6 because that had not gotten general review and I think it was a non-trivial change. Feel free to propose it for separate review.

PR #1808, opened 2024-04-26
[INFRA] Refactor metaschema file rules into definitions

effigies:
This should be reviewed as two separate commits. The first contains the change I want to make, the second just runs `prettier` to enforce a consistent style.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -409,15 +409,10 @@
                   "type": "object",
                   "patternProperties": {
                     "^[a-zA-Z0-9_]+$": {
-                      "type": "object",
-                      "properties": {
-                        "level": {"type": "string"},
-                        "path": {"type": "string"},
-                        "extensions": {"type": "array"},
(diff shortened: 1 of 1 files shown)


src/metaschema.json
```diff
@@ -712,16 +727,57 @@
...
+      "additionalProperties": false
+    },
+    "stemRule": {
+      "type": "object",
+      "properties": {
+        "level": { "enum": ["optional", "recommended", "required"] },
+        "datatypes": {
+          "type": "array",
+          "items": { "pattern": "^[a-z]+$" }
+        },
```
effigies, line 757: The main difference with what we currently allow is this entry, which I added to enable #1672. I can move this particular insertion to that PR, if preferred.

effigies: Actually, let me just remove it. We can have that discussion there.

```suggestion
```

bendichter (approved):
Looks great!

Issue #1771, opened 2024-04-11
BEP for audio/video capture of behaving subjects


bendichter:
I would like to create a BEP to store the audio and/or video recordings of behaving subjects.

While this would obviously be problematic for sharing human data, it would be useful to internal human data and for internal and shared data of non-human subjects.

Following the structure of the Task Events we will define types of files that can be placed in various data_type directories.

```
sub-<label>/[ses-<label>]
    <data_type>/
        <matches>_behcapture.mp3|.wav|.mp4|.mkv|.avi
        <matches>_behcapture.json
```

This schema will follow the standard principles of BIDS, listed here for clarity:
* If no relevant <data_type> exists, use `beh/`.
* Video or audio files that are continuous recordings split into files will use the `_split-` entity.
* Video or audio files that are recorded simultaneously but from different angles or at different locations would use the `_recording-` entity to differentiate. We will need to modify the definition of this term to generalize it a bit to accommodate this usage. This entity would also be used to differentiate if a video and audio were recorded simultaneously but from different devices. Not that simply using the file extension to differentiate would not work because it would not be clear which file the .json maps to.
* The start time of each audio or video recording should be noted in the `scans.tsv` file.

The JSON would define "streams" which would define each stream in the file.

The `*_beh.json` would looks like this:

```json
{
  "device": "Field Recorder X200",
  "streams": [
    {
      "type": "audio",
      "sampling_rate": 44100.0,
      "description": "High-quality stereo audio stream."
    },
    {
      "type": "video",
      "sampling_rate": 30.0,
      "description": "Standard 1080p video stream."
    }
  ]
}
```

To be specific, it would follow this JSON Schema structure:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "device": {
      "type": "string"
    },
    "streams": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "enum": ["audio", "video"]
          },
          "sampling_rate": {
            "type": "number",
            "format": "float"
          },
          "description": {
            "type": "string"
          }
        },
        "required": ["type", "sampling_rate"],
        "additionalProperties": false
      }
    }
  },
  "required": ["device", "streams"],
  "additionalProperties": false
}
```

This BEP would be specifically for audio and/or video, and would not include related data like eye tracking, point tracking, pose estimation, or behavioral segmentation. All of these would be considered derived and are reserved for another BEP.

[...]

Issue #153, opened 2019-02-15
Within-stimuli conditions


adelavega:
`stim_file` columns in `event` files allow users to specify which stimuli files are associated with an event onset:

> stim_file | OPTIONAL. Represents the location of the stimulus file  (image, video, sound etc.) presented at the given onset time. ...

However, what this does not allow for is the specification of sub-conditions that occur during a long-running stimulus.

For example, in [ds001545](https://openneuro.org/datasets/ds001545/versions/1.1.0/file-display/sub-01:func:sub-01_task-movie_run-01_events.tsv) a video file is presented which spans the entirety of the run. However, within each run/video there are 6 distinct conditions.

For example:

| onset  | duration | trial_type | stim_file |
| ------------- | ------------- | --- | --- |
| 6  | 90  | Intact A | cond1_run-01.mp4 |
| 105  | 90 | Scramble Fix C | cond1_run-01.mp4 |
| 204  | 90  | Scramble Rnd B V1 | cond1_run-01.mp4 |
| 303  | 90  | Scramble Fix C | cond1_run-01.mp4 |
| 402  | 90  | Intact A| cond1_run-01.mp4 |
| 501  | 90  | Scramble Rnd B V2 | cond1_run-01.mp4 |

IMO, the above example is invalid as the `stim_file` only has a single onset. 
The following is an event file which has all the necessary information (note I'm having to guess when the onset of the `stim_file` is, it could actually be `0`). 

| onset  | duration | trial_type | stim_file |
| ------------- | ------------- | --- | --- |
| 6  | 540  | n/a | cond1_run-01.mp4 |
| 6  | 90  | Intact A | n/a |
| 105  | 90 | Scramble Fix C |  n/a  |
| 204  | 90  | Scramble Rnd B V1 |  n/a  |
| 303  | 90  | Scramble Fix C | n/a  |
| 402  | 90  | Intact A|  n/a  |
| 501  | 90  | Scramble Rnd B V2 | n/a |

However, this is ambiguous as the conditions are only implied to occur during stimulus presentation due to the duration of the first row. 

@tyarkoni suggests adding optional but strongly encouraged `stim_onset` and `stim_offset` columns. These would denote onsets *within* a stimulus.

yarikoptic:
I would have made it

|         onset |      duration | trial_type        | stim_file        |
| ------------- | ------------- | ---               | ---              |
|             6 |           540 | Movie starts      | cond1_run-01.mp4 |
[...]

PR #1814, opened 2024-05-06
[ENH] Add stimulus and annotation entities

neuromechanist:
**stimulus** and **annotation** entities are necessary to distinguish various stimulus files and possible different annotations per each stimulus file.

A sample use-case of these entities is presented in bids-standard/bids-examples#433.

Both entities are being discussed in #153, and this [Google Doc](https://docs.google.com/document/d/1DoQghbJlQksCHEFs0boT816p3ejX2Hd9l-OWDTXFtV8/edit?usp=sharing).
The `stim-` entity was suggested first in #751.
The `annot-` entity might be of use in #1771 as well.

This pull request would likely require a minor version change.

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -436,7 +436,7 @@ annotation:
   display_name: Annotation
   description: |
     The `annot-<label>` entity can be used to distinguish different annotations
-    of the stimlus data.
+    of the stimulus data.
 
     This entity is only applicable to stimulus files.
   type: string
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -422,3 +422,22 @@ tracksys:
     may be longer and more human readable.
   type: string
   format: label
+stimulus:
+  name: stim
+  display_name: Stimulus
+  description: |
+    The `stim-<label>` entity can be used to distinguish stimulus files stored in the `stimuli/` directory.
+
+    This entity is only applicable to stimulus files.
+  type: string
+  format: label
+annotation:
+  name: annot
+  display_name: Annotation
+  description: |
+    The `annot-<label>` entity can be used to distinguish different annotations
+    of the stimlus data.
+
+    This entity is only applicable to stimulus files.
+  type: string
+  format: label
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -603,6 +603,14 @@ stim_file:
     For example `images/cat03.jpg` will be translated to `/stimuli/images/cat03.jpg`.
   type: string
   format: stimuli_relative
+stim_id:
+  name: stim_id
+  display_name: Stimulus ID
+  description: |
+    A stimulus identifier of the form `stim-<label>`,
+    matching a stimulus entity found in `stimuli/` directory.
+  type: string
+  pattern: ^stim-[0-9a-zA-Z]+$
 strain:
   name: strain
   display_name: Strain
--- a/src/schema/rules/entities.yaml
+++ b/src/schema/rules/entities.yaml
@@ -30,3 +30,5 @@
 - density
 - label
 - description
+- stimulus
+- annotation


src/schema/objects/entities.yaml
```diff
@@ -422,3 +422,22 @@ tracksys:
...
+
+    This entity is only applicable to stimulus files.
+  type: string
+  format: label
+annotation:
+  name: annot
+  display_name: Annotation
+  description: |
+    The `annot-<label>` entity can be used to distinguish different annotations
+    of the stimlus data.
```
dorahermes, line 439: Typo - stimlus should be stimulus.

neuromechanist: Done!

src/schema/objects/entities.yaml
```diff
@@ -422,3 +422,22 @@ tracksys:
...
+  type: string
+  format: label
+annotation:
+  name: annot
+  display_name: Annotation
+  description: |
+    The `annot-<label>` entity can be used to distinguish different annotations
+    of the stimulus data.
+
+    This entity is only applicable to stimulus files.
```
yarikoptic, line 441: A parallel to investigate here is `_events.tsv` files, i.e. those could also be created/annotated across various dimensions. ATM they would just "flatten" them into this single file with some column potentially disentangling those dimensions for different annotations (e.g. low level properties of visual stimuli vs higher level cognitive concepts vs responses).  So it feels to me it might be applicable there too.

But with that it somewhat begs me to think about some possible alternative name/concept but I find none and there be none ;)

src/schema/objects/entities.yaml
```diff
@@ -422,3 +422,22 @@ tracksys:
...
+  name: stim
+  display_name: Stimulus
+  description: |
+    The `stim-<label>` entity can be used to distinguish stimulus files stored in the `stimuli/` directory.
+
+    This entity is only applicable to stimulus files.
+  type: string
+  format: label
+annotation:
+  name: annot
```
yarikoptic, line 435:
```suggestion
  name: ann
```

<details>
<summary>although we do have some IMHO obnoxious length 9 "recording", most of the entities short names are 2-4 characters long</summary> 

```shell
❯ awk '/^ *name:/{print length($2) " " $2;}' src/schema/objects/entities.yaml | sort -n | nl
     1	2 ce
     2	2 mt
     3	3 acq
     4	3 den
     5	3 dir
     6	3 inv
     7	3 mod
     8	3 rec
     9	3 res
    10	3 run
    11	3 seg
    12	3 ses
    13	3 sub
    14	3 trc
    15	4 desc
    16	4 echo
    17	4 flip
    18	4 hemi
    19	4 part
    20	4 proc
    21	4 task
    22	5 chunk
    23	5 label
    24	5 space
    25	5 split
    26	5 stain
    27	6 sample
    28	8 tracksys
    29	9 recording
```
</details>

so I wonder if `ann` is enough descriptive/non-ambigous (would not collide with some possible sensible future entity)?  According to chatgpt here is the most popular words starting with `ann` (includes "annotation"):

Anniversary
Announcement
Annulment
Annotation
Annex
Annihilation
Announcer

so I think it would be safe to assume we would not collide

yarikoptic: probably scrape it -- `annot` is much more descriptive!

src/schema/objects/entities.yaml
yarikoptic: Since we already ran into discussion either `desc` could be in effect used as `desc` (and we have example in bids of `_desc-autoannot`), it would be valuable to describe here the difference from `desc`, e.g. `annot` is more specialized in providing annotations to the stimuli/events whenever `desc` is more about processing of the data.  So `_annot-chatgpt` would be preferable over `_desc-chatgpt` as would symbolize annotation over some other chatgpt driven processing (e.g. prompted chatgpt for some analysis of data/stimuli)

neuromechanist:
Closing this PR in favor of a complete for BEP044 (#153)

Issue #1812, opened 2024-05-06
update qMRI to align with the new definition of derivatives


Remi-Gau:
1. qmri appendix defines quantitative maps as derivatives regardless of how they were generated

   https://bids-specification.readthedocs.io/en/latest/appendices/qmri.html#quantitative-maps-are-derivatives

   > Regardless of how they are obtained (pre- or post-generated), qMRI maps are stored in the derivatives directory.

   One exception to this in the qMRI page is the recommendation to store scanner-generated UNIT1 files along with the raw data.

2. The schema files rules for quantitative maps under the rules for raw datasets.

   https://github.com/bids-standard/bids-specification/blob/0950f6df9e9c6d598fcc50e269f350066d78b7a0/src/schema/rules/files/raw/anat.yaml#L34

3. The legacy valiator (version 1.14.5) is completely fine with having those quantitative maps in a raw dataset (tested on several of the bids examples).

---

- **1** contradicts the operational definition of derivatives datasets that was agreed on in Copenhagen last year.

  https://bids-specification.readthedocs.io/en/latest/derivatives/introduction.html#bids-derivatives

  > Raw data are data that have been curated into BIDS from a non-BIDS source. If a dataset is derived from at least one other valid BIDS dataset, then it is a derivative dataset.

- **1** contradicts the behavior of the legacy validator (**3"") and, **1** and **2** tends to contradict each other as the schema seems to imply that quantitative maps are legit raw data (I would expect that the deno validator would behave the same as the legacy one given the schema).

Remi-Gau:
> [!NOTE]
> the location of this definition of derivatives should probably be moved to common principles (maybe in the definition section)

Remi-Gau:
## Suggestion

Adapt the qmri appendix to align with the definition of derivatives. 

As far as I can tell this would not break backward compatibility as it only expands things that can go in raw dataset and also because it seems that any invalid dataset under the current reading of the BIDS spec for qMRI would not have been detected by the validator.

Remi-Gau:
@agahkarakuzu 

pinging you, so you are not surprised if you get a PR to review.

agahkarakuzu:
@Remi-Gau, thank you so much for looking into this, and sorry for being late to the party!

> qMRI maps are stored differently depending on the process that generated them. Pre-generated qMRI maps MAY be stored as part of a raw BIDS dataset, where as they MUST be stored in a derivative BIDS dataset if they were post-generated.

[...]

PR #1815, opened 2024-05-07
[ENH] move definition of raw and derivatives datasets to common principles

Remi-Gau:
- closes None but relates to https://github.com/bids-standard/bids-specification/issues/1812#issuecomment-2095632418

--- 
- HTML
  - https://bids-specification--1815.org.readthedocs.build/en/1815/common-principles.html#definitions
  - note that those definitions are now also part of the glossary

---
- This makes the definition of raw and derivative both more official and easier to find.
- Add some semantic line break.

TODO

- [x] link to the definitions in the relevant derivate related sections

### DIFF
--- a/src/derivatives/introduction.md
+++ b/src/derivatives/introduction.md
@@ -1,12 +1,12 @@
 # BIDS Derivatives
 
-Derivatives are outputs of common processing pipelines, capturing data and
-meta-data sufficient for a researcher to understand and (critically) reuse those
-outputs in subsequent processing.
+[Derivatives datasets](../glossary.md#derivative-common_principles) are outputs of common processing pipelines,
+capturing data and meta-data sufficient for a researcher
+to understand and (critically) reuse those outputs in subsequent processing.
 Standardizing derivatives is motivated by use cases where formalized
 machine-readable access to processed data enables higher-level processing.
 
-The following sections cover additions to and divergences from raw BIDS.
+The following sections cover additions to and divergences from [raw BIDS datasets](../glossary.md#raw-common_principles).
 
 Examples:
 
--- a/src/derivatives/introduction.md
+++ b/src/derivatives/introduction.md
@@ -8,7 +8,7 @@ machine-readable access to processed data enables higher-level processing.
 
 The following sections cover additions to and divergences from [raw BIDS datasets](../glossary.md#raw-common_principles).
 
-[Raw BIDS datasets](../glossary.md#raw-common_principles) are data that have been curated into BIDS from a non-BIDS source.
+[Raw BIDS datasets](../glossary.md#raw-common_principles) are data that have been curated into BIDS from one or more non-BIDS sources.
 If a dataset is derived from at least one other valid BIDS dataset,
 then it is a [derivatives datasets](../glossary.md#derivative-common_principles).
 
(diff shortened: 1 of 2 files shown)


CPernet:
ok but will you keep the intro https://bids-specification.readthedocs.io/en/stable/derivatives/introduction.html with examples or will this go away?

Remi-Gau:
hum true that the examples look a bit weird without the previous sentence. I will put back that sentence.

src/schema/objects/common_principles.yaml
```diff
@@ -46,6 +46,9 @@ dataset:
   description: |
     A set of neuroimaging and behavioral data acquired for a purpose of a particular study.
     A dataset consists of data acquired from one or more subjects, possibly from multiple sessions.
+derivative:
+  display_name: derivative dataset
+  description: If a dataset is derived from at least one other valid BIDS dataset, then it is a derivative dataset.
```
yarikoptic, line 51: why to accent on "valid"? isn't it somewhat assumed since otherwise it is not a BIDS dataset per se? 
Also we are talking about BIDS derivative dataset, right?
Also I do not see need for a conditional here.

hence I would suggest

```suggestion
  description: A BIDS dataset derived from at least one other BIDS dataset.
```

src/schema/objects/common_principles.yaml
```diff
@@ -97,6 +100,9 @@ modality:
     the technique is sufficiently uniform to define the modalities `eeg`, `meg` and `ieeg`.
     When applicable, the modality is indicated in the **suffix**.
     The modality may overlap with, but should not be confused with the **data type**.
+raw:
+  display_name: raw dataset
+  description: A raw BIDS dataset is data that have been curated into BIDS from a non-BIDS source.
```
yarikoptic, line 105: Looking at other definitions, they are continuation of the phrase "{display_name} is ..." and hence I would adjust here too to

```suggestion
  description: A BIDS dataset that have been curated into BIDS from a non-BIDS source(s), for example data from acquisition hardware.
```

src/derivatives/introduction.md
```diff
@@ -1,30 +1,32 @@
...
 Standardizing derivatives is motivated by use cases where formalized
 machine-readable access to processed data enables higher-level processing.
 
-The following sections cover additions to and divergences from "raw" BIDS.
-Raw data are data that have been curated into BIDS from a non-BIDS source.
-If a dataset is derived from at least one other valid BIDS dataset, then it is a derivative dataset.
+The following sections cover additions to and divergences from [raw BIDS datasets](../glossary.md#raw-common_principles).
 
-Examples:
+[Raw BIDS datasets](../glossary.md#raw-common_principles) are data that have been curated into BIDS from a non-BIDS source.
```
yarikoptic, line 11:
```suggestion
[Raw BIDS datasets](../glossary.md#raw-common_principles) are data that have been curated into BIDS from one or more of non-BIDS sources.
```

yarikoptic (changes requested):
thank you!  proposed some wording tune ups

CPernet (approved):

src/schema/objects/common_principles.yaml
```diff
@@ -46,6 +46,9 @@ dataset:
   description: |
     A set of neuroimaging and behavioral data acquired for a purpose of a particular study.
     A dataset consists of data acquired from one or more subjects, possibly from multiple sessions.
+derivative:
+  display_name: derivative dataset
```
effigies, line 50: IMO, derivative and derivative dataset are not identical. If we want to define the generic term "derivative", then it should apply to files and datasets. Similarly with "raw".

PR #1817, opened 2024-05-08
Update motion.md

arnodelorme:


effigies:
I don't really understand the rationale for this. JSON files may always be inherited, so why specifically call them out here?

PR #1822, opened 2024-05-15
[FIX] Change hipCT microscopy suffix to XPCT

chourroutm:
Following the discussion at #1646, I am reverting @balbasty's PR #1686 with `XPCT` instead of `hipCT`: it is more suited for a broader audience.


effigies (approved):
This appears to be the consensus on #1646, and it covers the necessary pieces. Could you make a PR to the BIDS validator to add XPCT suffixes in the following locations?

* https://github.com/bids-standard/bids-validator/blob/43a7117/bids-validator/bids_validator/rules/top_level_rules.json#L254-L276
* https://github.com/bids-standard/bids-validator/blob/43a7117/bids-validator/bids_validator/rules/session_level_rules.json#L227-L247
* https://github.com/bids-standard/bids-validator/blob/43a7117/bids-validator/bids_validator/rules/file_level_rules.json#L582-L600

yarikoptic (approved):

PR #1774, opened 2024-04-12
[MISC] Add shellcheck checks to ensure that those few shell scripts we have are "robust"

yarikoptic:
TODO
- [x] remove TEMP commit and possibly a custom workflow if pre-commit picks it up too

Issue #147, opened 2019-02-12
Automatic enforcement of the 5 day rule


chrisgorgo:
According to https://github.com/bids-standard/bids-specification/blob/master/DECISION-MAKING.md#rules PRs should not be merged sooner than 5 days from the last commit. Right now this rule needs to be manually observed by members with write permissions.

We could automate this by creating a bot that will a) give a request changes review to PRs younger than 5 days and remove it when the 5 day period passes b) merge mergeable PRs.

emdupre:
Killing two birds with one stone -- I think a bot will clear up any ambiguity about when that 5-day clock starts ! :+1:

emdupre:
I wonder if we could write a rule for [Policy Bot](https://github.com/palantir/policy-bot).

chrisfilo:
Investigating if we can do this with mergify.io https://github.com/Mergifyio/mergify-engine/issues/349

emdupre:
Based on discussion in #150, it sounds like we should write rules for disapproval that will only be lifted after 5 days from oldest commit. That way the merging reviewer will still be able to ensure that the PR title is correct for the changelog. Unless anyone else has another suggestion for how to approach this issue !

sappelhoff:
>  That way the merging reviewer will still be able to ensure that the PR title is correct for the changelog.

+1, this would be more robust instead of **automatically** merging after 5 days if 2 approving reviews,

gallexi:
Hello! I'm just passing through while exploring OSS x Neuroscience, but I did a bit of research last night and found [Mergeable](https://probot.github.io/apps/mergeable/), a Probot plugin. I'm not sure if this will serve your needs completely, since it doesn't seem compatible with PushEvents (which, I'm pretty sure, is the particular event you need to be watching for/timing) but perhaps it could be modified slightly to handle those! If folks are interested, I can do some more exploring this weekend.

franklin-feingold:
[...]

PR #1823, opened 2024-05-16
Automerge PRs action

yarikoptic:
xref:
- https://github.com/bids-standard/bids-specification/issues/147

Inspired by 
- https://github.com/bids-standard/bids-specification/pull/1774#issuecomment-2115310934

TODOs
- [x] create automerge label (created, we didn't have any)
- [ ] fix that crap chatgpt produced... or just find a working example of such workflow elsewhere... 
- [ ] see if all desired outputs actually produce what is needed
- [ ] may be test on some other repo where we can trick the dates of commits etc
- [ ] uncomment and do provide GITHUB_TOKEN

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,63 @@
+name: automerge
+on:
+  schedule:
+    - cron: '4 5 * * *'  # Runs at 04:05 (random) UTC every day
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - unlocked
+  pull_request_review:
+    types:
+      - submitted
+      - dismissed
+  check_suite:
+    types:
+      - completed
+  status: {}
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0  # Ensures all history is fetched
+
+      - name: Get date of the last commit
+        id: last_commit_date
+        run: |
+          LAST_COMMIT_DATE=$(git log -1 --format=%cI)
+          echo "Last commit date: $LAST_COMMIT_DATE"
+          echo "::set-output name=last_commit_date::$LAST_COMMIT_DATE"
+
+      - name: Calculate days since last commit
+        id: days_since_last_commit
+        run: |
+          from datetime import datetime
+          last_commit_date = "${{ steps.last_commit_date.outputs.last_commit_date }}"
+          last_commit_datetime = datetime.strptime(last_commit_date, "%Y-%m-%dT%H:%M:%S%z")
+          now = datetime.now(last_commit_datetime.tzinfo)
+          delta = now - last_commit_datetime
+          days_since = delta.days
+          echo "Days since last commit: $days_since"
+          echo "::set-output name=days_since::$days_since"
+
+      - id: automerge
+        name: automerge
+        uses: "pascalgn/automerge-action@v0.16.3"
+        if: steps.days_since_last_commit.outputs.days_since > 5 && ((github.event_name == 'pull_request' && github.event.action == 'labeled') || github.event_name == 'schedule')
+        permissions:
+          contents: write
+        env:
+        # TODO: temporarily disabled to see if all above works
+        #  GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          MERGE_LABELS: automerge
+          MERGE_METHOD_LABEL_REQUIRED: true
+          MERGE_COMMIT_MESSAGE: automatic
+          MERGE_REQUIRED_APPROVALS: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
+name: automerge
+on:
+  schedule:
+    - cron: '4 5 * * *'  # Runs at 04:05 (random) UTC every day
```
jwodder, line 4: Actually, the expression is for 5:04.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+      - dismissed
+  check_suite:
+    types:
+      - completed
+  status: {}
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
jwodder, line 27:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          fetch-depth: 0  # Ensures all history is fetched
+
+      - name: Get date of the last commit
+        id: last_commit_date
+        run: |
+          LAST_COMMIT_DATE=$(git log -1 --format=%cI)
+          echo "Last commit date: $LAST_COMMIT_DATE"
+          echo "::set-output name=last_commit_date::$LAST_COMMIT_DATE"
```
jwodder, line 36:
```suggestion
          echo "last_commit_date=$LAST_COMMIT_DATE" >> "$GITHUB_OUTPUT"
```

The `::set-output` method is deprecated; you need to write to `$GITHUB_OUTPUT` now.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Calculate days since last commit
+        id: days_since_last_commit
+        run: |
```
jwodder, line 40: You need to add `shell: python` for this.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+          from datetime import datetime
+          last_commit_date = "${{ steps.last_commit_date.outputs.last_commit_date }}"
+          last_commit_datetime = datetime.strptime(last_commit_date, "%Y-%m-%dT%H:%M:%S%z")
+          now = datetime.now(last_commit_datetime.tzinfo)
+          delta = now - last_commit_datetime
+          days_since = delta.days
+          echo "Days since last commit: $days_since"
+          echo "::set-output name=days_since::$days_since"
```
jwodder, line 48:
```suggestion
          echo "days_since=$days_since" >> "$GITHUB_OUTPUT"
```

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  # Ensures all history is fetched
+
+      - name: Get date of the last commit
+        id: last_commit_date
+        run: |
+          LAST_COMMIT_DATE=$(git log -1 --format=%cI)
```
jwodder, line 34: This can be faked by the Git user.  It'd be better to fetch the `head.pushed_at` timestamp for the PR from the GitHub API.

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
...
+          now = datetime.now(last_commit_datetime.tzinfo)
+          delta = now - last_commit_datetime
+          days_since = delta.days
+          echo "Days since last commit: $days_since"
+          echo "::set-output name=days_since::$days_since"
+
+      - id: automerge
+        name: automerge
+        uses: "pascalgn/automerge-action@v0.16.3"
+        if: steps.days_since_last_commit.outputs.days_since > 5 && ((github.event_name == 'pull_request' && github.event.action == 'labeled') || github.event_name == 'schedule')
```
jwodder, line 53: Shouldn't you add some sort of check that tests have passed, or does the action take care of that?

jwodder (changes requested):

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,60 @@
+name: automerge
+on:
+  schedule:
```
jwodder, line 3: Problem: Scheduled runs only happen against the default branch, not against PRs.

jwodder:
Even if this succeeds in merging PRs after 5 days, it won't prevent merging before the 5 days are up, which seems like the more important qualification to me.

.github/workflows/automerge.yml
yarikoptic: ;-) thanks

```suggestion
    - cron: '4 5 * * *'  # Runs at 05:04 (random) UTC every day
```

.github/workflows/automerge.yml
yarikoptic: oh, that's a boomer... So the only way would be for a dedicated action to go through open PRs and see if any need to be triggered to be reran or do you see some other way?

jwodder: Do you mean "merged" instead of "reran"?

Aside from that, yes, that's pretty much the only way.

PR #1705, opened 2024-02-14
[ENH] Add BEP032 (microelectrode electrophysiology) specification

yarikoptic:
Replaces https://github.com/bids-standard/bids-specification/pull/1352 submitted from a fork outside of bids-specification.

Add specification for microelectrode electrohpysiology datasets based on the [BEP032 proposal](https://bids.neuroimaging.io/bep032)

- To use this WiP schema on sample datasets, see https://deno.land/x/bids_validator@v1.13.2-dev.0#modifying-and-building-a-new-schema on how to use "stock" `bids-validator` with a custom schema. (attn @TheChymera)

ToDos
- [ ] Please ensure your name is credited on our [Contributors appendix](https://github.com/bids-standard/bids-specification/blob/master/src/appendices/contributors.md).
  To add your name, please edit our [Contributors wiki](https://github.com/bids-standard/bids-specification/wiki/Contributors) and add your name with the type of contribution.
  For assistance, please tag @bids-standard/maintainers.
- [ ] After opening the PR, our continuous integration services will automatically check your contribution  for formatting errors and render a preview of the BIDS specification with your changes.
  To see the checks and preview, scroll down and click on the `show all checks` link.
  From the list, select the `Details` link of the `ci/circleci: build_docs artifact` check to see the preview of the BIDS specification.
- [ ] Add instructions here on how to run new `bids-validator` using schema in this PR
- [ ] Populate schema with specifications from the google doc ...
   - [ ] ... 
- [ ] Add CI action (likely github) to run `bids-validator` on sample datasets and this modified schema (@yarikoptic)

Issues this PR would likely to address
- Fixes #1375

Issues to see being addressed while working on this BEP (likely to move above) or not (moved below):
- #1634
- #1481

Other issues which relate but not in scope here and provided for reference/backreference
- #1669
- #1133
- #1165
- #197 
- https://github.com/bids-standard/bids-2-devel/issues/54

Issue #1634, opened 2023-10-18
Formalize participants' age to clarify the reference point


yarikoptic:
A satellite issue to
- https://github.com/bids-standard/bids-specification/issues/1633
and could be of interest to @bids-standard/bep032 .

ATM `age` column is assumed to be "since birth" but in some experiments, animal or human (fetal MRI) or some organoids, it would not really be "since birth".  To acknowledge the possibility of such different reference points e.g. in dandischema we [defined AgeReferenceType](https://github.com/dandi/dandi-schema/blob/HEAD/dandischema/models.py#L352) and ATM we just use `BirthReference` but allow for [GestationalReference](https://en.wikipedia.org/wiki/Gestational_age) happen we hit such a dataset.

I think it would be valuable for us to think in BIDS about such annotations and how we could handle them.  Since somewhat of a "niche" issue really, a simple workaround solution could potentially be to rely on `TermURL` side car field to point to specific ontology defining gestational age, e.g. http://www.ebi.ac.uk/efo/EFO_0005112 (attn @satra -- should we use this or some other ontology right away in dandischema?). That would allow to disambiguate.

satra:
it may also be helpful to separate out age concepts versus age related data elements. for example age_at_death from EFO can be useful for ex vivo datasets of structure and cannot be captured by age as a keyword alone.

lzehl:
@satra this ontology is not species independent defined correct? definitions are a bit sparse... I wondered if we could rather try to add the terms we need to the the UBERON life cycle ontology (or see if we can adapt to what they already provide):
https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0000000?lang=en

satra:
@lzehl - i don't know all the stage concepts well enough, but if they are there in uberon, we should just use it or add to it.

i think in the context of the original issue, for age specifically we do need a reference and how the reference is defined/measured. in humans it's different from animals.

stage can, and perhaps should, be added as an additional data element for certain species (e.g., c elegans) where the community defines them as relevant to scientific communication.

PR #1826, opened 2024-05-17
[ENH] Schema changes related to `particiapants.tsv` for BEP032

ree-gupta:
- [BIDS BEP032 Proposal](https://docs.google.com/document/d/1oG-C8T-dWPqfVzL2W8HO3elWK8NIh2cOCPssRGv23n0/edit#heading=h.yxuspyxo3mzr)
- Contributes to #1705 
- Related issue about `age` #1634 

- [ ] Clarify if columns can have arrays as value, if not, how do we handle the `surgery_date`

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -69,6 +69,13 @@ associated_brain_region_quality_type:
     The method used to identify the associated brain region (estimated|proof)
     depending on anatomical pictures proofing the location or indirect estimation of the location.
   type: string
+birthdate:
+  name: birthdate
+  display_name: Birthdate
+  description: |
+    Day of birth of the participant. Datetime format and their anonymization are described in [Units](SPEC_ROOT/common-principles.md#units).
+  type: string
+  format: datetime
 beta_rotation:
   name: beta_rotation
   display_name: Beta rotation
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -720,6 +720,14 @@ strain_rrid:
     of the strain of the species, for example: `RRID:IMSR_JAX:000664`.
   type: string
   format: rrid
+# TODO: Clarify if column values can be arrays.
+surgery_date:
+  name: surgery_date
+  display_name: Surgery date
+  description: |
+    Date(s) of surgery. Datetime format and their anonymization are described in [Units](SPEC_ROOT/common-principles.md#units). The details of the surgery should be described in the sidecar json file.
+  type: string
+  format: datetime
 time:
   name: time
   display_name: Time
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -47,8 +47,7 @@ alpha_rotation:
   display_name: Alpha rotation
   description: |
     Euler angle to match probe extension dimensions (width, height, depth) to global x, y, z coordinates.
-  unit:
-    degree
+  unit: degree
 associated_brain_region:
   name: associated_brain_region
   display_name: Associated brain region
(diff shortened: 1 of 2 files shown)


src/schema/objects/columns.yaml
```diff
@@ -42,13 +42,29 @@ age:
...
+age_type:
+  name: age_type
+  display_name: Age type
+  description: |
+    Type of age measurement (e.g. `postnatal`, `fetal`, `gestational`). If not specified, `postnatal` is assumed.
+  type: string
+  enum:
+    - $ref: objects.enums.postnatal.value
+    - $ref: objects.enums.fetal.value
+    - $ref: objects.enums.gestational.value
```
yarikoptic, line 61: Thank you @ree-gupta !  I wonder if we better propose this as a PR against current bids version, just keep `surgery_date` for the BEP032 but propose `age_category` and `age_type` . 

It also would complement
- https://github.com/bids-standard/bids-specification/pull/1791
which I am yet to finish.

Then it would give it a better chance for discussion etc.

Also I guess they should be added as recommended to
https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/tabular_data/modality_agnostic.yaml#L2

src/schema/objects/enums.yaml
```diff
@@ -1405,3 +1405,24 @@ microvascular:
...
+    This is the actual time that has passed since a subject's birth,
+    also known as “chronological age”.
+    If age_type is omitted, age is assumed to be this type.
+fetal:
+  value: fetal
+  display_name: fetal
+  description: |
+    Also known as "embryonic age," this measures the age of a
+    fetus or embryo from the time of conception. It's typically about two weeks
+    less than the gestational age.
```
bendichter, line 1421: For humans, but not other species, right? Or is this common across a broad spectrum of species?

src/schema/objects/columns.yaml
ree-gupta: Done! Also, `birthdate` is another column suggested by the BEP032.

yarikoptic:
@ree-gupta Thanks again for working on this!  Could you please "enable" `pre-commit` for you locally -- `pre-commit install` so then it would check/address problems before you even make/push a commit?

ree-gupta:
> @ree-gupta Thanks again for working on this! Could you please "enable" `pre-commit` for you locally -- `pre-commit install` so then it would check/address problems before you even make/push a commit?

Sure thing! Thank you for the suggestion.

src/schema/objects/enums.yaml
lzehl: If we want to be species independent here (which we should) we need to avoid time absolute times for life cycle stages.

"embryo stage" according to UBERON is defined as: "A life cycle stage that starts with fertilization and ends with the fully formed embryo." http://purl.obolibrary.org/obo/UBERON_0000068

src/schema/rules/tabular_data/microephys.yaml
```diff
@@ -0,0 +1,99 @@
...
+    y__probes: recommended
+    z__probes: recommended
+    manufacturer__probes: recommended
+    device_serial_number: recommended
+    electrode_count: optional
+    width__probes: optional
+    height__probes: optional
+    depth__probes: optional
+    dimension_unit: optional
+    alpha_rotation: recommended
```
Peyman-N, line 25: @ree-gupta can you please delete this so it shows the latest changes google docs?

yarikoptic:
Took me awhile, even sent complimentary #1963 to see the obvious gotcha! ;)   pushed the fix -- hopefully gets green now @ree-gupta ! Thanks again for working on this!!!

src/schema/objects/enums.yaml
yarikoptic: let's continue in #1839 on this fun activity

ree-gupta:
> Took me awhile, even sent complimentary #1963 to see the obvious gotcha! ;) pushed the fix -- hopefully gets green now @ree-gupta ! Thanks again for working on this!!!
> 
> edit: CI is green where we care.

Thank you for spotting and fixing this @yarikoptic! 🙏🏽  the schema validation check passes now.

PR #1830, opened 2024-05-21
[SCHEMA] MRS schema updates

effigies:
@markmikkelsen Here is the PR I promised in https://github.com/bids-standard/bids-specification/pull/1377#issuecomment-2123375356. A bit ahead of schedule.

@rwblair Heads up that we need to add this to the validator. We're going to need to make sure our NIfTI reader can handle NIfTI-2 and header extensions.

### DIFF
--- a/src/modality-specific-files/magnetic-resonance-spectroscopy.md
+++ b/src/modality-specific-files/magnetic-resonance-spectroscopy.md
@@ -179,7 +179,7 @@ A guide for using macros can be found at
 
 ### MRS-relevant fields
 
-Metadata fields that MUST, SHOULD or MAY be present:
+Metadata fields that MUST be present:
 
 <!-- This block generates a metadata table.
 These tables are defined in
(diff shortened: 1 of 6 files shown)


src/modality-specific-files/magnetic-resonance-spectroscopy.md
```diff
@@ -179,19 +179,7 @@ A guide for using macros can be found at
...
-  src/schema/rules/sidecars
-The definitions of the fields specified in these tables may be found in
-  src/schema/objects/metadata.yaml
-A guide for using macros can be found at
- https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
--->
-{{ MACROS___make_sidecar_table("mrs.MRSRequiredFields") }}
-
-SHOULD be present:
+Metadata fields that MUST, SHOULD or MAY be present:
```
effigies, line 182: Combined this into one table. Let me know if you think this is harder to read and we should split it back out into three.

src/schema/objects/metadata.yaml
```diff
@@ -3856,12 +3835,10 @@ VolumeTiming:
...
     It is described using a list of times referring to the onset of each volume
-    in the BOLD series.
-    The list must have the same length as the BOLD series,
+    in the series.
+    The list must have the same length as the series,
     and the values must be non-negative and monotonically increasing.
-    This field is mutually exclusive with `"RepetitionTime"` and `"DelayTime"`.
-    If defined, this requires acquisition time (TA) be defined via either
-    `"SliceTiming"` or `"AcquisitionDuration"` be defined.
+    This field is mutually exclusive with `"RepetitionTime"`.
```
effigies, line 3841: Updated this text to be less BOLD-specific.

src/modality-specific-files/magnetic-resonance-spectroscopy.md
markmikkelsen: I think having them in three makes it easier to parse/read.

effigies: Will revert.

effigies: Done.

PR #1834, opened 2024-05-24
Inheritance Principle: Advise against value overloading. Plan to deprecate support for overloading for BIDS 2.0

Lestropie:
Re-commit of b7a3ab625a7c1994d3984f3eb79c4811449e7e2b that was rejected as a late addition to #1807; re-posting here for separate review as suggested.

I've become progressively convinced that the ability to overload metadata fields via Inheritance is a major source of the disparity in opinions about the Inheritance Principle. Regardless of whether you like or hate it, where I think there should be consensus is that overloading should not be intentionally used, even if BIDS 1.x explicitly permits it and it was used as one of the key demonstrative examples.

The proposed change hopefully also helps to contextualise why the subsequent text is a complex but necessary set of rules.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -669,7 +669,10 @@ to which the metadata contained within are not applicable.
 The Inheritance Principle defines a systematized set of rules
 to determine which metadata files to associate with which data files.
 Further, because multiple metadata files may apply to an individual data file,
-the Principle defines the *order of precedence* of such metadata files contents.
+the Principle defines the *order of precedence* of such metadata content;
+this is necessary for resolution of conflicts if the same metadata field
+contains different values in different metadata files
+(though it is RECOMMENDED to avoid such overloading).
 
 ### Rules
 


effigies:
Seems reasonable to me. If there's consensus here, it would be ideal to raise a validator warning and to provide a tool to remove overrides from datasets automatically, with validation being the more important.

arokem (approved):

Lestropie:
> provide a tool to remove overrides from datasets automatically

Good idea; [added to my list](https://github.com/Lestropie/IP-freely/issues/8).

effigies:
Tagging as needs implementation. Recommendations without validator warnings will have essentially no impact.

effigies:
I created https://github.com/bids-standard/bids-validator/issues/117 to track implementation. I think this would be a good first issue for a new contributor to the validator.

effigies:
xref https://github.com/bids-standard/bids-2-devel/issues/65

src/common-principles.md
```diff
@@ -669,7 +669,10 @@ to which the metadata contained within are not applicable.
 The Inheritance Principle defines a systematized set of rules
 to determine which metadata files to associate with which data files.
 Further, because multiple metadata files may apply to an individual data file,
-the Principle defines the *order of precedence* of such metadata files contents.
+the Principle defines the *order of precedence* of such metadata content;
+this is necessary for resolution of conflicts if the same metadata field
+contains different values in different metadata files
+(though it is RECOMMENDED to avoid such overloading).
```
yarikoptic, line 675: since in BIDS 2.0 we make it not allowed (ERROR), we could also relax (completely remove?) the order of precedence statement since it would not matter, right?  or should we leave it in to disambiguate implementations robust operation and say that e.g. top level (most common) definition takes precedence?

yarikoptic (approved):
Small change for a human, big leap for humanity!

It would have been nice to 
- assess the impact of such a change on existing datasets (openneuro).  I guess the easiest way is to just run validator and see how many hit that new WARNING
- provide "upgrade" functionality (#1984) which would address the warning (help to move metadata down the hierarchy if encountered).

yarikoptic:
I adjusted title to provide more context when used in CHANGELOG.  Please review @Lestropie @effigies

effigies (approved):
LGTM. @yarikoptic Does your latest comment require a change to this text now, or is it more tagged for BIDS 2.0? I think the order of precedence statement should not be removed until it's an error to conflict.

Lestropie:
> It would have been nice to
> - assess the impact of such a change on existing datasets (openneuro). I guess the easiest way is to just run validator and see how many hit that new WARNING
> - provide "upgrade" functionality (https://github.com/bids-standard/bids-specification/issues/1984) which would address the warning (help to move metadata down the hierarchy if encountered).

https://github.com/ohbm/hackathon2025/issues/13 if successful would provide a more tailored tool for detecting these.

effigies:
Cc @ericearl

src/common-principles.md
Lestropie: The mechanisms relating to determination of the order of metadata files remains relevant, not for `.json` but for other metadata extensions. For formats where it is only the "nearest" metadata file that applies, there has to be no ambiguity about which is "nearest".

yarikoptic: well, while it is just a WARNING it remains relevant even for JSONs .  Even when it would be ERROR, it would be nice to describe expected behavior in case of non-compliance, but I think it would not be possible if we relax requirements and allow multiple files at the same level to provide it.

PR #1836, opened 2024-05-24
 Fix generation of BIDS changelog 

sappelhoff:
Replaces #1835, so that this is a PR **not from a fork**, but from the org itself.

This is an attempt to work around the "Pass secrets to builds from forked pull requests" CircleCI setting, which we'd rather leave "switched off" for security reasons.

EDIT: will migrate this to GH actions, as discussed in maintainers meeting.

**NOTE: authentification is working, but we run into secondary API limits, so the workflow may take a long time to run!**

... mentioned in #1844


sappelhoff:
Okay, I am confident that this PR now fixes our changelog situation. I have tested the last stage (after a merge commit) also in https://github.com/sappelhoff/testing and everything runs as expected.

All in all, this PR:

1. moves the changelog generation from circleci to GH actions
2. updates the changelog generation software to a newer version
3. fixes the ensuing bugs that were previously uncaught

with this in place, we should have time to tackle a migration of the changelog generation as discussed in:

- #1844 

... without stressing over it.

PR #1841, opened 2024-05-30
feat(schema): Add schema check for README file

effigies:
Restores issue `README_FILE_MISSING` to the schema-based validator.


rwblair (approved):
We could throw an error/warning if that length is more than 1:
> A BIDS dataset MUST NOT contain more than one README file (with or without extension) at its root directory.

Not sure if that has ever happened, anyone cares if it does happen, or if the legacy validator currently checks for it. Happy to merge with or without.

PR #1849, opened 2024-06-06
[ENH] Generalize Units definition to apply to any variable, including TSV columns

yarikoptic:


effigies:
This was written for qMRI, I think, where the units might be defined for the entire data array.

I do think variable is fine, but note that it now reads this way for an image:

image

I think it's probably clear enough.

tsalo (approved):

yarikoptic:
well, we also could just make it "Measurement unit." I am fine either way.

PR #1864, opened 2024-06-27
Multiple changes to DWI scanner derivatives

### LINKED ISSUE #1862: [BUG] Incorrect description of DWI "TRACE" image

Prompted by work happening in #1831, but is actually a consequence of #1725.

Having acquired data with all scanner-generated derivatives enabled, I can confirm that the "`*_TRACEW`" image produced by the Siemens product DWI sequence is *not* the trace of the diffusion tensor, as the specification currently states; rather, it is a "trace-weighted image", being the mean intensity of *b*>0 data.

(Unfortunately I did not think to obtain multi-shell data to find out whether the sequence does what one would naturally expect for multi-shell data, being a 4D series with a trace-weighted image volume per unique *b*-value. It's possible that whether or not it does this will depend on the scanner base software version / sequence used.)

_No response_

https://bids-specification.readthedocs.io/en/latest/modality-specific-files/magnetic-resonance-imaging-data.html#diffusion-imaging-data

Lestropie:
This PR is intended to supersede #1831 with additional content, as well as resolve #1862.

Rather than incrementing the spec with different aspects of scanner-generated DWI derivatives, as happened previously with #1725 and now #1831, it makes more sense to me to establish what that set is, and resolve the specification against the entirety of that set.

I have additionally created an exemplar dataset that shows all of the derivatives that can be generated by the product diffusion sequence on the Siemens XA platform:
https://github.com/bids-standard/bids-examples/pull/452
If anybody knows of other derivatives that can be generated by other sequences / platforms, it would make sense to add those as extra subjects within that dataset.

While the sequence exports can include data encoding the outcomes of the tensor fit itself, those are exported in a proprietary format, cannot be trivially converted to NIfTI, and are therefore excluded from this list.

-----

Some points requiring discussion (hence draft PR):

- [ ] The "exponential ADC" map is explained [here](https://mriquestions.com/exponential-adc.html). I've confirmed that if you compute the ratio of the trace-weighted image to the mean *b*=0 image you get the same result. This is the name that parameter is given in the protocol editor on the scanner console, but it's not a unique name; it can also be called "attenuation coefficient" or "attenuation factor". Most likely "`expADC`" is the most sensible suffix, but requires agreement.
- [ ] I've de-capitalised the existing `TRACE` suffix, and set these new suffices, based on my thinking [registered over on bids-2-devel](https://github.com/bids-standard/bids-2-devel/issues/26#issuecomment-2187692072) where capitalisation should apply to acronyms.
- [ ] I've not confirmed compatibility between the proposed exemplar dataset and the proposed changes to the spec. I presume this can be done, just not something I've done before.
- [ ] One of the scanner-generated derivatives is labelled as "TENSOR_B0". This is not precisely equivalent to the mean *b*=0 image; rather, it is the estimated *b*=0 signal intensity based on the tensor model fit ("*S0*" is more commonly used for this term in diffusion model space). So it's just a T2-weighted image, but it's arisen from a tensor model fit. Therefore in the exemplar dataset, I've named these data "`*_desc-tensor_T2w.*`". I presume this is not currently permissible; I would personally prefer this kind of encoding as opposed to giving it its own suffix, but if others dislike it we'll need to figure out some different encoding.

### DIFF
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -42,13 +42,6 @@ DIC:
   display_name: Differential interference contrast microscopy
   description: |
     Differential interference contrast microscopy imaging data
-colFA:
-  value: colFA
-  display_name: Colored Fractional Anisotropy image
-  description: |
-    Diffusion images reflecting the directionality of the diffusion tensor color-coded in red for
-    left-right oriented fibers, in blue for superior-inferior oriented fibers and green for
-    anterior-posterior oriented fibers.
 DF:
   value: DF
   display_name: Dark-field microscopy
@@ -58,7 +51,8 @@ FA:
   value: FA
   display_name: Fractional Anisotropy image
   description: |
-    Diffusion images reflecting the directionality of the diffusion tensor
+    Scanner-generated derivative parametric map
+    reflecting the anisotropy of the diffusion tensor
 FLAIR:
   value: FLAIR
   display_name: Fluid attenuated inversion recovery image
@@ -477,11 +471,6 @@ TEM:
   display_name: Transmission electron microscopy
   description: |
     Transmission electron microscopy imaging data
-TRACE:
-  value: TRACE
-  display_name: Trace diffusion weighted image
-  description: |
-    Diffusion images proportional to the trace of the diffusion tensor
 UNIT1:
   value: UNIT1
   display_name: Homogeneous (flat) T1-weighted MP2RAGE image
@@ -559,6 +548,15 @@ channels:
   display_name: Channels File
   description: |
     Channel information.
+colFA:
+  value: colFA
+  display_name: Colored Fractional Anisotropy image
+  description: |
+    Scanner-derived parametric map encoding both the anisotropy and principal direction of the diffusion tensor.
+    Image intensity is proportional to the Fractional Anisotropy (FA) measure;
+    image colour is determined by the principal eigenvector of the diffusion tensor,
+    where red encodes the magnitude of the left-right component of that orientation,
+    with green and blue similarly encoding the anterior-posterior and inferior-superior components respectively.
 coordsystem:
   value: coordsystem
   display_name: Coordinate System File
@@ -617,6 +615,12 @@ events:
   display_name: Events
   description: |
     Event timing information from a behavioral task.
+expADC:
+  value: expADC
+  display_name: Exponental ADC
+  description: |
+    Exponental ADC / Attenuation Coefficient / Attenuation Factor map,
+    as the ratio of trace-weighted to non-diffusion-weighted image intensity
 fieldmap:
   value: fieldmap
   display_name: Fieldmap
@@ -831,6 +835,11 @@ stim:
   display_name: Continuous recording
   description: |
     Continuous measures, such as parameters of a film or audio stimulus.
+trace:
+  value: trace
+  display_name: Trace-weighted diffusion image
+  description: |
+    Trace-weighted image, as the mean intensity of diffusion-weighted images
 uCT:
   value: uCT
   display_name: Micro-CT
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -42,6 +42,13 @@ DIC:
   display_name: Differential interference contrast microscopy
   description: |
     Differential interference contrast microscopy imaging data
+colFA:
+  value: colFA
+  display_name: coloured Fractional Anisotropy image
+  description: |
+    Diffusion images reflecting the directionality of the diffusion tensor color-coded in red for 
+    left-right oriented fibers, in blue for superior-inferior oriented fibers and green for 
+    anterior-posterior oriented fibers.
 DF:
   value: DF
   display_name: Dark-field microscopy
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -47,6 +47,11 @@ DF:
   display_name: Dark-field microscopy
   description: |
     Dark-field microscopy imaging data
+FA:
+  value: FA
+  display_name: Fractional Anisotropy image
+  description: |
+    Diffusion images reflecting the directionality of the diffusion tensor
 FLAIR:
   value: FLAIR
   display_name: Fluid attenuated inversion recovery image
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -46,8 +46,8 @@ colFA:
   value: colFA
   display_name: coloured Fractional Anisotropy image
   description: |
-    Diffusion images reflecting the directionality of the diffusion tensor color-coded in red for 
-    left-right oriented fibers, in blue for superior-inferior oriented fibers and green for 
+    Diffusion images reflecting the directionality of the diffusion tensor color-coded in red for
+    left-right oriented fibers, in blue for superior-inferior oriented fibers and green for
     anterior-posterior oriented fibers.
 DF:
   value: DF
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -44,7 +44,7 @@ DIC:
     Differential interference contrast microscopy imaging data
 colFA:
   value: colFA
-  display_name: coloured Fractional Anisotropy image
+  display_name: Colored Fractional Anisotropy image
   description: |
     Diffusion images reflecting the directionality of the diffusion tensor color-coded in red for
     left-right oriented fibers, in blue for superior-inferior oriented fibers and green for
--- a/src/schema/rules/files/raw/dwi.yaml
+++ b/src/schema/rules/files/raw/dwi.yaml
@@ -40,7 +40,7 @@ sbref:
     chunk: optional
 
 # Common scanner-generated derivatives need raw names
-isotropic:
+ScannerDerivatives:
   suffixes:
     - ADC
     - FA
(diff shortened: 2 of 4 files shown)


src/schema/objects/suffixes.yaml
```diff
@@ -547,6 +548,15 @@ channels:
   display_name: Channels File
   description: |
     Channel information.
+colFA:
+  value: colFA
+  display_name: Colored Fractional Anisotropy image
+  description: |
+    Scanner-derived parametric map encoding both the anisotropy and principal direction of the diffusion tensor.
+    Image intensity is proportional to the Fractional Anisotropy (FA) measure;
+    image colour is determined by the principal eigenvector of the diffusion tensor,
```
effigies, line 557: Appease the spell checking gods (who are American):

```suggestion
    image color is determined by the principal eigenvector of the diffusion tensor,
```

effigies:
Overall these changes look good to me. Made some comments over on https://github.com/bids-standard/bids-examples/pull/452.

effigies:
My takes (as someone who's never analyzed DWI data):

* `expADC`: Seems fine.
* `TRACE` -> `trace`: Seems fine, as long as it happens before 1.10.0 is released (which is why I'm bumping this issue).
* `desc-tensor_T2w`: As noted on https://github.com/bids-standard/bids-examples/pull/452, `desc-<label>` is not permitted for raw files. The T2w suffix could be added to the DWI file rule, if people agree. `dwi/sub-01_T2w.nii` would probably be indicative enough that it's a scanner-derived T2w image, so an additional required entity would not be needed (although `rec-tensor` might be a sensible convention). I also don't see a problem with introducing S0 if people would be more used to calling those images by that name.
* Spec/example comparison: The `desc-tensor_T2w` is the only obvious discrepancy. See below for testing.

If you want to test with schema validation (you need Python 3.8+ and [Deno](https://docs.deno.com/runtime/manual/getting_started/installation/) installed):

```console
# In spec directory
python -m venv .venv
source .venv/bin/activate
pip install -e tools/schemacode
bst export > src/schema.json
BIDS_SCHEMA=$PWD/src/schema.json deno run -A https://github.com/bids-standard/bids-validator/raw/master/bids-validator/src/bids-validator.ts /path/to/bids-examples/example
```

Note that you can add `--reload` to `deno run` to force Javascript and JSON files to be dropped from the cache. When fixes don't fix things, this is often a problem. And using `-e` in the `pip install` ensures that any changes you make to the schema will be immediately available to the next call to `bst`.

@CPernet @jhlegarreta @arokem As participants in the prior discussion (#1831) would you mind weighing in on the discu

jhlegarreta:
- `expADC`: I am OK with the term for the exponential ADC map. I see that "attentuation coefficient" and "attenuation factor" are mentioned in the docs where `expADC` is defined, so I'd say that it is enough.
- `TRACE` -> `trace` : definitely OK.
- `desc-tensor_T2w` / `dwi/sub-01_T2w` : I am fine with the latter as long as it is clearly explained. Not sure how `rec-tensor` would be used (i.e. `dwi/sub-01_rec-tensor` ?); same goes for `S0`: would it be like `dwi/sub-01_S0` ? In either case, I am less inclined towards introducing a new label (either `rec-tensor` or `S0`) for this, but my view on the breadth of the extension is also limited.

Thanks for doing all this.

effigies:
@jhlegarreta 

> Not sure how rec-tensor would be used

`sub-01/dwi/sub-01_rec-tensor_T2w.nii.gz`

`rec-<label>` is already permitted for most (if not all) MR images to indicate reconstruction method.

Lestropie:
After some time, I think I'm gravitating toward the `_S0` suffix:

-   While such images will almost always be T2-weighted, they're not acquired with *the intention* of obtaining T2-weighted contrast between different tissues; it just happens to be a spin echo sequence with a moderate TE. 
-   It would achieve consistency w.r.t. "here's the dedicated suffices to use for scanner-generated DWI derivatives".

jhlegarreta:
Re https://github.com/bids-standard/bids-specification/pull/1864#issuecomment-2275463035 Sound reasonable Robert. Thanks.

Lestropie:
Was not aware of the existing suffix [`S0map`](https://github.com/bids-standard/bids-specification/blob/v1.9.0/src/schema/objects/suffixes.yaml#L282-L293). I think that generalising that suffix to represent "the estimated signal intensity were the contrast of interest to be absent", such that it applies to both TE=0 for fMRI and b=0 for DWI, may be the right approach?

Lestropie:
Note that 7241c2b24669488f56a800d532270e3611233d55 also changes the definition of the "`S0map`" suffix:
```diff
< display_name: Observed signal amplitude (S0) image
> display_name: Projected baseline signal amplitude (S0) image
```

This is not just for compatibility with the augmentation proposed here. I consider that description to be erroneous for its existing usage in multi-echo fMRI. Because there, the signal intensity at TE=0 is *not* "observed"; it is *reconstructed* by fitting a mathematical model to the empirical data.

I have also updated the corresponding example dataset via https://github.com/bids-standard/bids-examples/pull/452/commits/2f23b503537d995b33f80e6c46a97f66e9930e10. But I'm yet to attempt the validation with the proposed schema update locally.

src/schema/rules/files/raw/dwi.yaml
```diff
@@ -40,10 +40,14 @@ sbref:
...
-isotropic:
+ScannerDerivatives:
   suffixes:
     - ADC
-    - TRACE
+    - FA
+    - S0map
+    - colFA
+    - expADC
+    - trace
```
effigies, line 50: Looking at the differences between this and the anatomical S0map:

https://github.com/bids-standard/bids-specification/blob/817454ba05d70893287e2d619230f1fc01b26b8d/src/schema/rules/files/raw/anat.yaml#L34-L66

* No `task` - makes sense
* No `ceagent` - makes sense as long as plain DWI does not permit it
* Has `direction` - Is that still applicable for these files?

effigies (approved):
This LGTM. Small question about the entities.

src/schema/objects/suffixes.yaml
```diff
@@ -286,16 +292,23 @@ RB1map:
...
 S0map:
   value: S0map
-  display_name: Observed signal amplitude (S0) image
+  display_name: Projected baseline signal amplitude (S0) image
   description: |
     In arbitrary units (arbitrary).
+    This represents a model-based estimate of signal intensity
+    in the absence of the contrast under investigation.
     For a multi-echo (typically fMRI) sequence, S0 maps index the baseline signal
     before exponential (T2-star) signal decay.
```
Lestropie, line 301:
```suggestion
    For a multi-echo sequence
    (eg. multi-echo spin echo T2 relaxometry or multi-echo fMRI),
    S0 maps index the baseline signal at a hypothetical echo time of zero
    (that is, in the absence of any T2 or T2-star decay).
```

Guessing from fact that `S0map` is a valid suffix for anatomical MRI that it's there for the context of T2 relaxometry? In which case that's arguably a *better* exemplar than multi-echo fMRI, where such a parametric map is much more likely to come from a processing pipeline than the scanner.

src/schema/rules/files/raw/dwi.yaml
Lestropie: RE `S0` more generally, I originally revised based on what I read in the description, where it exclusively referenced the projected TE=0 signal intensity in fMRI. If it's also used in anatomical imaging, maybe its description should be revised accordingly; see suggestion below.

RE `direction`: in a pipeline context one would expect data from all phase encoding directions to have been combined prior to estimation of such a map. But as a scanner-generated derivative (on Siemens at least), if an imaging experiment involves multiple phase encoding directions, they are run as separate sequence protocols, and so if this scanner-generated derivative is activated there will be one generated per phase encoding direction. So persisting there as optional is right I think.

src/schema/objects/suffixes.yaml
```diff
@@ -286,16 +292,23 @@ RB1map:
   unit: arbitrary
 S0map:
   value: S0map
-  display_name: Observed signal amplitude (S0) image
+  display_name: Projected baseline signal amplitude (S0) image
```
jhlegarreta, line 295: Projected -> Predicted? Estimated?

jhlegarreta (commented):
So `S0map` seems a reasonable choice then :+1:.

Do not know enough about fMRI data yet, so cannot comment on the rest of the points made. Also not familiar about how different scanners consider different phase encoding acquisitions :grimacing:.

src/schema/objects/suffixes.yaml
Lestropie: Don't feel particularly strongly for any choice.

effigies:
This PR needs to be completed ASAP to be included in 1.10.0, or we need to back out #1725 to avoid releasing something we intend to change.

src/schema/objects/suffixes.yaml
jhlegarreta: Agreed. I'm fine if others feel like "projected" is the right.

jhlegarreta (approved):
The PR is still in draft format, but approving @effigies.

tsalo (approved):
I discussed this with @mattcieslak and we're both good with it.

effigies:
In the absence of objection, I'm inclined to merge this next week. @Lestropie Given that you hadn't moved it out of draft, I want to make sure you've got time to respond if there were still open issues, to your mind.

Lestropie:
I'm happy to proceed content-wise, my only concern was the absence of confirmation of validator compatibility, which I wasn't successful at myself; sorry I haven't been monitoring updates here or over at https://github.com/bids-standard/bids-examples/pull/452. As long as that check produces sensible results I suggest merging both at the same time; I've flagged https://github.com/bids-standard/bids-examples/pull/452 as being ready for review also.

PR #1867, opened 2024-07-10
FIX: Declare context definition to be JSON schema, fix validation errors

effigies:
`meta.context.context` is a JSON schema document embedded in the BIDS schema. This establishes that in `metaschema.json` and resolves the bugs that would have been discovered sooner if we had done this.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -51,7 +51,13 @@
           "additionalProperties": false
         },
         "context": {
-          "type": "object"
+          "type": "object",
+          "properties": {
+            "context": {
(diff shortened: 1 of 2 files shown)


effigies:
@rwblair I think this should be a quick one.

effigies:
One question: Do we want to consider removing the extra `context:` level, and just start with `type: object`? That would modify the schema structure, but not a lot depends on this yet, so it's probably not a terrible time to do it.

rwblair:
Sounds good to me.

PR #1869, opened 2024-07-13
rf(metaschema): Validate object definitions that are JSON schema

effigies:
`schema.objects.metadata` entries are very nearly JSON schema definitions. To allow BIDS validators to stand on the shoulders of JSON schema validators, the goal of this PR is to define them as JSON schema with extra fields. By doing so, BIDS validators can pass these directly to JSON schema validators along with the pieces of JSON objects that they are intended to validate.

In order to do this, I've tried to formalize the different "kinds" of objects (or terms) we have in composable classes.

The main change to the schema is to rename `required_fields` to `required`, which has the same meaning in JSON schema. Similarly, we update `recommended_fields` to `recommended`, although this will have no impact on validation for now.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -199,6 +199,7 @@
             "allOf": [
               { "$ref": "#/definitions/generalTerm" },
               {
+                "type": "object",
                 "properties": {
                   "pattern": {
                     "$ref": "https://json-schema.org/draft/2020-12/meta/validation#/properties/pattern"
@@ -219,6 +220,7 @@
                 { "$ref": "#/definitions/generalTerm" },
                 { "$ref": "#/definitions/nameValueTerm" },
                 {
+                  "type": "object",
                   "properties": {
                     "recommended": {
                       "$ref": "https://json-schema.org/draft/2020-12/meta/validation#/properties/required"
@@ -250,6 +252,7 @@
                 { "$ref": "#/definitions/generalTerm" },
                 { "$ref": "#/definitions/valueTerm" },
                 {
+                  "type": "object",
                   "properties": {
                     "unit": { "type": "string" },
                     "anyOf": {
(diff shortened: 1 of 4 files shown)


effigies:
Validating `objects.metadata` as JSON-schema+ objects finds no failures. @rwblair This should be relatively straightforward to review, but LMK if you want to go over anything live.

I've got some more stuff that does catch bugs in the schema that I want to patch in after this is in.

src/metaschema.json
```diff
@@ -260,66 +180,54 @@
...
-                  "type": "string"
-                },
-                "description": {
-                  "type": "string"
+          "propertyNames": { "$ref": "#/definitions/formats" },
+          "additionalProperties": {
+            "allOf": [
+              { "$ref": "#/definitions/generalTerm" },
+              {
+                "properties": {
```
effigies, line 202:
```suggestion
                "type": "object",
                "properties": {
```

src/metaschema.json
```diff
@@ -260,66 +180,54 @@
...
-                  "type": "string"
-                },
-                "type": {
-                  "$ref": "http://json-schema.org/draft-07/schema#/properties/type"
+              "allOf": [
+                { "$ref": "#/definitions/schemaTerm" },
+                { "$ref": "#/definitions/generalTerm" },
+                { "$ref": "#/definitions/nameValueTerm" },
+                {
+                  "properties": {
```
effigies, line 222:
```suggestion
                  "type": "object",
                  "properties": {
```

src/metaschema.json
```diff
@@ -347,19 +246,21 @@
...
-                "display_name": {
-                  "type": "string"
-                },
-                "description": {
-                  "type": "string"
+              "allOf": [
+                { "$ref": "#/definitions/generalTerm" },
+                { "$ref": "#/definitions/valueTerm" },
+                {
+                  "properties": {
```
effigies, line 253:
```suggestion
                  "type": "object",
                  "properties": {
```

rwblair (approved):

PR #1874, opened 2024-07-25
[FIX] Replace defunct awaresystems.be and fslwiki links

effigies:
The link checker failing across the board. awaresystems.be has let their domain lapse, so switching to the Wayback Machine.


effigies:
We could also use http://www.bigtiff.org/...

effigies:
In the absence of feedback, I assume this won't outrage anybody. Feel free to propose alternative links in a new PR.

PR #1875, opened 2024-07-25
FIX: Add MRS events files, restrict sidecar metadata checks to data files

effigies:
@markmikkelsen With these patches, your examples in https://github.com/bids-standard/bids-examples/pull/425 pass.

### DIFF
--- a/src/schema/rules/files/raw/task.yaml
+++ b/src/schema/rules/files/raw/task.yaml
@@ -91,6 +91,16 @@ events__pet:
     reconstruction: optional
     run: optional
 
+events__mrs:
+  $ref: rules.files.raw.task.events
+  datatypes:
+    - mrs
+  entities:
+    $ref: rules.files.raw.task.events.entities
+    reconstruction: optional
+    nucleus: optional
+    volume: optional
+
 timeseries__func:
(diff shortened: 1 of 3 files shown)


effigies:
Also included a small patch to the rendering macro to pick up the events.tsv files:

image
image

src/schema/rules/files/raw/task.yaml
```diff
@@ -91,6 +91,16 @@ events__pet:
     reconstruction: optional
     run: optional
 
+events__mrs:
+  $ref: rules.files.raw.task.events
+  datatypes:
+    - mrs
+  entities:
+    $ref: rules.files.raw.task.events.entities
+    reconstruction: optional
```
tsalo, line 100: Isn't `reconstruction` a within-acquisition entity? Like you can have `rec-defaced` and `rec-refaced` versions of a file, but they come from the same acquisition, so you wouldn't have separate events files for each one.

effigies: It's currently permitted for MRI. I don't have a strong commitment to this, though.

markmikkelsen (approved):

PR #1879, opened 2024-07-31
fix(schema): Add hint to see if dataset name contains a visible character

rwblair:


src/schema/rules/checks/hints.yaml
```diff
@@ -34,6 +34,18 @@ TooFewAuthors:
...
+EmptyDatasetName:
+  issue:
+    code: EMPTY_DATASET_NAME
+    message: |
+      The Name field of dataset_description.json is present but empty of visible characters.
+    level: warning
+  selectors:
+    - path == '/dataset_description.json'
+  checks:
+    - match(json.name, '\S')
```
effigies, line 47: Needs to be `json.Name`. And we should not duplicate the error if `Name` is missing.

```suggestion
  selectors:
    - path == '/dataset_description.json'
    - type(json.Name) != 'null'
  checks:
    - match(json.Name, '\S')
```

rwblair:
I messed up the initial regex:
```
> new RegExp('\\S').test("test")
true
> new RegExp('\S').test("test")
false
```

effigies:
Single backslash should work. YAML will not interpolate them:

```python
>>> import yaml
>>> yaml.safe_load(r"\S")
'\\S'
```

So when you do a double `\\` it gets expanded to `\\\\` in JSON:

```console
❯ curl -sSL https://bids-specification--1879.org.readthedocs.build/en/1879/schema.json | jq .rules.checks.hints.EmptyDatasetName
{
  "issue": {
    "code": "EMPTY_DATASET_NAME",
    "message": "The Name field of dataset_description.json is present but empty of visible characters.\n",
    "level": "warning"
  },
  "selectors": [
    "path == '/dataset_description.json'",
    "type(json.Name) != 'null'"
  ],
  "checks": [
    "match(json.Name, '\\\\S')"
  ]
}
```

I would rather use raw backslashes, since YAML will let us. This way the implementation can know that it's been given what we actually want, and do any escaping of them itself when needed: https://github.com/bids-standard/bids-validator/pull/2042

effigies (approved):

PR #1880, opened 2024-08-02
fix(schema): Add issue definitions for all checks

effigies:
Checks need to define their own error codes. Updated the metaschema to enforce this and then fleshed them out.

@rob-luke If you wouldn't mind having a look over the NIRS checks? I'm pretty sure I got the intent correct, from re-reading the spec.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -315,7 +315,7 @@
                       "$ref": "#/definitions/ruleTypes/expressionList"
                     }
                   },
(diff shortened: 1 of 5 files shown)


effigies:
@bendichter Just want to say that the metaschema has been such a helpful tool. Thanks so much for starting us down this road!

rwblair (approved):

nellh (approved):

PR #1887, opened 2024-08-07
[SCHEMA] add non behavioral datatype assertion for warning about missing events.tsv

rwblair:
fixes bids-standard/bids-validator#2058

### DIFF
--- a/src/schema/rules/checks/events.yaml
+++ b/src/schema/rules/checks/events.yaml
@@ -10,6 +10,7 @@ EventsMissing:
     level: warning # could be an error with the proper selectors, I think
   selectors:
     - dataset.dataset_description.DatasetType == "raw"
+    - datatype != "beh"
     - '"task" in entities'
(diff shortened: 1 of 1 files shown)


effigies (commented):
I don't think this works. I think you want to use intersects.

src/schema/rules/checks/events.yaml
```diff
@@ -10,9 +10,10 @@ EventsMissing:
     level: warning # could be an error with the proper selectors, I think
   selectors:
     - dataset.dataset_description.DatasetType == "raw"
+    - datatype != "beh"
     - '"task" in entities'
     - '!match(entities.task, "rest")'
-    - suffix != "events"
+    - '!(suffix in ["events", "beh"])'
```
effigies, line 16:
```suggestion
    - '!intersects([suffix], ["events", "beh"])'
```

PR #1888, opened 2024-08-14
[FIX] Allow (but discourage) task, acquisition and run entities for coordsystem.json and electrodes.tsv

effigies:
We've run into a situation where the regular validator regular expressions were written more laxly than the specification, leading to datasets that use entities that are against the spec but perfectly interpretable. In particular, the `task` entity has been permitted for `coordsystem.json` across EEG, iEEG, MEG and NIRS.

Following the pattern established in #1722, instead of marking these datasets invalid with a schema-based validator, we relax the spec but add warnings.

cc @bids-standard/raw-electrophys and @rob-luke @lpollonini (we should make a NIRS team...).

### DIFF
--- a/src/modality-specific-files/electroencephalography.md
+++ b/src/modality-specific-files/electroencephalography.md
@@ -439,6 +439,12 @@ landmarks or fiducials during an EEG session/run, must be stored separately in
 the corresponding `*_T1w.json` or `*_T2w.json` file and should be expressed in
 voxels (starting from `[0, 0, 0]`).
 
+`*_coordsystem.json` files SHOULD NOT be duplicated for each data file,
+for example, across multiple tasks.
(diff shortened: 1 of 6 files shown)


Remi-Gau (approved):

effigies:
@dorahermes @VisLab WDYT?

dorahermes:
Just noting that if a `_task-` entity is allowed for an `_coordsystem.json` it should also be allowed for an `_electrodes.tsv`.

effigies:
Good catch. task, acquisition and run were all permitted in the legacy validator for both electrodes.tsv and coordsystem.json, and all six combinations exist on OpenNeuro. For consistency, I've updated this PR to treat these all the same.

Issue #1165, opened 2019-05-09
[ENH] Allow for - (dash) (and/or +) in <label>


yarikoptic:
I thought that we already had an issue filed for this, but may be it was just my comments in the google doc version of the BIDS specification, so filing anew.  Please close it referencing original one if one already exists.

ATM `<label>` is [as I formalized in yet to be finished PR](https://github.com/bids-standard/bids-specification/pull/152/files#diff-0478dbefc276b99dd4637e2cf83dc01fR11) is allowed to contain only alpha numerics.  That makes it really difficult in some cases to place multiple entries (e.g. details for `_acq-` or abbreviate details of preprocessing in `_desc-` of BIDS derivatives). In ReproIn heuristic for heudiconv, where we do allow for `-` we just replace them with an ugly `X` as a separator to just stay compatible with BIDS.

I have been suggesting to extend the list of allowed characters in `<label>` to include at least `-` and possibly others (e.g., `+`) which otherwise should not introduce any backward incompatibility with BIDS 1.0 -- all BIDS 1.0 names would be compatible with new syntax allowing `-` in the `<labels>`.

The original concern (probably at least a year back) which precluded any action was a possible breakage of parsing BIDS filenames for BIDS-supporting software, and was suggested to be included in coming some time in the future BIDS 2.0.

N.B.  I have failed to file and issue outlining plans for BIDS 2.0, so I initiated a new milestone (2.0.0) to which I am adding this new issue.

IMHO (and may be we should outline it somewhere) BIDS supporting applications should be coded adhering to [Postel's principle](https://en.wikipedia.org/wiki/Robustness_principle) which is driving TCP implementations to "be conservative in what you do, be liberal in what you accept from others.". And most likely, in the case of added `-`, they already are since they are more likely to be splitting based on `_` and then on `-` to separate `key` from `value` instead of full rigid regular expressions limiting to the allowed characters (not so obviously) mandated by the BIDS specification.  So the only catch here indeed might be splitting on `-` without restricting that only a single split should happen.  That is where breakage could happen if `-` is added.  The possibility of breakage in case of extending the list of allowed characters with `+` (or even `:` which I dislike to suggest due to messing up scp/ssh) unlikely to have negative ramifications.

Please share your thoughts/opinions

effigies:
I moved this back. Sorry, not sure exactly why it got moved; perhaps it seemed abandoned, or was lumped in with the BIDS-2 proposal linked above because of the `-` suggestion.

[...]

PR #1900, opened 2024-08-21
[ENH] Add volspace entity for CIFTI files

tsalo:
This PR adds a `volspace` entity to describe the volumetric space in combined surface-volume derivative files (i.e., CIFTIs). The volspace entity comes from BEP011: Structural derivatives (#518).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -16,5 +16,5 @@
 /src/derivatives/diffusion-derivatives.md @francopestilli @oesteban @Lestropie
 
 # The schema
-/src/schema/ @tsalo @erdalkaraca
-/tools/schemacode/ @tsalo @erdalkaraca
+/src/schema/ @erdalkaraca
+/tools/schemacode/ @erdalkaraca
--- a/src/derivatives/imaging.md
+++ b/src/derivatives/imaging.md
@@ -11,7 +11,7 @@ Template:
 <pipeline_name>/
     sub-<label>/
         <datatype>/
-            <source_entities>[_space-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.<extension>
+            <source_entities>[_space-<space>][_volspace-<space>][_res-<label>][_den-<label>][_desc-<label>]_<suffix>.<extension>
 ```
 
 Volumetric preprocessing does not modify the number of dimensions, and so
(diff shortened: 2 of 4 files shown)


effigies:
TBH I don't like it very much. I would rather revive https://github.com/bids-standard/bids-specification/issues/1165 than have two entities with such similar purposes.

A space can refer to a dictionary of references (https://bids-specification.readthedocs.io/en/stable/derivatives/common-data-types.html#spatial-references), so you can already just make up a name and define it. Permitting `+` would just be a convenience to not have to come up with a name for a new combined space.

tsalo:
I had hoped to use `+` for cohorts, but what would that look like with volspace as well? Something like `dhcpAsym+42+MNIInfant+2` seems a little unwieldy.

effigies:
I see, the concern is parsing filenames in order to look up associated files in something like templateflow. That makes me think that metadata is probably the better way to go. What if you had something like `space-dhcpAsym42+MNIInfant2_<suffix>.json`:

```json
{
  "SpatialReference": {
    "VolumeReference": "bids:MNIInfant:cohort-2/tpl-MNIInfant_cohort-2_res-1_T2w.nii.gz",
    "CIFTI_STRUCTURE_CORTEX_LEFT": "bids:dhcpAsym:cohort-42/tpl-dhcpAsym_cohort-42_hemi-L_den-32k_white.surf.gii",
    "CIFTI_STRUCTURE_CORTEX_RIGHT": "bids:dhcpAsym:cohort-42/tpl-dhcpAsym_cohort-42_hemi-R_den-32k_white.surf.gii"
  }
}
```

(I'm making up these filenames. They may not be right.)

tsalo:
According to the spec, space values must be defined in the schema's list of coordinate spaces. I imagine we could have something in the spec/schema about parsing the space values, though it would make the validator a bit more complicated I assume.

I'm fine switching to your proposal, but I'm a little worried about what the effects would be on downstream software (e.g., niworkflows, fMRIPrep, XCP-D). Adding new entities is straightforward, but changing how we structure space definitions might be tougher.

effigies:
> According to the spec, space values must be defined in the schema's list of coordinate spaces.

Reference? AFAIK this was never intended to be exhaustive but to provide common names for frequently-used spaces.

tsalo:
https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#space

> The `<label>` MUST be taken from one of the modality specific lists in the [Coordinate Systems Appendix](https://bids-specification.readthedocs.io/en/stable/appendices/coordinate-systems.html).

effigies:
Got it, thanks. That requirement was introduced in https://github.com/bids-standard/bids-specification/pull/734, and it appears to be for the sake of validating raw iEEG data (https://github.com/bids-standard/bids-validator/issues/743) and has been enforced for iEEG and EEG (https://github.com/bids-standard/bids-validator/pull/1190) only.

It has been widely flouted in the derivatives context, at least. fMRIPrep, for example, has used spaces `T1w`, `anat`, `boldref` and `func`, to denote spaces with explicit images. Further, the schema rules are:

https://github.com/bids-standard/bids-specification/blob/7b0f76d3c062363d1338084efcf979afc46cf50d/src/schema/rules/sidecars/derivatives/common_derivatives.yaml#L12-L28

This apparently doesn't have corresponding text, but it is what the schema validator is enforcing. The text I would write would be:

> If the `space-<label>` entity is not defined in the BIDS specification (Coordinate Systems Appendix),
> then the `SpatialReference` metadata field MUST be defined.

tsalo:
To summarize what I think the current state of the conversation is, we are going to try to support `+`s in entity labels (see #1926), and surface space, surface cohort, volume space, and volume cohort will all be encoded in the `space` entity.

Additionally, in some cases, we will use surface template names as shorthand for specific surface-volume combinations. For example, `space-fsLR` in a CIFTI means that the surface is fsLR and the volume is MNI152NLin6Asym.

I don't want to extend that to _every_ surface-volume combination though. For example, if the onavg surface template folks prefer to combine it with MNI152NLin2009cAsym, I'd like to see that labeled as `space-onavg+MNI152NLin2009cAsym` rather than just `space-onavg`, but I guess we can discuss that elsewhere.

In any case, I guess I can close this in favor of #1926.

PR #1905, opened 2024-08-22
fix(schema): Remove SLICE_TIMING_AND_DURATION_MUTUALLY_EXCLUSIVE

effigies:
This rule goes counter to the table in https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#timing-parameters_1.

### DIFF
--- a/src/schema/rules/checks/func.yaml
+++ b/src/schema/rules/checks/func.yaml
@@ -117,19 +117,6 @@ VolumeTimingDelayTimeMutex:
   checks:
     - type(sidecar.DelayTime) == "null"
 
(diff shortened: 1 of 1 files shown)


effigies:
While a review would be nice, I'm comfortable merging this without. I have another post RE `AcquisitionDuration` upcoming.

Remi-Gau:
> While a review would be nice, I'm comfortable merging this without. I have another post RE `AcquisitionDuration` upcoming.

post merge review: yup that checks out for me

PR #1907, opened 2024-08-22
[ENH] Provide baseline definition for behavioral files

effigies:
Part two of resolving https://github.com/bids-standard/bids-validator/issues/1376. Also provides a starting point for behavioral files, instead of being purely free-form. If someone's going to use columns that are defined for `events.tsv`, they should mean the same thing in `beh.tsv`.


Remi-Gau (approved):
good with me

nellh (approved):

PR #1910, opened 2024-08-27
[FIX] typo

Remi-Gau:
while checking #1905

### DIFF
--- a/src/schema/rules/checks/func.yaml
+++ b/src/schema/rules/checks/func.yaml
@@ -101,7 +101,7 @@ RepetitionTimeAcquisitionDurationMutex:
       (RepetitionTime - AcquisitionDuration).
     level: error
   selectors:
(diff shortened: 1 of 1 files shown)


tsalo (approved):

Remi-Gau:
will try to see if we can a have a script to catch things like this more systematically

effigies (approved):

effigies:
We have an expression parser in bidsschematools. We could write a test and validate that, given `sidecar.$SYMBOL`, `$SYMBOL` must exist in `schema.objects.metadata`.

Remi-Gau:
> We have an expression parser in bidsschematools. We could write a test and validate that, given `sidecar.$SYMBOL`, `$SYMBOL` must exist in `schema.objects.metadata`.

stop reading my mind. 😉

PR #1911, opened 2024-08-27
[STY] Use wikipedia links, separate URL definitions

effigies:
Should fix checklinks. We also don't need to be pointing to an ad-supported skin for wikipedia.

I moved the URL definitions out of the table so that URLs can be edited without modifying the width of the table in the future.


Remi-Gau (approved):

Remi-Gau:
I don't think anyone would be outraged if this was merged without a second review.

PR #1915, opened 2024-08-29
[ENH] Add human statement to emphasize that _desc is the last entity

### LINKED ISSUE #1895: Derivatives: `desc-` to be the last entity when used?
In a conversation about diffusion derivatives @tsalo pointed out that in our development of BEP16, we have a `desc-` entity, followed by `param-` entity (see: https://github.com/bids-standard/bids-bep016/pull/24/files). He thought this was inelegant and pointed out that in other places (e.g., https://bids-specification--519.org.readthedocs.build/en/519/derivatives/functional-derivatives.html#functional-derivatives-maps), there is an understanding that the `desc-` entity is the last one to be used before the suffix. I wonder whether others feel similarly and whether we want to be explicit about this, e.g., here:

https://github.com/bids-standard/bids-specification/blob/b91d0e355c69fbfa888577abc03fe8c5bdada592/src/schema/objects/entities.yaml#L68

yarikoptic:
Partially addresses #1895

attn @arokem

### DIFF
--- a/src/schema/rules/entities.yaml
+++ b/src/schema/rules/entities.yaml
@@ -31,4 +31,7 @@
 - resolution
 - density
 - label
+# NOTE: there is a current convention to keep description to be the last entity.
+# If to be changed - potentially might need to adjust text, e.g. now in
+# src/derivatives/common-data-types.md
 - description
--- a/src/schema/rules/entities.yaml
+++ b/src/schema/rules/entities.yaml
@@ -32,6 +32,7 @@
 - density
 - label
 # NOTE: there is a current convention to keep description to be the last entity.
-# If to be changed - potentially might need to adjust text, e.g. now in
+# If to be changed - potentially might need to adjust text in
 # src/derivatives/common-data-types.md
+# See https://github.com/bids-standard/bids-specification/pull/1915
 - description
--- a/src/derivatives/common-data-types.md
+++ b/src/derivatives/common-data-types.md
@@ -25,7 +25,7 @@ A guide for using macros can be found at
 Preprocessed `bold` NIfTI file in the original coordinate space of the original run.
 The location of the file in the original datasets is encoded in the `Sources` metadata,
 and [`_desc-<label>`](../appendices/entities.md#desc)
-is used to prevent clashing with the original filename.
+is used (the last entity before the suffix) to prevent clashing with the original filename.
 
 <!-- This block generates a file tree.
 A guide for using macros can be found at


arokem:
Thanks! I don't think this quite covers the issue raised in #1895 though, because this points to a specific example, without stating explicitly that this is the convention for other cases as well. Does that make sense?

yarikoptic:
Good observation @arokem !  our problem is that we do not really introduce `_desc` anywhere before.  For now I just adjusted original description that it doesn't really close that issue, but I think for this could be a quick enhancement.

arokem (approved):

effigies:
I realized the other day that I now wish that we had put `hemi-` last. e.g., this would be preferable:

```
sub-01_desc-custom_hemi-L_white.surf.gii
sub-01_desc-custom_hemi-R_white.surf.gii
sub-01_desc-freesurfer_hemi-L_white.surf.gii
sub-01_desc-freesurfer_hemi-R_white.surf.gii
```

to

```
sub-01_hemi-L_desc-custom_white.surf.gii
sub-01_hemi-L_desc-freesurfer_white.surf.gii
sub-01_hemi-R_desc-custom_white.surf.gii
sub-01_hemi-R_desc-freesurfer_white.surf.gii
```

It seems to me that `desc-` is not a grouping entity, like `echo-` or `part-` (or `hemi-`) and generally we probably want grouping entities toward the back and distinguishing entities toward the front.

So if we have something like components and a mixing matrix that are distinguished by entity, we might prefer:

```
sub-01_task-rest_desc-ICA_<key>-components_<suffix>.<ext>
sub-01_task-rest_desc-ICA_<key>-mixing_<suffix>.<ext>
```

This is not grounded in a particular BEP, but just a thing to consider before we codify this.

yarikoptic:
I think your motivation @effigies relates to fresh

- https://github.com/bids-standard/bids-specification/issues/1949

as a more generic rule of entities composition.

> generally we probably want grouping entities toward the back and distinguishing entities toward the front.

well, `sub-` is a very groupping entity ;)  overall it all depends on the case and there might be no universally best order.  May be could be in part or fully addressed as part of the 

- https://github.com/bids-standard/bids-2-devel/issues/54
- (some drafty beginning of work on it): https://github.com/bids-standard/bids-specification/pull/1809/files#diff-efe09d8f5bbcea3e5772ac9f06f13e149454fbdfc00b1aa286a25aac7c2d7303R185

as that one already would require change of the order of entities in the folders/filenames.  Hence could as well be extended to the customization of entities to fit "the best" order.

effigies:
To be clear, by "grouping entity", I mean one that can be varied without changing the overall meaning of the files. Different `part`s or `echo`s indicate multiple files that should be considered together. Different `task`s or `sub`s need separate consideration.

effigies:
Dismissing my review request, as I don't think this should be in the spec, and I don't have a suggestion to make that would get me to approve this. I won't block, however, so others are welcome to review.

IMO this should be part of contribution guidelines.

neuromechanist (approved):

neuromechanist (approved):
In terms of derivative ExG data that I have worked on, this pretty much makes sense.

src/schema/rules/entities.yaml
```diff
@@ -31,4 +31,7 @@
 - resolution
 - density
 - label
+# NOTE: there is a current convention to keep description to be the last entity.
+# If to be changed - potentially might need to adjust text, e.g. now in
+# src/derivatives/common-data-types.md
```
effigies: Apparently our latin phrase checker doesn't like e.g. in comments. Adding a link to this PR in case filenames change and this reference fails to get updated.

```suggestion
# If to be changed - potentially might need to adjust text in
# src/derivatives/common-data-types.md
# See https://github.com/bids-standard/bids-specification/pull/1915
```

PR #1926, opened 2024-09-20
[ENH] Allow plus signs in labels

### LINKED ISSUE #1165: [ENH] Allow for - (dash) (and/or +) in <label>

tsalo:
Closes #1165. Adds `+` to the valid characters for the "label" format, which is currently only used for a subset of entities.


tsalo:
In this PR, plus signs are allowed for any label-format value, but I believe some folks (@effigies?) were thinking of limiting them to special multi-label cases. If we want to go down that route, then we'll need to create a new "format" (e.g., "multilabel") that applies to specific entities. I'm not 100% which entities we would want the multilabel format for... perhaps `acq`, `desc`, and `space`?

effigies:
Reading https://github.com/bids-standard/bids-specification/issues/1165#issuecomment-1549699835 and the following discussion, I think the overall consensus was toward permissiveness; i.e., allow in any label instead of splitting out a new multi-label concept.

I feel like the discussion about "semantics" was not entirely clear. I personally feel we should refrain from implying a relation between `ent-X+Y` and `ent-X` or `ent-Y` that tooling would need to respect. Perhaps we want to say something like:

> Free-form labels with alphanumeric and plus (+) characters.
> Plus signs MAY be used to concatenate multiple applicable labels,
> but no relationship is established by a partial match.
> In particular, the inheritance principle does not connect files
> containing `ent-X+Y` and `ent-X` or `ent-Y`.

tsalo:
I like that!

effigies (approved):

CODEOWNERS
oesteban, line 1: I would leave the update of `CODEOWNERS` for a separate PR.

src/schema/objects/formats.yaml
```diff
@@ -9,8 +9,13 @@ index:
...
   display_name: Label
   description: |
-    Freeform labels without special characters.
-  pattern: '[0-9a-zA-Z]+'
+    Free-form labels with alphanumeric and plus (+) characters.
+
+    Plus signs MAY be used to concatenate multiple applicable labels,
+    but no relationship is established by a partial match.
+    In particular, the inheritance principle does not connect files
+    containing "ent-X+Y" and "ent-X" or "ent-Y".
```
oesteban, line 17:
```suggestion
    containing entities such as `<name>-x+y` with either `<name>-x` or `<name>-y`.
    For example, metadata stored in a file at the root of the dataset with name `/acq-6p+s2_T2w.json`
    does not apply to files with partially matching "acquisition" entity values
    such as `/sub-1/anat/sub-1_acq-6p_T2w`.
```

effigies: I would worry about people misinterpreting it in the opposite direction:

```suggestion
    containing entities such as `<name>-x+y` with either `<name>-x` or `<name>-y`.
    For example, metadata stored in a file at the root of the dataset with name `/acq-6p_T2w.json`
    does not apply to files with partially matching "acquisition" entity values
    such as `/sub-1/anat/sub-1_acq-6p+s2_T2w.nii`.
```

tools/schemacode/bidsschematools/tests/test_rules.py
```diff
@@ -50,8 +50,8 @@ def test_entity_rule(schema_obj):
     json_rule = rules._entity_rule(rule, schema_obj)
     assert json_rule == {
         "regex": (
-            r"(?:sub-(?P<subject>[0-9a-zA-Z]+)/)?"
-            r"(?:ses-(?P<session>[0-9a-zA-Z]+)/)?"
+            r"(?:sub-(?P<subject>[0-9a-zA-Z+]+)/)?"
+            r"(?:ses-(?P<session>[0-9a-zA-Z+]+)/)?"
```
yarikoptic, line 54: I am afraid that is not all ...  I will push a few commits on that end in a few minutes (I hope you don't mind).
Some other might need adjustment and I even start feeling that we might need to come up with some term (like "literal" but there might be better) to encompass "alphanumeric" and `+`.

yarikoptic:
Just a sidenote FTR: There is also a question of either to allow it in suffixes (e.g. would converge DANDI layout closer (https://github.com/dandi/dandi-cli/issues/1498) but since we do not have any "incident" of that in BIDS ATM, it is not appropriate to change ATM even though in general I see that suffixes also should be of the same nature as "labels", in particular that IIRC their values could "migrate" into `_mod-` entity.

tools/schemacode/bidsschematools/tests/test_rules.py
effigies: @yarikoptic Note this is a regression test that shows the specific output of a specific synthetic rule.

yarikoptic:
I pushed some changes which I think are due as well, although might need tune ups -- individual commits might have more information/reasoning in the commit messages. But there is also IMHO outstanding review and possibly tune up needed to [src/metaschema.json](http://github.com/bids-standard/bids-specification/blob/HEAD/src/metaschema.json)

<details>
<summary>it does have plenty of "alphanumeric" and _  allowances and I am not yet 100% sure none of them somehow overlaps with "alphanumeric" possible in filename etc... most likely none, but still worth looking at IMHO</summary> 

```shell
❯ git grep 'a-zA-Z0-9[^+]' -- src/metaschema.json
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9][a-zA-Z0-9_-]*$": {
src/metaschema.json:            "^_[a-zA-Z0-9_-]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                    "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                    "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                    "^[a-zA-Z0-9_]+$": { "$ref": "#/definitions/suffixRule" }
src/metaschema.json:                    "^[a-zA-Z0-9_]+$": { "$ref": "#/definitions/suffixRule" }
src/metaschema.json:            "^[a-zA-Z0-9_]+$": {
src/metaschema.json:        "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                "^[a-zA-Z0-9_]+$": {
src/metaschema.json:        "^[a-zA-Z0-9_]+$": {
src/metaschema.json:                "^[a-zA-Z0-9_]+$": {
src/metaschema.json:          "items": { "type": "string", "pattern": "^[a-zA-Z0-9]+$" }
```
</details>

tools/schemacode/bidsschematools/tests/test_rules.py
yarikoptic: my comment is not really about this test -- I meant that changes in this PR (just this test) aren't sufficient. pushed now

tools/schemacode/bidsschematools/tests/test_validator.py
```diff
@@ -86,12 +86,12 @@ def test_write_report(tmp_path):
...
-            "(|_ses-(?P=session))(|_acq-(?P<acquisition>[0-9a-zA-Z]+))"
-            "(|_ce-(?P<ceagent>[0-9a-zA-Z]+))"
-            "(|_rec-(?P<reconstruction>[0-9a-zA-Z]+))"
-            "(|_run-(?P<run>[0-9a-zA-Z]+))"
+            "regex": ".*?/sub-(?P<subject>[0-9a-zA-Z+]+)/"
+            "(|ses-(?P<session>[0-9a-zA-Z+]+)/)anat/sub-(?P=subject)"
+            "(|_ses-(?P=session))(|_acq-(?P<acquisition>[0-9a-zA-Z+]+))"
+            "(|_ce-(?P<ceagent>[0-9a-zA-Z+]+))"
+            "(|_rec-(?P<reconstruction>[0-9a-zA-Z+]+))"
+            "(|_run-(?P<run>[0-9a-zA-Z+]+))"
```
tsalo, line 94: This one doesn't fit. `run` takes an index, not a label. I'm not sure why it currently accepts an alphanumeric label.

effigies: This is not generated from anything, but was hand-written to test writing out a report.

yarikoptic: Good eye! I guess it was a bug, let's fix

```suggestion
            "(|_run-(?P<run>[0-9]+))"
```

?

Remi-Gau (changes requested):
For something like this I would much prefer that one bids example is modified to showcase this so that downstream validator and parsers have got some food to sink their digital teeth into.

tsalo:
@Remi-Gau that's a great point. Do you think it makes more sense to create a new example dataset or to modify an existing one?

I took a quick look at the BIDS examples and https://github.com/bids-standard/bids-examples/tree/master/ds004332 looks promising. There are complex `acq` entities that seem to convey multiple discrete pieces of information. For example, `acq-mpragePMCoff`, `acq-mpragePMCon`, `acq-t2starPMCoff`, and `acq-t2starPMCon` could be changed to `acq-mprage+pmc+off` etc.

Remi-Gau:
I am afraid that if we start adding a new dataset for every new aspect of the spec we need to validate we will end up with too many examples.

How about adding this to the synthetic example?

tsalo:
That works! I'll look into modifying https://github.com/bids-standard/bids-examples/tree/master/synthetic.

src/appendices/entity-table.md
```diff
@@ -12,7 +12,7 @@ For example, if a file has an acquisition and reconstruction label, the
 acquisition entity must precede the reconstruction entity.
 REQUIRED and OPTIONAL entities for a given file type are denoted;
 empty cells imply that entities MUST NOT be specified.
-Entity formats indicate whether the value is alphanumeric
+Entity formats indicate whether the value is alphanumeric (potentially including `+` character(s))
```
yarikoptic, line 15: just for uniformity
```suggestion
Entity formats indicate whether the value is alphanumeric (and possibly including `+` character(s))
```

or should we use some other similar phrase?

yarikoptic (approved):
FWIW, looks good and complete to me. Should we proceed with the merge?

tsalo:
I never got the bids example updated unfortunately. Maybe I could make time late next week to do it, but if someone else is willing to do it that would be great.

yarikoptic:
quick&dirty:
- https://github.com/bids-standard/bids-examples/pull/481

tsalo:
@yarikoptic thanks for opening the examples PR! I think it's ready to merge now.

tsalo:
@Remi-Gau could you check this over one last time? I think you're the only reviewer who hasn't approved.

Remi-Gau (approved):

tsalo:
Thanks @Remi-Gau! Alright, does anyone have any issue with me merging? Otherwise, I'll merge tomorrow morning.

PR #1861, opened 2024-06-20
Add DatasetType="project" and rework existing "layout" example into a proper BIDS dataset

yarikoptic:
- Rationale 1 (major): BIDS standard already provides reasonable structure to formalize organization of various components of a neuroscientific data project: where to place code, original (source) data, derivaitve data, README, CHANGES.  Many projects (e.g. nipoppy, YODA, etc) propose similar and often might be even "inspired" templates . If we explicitly allow for BIDS standard to prescribe study level organization, IMHO it would help many people and projects decide on how to organize their studies/projects.
- Rationale 2:  IMHO BIDS standard should describe only what standard prescribe and not recommend some potential "non-standardized" layouts.  That is why I "reworked" that example into a legitimate BIDS dataset merely by adding `dataset_description.json`.

TODOs:

- [x] provide more relation to existing approaches (attn @snastase)
- [started] craft example(s) for bids-example : https://github.com/bids-standard/bids-examples/pull/451
  - [x] ensure bids-validator with modified schema passes its validation 

## Some references

- edit 1: fresh related quote from [medium.com](https://medium.com/bitgrit-data-science-publication/10-best-practices-for-data-science-21a748a410e4) post (emphasis added):

>   DrivenData Labs, the team behind the popular Cookiecutter Data Science template with over 7.8k stars on GitHub, knows these frustrations well.
> 
> After working on over 100 data science projects with a range of organizations, from new startups to large foundations and Fortune 50 companies. **They’ve identified a major issue: the lack of standardization in project organization**, collaboration, and reproducibility.

- edit 2: paper "A practical guide to data management and sharing for biomedical laboratory researchers" (paywall) https://doi.org/10.1016/j.expneurol.2024.114815 -- organization of data at lab level but so also at a project level.  Check on more ideas and possibly comparison/transformation. Looking forward also for metadata to add to BIDS

PR #1928, opened 2024-09-21
Downgrade absence of sub- folders to warning

yarikoptic:
I could be wrong but I think there is no strict and agreed upon requirement to require having sub- subfolders in BIDS specification. Although odd, thus warranting a warning, I think it is not strictly forbidden, and e.g. a BIDS dataset only with stimuli/ or derivatives/ or sourcedata might still be well legit BIDS dataset.

Check was added in b9ea963df27ded2c4363199c4c030c0229f2d1e2 likely just copying from prior node version of bids-validator.  Asking @effigies for setting me straight ;)

### DIFF
--- a/src/schema/rules/checks/dataset.yaml
+++ b/src/schema/rules/checks/dataset.yaml
@@ -7,7 +7,7 @@ SubjectFolders:
     code: SUBJECT_FOLDERS
     message: |
       There are no subject directories (labeled "sub-*") in the root of this dataset.
-    level: error
+    level: warning
   selectors:
     - path == '/dataset_description.json'
   checks:


yarikoptic:
> I don't see the use-case for subject-less datasets. Calling a directory that contains opaque `sourcedata/` and `derivatives/` directories a BIDS dataset seems counter to the goal of BIDS, which is legible neuroimaging data. What's your goal?

immediate, although not mandatory if we do add dedicated DatasetType: 
- https://github.com/bids-standard/bids-specification/pull/1861
and inline with that -- it could be neuro*-related (BIDS is beyond MRI for a while already) dataset of `stimuli/`, or purely `phenotype/` data and IMHO could still be quite legit without "neuroimaging" or per subject data. Also group level results in derivatives etc.
 
But also it is a matter of consistency: if there is such a rule it should be clearly described also in the specification, which I think is not the case ATM.

Indeed it might require openneuro specific configuration, but I thought you already have one anyways.

effigies (approved):

effigies:
@arokem @barbarastrasser I see you gave thumbs ups on the initial post. Would either of you care to make a formal review?

effigies:
For the record, I agree with the argument that this error is not reflected in the text, so we should either update the text or relax the warning. OpenNeuro will continue to make it an error.

barbarastrasser:
> @barbarastrasser I see you gave thumbs ups on the initial post. Would either of you care to make a formal review?

Yes, this is great! This way we can remove the manual downgrade in the validator config :)

PR #1930, opened 2024-09-25
[FiX] Exclude emtpyroom sub when checking event files

monique2208:


effigies:
Thanks!

effigies:
@monique2208 Could you add yourself to https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors#new-contributors to be credited as a contributor?

PR #1939, opened 2024-10-03
[ENH] Add ContextURI to allow to define the context for the entity values

yarikoptic:
It is often desired to be able to determine that values used for entities in the dataset belong to some controlled vocabulary, or simply defined centrally within some "id" authority. E.g. could be unique scanning session IDs per scanning center, or similarly subject_ids defined per study or centrally for the center.

It is of particular interest for large studies where multiple datasets could be created, one per site or primary data modality, to later possibly be composed into a single dataset or just to become parts of the one larger multi-site one. In such cases it becomes quite important to annotate that particular entities (subject_id, session_id and possibly even _desc- or _acq- values) are defined in the scope of the specific  larger study and thus correspond to the "same" thing given the same contextURI and value.

TODOs:
- [ ] discuss definition of centrally defining context prefixes (see below)
- [ ] provide example to bids-examples datasets

## Context Prefixes

in .jsonld etc it is common to centrally define common [JSON-LD Contexts](https://niem.github.io/json/reference/json-ld/context/) which could even be defined externally and pointed via `@context` attribute. E.g. in https://dandiarchive.s3.amazonaws.com/dandisets/000003/draft/dandiset.jsonld we point to https://raw.githubusercontent.com/dandi/schema/master/releases/0.6.0/context.json which would tell within its `@context` that `"ORCID": "https://orcid.org/",` and `"spdx": "http://spdx.org/licenses/",`.  Now if we specify that `"license": "spdx:apache-2.0"` we know that license "identity" is really `http://spdx.org/licenses/apache-2.0` (actual URL does not even have to exist).

So, I wonder if we could/should define within `dataset_description.json` also `Context: dict[str, str]` which would provide similar mappings. So then I could 

- in `dataset_description.json` have `"Context": {"thelab": "http://thelab.example.com/term/"}`
- in `participants.json` for `participant_id` to have `"ContextURI": "thelab:subject"` which in turn for every `participant_id` ultimately get expanded into `http://thelab.example.com/term/subject/{participant_id}` if to map across datasets.

attn @satra and @tekrajchhetri who know "linked" stuff better and could express their recommendations how we could align even better


src/common-principles.md
```diff
@@ -592,6 +593,20 @@ like in the example below.
...
+In the example below it would establish that `participant_id` is shared across datasets of the center, which would allow to differentiate center's subjects apart from subject ids used by other datasets used in the lab.
+
+```JSON
+{
+    "participant_id": {
+        "Description": "Unique identifier used for study subjects in the lab across studies",
+        "ContextURI": "https://thelab.example.com/term/subject/"
+    }
+}
+```
```
yarikoptic, line 607: I guess here would also be nice to give example with `bids:` URI to point to this particular dataset to say that IDs are unique to this dataset?

yarikoptic:
@effigies WDYT about this idea ?  It came up again in the scope of the 

- #1705

as a way to define context for pointing to external resource for probe geometries.  I think it would overall be handy to complement `bids:` URIs.

Let me know if we need one more bbq to discuss it "interactively".

effigies:
Looked very briefly. Apologies if this is a dumb question, but isn't `TermURL` designed for this?

PR #1940, opened 2024-10-03
[FIX] Move description of "n/a" above "String values" to avoid any association

### LINKED ISSUE #1938: [BUG] clarify/re-consider use of "n/a" in numeric columns

ATM specification for Tabular files https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files states

> String values containing tabs MUST be escaped using double quotes. Missing and non-applicable values MUST be coded as `n/a`. Numerical values MUST employ the dot (`.`) as decimal separator and MAY be specified in scientific notation, using `e` or `E` to separate the significand from the exponent. TSV files MUST be in UTF-8 encoding.

So, in the best reading of it, it mandates use of explicit `n/a` for a missing value in **any** (not only "String values" column) column.
As `n/a` is not a standard placeholder, that unnecessarily complicates loading of such files using anything which expects numeric values for the column (e.g. `onset`).

I have not investigated this further yet and do not have any specific recommendation ATM (e.g. after looking how pandas would expect to have float.nan to be defined in tsv etc). Just raising a possible discussion point.

At least we might want to reorder sentences to remove possible misassociation with string only columns, i.e. to have it

> Missing and non-applicable values MUST be coded as `n/a`. String values containing tabs MUST be escaped using double quotes. Numerical values MUST employ the dot (`.`) as decimal separator and MAY be specified in scientific notation, using `e` or `E` to separate the significand from the exponent. TSV files MUST be in UTF-8 encoding.

https://bids-specification.readthedocs.io/en/latest/...

yarikoptic:
as they should be applicable also to numeric values.

- Fixes #1938

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -489,8 +489,8 @@ first letter in lower case (for example, `variable_name`, not `Variable_name`).
 Column names defined in the header MUST be separated with tabs as for the data contents.
 Furthermore, column names MUST NOT be blank (that is, an empty string) and MUST NOT
 be duplicated within a single TSV file.
-String values containing tabs MUST be escaped using double quotes.
 Missing and non-applicable values MUST be coded as `n/a`.
+String values containing tabs MUST be escaped using double quotes.
 Numerical values MUST employ the dot (`.`) as decimal separator and MAY be specified
 in scientific notation, using `e` or `E` to separate the significand from the
 exponent.


effigies (approved):

yarikoptic:
@bids-maintenance - this is an easy low hanging fruit PR.

effigies:
No semantic change. I'm going to go ahead and merge and let people yell at me if they feel very strongly about the second review.

PR #1956, opened 2024-10-15
[ENH] Allow MED format for iEEG data (`*_ieeg.medd/`)

### LINKED ISSUE #1931: Update (new version) of MEF3 format (MED).
Hello,

I am Matt Stead, the creator of all the MEF formats (1-3).  The new version MEF version (MED) is faster & produces better compression than its predecessors. It continues to be HIPAA & IRB compliant. A fuller description of MED is available on https://MEDFormat.org.

I currently have 4 groups asking me about official BIDS compatibility. 

MED data can be read in open source multiple tools (Matlab, FieldTrip, EEGLab, Python , Neo).  

These tools are freely available in compiled form for all major platforms (MacOS, Linux, Windows) as well as having source code and specification documents available on https://github.com/MEDFormat.  

There are also multiple free data converters (MEF, EDF, DAT, Neuralynx (NCS & NRD), & NATUS) available.

I would be happy to make the pull request for this.

@bids-standard/raw-electrophys-ieeg

MEDFormat:
Added MED format

- closes #1931

### DIFF
--- a/src/modality-specific-files/intracranial-electroencephalography.md
+++ b/src/modality-specific-files/intracranial-electroencephalography.md
@@ -38,6 +38,7 @@ stored in one of the following formats:
 | [EEGLAB](https://sccn.ucsd.edu/eeglab/index.php)                          | `.set`, `.fdt`           | The format used by the MATLAB toolbox [EEGLAB](https://sccn.ucsd.edu/eeglab/index.php). Each recording consists of a `.set` file with an OPTIONAL `.fdt` file.             |
 | [Neurodata Without Borders](https://nwb-schema.readthedocs.io/en/latest/) | `.nwb`                   | Each recording consists of a single `.nwb` file.                                                                                                                           |
 | [MEF3](https://osf.io/e3sf9/)                                             | `.mefd`                  | Each recording consists of a `.mefd` directory.                                                                                                                            |
+| [MED](https://medformat.org/)                                             | `.medd`                  | Each recording consists of a `.medd` directory.                                                                                                                            |
 
 It is RECOMMENDED to use the European data format, or the BrainVision data
 format. It is furthermore discouraged to use the other accepted formats over
--- a/src/schema/objects/extensions.yaml
+++ b/src/schema/objects/extensions.yaml
@@ -153,6 +153,14 @@ md:
   display_name: Markdown
   description: |
     A Markdown file.
+medd:
+  value: .medd/
+  display_name: Multiscale Electrophysiology Data Format
+  description: |
+    A directory in the [MED](https://medformat.org/) format.
+    Successor to the MEF 3.0 format.
+
+    Each recording consists of a `.medd` directory.
 mefd:
   value: .mefd/
   display_name: Multiscale Electrophysiology File Format Version 3.0
--- a/src/schema/rules/files/raw/ieeg.yaml
+++ b/src/schema/rules/files/raw/ieeg.yaml
@@ -3,6 +3,7 @@ ieeg:
   suffixes:
     - ieeg
   extensions:
+    - .medd/
     - .mefd/
     - .json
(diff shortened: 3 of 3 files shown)


effigies (approved):

Remi-Gau (commented):
This should not be considered a blocker but I would REALLY like to see an example dataset with this type of data added to the [BIDS examples repository](https://github.com/bids-standard/bids-examples).

Remi-Gau:
> This should not be considered a blocker but I would REALLY like to see an example dataset with this type of data added to the [BIDS examples repository](https://github.com/bids-standard/bids-examples).

Will ping all the people who reacted on the associated issue to see if anyone is up for this.

@dorahermes @cimbi @EmmanuelBarbeau @michaelpereira73 @KLehongre @tjr1

MEDFormat:
That's no problem.  I just need permission.

Cheers, Matt

￼

Matt Stead
***@***.***

Remi-Gau:
> I just need permission

Note that the example datasets are 'emptied' of any data: https://github.com/bids-standard/bids-examples/blob/master/CONTRIBUTING.md#why-do-we-only-host-truncated-data-with-0kb-size

This should also make it easier to 'share' them, though we prefer when possible to make sure that the full 'source' dataset we are using is also openly accessible, but this is not a requirement.

MEDFormat:
It is not possible to have a MED dataset with zero kb size (how is this possible at all in any format?).  
I can create a dataset with no samples, but it will still have a finite size (headers).  Is this what you would like?
I have small datasets that contain no protected info, which would be far more useful, if you prefer.

Cheers, Matt

￼

Matt Stead
***@***.***

Remi-Gau:
I meant the 'large' datafiles in the example datasets are truncated:

see here for how we do it.

https://github.com/bids-standard/bids-examples/blob/master/CONTRIBUTING.md#how-to-truncate-data-files-to-0kb

I guess header only files must be pretty light, so I think this could do as well. @effigies what do you think?

MEDFormat:
I think header only would not be too useful to new users.  Small anonymous data set seems like a better choice to me.


Cheers, Matt

￼

Matt Stead
***@***.***

effigies:
BIDS examples are strictly about directory structure, file naming and BIDS-specified metadata. This allows us to validate the examples to exercise the rules.

If we intend to duplicate format-internal metadata in BIDS metadata and validate that they are the same, it's worth including enough header information to check that. We have one example with NIfTI files truncated to the header to do this, but otherwise the NIfTI files are truncated to 0 bytes and the header checks skipped.

This is not to say that there shouldn't be a place with a working MED example, just that that's not what bids-examples is for. What's ideal is if there is a canonical example, we can strip it down with a script for inclusion in bids-examples and link back to the original.

MEDFormat:
OK I’ll make one with no data.  Not sure all tools (Python, Matlab, etc.) will work with it - never tested that situation.

This will take a little work because I’ll have to write some code to make a session like that.  Probably by tomorrow, if that’s OK.


Cheers, Matt

￼

Matt Stead
***@***.***

MEDFormat:
Hi Chris,

I received this (forwarded below) yesterday after uploading the test data set.

All I did to makes the other changes was copy the MEF3 entries in the other files & change them to MED versions.  This did not include all MED directories & file types, as MEF3 did not.  Is that what is causing this?  If so, should I add those?


Cheers, Matt

￼

Matt Stead
***@***.***

Remi-Gau:
> @Remi-Gau Are we good to go with [bids-standard/bids-examples#469](https://github.com/bids-standard/bids-examples/pull/469)?

yes it should be fine: I have some comments but they are minor and not related to the format so cna be tackled in follow up PR.

Remi-Gau (approved):

PR #1961, opened 2024-10-17
[SCHEMA] Allow physio files for anat datatype

rwblair:
Issue outlined in bids-standard/bids-validator#2164

This would fail to cover echo, part, and chunk entities which are legal for anatomical images and so too would be legal in the physio filenames.

### DIFF
--- a/src/schema/rules/files/raw/task.yaml
+++ b/src/schema/rules/files/raw/task.yaml
@@ -46,6 +46,7 @@ timeseries__mri_no_task:
     - .tsv.gz
     - .json
   datatypes:
+    - anat
     - dwi
     - perf
   entities:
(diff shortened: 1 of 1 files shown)


src/schema/rules/files/raw/task.yaml
```diff
@@ -101,6 +102,24 @@ events__mrs:
...
+timeseries__anat:
+  $ref: rules.files.raw.task.timeseries
+  datatypes:
+    - anat
+  entities:
+    $ref: rules.files.raw.task.timeseries.entities
+    subject: required
+    session: optional
+    task: optional
+    acquisition: optional
```
effigies, line 114: Duplicate

```suggestion
```

src/schema/rules/files/raw/task.yaml
```diff
@@ -101,6 +102,24 @@ events__mrs:
...
+    ceagent: optional
+    reconstruction: optional
+    run: optional
```
effigies, line 117:
```suggestion
```

src/schema/rules/files/raw/task.yaml
```diff
@@ -46,6 +46,7 @@ timeseries__mri_no_task:
     - .tsv.gz
     - .json
   datatypes:
+    - anat
```
effigies, line 49: I don't think `anat` has the `direction` entity, so I think we need to remove this.

```suggestion
```

src/schema/rules/files/raw/task.yaml
```diff
@@ -101,6 +101,19 @@ events__mrs:
...
+  entities:
+    $ref: rules.files.raw.task.timeseries.entities
+    ceagent: optional
+    reconstruction: optional
+    echo: optional
+    part: optional
+    chunk: optional
+
+
```
effigies, line 116: Style. Will commit it myself.
```suggestion
```

effigies:
Implements a change agreed upon in #513.

Remi-Gau (approved):
Good with me.

PR #1962, opened 2024-10-17
[ENH] Formalize presence of optional logs/ folder

yarikoptic:
It is quite often desired to store logs, e.g. simply stdout/stderr from invocation of BIDS converters or other tools which were used to produce/change content in current BIDS datasset.  They quite often provide ultimate provenance information to troubleshoot odd or incorrect results.  But researchers, in my personal opinion, undervalue logs!  But hinting them on "standard" location for them, I think we could inspire more of relevant to dataset provenance metadata being collected and shared.  

TODOs:
- ~~?Danger: might leak sensitive metadata, so may be wording should be adjusted to mention that they should be inspected/sensored? WDYT?~~ I think it is fine/assumed

Shameless plug: consider using con-duct (https://github.com/con/duct) for your "logging needs".

Note that I think that ideally the `logs/` should not include "derivative" data, such as e.g. output of `bids-validator`.  Those outputs should go under `derivatives/bids-validator-{version}/output.json` or alike.  (It might be the only reasonable "derivative" to recommend bundling with any "raw" or derived BIDS dataset).

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -387,7 +387,7 @@ Derivatives can be stored/distributed in two ways:
     that were used to generate the derivatives.
     Likewise, any code used to generate the derivatives from the source data
     MAY be included in the `code/` subdirectory.
-    Logs from running the code or other commands could be stored under `logs/` subdirectory.
+    Logs from running the code or other commands MAY be stored under `logs/` subdirectory.
 
     Example of a derivative dataset including the raw dataset as source:
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -387,6 +387,7 @@ Derivatives can be stored/distributed in two ways:
     that were used to generate the derivatives.
     Likewise, any code used to generate the derivatives from the source data
     MAY be included in the `code/` subdirectory.
+    Logs from running the code or other commands could be stored under `logs/` subdirectory.
 
     Example of a derivative dataset including the raw dataset as source:
 
--- a/src/schema/rules/directories.yaml
+++ b/src/schema/rules/directories.yaml
@@ -21,6 +21,7 @@ raw:
     subdirs:
       - code
       - derivatives
+      - logs
       - phenotype
       - sourcedata
       - stimuli
@@ -31,6 +32,10 @@ raw:
     opaque: true
   derivatives:
     name: derivatives
+    level: optional
+    opaque: true
+  logs:
+    name: logs
     level: optional
     opaque: true
   phenotype:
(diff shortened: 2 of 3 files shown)


src/common-principles.md
```diff
@@ -387,6 +387,7 @@ Derivatives can be stored/distributed in two ways:
     that were used to generate the derivatives.
     Likewise, any code used to generate the derivatives from the source data
     MAY be included in the `code/` subdirectory.
+    Logs from running the code or other commands could be stored under `logs/` subdirectory.
```
effigies, line 390:
```suggestion
    Logs from running the code or other commands MAY be stored under `logs/` subdirectory.
```

effigies (approved):
I'm fine with this. Up to now I probably would put them under `code/` but that's not really the right place.

This is, IMO low-priority to standardize, since the logs themselves are not standardized, so this doesn't make them substantially more legible to tooling. But it doesn't harm anything that I can see and it's unlikely to have another use.

yarikoptic:
ping @Remi-Gau @ericearl , WDYT?  I am seeking for more approvals ;)

ericearl (approved):
I read and understood the `src/common-principles.md` changes and they lgtm. I assume the rest of the changes to the schema which look probably right are fine.

One non-blocking question though: If they are logs from running code, why wouldn't you want the `logs/` subdirectory under the `code/` subdirectory? Imagine a dataset with no `code/` subdirectory, but a `logs/` subdirectory is present. It leaves the end-user wondering a bit what's going on. I personally usually put my logs from code into the `code/subdirectory` as-is.

yarikoptic:
I think it is valuable to `logs/` from `code/` for a number of reasons:

- `logs/` would primarily be machine generated (although could be some logs from "operator"), whenever `code/` is human (well, might be AI ;)) produced
- `logs/` might be produced by commands which are not under `code/` , e.g. straight invocations of bet, or bids-validator or anything else.
- could be `.gitignored`
  - I would cherish `code/` content and likely ignore `logs/` until I need to investigate. 
- could be made into git submodule and thus not part of the dataset for some reasons (size, leak of sensitive information, etc)
-  I, as a linux user with logs going under `/var/log` and some `~/.cache` while code under `/usr/bin` etc,  personally would not expect logs be mixed into folder with tools

yarikoptic:
We have 2 approvals. Need one more or could potentially proceed?

effigies:
Rules are 5 days and 2 approvals. Just fell off our radar.

PR #1963, opened 2024-10-18
Add a few more tests for the schema

yarikoptic:
- "explicit better than implicit" especially in case of tests

when test fails - very hard to debug. hopefully more of checks would help

### DIFF
--- a/tools/schemacode/bidsschematools/data/tests/test_rules.py
+++ b/tools/schemacode/bidsschematools/data/tests/test_rules.py
@@ -64,6 +64,7 @@ def test_rule_objects(schema_obj):
                 is_list = False
 
             for i_use, use in enumerate(instance):
+                assert isinstance(use, str)
                 if use == "derivatives":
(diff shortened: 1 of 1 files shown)


yarikoptic:
ping @effigies and @tsalo on this trivialish PR

tools/schemacode/bidsschematools/data/tests/test_rules.py
```diff
@@ -81,9 +82,21 @@ def test_rule_objects(schema_obj):
...
+                    "enums",
+                    "files",
+                    "formats",
+                    "metadata",
+                    "modalities",
+                ]:
                     # But other object types are referenced via their keys
                     object_values = list(schema_obj["objects"][object_type].keys())
+                else:
+                    raise AssertionError(f"Object type {object_type} not implemented.")
```
effigies, line 99: This makes more sense to me to put up right under `for object_type in schema_obj.objects:` rather than checking repeatedly.

PR #1966, opened 2024-10-21
[ENH] Add interpretation of "n/a" onsets in events.tsv files

### LINKED ISSUE #1943: Should "n/a" be allowed in the `onset` column of `events.tsv`
The current description of `onset` in [Task events](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/task-events.html) says that the onset column may have 0, positive and negative values.  It does not explicitly say it does not allow "n/a", although I believe that somewhere in the past that was in the spec.   The current description of `duration` explicitly says that it can be "n/a" and what that means.

My question is --- is "n/a" really allowed in the `onset` column.  If so, what does it mean and how are tools required to handle it?

This should be made explicit in the specification and in the schema.

This issue is related to issue #1938 and to issue hed-standard/hed-python#1026.

Remi-Gau:
- fix #1943

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -339,7 +339,7 @@ onset:
     Onset (in seconds) of the event, measured from the beginning of the acquisition
     of the first data point stored in the corresponding task data file.
     Negative onsets are allowed, to account for events that occur prior to the first
-    stored data point.
+    stored data point. A `n/a` value indicates the onset time is unknown or unavailable.
     For example, in case there is an in-scanner training phase that begins before
     the scanning sequence has started events from this sequence should have
     negative onset time counting down to the beginning of the acquisition of the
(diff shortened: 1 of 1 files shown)


Remi-Gau:
@VisLab

src/schema/objects/columns.yaml
```diff
@@ -339,7 +339,7 @@ onset:
     Onset (in seconds) of the event, measured from the beginning of the acquisition
     of the first data point stored in the corresponding task data file.
     Negative onsets are allowed, to account for events that occur prior to the first
-    stored data point.
+    stored data point. A `n/a` value indicates the onset time is unknown or unavailable.
```
effigies, line 342: I would move this to the bottom to avoid breaking up discussion of negatives. And maybe "A value of `n/a`..." to get around the a/an weirdness with consonants like n (I would use "an n/a").

Remi-Gau: moved actually at the end of the whole section

VisLab: Thanks for addressing this.  I think `n/a` is fine.  The description for `onset` now seems to be parallel to the description for `duration`.

effigies (approved):

yarikoptic (approved):

Issue #1906, opened 2024-08-22
[BUG] Inconsistent definition of AcquisitionDuration


effigies:
The current definition of [`AcquistionDuration`](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.metadata.AcquisitionDuration) is 

> Duration (in seconds) of volume acquisition. Corresponds to DICOM Tag 0018, 9073 Acquisition Duration. This field is mutually exclusive with "RepetitionTime".

The definition of [DICOM tag 0018, 9073](https://dicom.innolitics.com/ciods/enhanced-mr-image/enhanced-mr-image/00189073) is:

> The time in seconds needed to run the prescribed pulse sequence.

They also helpfully provide a graphic to clarify the definition:

![](http://dicom.nema.org/medical/dicom/current/output/html/figures/PS3.3_C.7.6.16-2.svg)

So this is the duration of the full scan, which only overlaps for single-volume acquisitions. Our definition more closely aligns with 
`FrameAcqusitionDuration` [0018,9220](https://dicom.innolitics.com/ciods/enhanced-us-volume/enhanced-us-volume-multi-frame-functional-groups/52009230/00209111/00189220):

>  The actual amount of time [in milliseconds] that was used to acquire data for this frame.

---

I propose the following (re)definitions:

> `AcquisitionDuration`: Duration, in seconds, of scan acquisition, including all volumes for multi-volume scans. Corresponds to DICOM Tag 0018, 9073 Acquisition Duration.
> `FrameAcquisitionDuration`: Duration, in seconds, of volume acquisition. Corresponds to DICOM Tag 0018,9220 Frame Acquisition Duration.

This change needs to be communicated somehow. While upstream tools are going to refer to DICOM, downstream tools may only refer to BIDS. Any suggestions for wording?

Lestropie:
Even the DICOM definitions might not be ideal:

-   "Time in seconds needed to run the prescribed pulse sequence" is very MR-centric
-   "The time in seconds needed to complete the acquisition of data" (found from [same DICOM tag but for CT](https://dicom.innolitics.com/ciods/enhanced-ct-image/enhanced-ct-image/00189073)) might be misconstrued as commencing on the first TR where data corresponding to the output DICOM series are "acquired", ie. excluding establishing steady-state / calibration volumes

[...]

PR #1969, opened 2024-10-24
fix(checks): Enforce timing mutual exclusions on BOLD data only

effigies:
Begins to address #1906 by at least containing the metadata complaints to BOLD files, which were the data that were intended to be checked. @tsalo Should this be expanded to include ASL as well?


tsalo:
It seems like it could apply to ASL, but TBH I don't know for sure. I say we include ASL and change it if an ASL user raises a valid concern.

tsalo (approved):

PR #1972, opened 2024-10-29
Enh project datasettype

yarikoptic:
This PR was initially submitted as #1861 but I made a mistake to combine it with a discussion of transformations of existing projects' layouts into such BIDS project dataset. Please refer to that PR for examples but otherwise let's concentrate here on the discussion of this specific proposed change.

- Rationale 1 (major): BIDS standard already provides reasonable structure to formalize organization of various components of a neuroscientific data project: where to place code, original (source) data, derivative data, README, CHANGES.  Many projects (e.g. nipoppy, YODA, etc) propose similar and often might be even "inspired" templates . If we explicitly allow for BIDS standard to prescribe study level organization, IMHO it would help many people and projects decide on how to organize their studies/projects. 
   - See #1861 on some examples and initial discussions on possible "transformations". But this PR has merit *regardless* whether individual existing projects adopt it partially or fully. 
- Rationale 2:  IMHO BIDS standard should describe only what standard prescribe and not recommend some potential "non-standardized" layouts.  That is why I "reworked" that example into a legitimate BIDS dataset merely by adding `dataset_description.json`.

TODOs:

- [x] provided more relation to existing approaches (mostly by @snastase in #1861)
- [x] crafted example for bids-examples : https://github.com/bids-standard/bids-examples/pull/451
  - [x] ensure bids-validator with modified schema passes its validation


yarikoptic:
@effigies I wonder if we should extend `DatasetType` to become a list or replace or compliment with `DatasetTypes` which could then indeed be just "project" when containing other derivative bids-datasets or sourcedata, or `["raw", "project"]` when also a raw and project, or could be also `["derivative", "project"]` when derivative but points to `sourcedata/` original "raw" ... ?

effigies:
I think this overall needs more specification. What are valid directories in a `project`-type dataset? We should add them to https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/directories.yaml.

I think a project dataset is barely worth specifying if we don't validate at least the raw data subdataset. Possibly we should have rules for indicating where validators should look for subdatasets.

In [OpenNeuroDerivatives](https://github.com/OpenNeuroDerivatives), we use `sourcedata/*` as a mixture of BIDS and non-BIDS (FreeSurfer) inputs. [Nipoppy](https://nipoppy.readthedocs.io/en/latest/) uses `derivatives/<pipeline>/<version>/<output>` as derivative datasets; presumably this could also be a mixture of BIDS and non-BIDS.

yarikoptic:
yet to "process" but a quick side idea inspired by #1928  --- I wonder if there is a hierarchy here: project (everything common) -> raw (current default, requires having sub- folder(s)) -> derivative (more stuff could be added), as every next level adds capabilities but includes all of the prior one as derivative could include raw in it?  or we have already something which invalidates that?

yarikoptic:
@effigies Following your idea, I have now added a "warning" (to reflect level of the analogous SubjectFolders check in "raw" BIDS). I guess, in principle, we could take this as an opportunity to revert `SubjectFolders` check back to "error" and demand making it "project" type, and then make it also "error" in project type. WDYT?

yarikoptic:
While discussing with @jbpoline we wondered, if may be `study` would be a better descriptor to use here in favor of `project`.  One of the rationales, is that e.g. in [BEP035](https://bids.neuroimaging.io/extensions/beps/bep_035.html) (attn @bids-standard/bep035) on Mega-analysis they introduce `study-` entity as a groupping element.  It kinda then would match natively.

<details>
<summary>we also  mention "study" in various places in BIDS which seems to align nicely here</summary> 

```shell
❯ git grep study
src/CHANGES.md:-   \[FIX] update physio bids name in longitudinal study page examples [#863](https://github.com/bids-standard/bids-specification/pull/863) ([Remi-Gau](https://github.com/Remi-Gau))
src/appendices/coordinate-systems.md:The following template identifiers are RECOMMENDED for individual- and study-specific reference
src/appendices/coordinate-systems.md:In the case of multiple study templates, additional names may need to be defined.
src/appendices/coordinate-systems.md:| study                 | Custom space defined using a group/study-specific template. This coordinate system requires specifying an additional file to be fully defined.                                                                                                                         |
src/appendices/hed.md:numerical values that are similar across the recordings in the study.
src/appendices/hed.md:repository on GitHub should be used to validate the study event annotations.
src/common-principles.md:    unless when appropriate given the study goals, for example, when scanning babies.
src/introduction.md:> The data used in the study were organized using the
src/modality-specific-files/genetic-descriptor.md:     "Dataset": "https://www.ncbi.nlm.nih.gov/projects/gap/cgi-bin/study.cgi?study_id=phs001364.v1.p1",
src/modality-specific-files/intracranial-electroencephalography.md:Note that the date and time information SHOULD be stored in the study key file
src/modality-specific-files/magnetic-resonance-spectroscopy.md:acquisition parameters in filenames is helpful or necessary to distinguish datasets in a given study.
src/modality-specific-files/motion.md:Note that the onsets of the recordings SHOULD be stored in the study key file [(`scans.tsv`)](../modality-agnostic-files.md#scans-file).
src/modality-specific-files/positron-emission-tomography.md:This entity is OPTIONAL if only one tracer is used in the study,
src/modality-specific-files/task-events.md:Please mind that this does not imply that only so called "event related" study designs
src/schema/objects/common_principles.yaml:    A set of neuroimaging and behavioral data acquired for a purpose of a particular study.
src/schema/objects/common_principles.yaml:    Session can (but doesn't have to) be synonymous to a visit in a longitudinal study.
src/schema/objects/common_principles.yaml:    A person or animal participating in the study.
src/schema/objects/entities.yaml:    For example, this should be used when a study includes two T1w images -
src/schema/objects/entities.yaml:    Session can (but doesn't have to) be synonymous to a visit in a longitudinal study.
src/schema/objects/entities.yaml:    A person or animal participating in the study.
src/schema/objects/enums.yaml:study:
src/schema/objects/enums.yaml:  value: study
src/schema/objects/enums.yaml:  display_name: study
src/schema/objects/enums.yaml:    Custom space defined using a group/study-specific template.
src/schema/objects/metadata.yaml:    Reference to the study/studies on which the implementation is based.
src/schema/objects/metadata.yaml:    The version of the HED schema used to validate HED tags for study.
tools/schemacode/src/bidsschematools/tests/data/broken_dataset_description.json:"EthicsApprovals": ["The original study from which this BIDS example dataset was derived was approved by the Ethics committee of Ghent University Hospital with identifier EC 2017/1103."]

```
</details>

and "project" mentionings are not particularly  aligned.  So, I think, we should just make it a "study", hence renaming accordingly.

src/schema/objects/metadata.yaml
```diff
@@ -583,6 +583,7 @@ DatasetType:
   enum:
     - $ref: objects.enums.raw.value
     - $ref: objects.enums.derivative.value
+    - $ref: objects.enums.study__datasetype.value
```
effigies:
```suggestion
    - $ref: objects.enums.study__datasettype.value
```

src/schema/rules/checks/dataset.yaml
```diff
@@ -6,13 +6,26 @@ SubjectFolders:
   issue:
     code: SUBJECT_FOLDERS
     message: |
-      There are no subject directories (labeled "sub-*") in the root of this dataset.
+      There are no subject directories (labeled "sub-*") in the root of this raw BIDS dataset.
     level: warning
   selectors:
     - path == '/dataset_description.json'
+    - dataset.dataset_description.DatasetType == "raw"
```
effigies: This drops the warning for derivatives. Intentional?

yarikoptic: I think it was but you made me doubt it since it changes behavior for derivative datasets. Regardless whether it is a right thing to do, I think it is an orthogonal issue so I would need to make it for any dataset which is not a "pure" study one. Pushed a fix with similar reasoning in commit msg.

effigies (approved):
Works for me.

effigies:
A functional question for the validator: Should we error if there are no valid BIDS datasets in `sourcedata/`?

cc @rwblair @nellh

yarikoptic:
@effigies who, among maintainers, do you think  might also be interested to review this PR?

effigies:
But what are you validating if there are no valid subdatasets? Why are you running the validator?

yarikoptic:
@effigies could you invite or nominate a few other PR reviewers who might potentially be interested? ;-)

src/schema/objects/enums.yaml
```diff
@@ -360,14 +360,19 @@ individual:
...
   description: |
     Custom space defined using a group/study-specific template.
     This coordinate system requires specifying an additional file to be fully defined.
 
     In order for this space to be interpretable, `SpatialReference` metadata MUST be provided.
+study__datasettype:
+  value: study
+  display_name: study
+  description: |
+    A study BIDS dataset.
```
kabilar:
```suggestion
    A study BIDS dataset to organize the data for the entire study (original source data, raw BIDS, derivatives) as a valid BIDS dataset.
```

yarikoptic: eh, greedy me saw this as a potential for "common denominator" as e.g. even an empty BIDS dataset before any subject data is collect could be a valid "study" one and then re-assigned to "raw".  But you are right, we better make it cleaner and following the example in the text aim to cover the entire study (only) in this description instead of staying vague. Let's do it, thanks

julia-pfarr (approved):
sounds good to me for a start!

PR #1974, opened 2024-11-01
[ENH] Update AcquisitionDuration definition to match DICOM, define FrameAcquisitionDuration for sparse sequences

### LINKED ISSUE #1906: [BUG] Inconsistent definition of AcquisitionDuration

effigies:
This PR brings the AcquisitionDuration definition in line with DICOM and separates it from checks for sparse sequence descriptions. We already narrowed the AcquisitionDuration mutual exclusion rules to only apply to BOLD and ASL data.

This defines `FrameAcquisitionDuration`, which corresponds to [DICOM tag 0018, 9220](https://dicom.innolitics.com/ciods/enhanced-mr-image/enhanced-mr-image/00189220).

@neurolabusc I wonder if this tag is used in any of your QA datasets yet?
@Lestropie Are you okay with deferring a CT-applicable definition to when CT data is accepted in BIDS? As noted in https://github.com/bids-standard/bids-specification/issues/1906#issuecomment-2436578444, we do have mechanisms in the schema to provide multiple definitions for terms that are used in multiple contexts.

Closes #1906.

----

I would *also* like to relax the mutual exclusion constraints that are indicated in [Task Imaging Data - Timing Parameters](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#timing-parameters_1):

image

Given that `RepetitionTime`, `SliceTiming` and `FrameAcquisitionDuration` could all be present in the DICOM metadata, I think a better check would be for consistency among these, instead of enforcing mutual exclusion. `DelayTime` and `VolumeTiming` cannot be derived from DICOM tags (I don't think), so I would be okay with continuing to validate mutual exclusion or to validate consistency.

`RepetitionTime` and `VolumeTiming` need to remain mutually exclusive, as they are distinct ways of declaring the onsets of each volume.

I am happy to enact this change in this PR or another, if people agree that it's worth doing.

### DIFF
--- a/src/modality-specific-files/magnetic-resonance-imaging-data.md
+++ b/src/modality-specific-files/magnetic-resonance-imaging-data.md
@@ -600,13 +600,13 @@ to be populated for functional sequences. Note that all these options can be
 used for non sparse sequences but that only options B, D and E are valid for
 sparse sequences.
 
-|          | **`RepetitionTime`** | **`SliceTiming`** | **`AcquisitionDuration`** | **`DelayTime`** | **`VolumeTiming`** |
-| -------- | -------------------- | ----------------- | ------------------------- | --------------- | ------------------ |
-| option A | \[ X ]               |                   | \[ ]                      |                 | \[ ]               |
-| option B | \[ ]                 | \[ X ]            |                           | \[ ]            | \[ X ]             |
-| option C | \[ ]                 |                   | \[ X ]                    | \[ ]            | \[ X ]             |
-| option D | \[ X ]               | \[ X ]            | \[ ]                      |                 | \[ ]               |
-| option E | \[ X ]               |                   | \[ ]                      | \[ X ]          | \[ ]               |
+|          | **`RepetitionTime`** | **`SliceTiming`** | **`FrameAcquisitionDuration`** | **`DelayTime`** | **`VolumeTiming`** |
+| -------- | -------------------- | ----------------- | ------------------------------ | --------------- | ------------------ |
+| option A | \[ X ]               |                   | \[ ]                           |                 | \[ ]               |
+| option B | \[ ]                 | \[ X ]            |                                | \[ ]            | \[ X ]             |
+| option C | \[ ]                 |                   | \[ X ]                         | \[ ]            | \[ X ]             |
+| option D | \[ X ]               | \[ X ]            | \[ ]                           |                 | \[ ]               |
+| option E | \[ X ]               |                   | \[ ]                           | \[ X ]          | \[ ]               |
 
 **Legend**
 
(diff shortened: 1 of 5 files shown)


Lestropie:
> Are you okay with deferring a CT-applicable definition to when CT data is accepted in BIDS?

I've no issue proceeding with a definition that's not perfect for CT if there is not yet CT support in BIDS. I only mentioned it in #1906 to show that the inconsistency in definition of this parameter is not exclusively between DICOM and BIDS, but even within DICOM itself.

neurolabusc:
I think this is fine. As the BIDS specifcation notes for [SliceTiming](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#timing-parameters) `Without this parameter slice time correction will not be possible`, so that provides more granularity than the other options. I do wonder if we should include a graphic or at lest a link to the DICOM [Figure C.7.6.16-2](https://dicom.nema.org/medical/dicom/current/output/html/part03.html#figure_C.7.6.16-2). Hopefully the fence post problem is clear with the existing definitions, e.g. the final slice time is the onset time for the final slice in a 3D volume relative to the onset time of the first slice in the volume. In contrast, Acquisition Duration spans the start of the first temporal slice to the completion of the final slice.

effigies:
This has two approvals and the critiques have gone silent. Will merge, and we can back it out if that motivates critics to speak up again.

PR #1981, opened 2024-11-11
[SCHEMA] harmonize into dirs, ids, ids_phenotype for "subjects" and "sessions"

yarikoptic:
Current names duplicate their "domain" (subject_ or session_), inconsistent in plurality (_id although for all ids), and not clear really what 'phenotype' corresponds to without reading the description.

With proposed change there is no duplication of the domain, consistency in plurality (albeit 'id' is an abbreviation so 'ids' is a non-word), and IMHO clearer meaning in `ids_phenotype`.

This would also allow for generalization across other entities in a perspective https://github.com/bids-standard/bids-2-devel/issues/54 - where then any entity with folders for its level could have `dirs`. Also it could come handy to determine `ids` for some other entities in tests.

Ref: https://github.com/bids-standard/bids-validator/pull/94#discussion_r1836876737

TODOs
- [ ] introduce corresponding changes to bids-validator
- [ ] since it is a breaking change, I will boost SCHEMA_VERSION to next MINOR version.

### DIFF
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -67,20 +67,20 @@ properties:
         description: 'Collections of subjects in dataset'
         type: object
         required:
-          - sub_dirs
+          - dirs
         additionalProperties: false
         properties:
-          sub_dirs:
+          dirs:
             description: 'Subjects as determined by sub-* directories'
             type: array
             items:
               type: string
-          participant_id:
+          ids:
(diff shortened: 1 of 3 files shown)


effigies:
Blocked on bids-standard/bids-validator#102.

effigies:
This will be impacted by #2107, which removes the phenotype summary fields.

PR #1987, opened 2024-11-15
chore(rtd): Speed up doc build with uv

effigies:
This PR is aimed at speeding up the build time, which currently clocks in at 2.5m. Let's see if we can't shave off at least 30s.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,12 +1,30 @@
 version: 2
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-lts-latest
   apt_packages:
     - jq
   tools:
-    python: "3.11"
+    python: latest
   jobs:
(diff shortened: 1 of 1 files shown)


effigies:
| | Pre-build | Build | Total |
|--|--|--|--|
| [Before](https://app.readthedocs.org/projects/bids-specification/builds/26278846/)| 88s | 60s | 148s |
| [After](https://app.readthedocs.org/api/v2/build/26292578.txt) | 26s | 72s | 98s |

So we can cut off about a minute (~65%) of pre-build time. Without investigating the range of build times, we're probably going from 2-3 minutes to 1-2 minutes, so a 30-50% saving.

Will leave it to other maintainers to decide if the complexity of the RTD config isn't justified. It's always possible to go back if this is found difficult to maintain, though.

effigies:
Nope.

PR #1990, opened 2024-11-22
[SCHEMA] Allow `acq-*` to relate physio data to MRI data in `anat/` as described in #2164

tstoeter:
The deno based validator did not accept files like `sub-01_acq-T1w_physio.tsv.gz` located in `anat/` and related to the T1w anatomy. This was, because a task was required, which was inherited from `rules.files.raw.task.timeseries.entities`. The key task is now overridden to optional. Please also see issue #2164 for more information and related patches.


tsalo:
I think [`mod`](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#mod) might make more sense for this.

tstoeter:
Thank you, @tsalo, for your suggestion, indeed from reading the specs `mod` seems more appropriate. I have tested it and received another error:
```
[ERROR] ENTITY_NOT_IN_RULE Entity not listed as required or optional for files with this suffix
		/sub-11/anat/sub-11_mod-T1w_physio.tsv.gz - mod not in rule rules.files.raw.task.timeseries__anat
```
I believe, this requires another modification of the schema to fix it.

effigies:
The task entity is already inherited via

https://github.com/bids-standard/bids-specification/blob/c81144bb23afb1ede6499c2ef751d3db711c3d4b/src/schema/rules/files/raw/task.yaml#L108-L109

PR #1991, opened 2024-11-22
[SCHEMA] Allow `mod-*` to relate physio data to MRI data in `anat/`

tstoeter:
`mod` seems to be more appropriate, but was also not yet supported.

### DIFF
--- a/src/schema/rules/files/raw/task.yaml
+++ b/src/schema/rules/files/raw/task.yaml
@@ -37,6 +37,7 @@ timeseries:
     acquisition: optional
     run: optional
     recording: optional
+    modality: optional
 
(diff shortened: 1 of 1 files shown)


src/schema/rules/files/raw/task.yaml
```diff
@@ -37,6 +37,7 @@ timeseries:
     acquisition: optional
     run: optional
     recording: optional
+    modality: optional
```
effigies, line 40: I think this needs to go on the `timeseries__anat` rule below, to achieve what you're trying to accomplish.

effigies (approved):

julia-pfarr (approved):

PR #1992, opened 2024-11-26
[ENH] Do explicitly "allow" for having dotfiles and explicitly ignore them

yarikoptic:
Up to this point it was up for a validator to "code" that up. E.g. current new deno based validator in
https://github.com/bids-standard/bids-validator/blob/main/src/files/ignore.ts has

    const defaultIgnores = [
      '.git**',
      '.*',
      'sourcedata/',
      ...

thus ignores all dotfiles from validation.  But this is inconsistent e.g.  with shipped in bids-specification schematools treatment, where files in a dotdirectory were reported as "invalid":

	❯ cat /tmp/input.txt | python -m bidsschematools pre-receive-hook
	.datalad/config
	.gitattributes
	sub-01/anat/sub-01_unknown.nii.gz

And as demonstrated above there could be legit "system" dotfiles (VCS related etc) which are part of the folder containing BIDS dataset, I think we should **explicitly** describe position of BIDS in relation to the dotfiles.  As it is pretty much "ignore", that is what this PR is intended to state.

In this PR I also adjusted `pre_receive_hook` to ignore dotfiles. The `find_files` was already ignoring them.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -202,6 +202,11 @@ as the labels would collide on a case-insensitive filesystem.
 Additionally, because the suffix `eeg` is defined,
 then the suffix `EEG` will not be added to future versions of the standard.
 
+### Dotfiles
+
+Files and directories starting with a dot (`.`) are reserved for system use and no valid recognized BIDS file or directory can start with a `.`.
+Any file or directory starting with a `.` present in a BIDS dataset is considered hidden and not subject to BIDS validation.
+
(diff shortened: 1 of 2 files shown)


effigies (approved):
I think this is reasonable.

Remi-Gau (approved):

effigies:
Do we want to consider a schema addition, or leave it as an instruction to implementations?

PR #1994, opened 2024-11-28
[SCHEMA] Add v1.10.0 to known BIDS versions list

jeffersoncasimir:
This PR adds BIDS version 1.10.0 to `src/schema/meta/versions.yaml` to address the following bug:

The `bids-validator` throws an `UNKNOWN_BIDS_VERSION` warning for datasets with `"BIDSVersion" : "1.10.0"`

### DIFF
--- a/src/schema/meta/versions.yaml
+++ b/src/schema/meta/versions.yaml
@@ -1,5 +1,6 @@
 ---
 # A list of released BIDS versions as of the next schema release
+- '1.10.0'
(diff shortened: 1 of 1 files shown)


effigies (approved):

effigies:
Yes. We need to add this to the [Release protocol](https://github.com/bids-standard/bids-specification/blob/master/Release_Protocol.md) so that this is part of the released schema.

Issue #1968, opened 2024-10-23
add field for timezone


JojoVh:
### Your idea

Dear BIDS developers

I would like to raise the issue that there is a need for timezone specification.
If the time indicator 'Z' for UTC time is lacking, the time zone is assumed to be local time. But this is not necessarily true when working internationally, and causes time zoning issues.

I would like to suggest adding this in the corresponding json file with the field "timezone", which SHOULD be specified.

{
	"acq_time":{
		"Description":"date and time of the scan acquisition in the format YYYY-MM-DDThh:mm:ss",
		"Units":"seconds"
		"Timezone":"CET"
	}
}

as specified in this list: https://www.utctime.net/time-zone-abbreviations
Would that be possible to implement?

I am referring to an adaptation of the current section:
https://bids-specification.readthedocs.io/en/stable/common-principles.html#units
Describing dates and timestamps:

Date time information MUST be expressed in the following format YYYY-MM-DDThh:mm:ss[.000000][Z] (year, month, day, hour (24h), minute, second, optional fractional seconds, and optional UTC time indicator). This is almost equivalent to the RFC3339 "date-time" format, with the exception that UTC indicator Z is optional and non-zero UTC offsets are not indicated. If Z is not indicated, time zone is always assumed to be the local time of the dataset viewer. No specific precision is required for fractional seconds, but the precision SHOULD be consistent across the dataset. For example 2009-06-15T13:45:30.

Remi-Gau:
So this would be a recommended field for the description acq_time in scans.json or sessions.json files?

effigies:
What is the use case for specifying time zone? The two use cases we considered are wanting absolute time coordinates and wanting to preserve time of day information. For the latter, the time zone is not generally relevant.

If we do add timezones, I would probably prefer just to use full RFC3339.

JojoVh:
Dear @Remi-Gau and @effigies 

Thank you very much for your enthusiasm and feedback.

I would recommend the field in the scans.json the most, because the timepoint of scanning matters most, but it makes sense to also recommend it in the sessions.json

The **use-case** for this is the following:
For a BIDS dataset recorded at 1 center, it is logical to us to keep the local time in the files, as it corresponds to time-sensitive data such as medication intake, artifact annotation, and the curation of the file with the people outside of research, such as our collaborators at the clinic, who are using local time.

[...]

PR #2001, opened 2024-12-10
[ENH] Accept UTC offsets in datetimes

effigies:
As discussed in #1968, non-UTC timezones can be useful for coordinating acquisitions across time zones. The spec currently adopts RFC3339 with the exception that nonzero UTC offsets are not permitted, and the `Z` offset is optional to allow indicating local time.

For backwards compatibility, I propose allowing the offsets to remain optional, but permitting the full range of RFC3339 offsets.

Separately, the current regular expression does not express the current rule, so I will open another PR against the maintenance branch for that.

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -1014,16 +1014,15 @@ For additional rules, see below:
 Describing dates and timestamps:
 
 -   Date time information MUST be expressed in the following format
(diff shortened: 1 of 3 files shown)


nellh (approved):
Verified the structure of the regular expression and this looks good to me.

christinerogers (approved):
lgtm, especially since @nellh verified the regex

PR #2002, opened 2024-12-10
fix: Update datetime definition/regex to match specification

effigies:
As noted in #1968 and https://github.com/bids-standard/bids-validator/issues/131, we fail to validate timestamps with the `Z`, but instead permit arbitrary 2-4 character upper-case strings. This goes against what the spec actually says.

This PR narrowly brings the schema into line with the text of BIDS 1.10.0. In #2001, I propose permitting the full range of RFC3339 values to handle needs for non-zero offsets.


tsalo (approved):
It looks like we were previously allowing any timezone code. The change seems to fix that.

PR #2005, opened 2024-12-12
[ENH] use snippets to link to definitions via tooltips

### LINKED ISSUE #954: [ENH] Extensive use of definition hyperlinks
Originally listed in Lestropie/bids-specification#1, but warrants its own listing.

The theory is that if there exists within the specification a strict list of definitions, with centralised precise definitions of terms, then any usages of those terms elsewhere in the specification could use a hyperlink back to the precise definition.

It would be preferable if there were some solution that would minimise the length of code necessary within the raw `.md` files every time such a link is created; i.e. not having to do a full-length relative-path hyperlink every time the word "subject" or "session" or "entity" is used.

(Even better would be e.g. a tooltip that would pop up when hovering the mouse over the term; but don't know if there exists a solution for such that would be portable across different rendering targets)

Remi-Gau:
- closes https://github.com/bids-standard/bids-specification/issues/954

this is me trying things out, see if this is a viable option

image

### DIFF
--- a/includes/abbreviations.md
+++ b/includes/abbreviations.md
@@ -1,5 +1,4 @@
-<!-- *[EEG]:  -->
-<!-- *[MEG]:  -->
+<!-- auto appended content  -->
 
 *[ASL]: Arterial Spin Labeling
 *[BEP]: BIDS Extension Proposal
--- a/includes/abbreviations.md
+++ b/includes/abbreviations.md
@@ -12,3 +12,5 @@
 *[TSV]: Tab Separated Values
 *[URI]: Uniform Resource Identifier
 *[URL]: Uniform Resource Locator
+*[subject]: A person or animal participating in the study. Used interchangeably with term **Participant**.
+*[task]: A set of structured activities performed by the participant (see [glossary](https://bids-specification.readthedocs.io/en/latest/glossary.html#task-common_principles))
--- a/includes/abbreviations.md
+++ b/includes/abbreviations.md
@@ -1,0 +1,15 @@
+<!-- *[EEG]:  -->
+<!-- *[MEG]:  -->
+
+*[ASL]: Arterial Spin Labeling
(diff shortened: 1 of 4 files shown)


Remi-Gau:
@tsalo @Lestropie 

- this seems to be a possible easy pure mkdocs solution to #954 using https://squidfunk.github.io/mkdocs-material/reference/tooltips/#tooltips

- in brief we can create a generic list of 'links' that will be appended to every page in the spec
- if those are prefixed with `*` then they are a tooltip

demo

https://www.dropbox.com/scl/fi/h1vmjihby032jdqismplg/Screencast-from-12-12-2024-15-45-05.webm?rlkey=rgm5qkjvlvtuj18e3fclchl11&dl=0

@effigies 
maybe also useful to have consistent set of links to external things across all pages: 

```bash
$ grep -rni https://www.json.org/json-en.html src

src/modality-agnostic-files.md:549:[object]: https://www.json.org/json-en.html
src/modality-specific-files/genetic-descriptor.md:137:[object]: https://www.json.org/json-en.html
src/modality-specific-files/motion.md:240:[object]: https://www.json.org/json-en.html
src/derivatives/common-data-types.md:343:[object]: https://www.json.org/json-en.html
src/schema/objects/metadata.yaml:3360:    [Object](https://www.json.org/json-en.html)
src/common-principles.md:481:    [JSON object](https://www.json.org/json-en.html) described below; and
src/common-principles.md:515:in the form of a JSON [object](https://www.json.org/json-en.html)
src/appendices/hed.md:58:The JSON sidecar provides a [JSON object](https://www.json.org/json-en.html) of annotations for these categorical values.
```

Though I wonder how this part would play with the PDF generation.

effigies:
IMO links are less useful in a PDF, so if this only affects the HTML spec, that seems fine.

PR #1794, opened 2024-04-18
[ENH] Render code blocks with language "tsv" as tables

effigies:
Rendered: https://bids-specification--1794.org.readthedocs.build/en/1794/

This is another go at https://github.com/bids-standard/bids-specification/issues/644 along the lines of https://github.com/bids-standard/bids-specification/pull/1783#issuecomment-2062890572.

~~It's not working 100% for me, but I've maxed out my time on this. Pushed to upstream so others can take it up or open PRs against it.~~

The code for this does not need to live in bidsschematools, but that's the quickest way to get something into the import namespace.

cc @Remi-Gau 

- Closes #644
- Closes #1756 
- Closes #1757
- Closes #1783

Issue #807, opened 2021-05-23
inconsistent style for the folder structure in the examples


Remi-Gau:
Most examples folder structure in the specs are of the form:

```
sub-01/
    ses-test/
        sub-01_ses-test_task-overtverbgeneration_bold.json
        sub-01_ses-test_task-overtverbgeneration_run-2_bold.json
        anat/
            sub-01_ses-test_T1w.nii.gz
        func/
            sub-01_ses-test_task-overtverbgeneration_run-1_bold.nii.gz
            sub-01_ses-test_task-overtverbgeneration_run-2_bold.nii.gz
```

But I realized that those in the qMRI page are more of the "tree" type:

```
└── sub-01/
     └── anat/
         ├── sub-01_inv-1_part-mag_MP2RAGE.nii.gz
         ├── sub-01_inv-1_part-phase_MP2RAGE.nii.gz
         ├── sub-01_inv-1_MP2RAGE.json
         ├── sub-01_inv-2_part-mag_MP2RAGE.nii.gz
         ├── sub-01_inv-2_part-phase_MP2RAGE.nii.gz
         └── sub-01_inv-2_MP2RAGE.json
```

This is a minor difference that it would be nice to standardize.

Even though it represents more work, I am tempted to make the "tree" version the default for the example.

The main reason is that it makes them even more visually distinct than the filename templates.

Any thoughts?

effigies:
As long as they look okay in HTML and PDF, I have no problem with the tree style. I think they're mostly going to be auto generated from schema, so the additional work should be minimal.

Remi-Gau:
wait... the examples are going to be auto-generated or the filename templates ? It is the latter, right?

then in this case we would just need to make sure that all the examples have a non-tree style.

sappelhoff:
I like the tree style pipes, but they break in the PDF, see https://github.com/bids-standard/bids-specification/issues/766

Although that issue does not seem to reduce legibility, so it may be fine to change all examples to use the tree style pipes.

In any case, +1 for consistency.

And I was not aware that we'd turn examples into schemas?

effigies:
I misread and thought we were talking templates...

tsalo:
This might be useful for making tree-like templates: https://realpython.com/directory-tree-generator-python/

Remi-Gau:
Thanks @tsalo 
Will have a look at that

Remi-Gau:
[...]

PR #2014, opened 2024-12-18
NF: tools/text2filetree.py helper to convert some indented text into our filetree example spec

yarikoptic:
See included within script tests

It was produced pretty much fully by just interacting with Claude . chatgpt got stuck on inability to run tests itself (although it did try!). 

Original demand came while working on

- #1705 

while transferring examples from google doc (yet to actually use the script on them fully ;) )

### DIFF
--- a/tools/text2filetree.py
+++ b/tools/text2filetree.py
@@ -1,0 +1,287 @@
+#!/usr/bin/env python3
+import argparse
+import json
+import sys
+
+
+def parse_file_tree(input_data, tab_width=4):
+    """
+    Parse a file tree hierarchy from an input stream and convert it to a nested dictionary.
+
+    :param input_data: Input string or file-like object representing the file tree
+    :param tab_width: Number of spaces to replace tabs with
+    :return: A nested dictionary representing the file tree
+    """
+    # If input is a string, convert to a list of lines
+    if isinstance(input_data, str):
+        lines = input_data.splitlines()
+    else:
+        lines = [line.rstrip() for line in input_data.readlines()]
+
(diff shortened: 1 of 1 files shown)


effigies:
It seems like a problem that we're round-tripping a text tree into a Python dict and back. It looks like (#807, #837) the goal was just to convert to tree-style and use it consistently.

I don't think the dict approach is very pleasant, and I assume that's why this tool was desired, so I wonder if it would be better to make a custom fence (a la #1794):

````markdown
```filetree
root
  subdir1
    file1.txt
    subsubdir
      file2.txt
  subdir2
    file3.txt
```
````

And convert that to:

```html
<pre>
root
├── subdir1
│   ├── file1.txt
│   └── subsubdir
│       └── file2.txt
└── subdir2
    └── file3.txt
</pre>
```

We could also handle creating a tree from a list of files, e.g.:

````markdown
```filetree
root/subdir1/file1.txt
root/subdir1/subsubdir/file2.txt
root/subdir2/file3.txt
```
````

yarikoptic:
good point.  I guess I did not mind python way in general because it was reminding me our [`@with_tree`](https://github.com/yarikoptic/datalad/blob/HEAD/datalad/core/local/tests/test_create.py#L426) tests decorator to provide sample cases. But there dictionary was really useful so that we could provide content for the files as well.  Here indeed there is clutter to a degree.  Your `filetree` fence example sounds really nice.  So overall it would indeed make more sense than having this helper... although some auto-formatting of inconsistent `filetree`s might be desired too.

PR #2030, opened 2025-01-16
[INFRA] Ignore www.incf.org link check.. unlikely to go away (hopefully)

yarikoptic:
Currently check fails

    URL        `https://www.incf.org/'
    Name       `INCF'
    Parent URL file:///home/circleci/project/site/introduction.html, line 1795, col 2
    Real URL   https://www.incf.org/
    Check time 0.084 seconds
    Result     Error: SSLError: HTTPSConnectionPool(host='www.incf.org', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certifica...

possibly due to outdated collection of root certificates in the env where testing?

dunno, but we better address it since it brings overall CI to red quite often, causing fatigue and increasing likeliness to skip important issues.

edit: also sorted lines there for consistency etc

PR #2032, opened 2025-01-21
[INFRA] Do check for being a pull request not at job level

yarikoptic:
Since this way it keeps it "grey" on PRs, which causes all other
matrix runs and workflows to be expanded and thus requiring more
mental cycles to understand that it is actually all good.

This way it should be collapsed.  This commit sits on top of
https://github.com/bids-standard/bids-specification/pull/2030
to ensure no other workflows fail ATMSince this way it keeps it "grey" on PRs, which causes all other
matrix runs and workflows to be expanded and thus requiring more
mental cycles to understand that it is actually all good.

This way it should be collapsed.  This commit sits on top of

- https://github.com/bids-standard/bids-specification/pull/2030

to ensure no other workflows fail ATM


yarikoptic:
with my luck I do get a link fail ATM and it seems legit site down

```

URL        `https://www.encodeproject.org/profiles/biosample_type'
Name       `ENCODE Biosample Type'
Parent URL file:///home/circleci/project/site/modality-agnostic-files.html, line 2210, col 31
Real URL   https://www.encodeproject.org/profiles/biosample_type
Check time 60.250 seconds
Result     Error: ReadTimeout: HTTPSConnectionPool(host='www.encodeproject.org', port=443): Read timed out. (read timeout=60)
```

filed
- https://github.com/ENCODE-DCC/encoded/issues/4441

DimitriPapadopoulos:
https://www.encodeproject.org/profiles/biosample_type appears to be back online.

effigies (approved):

effigies:
I've approved this on the grounds that it's not a big deal and it doesn't cause problems, but I also have not experienced the confusion that this addresses.

I would appreciate it if a second approval comes with a concurrence that this actually solves a problem.

yarikoptic:
FWIW, I extended original description  with an example/comparison .

effigies:
@yarikoptic This is what I'm seeing:

image

PR #2033, opened 2025-01-21
[FIX] Run codespell 2.4.0 throughout fixing few left typos automagically

yarikoptic:
CI started to error out codespell as I saw in #2032 and it is due to new codespell 2.4.0.

So this PR fixes those detected with codespell 2.4.0 and indeed we had inconsistencies on GB vs US forms for "localization" so it is harmonized now.

I have done `HEP` exclusion via regex which might slow things down but would keep it case specific

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,7 +1,7 @@
 [codespell]
 skip = *.js,*.svg,*.eps,.git,node_modules,env,venv,.mypy_cache,package-lock.json,CITATION.cff,tools/new_contributors.tsv,./tools/schemacode/docs/build,venvs
 ignore-regex = \bHEP\b
-ignore-words-list = fo,te,als,Acknowledgements,acknowledgements,weill,bu,winn,manuel
+ignore-words-list = acknowledgements,als,bu,fo,te,weill,winn
 builtin = clear,rare,en-GB_to_en-US
 # this overloads default dictionaries and I have not yet figured out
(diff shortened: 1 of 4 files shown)


.codespellrc
```diff
@@ -1,5 +1,6 @@
 [codespell]
-skip = *.js,*.svg,*.eps,.git,node_modules,env,venv,.mypy_cache,package-lock.json,CITATION.cff,tools/new_contributors.tsv,./tools/schemacode/docs/build
+skip = *.js,*.svg,*.eps,.git,node_modules,env,venv,.mypy_cache,package-lock.json,CITATION.cff,tools/new_contributors.tsv,./tools/schemacode/docs/build,venvs
+ignore-regex = \bHEP\b
 ignore-words-list = fo,te,als,Acknowledgements,acknowledgements,weill,bu,winn,manuel
```
DimitriPapadopoulos, line 4:
```suggestion
ignore-words-list = acknowledgements,als,bu,fo,hep,te,weill,winn
```

DimitriPapadopoulos (changes requested):
* Don't need both `acknowledgements` and `Acknowledgements`.
* Also `manuel` has been removed from the main dictionary (https://github.com/codespell-project/codespell/pull/3204).

DimitriPapadopoulos (approved):

effigies (approved):

Issue #914, opened 2021-10-28
Extra participants in phenotype TSVs, but not in data set folders should not be an ERROR in validation


ericearl:
## Introduction

In [this section about the modality agnostic phenotypic data](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#phenotypic-and-assessment-data), there is a sentence that I don't agree with (after preparing A LOT of phenotypic data):

> ... the entries of [the `participant_id`] column MUST correspond to the subjects in the BIDS dataset and `participants.tsv` file.

In our case there are more subjects in the `phenotype/` TSV data who were screened out (but can still be shared) before MRI or MEG data collection than there are MEG and MRI participant data combined.  And our `participants.tsv` file contained every subject across all `phenotype/` TSV data so I would think perhaps from the above sentence that we could be in compliance aside from the *"BIDS dataset and `participants.tsv` file"* part of that sentence.

I discovered this behavior by way of running the validator, but I don't think my proposal is a particularly difficult one.  I propose one of three changes.  They are listed in my order of preference, so I would prefer Option A over the others.

## Option A: Inclusive OR/Union

Change the sentence explicitly to accept a union (an inclusive OR) of both the BIDS dataset and the `participants.tsv` file:

> ... the entries of [the `participant_id`] column MUST correspond to the subjects in the BIDS dataset **unioned with (inclusive or) the** `participants.tsv` file.

This way `phenotype/` TSV data can correspond to either subjects in the `participants.tsv` or the BIDS dataset all-inclusive.

## Option B: Exclusive `participants.tsv`

Change the sentence to accept only the `participants.tsv` and ignore the BIDS dataset:

> ... the entries of [the `participant_id`] column MUST correspond to the subjects in the ~~BIDS dataset and~~ `participants.tsv` file.

This is a little stricter to really require that if `phenotype/` TSV data is prepared, then a valid `participants.tsv` is prepared as well.

## Option C: Soften requirement

Change the sentence to make it a soft and optional thing rather than a hard and rigid rule:

> ... the entries of [the `participant_id`] column **SHOULD** correspond to the subjects in the BIDS dataset and `participants.tsv` file.

I don't like this option because it is less prescriptive and could be more confusing to someone receiving a BIDS dataset, but it is an option.  This would imply `phenotype/` TSV data could be asynchronous with the BIDS dataset and `participants.tsv`, but the validator would only WARN about this being a problem instead of giving an ERROR.

## References

[...]

PR #2044, opened 2025-02-05
[ENH] Allow participants.tsv to contain a superset of subject directories and subjects listed in phenotype files

ericearl:
The `participants` description in `src/schema/objects/files.yaml` now contains the comprehensive superset rule from #914. This change allows phenotype-only participant_ids (participants not present in the `sub-XX` folders) to be included in the participants.tsv file. @effigies I believe has a plan to integrate this change into the next BIDS release for the BIDS schema validator.


src/schema/objects/files.yaml
```diff
@@ -72,6 +72,11 @@ participants:
     followed by a list of optional columns describing participants.
     Each participant MUST be described by one and only one row.
 
+    If this file exists and a `phenotype/` folder is present, then
+    the `participant_id` entries MUST be the union of all subject directories
+    and all `participant_id` entries found among phenotypic and assessment data
+    in the `phenotype/` folder.
```
effigies, line 78: I would get rid of the conditional. If it doesn't exist, any rules don't apply.

I would also just say superset. $A \supseteq B \iff (A \cap B = B)$

```suggestion
    The `participant_id` entries MUST be a superset of all subject directories
    and all `participant_id` entries found among phenotypic and assessment data
    in the `phenotype/` folder.
```

If it's strongly desired, we can implement unions, but we can do this with a minimal change right now.

ericearl:
Yes, that looks like it satisfies our need. Thanks for the suggestion @effigies!

effigies:
@rwblair We pre-load all phenotype files at the beginning of the run in order to populate `dataset.subjects.phenotype`. What if we dropped that and instead used the rule:

```yaml
RuleName:
  selectors:
    - datatype == 'phenotype'
    - extension == '.tsv'
  checks:
    - |
      allequal(
        sorted(intersects(dataset.subjects.participant_id, columns.participant_id)),
        sorted(columns.participant_id)
      )
```

I'm curious which one would be more inefficient:

1) Load all phenotype files, take the union, and run the intersection once. Then load all phenotype files when validating them. (current)
2) When validating each phenotype file, run the intersection with this file's column only. (proposal)

It would also be worth considering which one could be optimized under the hood. While it is simplest if the context continues to be serializable to a JSON object, we could consider set-like structures that make it more efficient to run `intersects()` when encountered.

yarikoptic:
note (please correct me if I am wrong): this rule ATM could only be stated in human language and cannot be operationalized  as a "schema rule" for the validator since placement of `participant_id`s within `phenotype/` is not formalized.

yarikoptic:
I crossed my prior note, but reflecting on the rule by @effigies above, do we already provide top level directory  `phenotype/` as "datatype" ?

<details>
<summary>ATM no rule mentions it as a datatype, here is the list/counts
</summary> 


```shell
❯ git grep -h 'datatype ==' | sed -e 's,^ *,,g' | sort | uniq -c | sort -n
      1 - datatype == 'fmap'
      2 - datatype == "beh"
      2 - datatype == "dwi"
      2 - datatype == "mrs"
      3 - datatype == "anat"
      6 - datatype == "motion"
      7 - datatype == "micr"
      9 - datatype == "fmap"
      9 - datatype == "func"
     13 - datatype == "ieeg"
     17 - datatype == "eeg"
     18 - datatype == "pet"
     20 - datatype == "perf"
     21 - datatype == "meg"
     24 - datatype == "nirs"
```

</details>

Would we similarly define `stimuli` and potentially other data types for other possible top level directories?

src/schema/objects/files.yaml
```diff
@@ -72,6 +72,10 @@ participants:
     followed by a list of optional columns describing participants.
     Each participant MUST be described by one and only one row.
 
+    The `participant_id` entries MUST be a superset of all subject directories
+    and all `participant_id` entries found among phenotypic and assessment data
+    in the `phenotype/` folder.
```
effigies, line 77:
```suggestion
    in the `phenotype/` directory.
```

effigies:
@ericearl I took a quick pass at updating the schema. Would you mind putting together a small example for bids-examples? Maybe one with `sub-01/` and `sub-02/` directories, phenotype data for `sub-01` and `sub-03`, and a `participants.tsv` that contains subs 1-4?

ericearl:
@effigies I made our draft PR ready for review over on bids-examples at https://github.com/bids-standard/bids-examples/pull/465. You'll want pheno004 for the example you're asking for.

ericearl:
@effigies What else needs to happen next to finish off this PR? I know there's got to be the two reviews that aren't done by you or I.

ericearl:
@effigies All 4 or just `pheno004`?

yarikoptic:
> My understanding was your preference was to classify stimuli as a new dataset type and validate its contents separately?

In the scope of ongoing https://bids.neuroimaging.io/bep044 -- nope. Just a folder with some formalization.

Could then become a BIDS 1.0 dataset with such `stimuli/` folder and that is it I guess ;)

In the scope of [BIDS 2.0's hopeful flexible layout](https://github.com/bids-standard/bids-2-devel/issues/54) -- could then just become without such a leading folder and potentially with `stim-*/` subfolders. (indeed `stimuli/` was what inspired that in https://github.com/bids-standard/bids-specification/issues/751 )

ericearl:
I added just the `pheno004/` example dataset to a fresh PR: https://github.com/bids-standard/bids-examples/pull/483.

effigies (approved):

nellh (approved):

Remi-Gau (approved):

effigies:
2 independent reviews and more than a week since substantive changes. Merging.

Issue #2045, opened 2025-02-06
columns.yml: unclear difference in schema in defining within record or within "definition" sub-record


yarikoptic:
In https://github.com/bids-standard/bids-specification/blob/HEAD/src/schema/objects/columns.yaml we have 

```yaml
duration:
  name: duration
  display_name: Event duration
  description: |
    Duration of the event (measured from onset) in seconds.
    Must always be either zero or positive (or `n/a` if unavailable).
    A "duration" value of zero implies that the delta function or event is so
    short as to be effectively modeled as an impulse.
  type: number 
  unit: s
  minimum: 0
```

where there is `type` and `unit` within the record but then there is 

```yaml
age:
  name: age
  display_name: Subject age
  description: |
    Numeric value in years (float or integer value).
  
    It is recommended to tag participant ages that are 89 or higher as 89+,
    for privacy purposes.
  definition: {
    "LongName": "Subject age",
    "Description": "Subject age in postnatal years",
    "Units": "year",
  }
```

where there is `definition` field with `"Units"` in it.  I found no mentioning of `definition:` within https://github.com/bids-standard/bids-specification/tree/master/src/schema README.

- is such difference warranted or should we seek to unify by either abolishing `definition` and moving all into the record or unifying into everything having `definition`?
  - why `unit` vs `Units` difference? very loosely related https://github.com/bids-standard/bids-2-devel/issues/15  
  - and overall why in "definition" attributes are Capitalized?
- why `definition` is in JSON syntax (subclass of YAML, I know) and not regular YAML?

If not to be changed anyhow, should be addressed by clarifications within https://github.com/bids-standard/bids-specification/blob/master/src/schema/README.md .

effigies:
https://github.com/bids-standard/bids-specification/pull/1838

effigies:
I agree we should update the README, and that should have been part of that PR.

effigies:
Reopening because this was *not* documented, even in the schema README, in #2162. For reference, in #2048 we had a detailed discussion and I will briefly recap the two kinds of column definitions at the bottom of this comment.

[...]

PR #2046, opened 2025-02-06
doc(schema): Add definition field for objects.columns entries

effigies:
Addresses #2045.

### DIFF
--- a/src/schema/README.md
+++ b/src/schema/README.md
@@ -399,6 +399,13 @@ These objects additionally have the field:
 | `type`   | The type (such as `string`, `integer`, `object`) of values the term describes                                                                                                                                   |
 | `format` | The format of the term (defined in `objects.formats`)                                                                                                                                                           |
 
+`objects.columns` additionally permit a `definition` field that may take any value permissible
+for the definition of a column in a JSON sidecar:
+
+| Field        | Description                                                                                                                                                                                                   |
+| ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `definition` | A JSON object that describes the column as in [Common principles - Tabular files][tabular files]. This is mutually exclusive with the `type` field, and is used for fields that have overridable definitions. |
+
 Value terms groups (`datatypes`, `suffixes`, `extensions`) define terms where a field
 can take on multiple values.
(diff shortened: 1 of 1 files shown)


src/schema/README.md
```diff
@@ -399,6 +399,13 @@ These objects additionally have the field:
...
 | `format` | The format of the term (defined in `objects.formats`)                                                                                                                                                           |
 
+`objects.columns` additionally permit a `definition` field that may take any value permissible
+for the definition of a column in a JSON sidecar:
+
+| Field        | Description                                                                                                                                                                                                   |
+| ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `definition` | A JSON object that describes the column as in [Common principles - Tabular files][tabular files]. This is mutually exclusive with the `type` field, and is used for fields that have overridable definitions. |
```
yarikoptic, line 407: not clear to me what "overridable definitions" -- overridable by a BIDS user to override Units etc?  so not all columns can have Units overridden?

effigies: Does reading #1838 clarify? I tried to be clear here, and I'm happy to improve the wording, but if we're not starting from the same baseline understanding, we're in for many rounds of back and forth.

yarikoptic: Thanks! It does provide more background indeed. I formulated separate
- https://github.com/bids-standard/bids-specification/issues/2048
for potential further "clarification" which should not be  a part of this PR which just to improve documentation. I think your proposed formulation is as good as it gets until we address #2048 one way or another and then either adjust schema or description here to use consistent terminology (e.g. "fully-defined", "conventional" or alike).

yarikoptic: btw -- how does it relate to

```
`*`            | JSON-schema fields to further constrain values                      |
```
?  is JSON-schema here the https://json-schema.org/ or the "JSON sidecar schema" or ... ?

yarikoptic (approved):
as it clarifies current state of affairs, already would address the entirely missing documentation on that

src/schema/README.md
effigies: Yes, we use actual JSON schema, and the JS validator uses the AJV package to perform validation. We have tried to adopt JSON schema as much as is practical.

PR #2050, opened 2025-02-12
[ENH] Define "phenotype" as a data type not nested inside subjects

### LINKED ISSUE #1828: Classify "phenotype/" as a datatype directory with no subject/session parent
`phenotype/` is a bit of an outlier in BIDS terms. Other folders at the top level are either entities (`sub-<label>`) or their contents are opaque to BIDS. `phenotype/`, on the other hand, is a collection of `.tsv`/`.json` files that are to be validated on the same terms as `participants.tsv`/`participants.json`.

I would suggest classifying `phenotype` as a datatype, distinct from other datatypes only in that it spans multiple subjects, and so the subject and session entities do not apply. [BEP036](https://docs.google.com/document/d/1WTkfES8L0vItZVyyR68fc-9cO03jS-kCnMnw6602pbc/edit) seems to go some way in a similar direction, permitting a `pheno/` datatype within subjects/sessions.

In the (unmerged) PR #1672, I suggest using phenotype as a datatype for the purposes of filename validation, and then carve out some exceptions that allow us to use it that way without it being an official datatype. If we make it a datatype, then the exception can be removed. That it fits with very little modification to the schema and validation (https://github.com/bids-standard/bids-validator/pull/1957), seems to me to be an argument for this classification.

The alternative, as I see it, is to consider phenotype a completely unique category of thing, and all implementations will need to have special code for handling it.

effigies:
Closes #1828. Please continue the conversation in that issue, and reserve this PR for discussing concrete changes.

### DIFF
--- a/src/schema/objects/common_principles.yaml
+++ b/src/schema/objects/common_principles.yaml
@@ -13,7 +13,6 @@ data_type:
   description: |
     A functional group of different types of data.
     Data files are contained in a directory named for the data type.
-    In raw datasets, the data type directory is nested inside subject and (optionally) session directories.
     BIDS defines the following data types:
 
         1.  `func` (task based and resting state functional MRI)
@@ -43,6 +42,11 @@ data_type:
         13. `motion` (motion)
 
         14. `mrs` (magnetic resonance spectroscopy)
+
+        15. `phenotype` (participant-level measurement and survey data)
+
+    For files that are subject- and session- specific, the data type directory
+    is nested inside those directories.
 
 dataset:
   display_name: Dataset
--- a/src/schema/objects/common_principles.yaml
+++ b/src/schema/objects/common_principles.yaml
@@ -43,10 +43,12 @@ data_type:
 
         14. `mrs` (magnetic resonance spectroscopy)
 
-        15. `phenotype` (participant-level measurement and survey data)
+        15. `phenotype` (measurement and survey data)
 
     For files that are subject- and session- specific, the data type directory
     is nested inside those directories.
+    `phenotype` data files are aggregated across subjects and sessions,
+    and so the `phenotype/` directory is placed in the dataset root.
 
 dataset:
   display_name: Dataset
--- a/tools/schemacode/src/bidsschematools/__main__.py
+++ b/tools/schemacode/src/bidsschematools/__main__.py
@@ -149,9 +149,6 @@ def pre_receive_hook(schema, input_, output):
         )
 
     regexes = [rule["regex"] for rule in all_rules]
-    # XXX Hack for phenotype files - this can be removed once we
-    # have a schema definition for them
-    regexes.append(r"phenotype/.*\.(tsv|json)")
 
     output = sys.stdout if output == "-" else open(output, "w")
 
--- a/src/schema/objects/datatypes.yaml
+++ b/src/schema/objects/datatypes.yaml
@@ -61,6 +61,13 @@ pet:
   display_name: Positron Emission Tomography
   description: |
     Positron emission tomography data
+phenotype:
+  value: phenotype
+  display_name: Phenotype
+  description: |
+    Participant level measurement data
+    (for example, responses from multiple questionnaires)
+    split into individual files separate from `participants.tsv`.
 nirs:
   value: nirs
   display_name: Near-Infrared Spectroscopy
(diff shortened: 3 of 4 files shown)


yarikoptic:
attn @bids-standard/bep036

src/schema/objects/common_principles.yaml
```diff
@@ -44,6 +43,11 @@ data_type:
 
         14. `mrs` (magnetic resonance spectroscopy)
 
+        15. `phenotype` (participant-level measurement and survey data)
+
+    For files that are subject- and session- specific, the data type directory
+    is nested inside those directories.
```
ericearl, line 49: Does this mean `phenotype/` is permitted, but not required under the `subject/[session/]` directory? BEP036 is leaning into the aggregated model of tabular phenotypic data file storage at the root-level `phenotype/` subdirectory.

effigies: To permit it, we would need a file rule where `subject` and `session` are optional, so this describes the current situation reasonably well, as it can't be subject- or session- specific.

I do not have a specific objection to permitting it to be used inside subject/session directories, but if BEP036 is not proposing it, it won't happen.

Happy to consider alternative wordings that make this clearer.

effigies:
@yarikoptic @ericearl @surchs I believe that this and the issue (#1828) have had enough time to garner negative comments, so I would like to proceed by consensus. I think this would be a useful addition for consistency within the schema and would appreciate a review.

ericearl (approved):
lgtm

src/schema/objects/common_principles.yaml
```diff
@@ -44,6 +43,11 @@ data_type:
 
         14. `mrs` (magnetic resonance spectroscopy)
 
+        15. `phenotype` (participant-level measurement and survey data)
```
yarikoptic, line 46: so it would be coming under a session folder if it is a dataset where there is a session folder for a subject, right?  so it would not be "participant-level" but "session-level"?  then I would just KISS:

```suggestion
        15. `phenotype` (measurement and survey data)
```

tools/schemacode/src/bidsschematools/__main__.py
```diff
@@ -149,9 +149,6 @@ def pre_receive_hook(schema, input_, output):
         )
 
     regexes = [rule["regex"] for rule in all_rules]
-    # XXX Hack for phenotype files - this can be removed once we
-    # have a schema definition for them
-    regexes.append(r"phenotype/.*\.(tsv|json)")
```
yarikoptic, line 154: and we do not need any extra rules because there is already
https://github.com/bids-standard/bids-specification/blob/HEAD/src/schema/rules/files/common/tables.yaml#L34
?

src/schema/objects/common_principles.yaml
effigies: No, the whole idea is that this doesn't change the meaning of phenotype, it slightly broadens the notion of datatype. This does not permit phenotype directories in sessions.

yarikoptic: but they are becoming (or already?) allowed at subject level so I could have `sub-1/phenotype/` in addition to e.g. `sub-1/ses-1`?  if not, then it is still located at "dataset level" (as in https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#phenotypic-and-assessment-data) and "participant-level" is confusing here... then

```suggestion
        15. `phenotype` (participant-level measurement and survey data at dataset level phenotype/)
```

or alike to clear it up?

tools/schemacode/src/bidsschematools/__main__.py
effigies: Correct

src/schema/objects/common_principles.yaml
```diff
@@ -44,6 +43,13 @@ data_type:
 
         14. `mrs` (magnetic resonance spectroscopy)
 
+        15. `phenotype` (measurement and survey data)
+
+    For files that are subject- and session- specific, the data type directory
+    is nested inside those directories.
+    `phenotype` data files are aggregated across subjects and sessions,
+    and so the `phenotype/` directory is placed in the dataset root.
```
effigies, line 51: @yarikoptic @ericearl Hopefully this is clearer.

rwblair (approved):

PR #2053, opened 2025-02-13
chore(schematools): Add configuration for bump-my-version

effigies:
It turns out that the tool [bump-my-version](https://callowayproject.github.io/bump-my-version/) can handle our tagging strategy for the bidsschematools. Here is a configuration that captures the workflow I've been doing by hand.

I did recently add `tool.bumpver`, but it couldn't handle tags due to the need for a `schema-` prefix.

I added the release process in a comment, but I'm happy to update a doc if there is one.

### DIFF
--- a/tools/schemacode/pyproject.toml
+++ b/tools/schemacode/pyproject.toml
@@ -102,6 +102,7 @@ markers = [
 ]
 minversion = "6.0"
 xfail_strict = true
+
 [tool.coverage]
 
 [tool.coverage.paths]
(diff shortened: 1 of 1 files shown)


tools/schemacode/pyproject.toml
```diff
@@ -102,6 +102,7 @@ markers = [
 ]
 minversion = "6.0"
 xfail_strict = true
+[tool.coverage]
```
effigies, line 105:
```suggestion

[tool.coverage]
```

PR #2054, opened 2025-02-13
[ENH] Allow additional columns if defined for various TSV files

### LINKED ISSUE #1993: TSV lookup tables for segmentation do not allow extra columns
See here:

https://bids-specification.readthedocs.io/en/latest/derivatives/imaging.html#common-image-derived-labels

Image

https://github.com/bids-standard/bids-specification/blob/6a40161ed258ca87fa48365f4f9098c84c134200/src/schema/rules/tabular_data/derivatives/common_derivatives.yaml#L2

Is there a reason for this?

For context I am trying to use this to help standardize the atlases in nilearn and adding some extra columns would be handy.

effigies:
Several table definitions were missing an indication of whether additional columns were allowed, which defaults to `not_allowed`. This PR updates the metaschema to require the field to be explicit, and sets its value to `allowed_if_defined` for all cases that were missing.

cc @melanieganz @CPernet @mnoergaard for your thoughts on the PET cases. This is strictly speaking a relaxation, but IMO if people provide definitions, that's good?

Closes #1993.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -650,7 +650,8 @@
               "additionalProperties": false
             },
             "additional_columns": {
-              "type": "string"
+              "type": "string",
+              "enum": ["allowed", "allowed_if_defined", "not_allowed", "n/a"]
             },
             "index_columns": { "type": "array", "items": { "type": "string" } },
             "initial_columns": {
@@ -658,7 +659,7 @@
               "items": { "type": "string" }
             }
           },
-          "required": ["selectors", "columns"],
+          "required": ["selectors", "columns", "additional_columns"],
           "additionalProperties": false
         }
       }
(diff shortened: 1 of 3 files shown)


Remi-Gau (approved):

src/metaschema.json
```diff
@@ -634,15 +634,16 @@
...
+              "enum": ["allowed", "allowed_if_defined", "not_allowed"]
             },
             "index_columns": { "type": "array", "items": { "type": "string" } },
             "initial_columns": {
               "type": "array",
               "items": { "type": "string" }
             }
           },
-          "required": ["selectors", "columns"],
+          "required": ["selectors", "columns", "addtional_columns"],
```
effigies, line 646:
```suggestion
          "required": ["selectors", "columns", "additional_columns"],
```

effigies:
@yarikoptic Could I trouble you for a review? I suspect most of our European friends are on holiday, you seem active, and I think this should be straightforward.

rwblair (approved):

PR #2058, opened 2025-02-19
[BEP032] Schema changes for streotaxic coordinates

ree-gupta:
~Add enums for MicroephysCoordSys - similar to what is used already [here](https://github.com/bids-standard/bids-specification/blob/fa2b5d85f7e0eb269c7c935819e29a616d39ab60/src/schema/objects/enums.yaml#L29)~

Update from meeting on 2025.07.30: Coordystem to be removed entirely

- Adds rule to use column entities to describe `AP_angle`, `ML_angle`, `rotation_angle` and corresponding `AP`, `ML`, `DV` positions of the probe from #2160


yarikoptic:
I see TODOs yet to be addressed, so I will mark it as "draft" for now . Please mark "ready for review" when ready

robertoostenveld:
for reference (wrt the meeting just now): the coordinate system enums would be a subset of https://openminds-documentation.readthedocs.io/en/latest/instance_libraries/commonCoordinateSpaceVersions.html, corresponding to the non-human ones

bendichter (approved):

PR #2064, opened 2025-02-22
[MAINT] deal with links redirect to the new website

Remi-Gau:


yarikoptic (approved):
Look reasonable and few I checked seems ok.  But may be we first merge

- #2073 

first and rerun the check here?

effigies (approved):

PR #2065, opened 2025-02-23
Defaced derivative example should be in a pipeline subdirectory

### LINKED ISSUE #2134: [BUG] Strange example in the derivatives documentation
## Description

In the BIDS derivatives [section](https://bids-specification.readthedocs.io/en/stable/derivatives/introduction.html) (at the top) there seem to be an error in the example of how derivatives folders should be used with defaced images:

Current example:

A defaced T1w image could also, in theory, be derived from a BIDS dataset and would thus be under derivatives

```
sub-01/anat/sub-01_T1w.nii.gz
derivatives/sub-01/anat/sub-01_desc-defaced_T1w.nii.gz <-- missing directory name ?
```

Corrected example:

A defaced T1w image could also, in theory, be derived from a BIDS dataset and would thus be under derivatives

```
sub-01/anat/sub-01_T1w.nii.gz
derivatives/defaced/sub-01/anat/sub-01_desc-defaced_T1w.nii.gz
```

Shouldn't we add a subdirectory (here `defaced`) with the name of the data processing pipeline used ? And explained in [Storage of derived datasets](https://bids-specification.readthedocs.io/en/stable/common-principles.html#storage-of-derived-datasets). 

 

_No response_

https://bids-specification.readthedocs.io/en/latest/...

robertoostenveld:
The [derivatives introduction](https://bids-specification.readthedocs.io/en/stable/derivatives/introduction.html) demonstrates a defaced T1w file as `derivatives/sub-01/sub-01_desc-defaced_T1w.nii`, which places it directly under the derivatives folder.  According to [this](https://bids-specification.readthedocs.io/en/stable/common-principles.html#storage-of-derived-datasets) it should however be in a subfolder corresponding to the pipeline name, so a directory name like `derivatives/<pipeline>/sub-01/sub-01_desc-defaced_T1w.nii`.

I named the pipeline "defaced" and inserted it in the example file path. I also renamed the `sourcedata/private` directory into `sourcedata/defaced`. 

Closes #2134.

### DIFF
--- a/src/derivatives/introduction.md
+++ b/src/derivatives/introduction.md
@@ -15,7 +15,7 @@ Examples:
 A defaced T1w image would typically be made during the curation process and is thus under raw
 
 ```Text
-sourcedata/defaced/sub-01/anat/sub-01_T1w.nii.gz
+sourcedata/pre-deface/sub-01/anat/sub-01_T1w.nii.gz
 sub-01/anat/sub-01_T1w.nii.gz
 ```
 
--- a/src/derivatives/introduction.md
+++ b/src/derivatives/introduction.md
@@ -15,7 +15,7 @@ Examples:
 A defaced T1w image would typically be made during the curation process and is thus under raw
 
 ```Text
-sourcedata/private/sub-01/anat/sub-01_T1w.nii.gz
+sourcedata/defaced/sub-01/anat/sub-01_T1w.nii.gz
 sub-01/anat/sub-01_T1w.nii.gz
 ```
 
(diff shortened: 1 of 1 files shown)


src/derivatives/introduction.md
```diff
@@ -15,15 +15,15 @@ Examples:
 A defaced T1w image would typically be made during the curation process and is thus under raw
 
 ```Text
-sourcedata/private/sub-01/anat/sub-01_T1w.nii.gz
+sourcedata/defaced/sub-01/anat/sub-01_T1w.nii.gz
```
effigies, line 18: The idea here was that the undefaced image is in `sourcedata/private`, while the defaced image is in `sub-01/`.

robertoostenveld: Sorry, my bad. I had meant to write `original` or `with_face` there, to contrast to the `defaced`. Or `to_be_kept_private` would also work. 

This change was not crucial to the PR, but I found `private` a bit unclear. Feel free to keep `private` if you prefer that.

effigies: How about:

```suggestion
sourcedata/pre-deface/sub-01/anat/sub-01_T1w.nii.gz
```

robertoostenveld: oh yes, that would be good

effigies (approved):

christinerogers (approved):
lgtm and thanks @robertoostenveld

PR #2075, opened 2025-03-03
Event onsets SHOULD NOT be negative or extremely large

yarikoptic:
I found not a single openneuro datasets with negative onsets.

    smaug:/mnt/datasets/datalad/crawl/openneuro
    *$> for ds in ds*; do git grep '^ *-' '**_events.tsv' | head ; done

But ATM we are looking at a dataset with some buggy _events.tsv which has huge negative values sorted within onsets column.  I think in general/theory in some .1% of the cases there could be point to have onsets with negative onsets, e.g. due to desire to encode some preparatory to the data scanning activities  (e.g. during dummy scans etc).  So for those -- it would be nice to have a dedicated ERROR code to ignore/overwrite.  But in general, it just signals a problem with the file and thus IMHO should be an ERROR.

TODOs (@effigies @rwblair @tsalo - please guide)

- [x] how to handle `n/a` values there? (should be ok)
- [ ] how/where to add a test?

### DIFF
--- a/src/schema/rules/checks/events.yaml
+++ b/src/schema/rules/checks/events.yaml
@@ -44,9 +44,9 @@ SortedOnsets:
     # n/a values will likely cause false alarms if encountered. Consider alternatives.
     - allequal(sorted(columns.onset, "numeric"), columns.onset)
 
-ImplausibleNegativeEventOnset:
+SuspiciousNegativeEventOnset:
   issue:
-    code: IMPLAUSIBLE_NEGATIVE_EVENT_ONSET
+    code: SUSPICIOUS_NEGATIVE_EVENT_ONSET
     message: |
       Event onsets were found more than 60s before the start of recording, indicated by negative values.
       If this is not a mistake, remove or consider an alternative method of representing these events.
@@ -57,9 +57,9 @@ ImplausibleNegativeEventOnset:
   checks:
     - min(columns.onset) >= -60
 
-ImplausiblePositiveEventOnset:
+SuspiciousPositiveEventOnset:
   issue:
-    code: IMPLAUSIBLE_POSITIVE_EVENT_ONSET
+    code: SUSPICIOUS_POSITIVE_EVENT_ONSET
     message: |
       Event onsets were found more than a month after the start of recording.
       If this is not a mistake, remove or consider an alternative method of representing these events.


rwblair:
```
- expression: min([-1, "n/a", 1])                                                                                        
  result: -1
```
In the expression_tests.yaml negative values are treated as less than "n/a". I believe the deno validator handles this case in its implementation.

Remi-Gau:
that would be breaking backwards compatibility, no?

for as far as I can remember negative onsets have been explicitly allowed: https://bids-specification.readthedocs.io/en/stable/glossary.html#onset-columns


Description: Onset (in seconds) of the event, measured from the beginning of the acquisition of the first data point stored in the corresponding task data file. **Negative onsets are allowed, to account for events that occur prior to the first stored data point**. For example, in case there is an in-scanner training phase that begins before the scanning sequence has started events from this sequence should have negative onset time counting down to the beginning of the acquisition of the first volume.

effigies (changes requested):
This would be a breaking change: 

https://bids-specification.readthedocs.io/en/stable/modality-specific-files/task-events.html

| Column name | Level | Data type | Description |
|--|--|--|--|
| [onset](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.columns.onset) | REQUIRED | [number](https://www.w3schools.com/js/js_json_datatypes.asp) | Onset (in seconds) of the event, measured from the beginning of the acquisition of the first data point stored in the corresponding task data file. **Negative onsets are allowed, to account for events that occur prior to the first stored data point.** [...] |

There's very little point in including an onset <-10s. We could set a threshold at, say, -30s or -60s to allow for unconsidered cases while identifying probable errors.

If we are going to set a lower bound in the schema, it should also be indicated in the text.

src/schema/rules/checks/events.yaml
```diff
@@ -43,3 +43,16 @@ SortedOnsets:
...
+    code: EVENT_ONSET_NONNEGATIVE
+    message: |
+      The onsets should be positive.
+    level: error
+  selectors:
+    - suffix == "events"
+    - extension == ".tsv"
+  checks:
+    # TODO: how to handle n/a values?
+    - min(columns.onset) < 0
```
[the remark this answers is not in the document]
VisLab: Why not make a warning rather than an error if the onset is negative.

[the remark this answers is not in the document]
yarikoptic: have you encountered legit cases requiring negative onsets @VisLab ?

yarikoptic:
d'oh @effigies , and I thought I checked for that being spelled out!

I am Ok with making it "up to 60 seconds" or any other reasonable amount for an ERROR.  I wonder if it would still make sense to issue a warning if any negative is encountered given that none of the datasets seems to use them -- WDYT?

Remi-Gau:
Personally I'd prefer to be conservative not make it an error but just a warning as suggested by @VisLab 
https://github.com/bids-standard/bids-specification/pull/2075#discussion_r1977991718

yarikoptic:
ok, made it into a warning and extended description -- or would such wording be not advisable there?

src/schema/rules/checks/events.yaml
```diff
@@ -43,3 +43,15 @@ SortedOnsets:
...
+  issue:
+    code: EVENT_ONSET_NONNEGATIVE
+    message: |
+      The onsets are typically positive: negative ones are allowed but rarely encountered.
+    level: warning
+  selectors:
+    - suffix == "events"
+    - extension == ".tsv"
+  checks:
+    - min(columns.onset) >= 0
```
effigies, line 57: If we want to consider making it an error in a future release (e.g., 2.0), I would write it in a way that it can be promoted for testing and then switched in a future release. For example:

```suggestion
    code: IMPLAUSIBLE_EVENT_ONSET
    message: |
      Event onsets were found more than 60s before the start of recording, indicated by negative values.
      If this is not a mistake, remove or consider an alternative method of representing these events.
    level: warning
  selectors:
    - suffix == "events"
    - extension == ".tsv"
  checks:
    - min(columns.onset) >= -60
```

yarikoptic: Fine with me... I guess we better also then tune up the test "name" (NonNegativeOnsets) to match the error code?  BTW -- why they are not harmonized: do we ever refer to the names somewhere?

src/schema/rules/checks/events.yaml
```diff
@@ -43,3 +43,29 @@ SortedOnsets:
...
+    code: IMPLAUSIBLE_POSITIVE_EVENT_ONSET
+    message: |
+      Event onsets were found more than a month after the start of recording.
+      If this is not a mistake, remove or consider an alternative method of representing these events.
+    level: warning
+  selectors:
+    - suffix == "events"
+    - extension == ".tsv"
+  checks:
+    - max(columns.onset) < 2678400
```
yarikoptic, line 71: ok, got "inspired" and added one for implausible positive one.  I think not even .1% of users should stick more than a month worth of events into one file. WDYT?

Remi-Gau:
Should the definition of onsets be amended to mention those warnings?

src/schema/rules/checks/events.yaml
VisLab: I'm not sure these checks are necessary.  On the negative onsets, I think I have seen driving simulation experiments where there were simulator events and driver response events that were recorded in the events file before the EEG recording was turned on after the person was driving a while.  The behavioral events were recorded through the whole thing, but the onset was zeroed at the point the EEG started.

On the upper end...   I am not sure why this would be necessary either.   Consider a wearable EKG recording --- it is possible to have long term monitoring....

yarikoptic:
> Should the definition of onsets be amended to mention those warnings?

Lazy me: I think they are of the kind of SUSPICIOUSLY_LONG_EVENT_DESIGN and SUSPICIOUSLY_SHORT_EVENT_DESIGN hint types.. I do not think we document for them.

Doer me: I pushed change to use `SUSPICIOUS_` instead of `IMPLAUSIBLE_` prefix here for consistency

effigies (approved):
Just to verify, this is just a hint for the validator, not something you want to add a SHOULD rule in the spec for?

Edit: I see this was the last exchange with Remi. Sounds good. A second review is needed.

VisLab:
I agree with the not allowing extremely large onsets as it likely means that the onsets are in terms of milliseconds rather than seconds.

I still would prefer that negative onsets just give a warning as there are legitimate use cases for negative onsets.

PR #2077, opened 2025-03-04
[FIX] Enable format validation in validating BIDS schemas

candleindark:
The stable version of [`schema.json`](https://bids-specification.readthedocs.io/en/stable/schema.json) fails to validate against the stable version of [`metaschema.json`](https://bids-specification.readthedocs.io/en/stable/metaschema.json) if format validation is enabled.

Specifically, running `check-jsonschema --schemafile metaschema.json schema.json` produces the following output

```console
Schema validation errors were encountered.
  schema.json::$.objects.formats.participant_relative.pattern: '(?!/)(?!sub-)[0-9a-zA-Z/\\_\\-\\.]+' is not a 'regex'
  schema.json::$.objects.formats.participant_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  schema.json::$.objects.formats.bids_uri.pattern: 'bids:[0-9a-zA-Z/#:\\?\\_\\-\\.]+' is not a 'regex'
  schema.json::$.objects.formats.bids_uri: Unevaluated properties are not allowed ('pattern' was unexpected)
  schema.json::$.objects.formats.stimuli_relative.pattern: '(?!/)(?!stimuli/)[0-9a-zA-Z/\\_\\-\\.]+' is not a 'regex'
  schema.json::$.objects.formats.stimuli_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  schema.json::$.objects.formats.file_relative.pattern: '(?!/)[0-9a-zA-Z/\\_\\-\\.]+' is not a 'regex'
  schema.json::$.objects.formats.file_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  schema.json::$.objects.formats.dataset_relative.pattern: '(?!/)[0-9a-zA-Z/\\_\\-\\.]+' is not a 'regex'
  schema.json::$.objects.formats.dataset_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  schema.json::$.objects.formats.hed_version.pattern: '^(?:[a-zA-Z]+:)?(?:[a-zA-Z]+_)?(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\ (?:-(?:(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?\\ (?:\\+(?:[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$' is not a 'regex'
  schema.json::$.objects.formats.hed_version: Unevaluated properties are not allowed ('pattern' was unexpected)
```
where `check-jsonschema` is at `0.31.2`, and by default, it validates with format validation enabled.

My guess is that BIDS schemas are intended to be validated against the meta schema with format validation enabled. This PR provides a solution to ensure validations of BIDS schemas are done with format validation enabled. Specifically, this PR includes the following changes.

1. Provide a helper function to create jsonschema validators
2. Provide a helper function to create the jsonschema validator for validating BIDS schemas with format validation enabled. Additionally, this function also caches the validator upon the return of the first execution to improve performance.
3. Use the validator from previous step to validate BIDS schemas

### DIFF
--- a/tools/schemacode/src/bidsschematools/tests/test_utils.py
+++ b/tools/schemacode/src/bidsschematools/tests/test_utils.py
@@ -1,0 +1,166 @@
+from contextlib import nullcontext
+from typing import cast
+
+import pytest
+from jsonschema.exceptions import SchemaError, ValidationError
+from jsonschema.protocols import Validator as JsonschemaValidator
+from jsonschema.validators import Draft7Validator, Draft202012Validator
+
+from bidsschematools.utils import jsonschema_validator
+
+
+@pytest.fixture
+def draft7_schema() -> dict:
+    """
+    A minimal valid Draft 7 schema requiring a 'name' property of type 'string'.
+    """
+    return {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "type": "object",
+        "properties": {"name": {"type": "string"}},
+        "required": ["name"],
+    }
+
+
+@pytest.fixture
+def draft202012_schema() -> dict:
+    """
+    A minimal valid Draft 2020-12 schema requiring a 'title' property of type 'string'.
+    """
+    return {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {"title": {"type": "string"}},
+        "required": ["title"],
+    }
+
+
+@pytest.fixture
+def draft202012_format_schema() -> dict:
+    """
+    Draft 2020-12 schema that includes a 'format' requirement (e.g., 'email').
+    Used to test the 'check_format' parameter.
+    """
+    return {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {"email": {"type": "string", "format": "email"}},
+        "required": ["email"],
+    }
+
+
+@pytest.fixture
+def schema_no_dollar_schema() -> dict:
+    """
+    Schema that lacks the '$schema' property altogether.
+    Used to test that 'default_cls' is applied.
+    """
+    return {
+        "type": "object",
+        "properties": {"foo": {"type": "string"}},
+        "required": ["foo"],
+    }
+
+
+class TestJsonschemaValidator:
+    @pytest.mark.parametrize(
+        ("fixture_name", "expected_validator_cls"),
+        [
+            pytest.param("draft202012_format_schema", Draft202012Validator, id="Draft202012"),
+            pytest.param("draft7_schema", Draft7Validator, id="Draft7"),
+        ],
+    )
+    @pytest.mark.parametrize("check_format", [True, False])
+    def test_set_by_dollar_schema(
+        self,
+        request: pytest.FixtureRequest,
+        fixture_name: str,
+        expected_validator_cls: type,
+        check_format: bool,
+    ) -> None:
+        """
+        Test that the correct validator class is returned for different '$schema' values
+        """
+        # Dynamically retrieve the appropriate fixture schema based on fixture_name
+        schema = request.getfixturevalue(fixture_name)
+
+        validator = jsonschema_validator(schema, check_format=check_format)
+
+        assert isinstance(validator, expected_validator_cls)
+
+    @pytest.mark.parametrize(
+        ("check_format", "instance", "expect_raises"),
+        [
+            (True, {"email": "test@example.com"}, False),
+            (True, {"email": "not-an-email"}, True),
+            (False, {"email": "not-an-email"}, False),
+        ],
+        ids=[
+            "check_format=True, valid email",
+            "check_format=True, invalid email",
+            "check_format=False, invalid email",
+        ],
+    )
+    def test_check_format_email_scenarios(
+        self,
+        draft202012_format_schema: dict,
+        check_format: bool,
+        instance: dict,
+        expect_raises: bool,
+    ) -> None:
+        """
+        Parametrized test for check_format usage on valid/invalid email addresses under
+        Draft202012Validator.
+        """
+        validator = jsonschema_validator(draft202012_format_schema, check_format=check_format)
+
+        # If expect_raises is True, we use pytest.raises(ValidationError)
+        # Otherwise, we enter a no-op context
+        ctx = pytest.raises(ValidationError) if expect_raises else nullcontext()
+
+        with ctx:
+            validator.validate(instance)  # Should raise or not raise as parametrized
+
+    @pytest.mark.parametrize(
+        ("schema_fixture", "expected_validator_cls"),
+        [
+            # Scenario 1: no $schema => we expect the default_cls=Draft7Validator is used
+            pytest.param("schema_no_dollar_schema", Draft7Validator, id="no-$schema"),
+            # Scenario 2: has $schema => draft 2020-12 overrides the default_cls
+            pytest.param("draft202012_schema", Draft202012Validator, id="with-$schema"),
+        ],
+    )
+    def test_default_cls(
+        self,
+        request: pytest.FixtureRequest,
+        schema_fixture: str,
+        expected_validator_cls: type,
+    ) -> None:
+        """
+        If the schema has no '$schema' property, and we provide a 'default_cls',
+        the returned validator should be an instance of that class.
+
+        If the schema *does* have '$schema', then the default_cls is ignored, and
+        the validator class is inferred from the schema's '$schema' field.
+        """
+        # Dynamically grab whichever fixture is specified by schema_fixture:
+        schema = request.getfixturevalue(schema_fixture)
+
+        # Provide default_cls=Draft7Validator
+        validator = jsonschema_validator(
+            schema,
+            check_format=False,
+            default_cls=cast(type[JsonschemaValidator], Draft7Validator),
+        )
+        assert isinstance(validator, expected_validator_cls)
+
+    def test_invalid_schema_raises_schema_error(self) -> None:
+        """
+        Provide an invalid schema, ensuring that 'SchemaError' is raised.
+        """
+        invalid_schema = {
+            "$schema": "https://json-schema.org/draft/2020-12/schema",
+            "type": 123,  # 'type' must be string/array, so this is invalid
+        }
+        with pytest.raises(SchemaError):
+            jsonschema_validator(invalid_schema, check_format=False)
--- a/tools/schemacode/src/bidsschematools/schema.py
+++ b/tools/schemacode/src/bidsschematools/schema.py
@@ -9,7 +9,7 @@ from copy import deepcopy
 from functools import cache, lru_cache
 from importlib.resources import files
 
-from jsonschema import ValidationError, validate
+from jsonschema import ValidationError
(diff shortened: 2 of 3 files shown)


yarikoptic:
ping @effigies and @tsalo -- let's get "kosher" again?

candleindark:
> There aren't any changes to the schema or metaschema being proposed, so I don't see how this addresses the stated problem.

You are right. This PR doesn't correct the problems in the BIDS schema. It only ensures subsequent validations of the BIDS schemas are done with format validation enabled.

I don't think I am the right person to correct any existing problems in the BIDS schema at this point.

effigies:
I ran `check-jsonschema` locally, and this reproduces now. I believe that these are new errors as of https://github.com/python-jsonschema/check-jsonschema/pull/511.

Given that you're not fixing the regexes, shouldn't the goal be for this PR be for CI to fail until they're fixed?

I have a patch that will fix the regexes, but it doesn't make much sense to apply it before we have a test that fails.

effigies:
(Clicked the wrong comment button.)

candleindark:
> I ran `check-jsonschema` locally, and this reproduces now. I believe that these are new errors as of [python-jsonschema/check-jsonschema#511](https://github.com/python-jsonschema/check-jsonschema/pull/511).

You are right. The errors are result of 'regular expression interpretation in "pattern", "patternProperties", and "format": "regex" usages now uses unicode-mode JS regular expressions by default' in `check-jsonschema`. The following commands produced no error.

```console
(tmp) ➜  tmp check-jsonschema --regex-variant python --schemafile metaschema.json schema.json
ok -- validation done
(tmp) ➜  tmp check-jsonschema --regex-variant nonunicode --schemafile metaschema.json schema.json
ok -- validation done
```

> Given that you're not fixing the regexes, shouldn't the goal be for this PR be for CI to fail until they're fixed?

I just tried using the `jsonschema` package to validate the BIDS schema against the metaschema, and I was not able to reproduce the errors output by the `check-jsonschema` CLI. The validation was successful and didn't produce any error at all. In fact, according to https://github.com/python-jsonschema/check-jsonschema/pull/511, to get the default behavior of the `check-jsonschema`, one has to build a customer validators using the `jsonschema` package's `extend()` API. Since this is just a matter of choice regex variant, I don't think it is a good idea, or add much value, to deviate from the default behavior of `jsonschema` package.

There is an existing test that tests the validation of the BIDS schema against the metaschema, [`test_valid_schema`](https://github.com/bids-standard/bids-specification/blob/3b6d2f7c8bd060e9b7efb15ecb93011c500df72c/tools/schemacode/src/bidsschematools/tests/test_schema.py#L362C13-L365). It will fail should the BIDS schema fail to validate (now including format validation with this PR).

Though the observed errors that initiated this PR are no longer applicable. I still think the changes in this PR should be included since without them format validation would not run at all in validation of BIDS schemas.

effigies:
> Though the observed errors that initiated this PR are no longer applicable.

They are still applicable:

```
󰅂 uv run bst export | uvx check-jsonschema --schemafile src/metaschema.json -
Schema validation errors were encountered.
  -::$.objects.formats.participant_relative.pattern: '(?!/)(?!sub-)[0-9a-zA-Z+/\\_\\-\\.]+' is not a 'regex'
  -::$.objects.formats.participant_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  -::$.objects.formats.stimuli_relative.pattern: '(?!/)(?!stimuli/)[0-9a-zA-Z+/\\_\\-\\.]+' is not a 'regex'
  -::$.objects.formats.stimuli_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  -::$.objects.formats.bids_uri.pattern: 'bids:[0-9a-zA-Z/#:\\?\\_\\-\\.]+' is not a 'regex'
  -::$.objects.formats.bids_uri: Unevaluated properties are not allowed ('pattern' was unexpected)
  -::$.objects.formats.dataset_relative.pattern: '(?!/)[0-9a-zA-Z+/\\_\\-\\.]+' is not a 'regex'
  -::$.objects.formats.dataset_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
  -::$.objects.formats.hed_version.pattern: '^(?:[a-zA-Z]+:)?(?:[a-zA-Z]+_)?(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\ (?:-(?:(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?\\ (?:\\+(?:[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$' is not a 'regex'
  -::$.objects.formats.hed_version: Unevaluated properties are not allowed ('pattern' was unexpected)
  -::$.objects.formats.file_relative.pattern: '(?!/)[0-9a-zA-Z+/\\_\\-\\.]+' is not a 'regex'
  -::$.objects.formats.file_relative: Unevaluated properties are not allowed ('pattern' was unexpected)
```

That's why I was hoping we could get to the point of having them appear before I fixed them, so we know the check works.

candleindark:
I will give it another shot, possibly sometime next week though.

candleindark:
@effigies I added the test that invokes `json-checkschema` to validate the BIDS schema against the metaschema. Currently, all supported regex variants in `check-jsonschema` are tested. If you prefer only the default/unicode variant, please let me know, and I will remove the others.

yarikoptic:
@effigies -- we are quite red now.  Should you add fixes to this PR to make it green again before the merge? (I added TODO in original description)  or how would you like to proceed?

tools/schemacode/src/bidsschematools/tests/test_utils.py
```diff
@@ -0,0 +1,166 @@
...
+def draft7_schema() -> dict:
+    """
+    A minimal valid Draft 7 schema requiring a 'name' property of type 'string'.
+    """
+    return {
+        "$schema": "http://json-schema.org/draft-07/schema#",
+        "type": "object",
+        "properties": {"name": {"type": "string"}},
+        "required": ["name"],
+    }
```
effigies, line 22: I think these fixtures could just be constants. They're not actually used as fixtures anywhere, right?

effigies (commented):
My main concern is that there are strangely organized fixtures providing constants. I will push the patch to fix the test.

tools/schemacode/src/bidsschematools/tests/test_utils.py
candleindark: They are called as fixtures, but you are right, they can be defined as constants, which are better. 

The requested changes are made and pushed.

effigies (approved):

src/schema/objects/formats.yaml
```diff
@@ -47,9 +47,15 @@ hed_version:
...
-    (?:\+(?:[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$'
+  pattern: "\
+    (?:[a-zA-Z]+:)?\
+    (?:[a-zA-Z]+_)?\
+    (?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)\
+    (?:-(?:\
+      (?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z\\-]*)\
+      (?:\\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z\\-][0-9a-zA-Z\\-]*))*\
+    ))?\
+    (?:\\+(?:[0-9a-zA-Z\\-]+(?:\\.[0-9a-zA-Z\\-]+)*))?"
```
effigies, line 58: @happy5214 Can you verify that this pattern is correct?

Changes:

1) `\d` -> `[0-9]`
2) Lone `-` inside `[]` becomes `\\-`
3) Extra newlines to help break up logical chunks.

[Renders](https://bids-specification--2077.org.readthedocs.build/en/2077/schema.json) as:

```json
"(?:[a-zA-Z]+:)?(?:[a-zA-Z]+_)?(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)(?:-(?:(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z\\-]*)(?:\\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z\\-][0-9a-zA-Z\\-]*))*))?(?:\\+(?:[0-9a-zA-Z\\-]+(?:\\.[0-9a-zA-Z\\-]+)*))?"
```

happy5214: I couldn't even verify the *original* was correct. I only contributed the first two groups (the ones ending with the colon and underscore) and copied the rest from https://semver.org/. @VisLab Can a HED schema have any valid semantic version, or are there limits within that?

VisLab: We should only use plain x.y.z.   I don't think the specification says anything to the contrary. There are a few odd things in the deprecated folder like HED8.0.0-alpha.1.xml.  These have never been released and no known datasets use them.  (Validators don't support these preliminary formats.) I think we should remove them. 

Schemas in the prerelease folder also conform to x.y.z versioning.

happy5214: To clarify, the first two "lines" (in the updated version) are the prefix and library name, and those should be kept even if the version number is simplified.

effigies: Just reading this, the following is permissible:

```
xYz:HjK_10.0.3-7a4Q-0-0-0-.8b5R-1-1-1-+asduijkqr98926ah1.asduijkqr98926ah1.asduijkqr98926ah1
|--||--||----|||---------||----------||----------------------------------------------------|  
 L1  L2   L3  4    L5          L6                            L7
```

But I agree that the semver regex appears to allow this, for everything after `xYz:HjK_`, so if you're happy with this, we can call it good.

happy5214: Our rule is that "official" schema releases use only `x.y.z` version numbering (pre-releases, which pre-8.0.0 followed the `-alpha` tagging scheme mentioned by @VisLab, now share the same version numbering as official releases; either way, they should not be used). Therefore, anything after the `<prefix>:<library>_x.y.z` should be removed entirely.

effigies:
```suggestion
  pattern: "(?:[a-zA-Z]+:)?(?:[a-zA-Z]+_)?(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)\\.(?:0|[1-9][0-9]*)"
```

effigies:
Okay, I think this is all set. @candleindark @yarikoptic Do you want to look over my changes before merging?

candleindark:
> Okay, I think this is all set. @candleindark @yarikoptic Do you want to look over my changes before merging?

I think it is good to go.

PR #2079, opened 2025-03-06
feat(schema): Add selectors to file rules

### LINKED ISSUE #2052: [SCHEMA] Proposal: Add selectors to file rules
> We could consider putting `selectors` in filename rules to achieve this. Something like:
> 
> ```yaml
> stimuli_default:
>   selectors:
>     - '!exists("stimuli/stimuli.tsv", "dataset")'
>   datatypes:
>     - stimuli
>   stem:
>     - '*'
>   extensions:
>     - '*'
> 
> audio:
>   selectors:
>     - exists("stimuli/stimuli.tsv", "dataset")
>   suffixes:
>     - audio
>   extensions:
>     - .wav
>     - .mp3
>     - .aac
>     - .ogg
>   datatypes:
>     - stimuli
>   entities:
>     - stim: required
>     - part: optional
> 
> ...
> ``` 

 _Originally posted by @effigies in [#1828](https://github.com/bids-standard/bids-specification/issues/1828#issuecomment-2653926159)_

The above proposal considers the case where the contents of `stimuli/` may be free-form or structured, depending on a condition. This is necessary to preserve backwards compatibility while extending BIDS principles into previously un-managed areas.

This is also the basic problem we had with derivatives, which we resolved by creating the `rules.files.derivatives` branch. With this proposal, we could use `dataset.description.DatasetType != 'raw'` (or similar) to indicate rules that are invalid in raw but permissible in derivative datasets.

cc @rwblair @nellh @tsalo @bids-standard/schema-users

effigies:
Closes #2052.

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -650,6 +650,7 @@
     "pathRule": {
       "type": "object",
       "properties": {
+        "selectors": { "$ref": "#/definitions/ruleTypes/expressionList" },
         "level": { "enum": ["optional", "required"] },
         "path": { "type": "string" }
       },
@@ -659,6 +660,7 @@
     "stemRule": {
       "type": "object",
       "properties": {
+        "selectors": { "$ref": "#/definitions/ruleTypes/expressionList" },
         "level": { "enum": ["optional", "recommended", "required"] },
         "datatypes": {
           "type": "array",
@@ -673,6 +675,7 @@
     "suffixRule": {
       "type": "object",
       "properties": {
+        "selectors": { "$ref": "#/definitions/ruleTypes/expressionList" },
         "level": { "enum": ["optional", "recommended", "required"] },
         "datatypes": {
           "type": "array",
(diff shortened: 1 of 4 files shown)


effigies:
@yarikoptic Q: Do you think it would be better to target the `maint/1.10.0` branch so that there's a version of the 1.10.0 schema that has this feature, or should we just aim to include it in the next BIDS release?

yarikoptic:
filed
- https://github.com/bids-standard/bids-specification/issues/2080

and sure -- looks like a patch-worthy fix, so why not into `maint/`?

effigies:
@yarikoptic @rwblair Review would be appreciated.

rwblair (approved):
Looks good to me.

For future consideration:

- selectors on directory naming rules
- inheritance principle for selectors 😈😈😈

effigies:
> inheritance principle for selectors 😈😈😈

I don't know what that means?

PR #2083, opened 2025-03-10
[WIP][MAINT] use jinja in the macros

Remi-Gau:
this is just a test to see if using jinja templates in the macros could help with our rendering of HTML / PDF

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-.PHONY:  tools/contributors.tsv
-
 validate_citation_cff: CITATION.cff
 	cffconvert --validate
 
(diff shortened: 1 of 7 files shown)


Makefile
Remi-Gau, line 1: unrelated: could be extracted in separate PR

Remi-Gau:
a few oddities:

ordered list are a quote blocked in the html version:

image

or don't have the proper 'order' in pdf (item 1. appears twice)

image

Remi-Gau:
Also see https://bids-specification--2083.org.readthedocs.build/en/2083/glossary.html#stimuluspresentation-metadata

image

or similarly:

- https://bids-specification--2083.org.readthedocs.build/en/2083/glossary.html#deidentificationmethodcodesequence-metadata

- https://bids-specification--2083.org.readthedocs.build/en/2083/glossary.html#genetics-metadata

oesteban:
> ordered list are a quote blocked in the html version:

I believe this is not a bug - we need to remove the indentation from the common principles yaml (it was added precisely because the current parser would otherwise render something funky).

https://github.com/bids-standard/bids-specification/blob/7f5e0a778e8aec08e1d6743f19626c929d91e69c/src/schema/objects/common_principles.yaml#L11-L48

PR #2085, opened 2025-03-13
fix(bst): Do not alter logger in library mode, log to stderr not stdout

yarikoptic:
We tried to upgrade to use recent bidsschematools in dandi-cli (https://github.com/dandi/dandi-cli/pull/1591) but tests started to fail since output was polluted with log messages from bidsschematools, which prompted me to look inside.

I have tried to keep my changes minimal (but did rename `logger` to `lgr` for consistency) and individual commits provide a little more information, but overall point - do not configure logger unless it is an "end user application". And it is typical for logging to just go into stderr to not interfere with potential "useful" output on stdout.

### DIFF
--- a/tools/schemacode/src/bidsschematools/__main__.py
+++ b/tools/schemacode/src/bidsschematools/__main__.py
@@ -14,7 +14,10 @@ else:
 
 from .rules import regexify_filename_rules
 from .schema import export_schema, load_schema
+from .utils import configure_logger, get_logger
 from .validator import _bidsignore_check
+
+lgr = get_logger()
 
 
 @click.group()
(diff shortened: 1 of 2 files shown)


effigies:
Can you target the maint branch so we can release this without a BIDS release?

effigies:
Thanks, if this works for your use case, that's great. I don't have a great feel for logging best practices for libraries vs apps, so I haven't imposed much of a structure here. As far as I'm concerned, you're welcome to refactor this entirely, if you have further suggestions.

Do you have a doc that you have written or use as a guideline? It would be nice to have a reference I can go back to.

effigies (approved):
LGTM. Small nit.

PR #1998, opened 2024-12-06
[ENH] extension for electromyography (EMG) - BEP42

drammock:
This is a very early WIP implementation to add EMG support. CIs are not expected to pass yet.

cc @neuromechanist @jwelzel @larsoner @arnodelorme @robertoostenveld feel free to push directly to this branch, I'll add you as repo collaborators on my fork

> [!Note]
>
> **We meet regularly to discuss this BEP**
>
> Next meeting: 18 Dec 2024 on https://ucsd.zoom.us/j/96433382377
>
> Communication channel on github repo / matrix / slack / discord : https://github.com/bids-standard/bids-specification/issues/1371
>

Issue #2027, opened 2025-01-13
device/dev- entity + devices.{tsv,json}


yarikoptic:
### Your idea

Relates to
 
- https://github.com/bids-standard/bids-specification/issues/1771

by @bendichter , and overall to cover necessary disambiguation between acquisition across different devices of the same kind. In our case it would immediately be multiple video cameras, so we could have `_dev-camera1`, `_dev-camera2` with `devices.tsv` providing at least the `device_id` and `description` columns and then any other additional metadata fields.

It would reflect the addition of `_camera-` entity in neuroblueprint (ref: https://neuroblueprint.neuroinformatics.dev/latest/specification.html#example-project-folder ; attn @niksirbi) .

We should also see on extensive metadata schema already in AIND (https://github.com/bids-standard/bids-2-devel/issues/60) listed in https://github.com/AllenNeuralDynamics/aind-data-schema to potentially generalize a few other generic device description metadata fields, and/or provide more specialization for some common device types (audiovideo recording etc)

Issue #1675, opened 2023-12-19
BEP about non-neuronal physiological (cardiac, respiratory, skin conductance, gastro, ...) data and physiological data derivatives


smoia:
### Your idea

Hello all,

within the physiopy community ([website](https://physiopy.github.io/) and [github](https://github.com/physiopy)) we are facing the issue of talking about physiological data standardisation and physiological data derivatives.

To be clear: the data we're talking about are not M/EEG (no elettrophysiology/neurophysiology). Instead, we refer to data like cardiac or respiratory fluctuations, skin conductance, blood pressure, gas analyses, ...

We know there is already a start of standardisation in the BIDS specification, but having dealt with such data we recognise it might be nice to add more standardisation and information to the specification, including specifications on this data derivatives.

We also know there is a BEP about one type of such data, eyetracking, the [BEP020](https://bids.neuroimaging.io/bep020), but we aim at discussing other types of data instead.

I wanted to start sending this message to see if there might be the interest for such a BEP, if there is conflict with other BEPs I'm not aware of, and to see if anyone is interested in joining the effort - we'll discuss about this plenty in our monthly physiopy community meetings starting with the new year to reach a BEP submission.

I hope it's appreciated!

smoia:
There are a few parts that came up already by talking within the community.

**1. Precision**: there are at least four different types of data to extract from respiratory-related information (respiratory effort, ventilation, PCO2, PO2, ...); when measuring "cardiac fluctuations", those could come in the form of pulse or heartbeat (PPG vs ECG) - those are very different types of data.

**2. Acquisition information**: more information should be reported on the type of instrument used for the acquisition, since each of them has its own peculiarity (e.g. MRIs vs specific manufacturers).

**3. Peculiar signals with current limitations**: another example is continuous blood pressure recordings: at the moment the instruments we have to acquire such data in the MRI is with instruments that do not have a consistent sampling rate - something that is not considered at all by BIDS but is faced by the physio community.

**4. User/field customs and possible signal transformations**: another thing is the unit of measure - for instance, in the field of cerebrovascular reactivity mapping, the custom is to express gas changes in millimetres of Mercury (mmHg), which is not a SI unit, although all raw signals are measured in (Delta) Volts. Most manufacturers either make the transformation automatically or provide the formula to convert from Volts to whatever other measure is required, and that would be precious information to store. Beside that, it's about enforcing the use of certain unit of measures and alienate a group of users or to ask them to treat their data differently (indicating how).

[...]

PR #2090, opened 2025-03-24
[ENH] Clarify that recording entity may be used to distinguish recording instruments

drammock:
In #1998 we're working on a BEP for EMG data.  We're hoping to use the `acq` entity for cases where multiple EMG devices were used simultaneously, but they were recording through different amplifiers, into different data files, possibly with different sampling frequencies, etc.  This PR proposes some slight wording changes to ~~the `acq` entity definition to make it clearer that such uses of `acq` are acceptable.~~ *EDIT: following https://github.com/bids-standard/bids-specification/pull/2090#issuecomment-2748848197 we're going to use `recording` instead, and this PR modifies the definition of `recording` accordingly.*

(if they're deemed *not* acceptable, better that we know that now so we can adapt #1998 accordingly).

One possible alternative would be a new `sys` entity, analogous to the `tracksys` entity introduced for BIDS-Motion, with the (likely) end result that in future releases `tracksys` would be replaced by `sys` in motion datasets.

cc @JuliusWelzel @neuromechanist


effigies:
This sounds like the exact use case for the [`recording-<label>`](https://bids-specification.readthedocs.io/en/stable/glossary.html#recording-entities) entity. Was that considered?

At least in the MRI world, two files that differed by `acq-<label>` would not be expected to have been collected simultaneously.

neuromechanist:
We can expand the definition both as a separate PR or directly on #1998.
@bids-maintenance, please let us know which one you prefer.

drammock:
> > > `recording-<label>` is being currently used to attach `physio` and `stim` data to the main recordings ([link](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/physiological-recordings.html#physiological-recordings)),
> > 
> > 
> > Does that prevent using it in other ways though?
> 
> I don't think so. 

agreed

> If we're considering expanding `acq-` to permit simultaneous recordings, it seems at least as valid to expand `recording-` to apply to data files and not just auxiliary recordings.

agreed.

> I think we could update the `recording` definition as follows (additions in bold):
> 
> > The recording- entity can be used to distinguish **simultaneously acquired,** continuous recording files.
> > This entity may be applied when continuous recordings have different sampling frequencies or start times,
> > **or are acquired with different instruments**.
> > For example, physiological recordings with different sampling frequencies may be distinguished using labels like `recording-100Hz` and `recording-500Hz`.
> 
> Would that cover your use case?

Yep I think so. @neuromechanist do you agree that this would suffice?

neuromechanist:
TL, DR: BIDS had provided the [`data type`](https://bids-specification.readthedocs.io/en/stable/common-principles.html#definitions) facility to accommodate each functional group of data live under their own directory.
Along the way, however, the "auxiliary" `modalities` (such as `_physio` and `_stim`) were allowed to be at the host `data type` directory, using the additional `_recording-<label>` entity.
While we very well could extend the definition of the `recording` to also include concurrent data of the intended `data type`, I believe we would benefit from making a distinction between data files intended in the `data type` directory and the so-called auxiliary files.

_________________
Interestingly, the issue's scope is becoming much larger than anticipated, so I appreciate making it a separate pull request.

I see that either entities would require expansion, and with @drammock's example, using `recording` could work quite perfectly. But:

> If we're considering expanding acq- to permit simultaneous recordings, it seems at least as valid to expand recording- to apply to data files and not just auxiliary recordings. One person's auxiliary recording is another person's primary data.

I have reservations about this statement. Auxiliary recordings are auxiliary not in the sense that it might be important for someone or not, but in the sense that which `data type` is hosting the data.

IMO, we would benefit from using two separate entities to distinguish if the data file is auxiliary (as they live under a different `datatype` than their intended `modality`) or if they are native under their `data type`. It does not need to be `acq` or `recording`, it could be a new entity, like `sys` as suggested by @JuliusWelzel at #1998.

The use-case could become even more potent when considering the new BEP045 (physio-BIDS, #1675, cc @smoia), where the `physio` data would claim its own `data type`. I don't think under BIDS 1.x we can deprecate the use of auxiliary `_physio` files, in other directories, but we might be able to make a clear distniction using two separate entities based on what is being recorded and saved under the intended `data type` and what is being hosted out of its place.

Currently, Motion-BIDS (cc @JuliusWelzel, @sjeung) use `tracksys` to (among other uses) differentiate concurrent recordings under the same `data type` (hence, the new `sys` entity to keep the resemblance).

I also understand that this change has BIDS2.0 implications, both about scans/recordings.tsv and modality independence https://github.com/bids-standard/bids-2-devel/issues/29 and https://github.com/bids-standard/bids-2-devel/issues/45 (cc @tsalo, @yarikoptic) and also potentially harmonizing the `data type` and `modality` and also https://github.com/bids-standard/bids-2-devel/issues/54 (as well as probably a couple more issues).

To sum up, I think any or all will work ok, so I won't have any objections against them. Using different entities may make the file structures more clear down the road, and I hope that I could lay out the case and involve all potential stakeholders.

drammock:
> TL, DR: BIDS had provided the [`data type`](https://bids-specification.readthedocs.io/en/stable/common-principles.html#definitions) facility to accommodate each functional group of data live under their own directory.
> Along the way, however, the "auxiliary" `modalities` (such as `_physio` and `_stim`) were allowed to be at the host `data type` directory, using the additional `_recording-<label>` entity.

Yes, but the use of `*_recording-<label>` is optional in those cases, so IIUC it's really the `_physio` and `_stim` parts of the filenames that marks them as "this is a different datatype then the parent folder would indicate". Thus it's not clear (to me at least) that there's anything about the definition of `recording` that makes it unsuitable for our needs in #1998.

oesteban:
@drammock would it be possible to add some examples of this definition extension of `acq-` in the appropriate part of the spec? Without examples, it's hard to see why (or why not) recording would be adequate.

drammock:
@oesteban example added in 034f62ae3fcc954e63d889959d2b6af00d8d75d9. Does that help?

smoia:
@neuromechanist @drammock in #1675 and BEP045 we're proposing to add an optional metadata field to indicate what data are collected together and hence are "auxiliary" (without necessarily forcing a hierarchy). IDK if that could be a valid option for you as well?

Also cc'ing other BEP045 leads startign with @m-miedema

yarikoptic:
re `sys` entity -- sounds analogous to 
- #2027 

Left a comment with my preference on https://github.com/bids-standard/bids-specification/pull/1998#issuecomment-2766285223.

As for "auxiliary" data types -- I think we should at large strive for making all data types "1st class citizens" since discrimination between them is mostly "an intent" of an experimenter and thus IMHO should be just noted at the level of the dataset, and otherwise result in identical layouts regardless of the intent: users looking for datatype X should find it in a consistent location regardless of the intent of the experimenter, and then some metadata structure (e.g. within filenames and `_scans.tsv`) should establish the relationship across subfolders if we are to have them at all (where BIDS 2.0 could potentially dissolve them).  I know we have some fundamental divergences from such a potential principle (e.g. going back to `_events.tsv` nearby functional data vs `beh/` folder ), but IMHO it is something to keep in mind for while working out new data types.

drammock:
@smoia says:

> in #1675 and BEP045 we're proposing to add an optional metadata field to indicate what data are collected together and hence are "auxiliary" (without necessarily forcing a hierarchy). IDK if that could be a valid option for you as well?

That looks useful, but I don't think it captures/addresses the use case we're facing. We're dealing with EMG as the primary modality, but with multiple separate devices making recordings simultaneously (from different muscles/locations on the body). So we need multiple `*_emg.bdf` files in the same `sub-*/ses-*/emg/` folder, and need to differentiate them.

@yarikoptic says that we should consider the `dev` / `device` entity instead (cf. #2027). I strongly agree, `device` is semantically a better entity name than `acq` (or `sys` or `recording`) for what we're trying to capture.  Shall I close this PR and open a new one to implement that?  My only hesitation is that we need to pick *something* to make the EMG spec workable, and since `acq` already exists it seems like an easier lift to slightly expand its definition, rather than getting a whole new entity added.

oesteban:
> @oesteban example added in [034f62a](https://github.com/bids-standard/bids-specification/commit/034f62ae3fcc954e63d889959d2b6af00d8d75d9). Does that help?

It does help, however the suffix is `_emg` which doesn't exist in BIDS to date. I believe the example given would perfectly be described as:

```
sub-01_recording-bipolar_physio.tsv.gz
sub-01_recording-grid_physio.tsv.gz
```

assuming that the two recordings have different sampling frequencies. Otherwise, they can be sibling columns in a single `_physio.tsv.gz` file.

With the acceptance of #1128 (fingers crossed), we could quickly define a new `PhysioType` `electromyogram` which would allow for specific metadata and rules for recordings names, as we are proposing for the new `PhysioType` `eyetrack`.

drammock:
> > @oesteban example added in [034f62a](https://github.com/bids-standard/bids-specification/commit/034f62ae3fcc954e63d889959d2b6af00d8d75d9). Does that help?
> 
> It does help, however the suffix is `_emg` which doesn't exist in BIDS to date. 

@oesteban Please see the original description of this PR, where in the first sentence I explain that:

> In https://github.com/bids-standard/bids-specification/pull/1998 we're working on a BEP for EMG data.

I intentionally didn't include an example here at first, since I know that it wouldn't make sense to merge an example that illustrates the EMG modality *before* that BEP is merged; I added such an example later at your request. So if you assume for a moment that an EMG modality does exist (or will soon) and interpret the example in that context, what do you think? Is expanding the definition of `acq` good (as in this PR), or do you prefer one of the other suggestions that have emerged:

- use `recording`
- use a new entity `device`
- use a new entity `sys` (based on `tracksys`)

oesteban:
I did read the initial message. But rather than a solution to a problem
within that BEP, it is presented as a current problem of BIDS.

Unless an example fully compliant with current BIDS is given, I can only
see this as a solution to an issue we don't have, and we may not have it in
the future either.

The eye tracking BEP020 avoided creating new suffixes and would be solid
alternative to represent physio data with minimal changes to BIDS (no new
entities).

drammock:
> I did read the initial message. But rather than a solution to a problem within that BEP, it is presented as a current problem of BIDS.

OK, that was not my intent, so apologies that the original message wasn't clear. Hopefully this clarifies:

*We have enountered a problem in preparing a BEP for electromyography (cf #1998) that we want to get wider input on. It is common to record EMG at different sites from multiple devices simultaneously. Sometimes the devices feed into the same amplifier and the data end up in the same file, sometimes that is not the case. When it's not, we need an entity to distinguish data files from devices recording simultaneously. We're considering using `acq` but its current definition doesn't clearly reflect our use case. Here's a proposed definition update that would meet our needs. Open to other suggestions.*

> The eye tracking BEP020 avoided creating new suffixes and would be solid alternative to represent physio data with minimal changes to BIDS (no new entities).

Perhaps. But this PR as written *isn't proposing any new entities*, it's proposing to (slightly) expand the definition of an existing entity. @effigies then suggested using `recording` (also not new). @JuliusWelzel had previously suggested `sys` (as analogous to `tracksys`) which is technically new, but with the understanding that `tracksys` would be replaced with the more general `sys` afterwards. The only *truly* new proposed entity was @yarikoptic's idea to use `device` (which was already proposed elsewhere in #2027).

If your feedback is "there is no need for an EMG modality, just use `_physio` analogous to what we're proposing in BEP020" then I think that feedback belongs over at #1998, not here. Reading between the lines, however, your example filenames are using `recording` so I guess for the question at hand, you're voting for `recording` over the other proposed entities?

oesteban:
If this PR is merged but BEP042 finally doesn't use it or uses it differently, then we would be adding something already abandoned with this PR. I believe this should be discussed there already.

Happy to join the discussion of BEP042 and make a case for the BEP020 rationale. That said, I'll try to push BEP020 over the finish line before I can dedicate any time to another physiology BEP.

neuromechanist:
FWIW, I also support the `dev`/device entity. It is general enough to cover many cases (unless there is a multimodal recording device, such as a wearable, in which I think different channels will have their channel `type` anyways). This entity also provides the opportunity to also introduce `devcies.tsv/json` down the road to specifically describe the list of recording instruments used across modalities. 

@oesteban, You are more than welcome to join the BEP042 #1998 or #1371 discussion. I look forward to discussing the possibility of including electromyography on different limbs as a part of Eye-tracking specifications.

IMO, the rationale for this PR is quite clear and beyond #1998 as comments provided by the stakehodes from other BEPs also point out: Some modalities might have concurrent data recordings that, for some technical reasons, can't be part fo the same file. This can be different tracking systems in Motion-BIDS, concurrent video recordings in #1771, multiple Physio recordings in BEP045 (PhysioBIDS, #1675) and also recording EMG using more than one device.

Since these efforts are going quite in parallel, it made sense to create this PR to coordinate with all stakeholders to harmonize the use of an entity that refers to such concurrent recording across all data type/modality.

>Is there a risk that this PR goes through and #1998 does not?

Yes, but that is quite minimal. Even in that case, there are still several other potential BEPs that could use this PR for their specifications.

oesteban:
> FWIW, I also support the `dev`/device entity. It is general enough to cover many cases (unless there is a multimodal recording device, such as a wearable, in which I think different channels will have their channel `type` anyways). This entity also provides the opportunity to introduce `devices.tsv/json` down the road to describe the list of recording instruments used across modalities.

Sure, that's not what's being proposed here, and considering the volume of such a change, a device entity would likely require its own BEP because it conflicts with current BIDS foundations (which may be wrong, I'm not entering that debate). Let's focus on this particular PR.

> @oesteban, You are more than welcome to join the BEP042 #1998 or #1371 discussion. I look forward to discussing the possibility of including electromyography on different limbs as a part of Eye-tracking specifications. An argument might be whether there is a need for separate Eye-tracking or EMG BEPs, and shouldn't all be consolidated into BEP045 (Physio-BIDS, #1675). Although this discussion has been done here and there, it might be worth consolidation in a separate Issue.

Thank you for the invitation—I certainly appreciate the openness to broader discussions. My main point here was procedural clarity: the BEP mechanism is designed precisely to handle these broader, cross-cutting discussions.

BIDS established the BEP mechanism for large changes so those iterations over the spec are made consistently. Splitting this from those BEPs is inconsistent and risks introducing a problem. If there's a dependency relationship between BEPs (and that sounds reasonable and likely), then the solution is to make BEP045 contingent on the outcome of BEP042, so if this change to acq- is accepted there, then BEP045 would just piggyback on that change.

As you indicate at the end, I would not call for caution if this were just a *separate issue*—however, it's a PR that may introduce a change to the spec. 
 
> IMO, the rationale for this PR is quite clear and beyond #1998 as comments provided by the stakeholders from other BEPs also point out: Some modalities might have concurrent data recordings that, for some technical reasons, can't be part fo the same file. This can be different tracking systems in Motion-BIDS, concurrent video recordings in #1771, multiple Physio recordings in BEP045 and also recording EMG using more than one device.

I understand and appreciate that rationale. The issue I'm raising here is one of scope and procedure. The PR mechanism in BIDS is typically reserved for incremental, less controversial changes or bug fixes. Given the cross-cutting nature of the proposed change (affecting multiple BEPs), addressing this change first as an issue calling for opinions or as part of BEP042 seems more appropriate. This would ensure that all stakeholders have a clear understanding and agreement before making permanent updates to the specification. 

In general, BIDS evolves the spec BEP-by-BEP. The PR mechanism is reserved to fix uncontroversial, safe changes.

> Since these efforts are going quite in parallel, it made sense to create this PR to coordinate with all stakeholders to harmonize the use of an entity that refers to such concurrent recording across all data types/modalities.

I completely agree and support the idea of coordination and harmonization across BEPs—it is indeed essential. My intention isn't to discourage coordination but rather to ensure it happens in a structured way that respects established BIDS procedures. Perhaps an issue explicitly aimed at harmonizing these parallel discussions, gathering input from all stakeholders upfront, might serve better than immediately proceeding with a specification-changing PR.

> 
> > Is there a risk that this PR goes through and #1998 does not?
> 
> Yes, but that is quite minimal. Even in that case, there are still several other potential BEPs that could use this PR for their specifications.

I appreciate your assessment here. My perspective is simply that any risk—minimal as it may seem—of prematurely committing to a specification change that could affect ongoing BEPs warrants caution. It's precisely to manage this kind of risk that BIDS established the BEP mechanism.

Overall, my recommendation is procedural rather than conceptual: let's ensure we're clear on dependencies, implications, and cross-BEP coordination before committing to this PR. I believe that will strengthen our collective efforts, reduce potential friction, and better serve the community in the long run.

drammock:
I appreciate you taking the time to clearly explain your rationale @oesteban.

> Overall, my recommendation is procedural rather than conceptual: let's ensure we're clear on dependencies, implications, and cross-BEP coordination before committing to this PR. I believe that will strengthen our collective efforts, reduce potential friction, and better serve the community in the long run.

That makes sense to me, and I think the root of this dialog is that I am used to a different set of norms around the distinction between "issue" and "pull request".  In the dev communities  where I spend most of my time, it's OK to open pull requests as a conversation-starter/request-for-input if it makes it easier to see exactly what is being suggested (via the diff view on GitHub). I probably should have made it clearer that such was my intent (at a minimum marking the PR as draft, which I'll do now).

> The PR mechanism in BIDS is typically reserved for incremental, less controversial changes or bug fixes. Given the cross-cutting nature of the proposed change (affecting multiple BEPs), addressing this change first as an issue calling for opinions or as part of BEP042 seems more appropriate.

I'm happy to work within the norms of this community, now that I know what some of them are.
That said, I do feel you're being a bit harsh here, given that the discussion about this *was* started within the BEP, and it was [suggested that we get wider input **via a PR**](https://github.com/bids-standard/bids-specification/pull/1998#issuecomment-2723935232). I'm not trying to shift blame here, rather just pointing out that I'm apparently not the only one who thought a proposal-as-PR would be OK (and AFAICT nobody else who commented before you was bothered by its status as PR vs issue).

oesteban:
> That makes sense to me, and I think the root of this dialog is that I am used to a different set of norms around the distinction between "issue" and "pull request". In the dev communities where I spend most of my time, it's OK to open pull requests as a conversation-starter/request-for-input if it makes it easier to see exactly what is being suggested (via the diff view on GitHub). I probably should have made it clearer that such was my intent (at a minimum marking the PR as draft, which I'll do now).

Thanks for clarifying! I completely understand where you're coming from. BIDS BEPs function similarly to Python's PEPs (and other large-scale projects), guiding broader discussions through established processes to ensure changes have solid consensus. Given how critical backward compatibility is for BIDS, each PR tends to be carefully evaluated. Even changes that appear straightforward at first glance can have unforeseen implications later on, which is why careful review is so important.

> That said, I do feel you're being a bit harsh here, given that the discussion about this was started within the BEP, and it was https://github.com/bids-standard/bids-specification/pull/1998#issuecomment-2723935232.

I’m sorry if my feedback came across as harsh—this was definitely not my intent. I understand your approach better now after seeing how this discussion started within the BEP. For those like me who had not engaged in the BEP but received notifications of this PR, it seemed to represent an effort to push a change rather than a request for discussion. And as I mention above, changes to BIDS require rigorous assessment even if they seem innocuous.  Marking it as a draft now does help clarify your intent, and I appreciate you doing that.

I will try to look at BEP042 asap and provide feedback. That said, we are in the final stages of BEP020 and I think that could definitely change the way BEP042 and BEP045 are being approached. I would invite you, @neuromechanist, @JuliusWelzel and others to join our meeting on Thursday (info to join is given on the PR's message #1128).

neuromechanist:
I presented this idea during today's @bids-standard/maintainers. It seems that `recording-<label>` would require the least changes to the specifications to accommodate concurrent recordings.
This is of course in the context of EMG being a modality and data type, which is being discusses in #2108.

Maybe harmonizing the entity to indicate recordings from different instruments (`tracksys`, `recording`, and the proposed `dev` #2027) would be a topic for BIDS2.0

julia-pfarr:
@effigies and @neuromechanist I thought yesterday in the maintainers call we discussed that there will be a working group on this? Wouldn't approving and merging this PR interfere with that? We still didn't decide how to handle the two proposals (e.g., merging or not)

PR #2091, opened 2025-03-24
[FIX] Fix typos in specification about iEEG photo

qian-chu:
In the example:
1. Modality `ieeg` was missing from the filetree
2. `sub` and `ses` in the photo prefix didn't match the actual names
3. Remove unintended break

### DIFF
--- a/src/modality-specific-files/intracranial-electroencephalography.md
+++ b/src/modality-specific-files/intracranial-electroencephalography.md
@@ -495,11 +495,11 @@ A guide for using macros can be found at
 -->
 {{ MACROS___make_filetree_example(
    {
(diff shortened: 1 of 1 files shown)


effigies (approved):
LGTM. Thanks!

qian-chu:
Does anyone else want to take a look at the correction? Thanks!
(PS Error seems to come from an unrelated link issue shared with other PRs.)

effigies:
I think we should proceed by lazy consensus here, as this is a minor fix, not a new feature. I'm going to waive the second review requirement and merge, but others should feel free to back this out if it turns out they disagree.

PR #2097, opened 2025-04-04
[ENH] Specify that dir-<label> should correspond to PhaseEncodingDirection

effigies:
The BIDS convention is to use `dir-{AP,PA,LR,RL}` for phase-encoding directions in MRI filenames, as well as `PhaseEncodingDirection` in JSON sidecars. This duplication has the potential for inconsistency or multiple interpretations of the meaning of AP.

This PR aims to clarify the meaning of `dir-<label>` when its value matches a cardinal axis, and clearly state that this meaning SHOULD be consistent with the `PhaseEncodingDirection`.

### DIFF
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -316,6 +316,7 @@ properties:
       - xyzt_units
       - qform_code
       - sform_code
+      - axis_codes
     additionalProperties: false
     properties:
       dim_info:
@@ -404,6 +405,15 @@ properties:
         name: 'sform code'
         description: 'Use of the affine fields.'
         type: integer
+      axis_codes:
+        name: 'axis codes'
+        description: 'Orientation labels indicating primary direction of data axes.'
+        type: array
+        minItems: 3
+        maxItems: 3
+        items:
+          type: string
+          enum: ['R', 'L', 'A', 'P', 'S', 'I']
       mrs:
         name: 'NIfTI-MRS extension'
         description: 'NIfTI-MRS JSON fields'
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -407,7 +407,9 @@ properties:
         type: integer
       axis_codes:
         name: 'axis codes'
-        description: 'Orientation labels indicating primary direction of data axes defined with respect to the object of interest.'
+        description: >
+          Orientation labels indicating primary direction of data axes defined
+          with respect to the object of interest.
         type: array
         minItems: 3
         maxItems: 3
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -407,7 +407,7 @@ properties:
         type: integer
       axis_codes:
         name: 'axis codes'
-        description: 'Orientation labels indicating primary direction of data axes.'
+        description: 'Orientation labels indicating primary direction of data axes defined with respect to the object of interest.'
         type: array
         minItems: 3
         maxItems: 3
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -84,11 +84,16 @@ direction:
     Therefore, if the `dir-<label>` entity is present in a filename,
     `"PhaseEncodingDirection"` MUST be defined in the associated metadata.
 
-    Labels need not follow any particular convention,
-    and JSON metadata MUST take precedence in interpreting an image.
-    Generic labels, such as `dir-reference` and `dir-reversed` avoid any possible inconsistency.
-    `dir-j0` and `dir-j1` can indicate `"PhaseEncodingDirection": "j"` and `"j-"` (or vice versa),
+    Labels need not follow any particular convention and in case of conflict,
+    JSON metadata MUST take precedence in interpreting an image.
+
+    For example, `dir-j0` and `dir-j1` can indicate `"PhaseEncodingDirection": "j"` and `"j-"` (or vice versa),
     identifying the phase-encoding axis while avoiding indicating polarity.
+    However, in case of a dataset indicating `dir-j0` with `"PhaseEncodingDirection": "i"`, the
+    JSON metadata overrides the entity `dir-` setting.
+
+    The use of generic labels, such as `dir-reference` and `dir-reversed`, is RECOMMENDED
+    to avoid any possible inconsistency.
 
     To avoid inconsistency, if the `<label>` is `AP`, `PA`, `LR`, `RL`, `SI`, or `IS`,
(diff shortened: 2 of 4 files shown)


src/schema/objects/entities.yaml
```diff
@@ -83,7 +83,12 @@ direction:
     This entity represents the `"PhaseEncodingDirection"` metadata field.
     Therefore, if the `dir-<label>` entity is present in a filename,
     `"PhaseEncodingDirection"` MUST be defined in the associated metadata.
-    Please note that the `<label>` does not need to match the actual value of the field.
+
+    If the `<label>` is `AP`, `PA`, `LR`, `RL`, `SI`, or `IS`,
+    then the `"PhaseEncodingDirection"` SHOULD correspond to the data
+    axis that most closely aligns with the cardinal direction and orientation.
+    For example, `dir-AP` for an image in RAS+ orientation SHOULD have
+    `"PhaseEncodingDirection": "j-"`.
```
oesteban, line 91:
```suggestion
    It is RECOMMENDED not to encode the actual phase-encoding direction
    with the `<label>` to preempt conflicts with the `"PhaseEncodingDirection"`
    metadata.
    For example, this can be achieved by using indexes: `dir-0`, `dir-1` or
    labels without specifying the direction such as `dir-reference` and `dir-reversed`.
    Nonetheless, if `<label>` encodes the metadata by taking values such as 
    `AP`, `PA`, `LR`, `RL`, `SI`, or `IS`,
    then the `"PhaseEncodingDirection"` SHOULD correspond to the data
    axis that most closely aligns with the cardinal direction and orientation.
    For example, `dir-AP` for an image in RAS+ orientation SHOULD have
    `"PhaseEncodingDirection": "j-"`.
```

src/schema/rules/checks/nifti.yaml
```diff
@@ -49,3 +49,18 @@ XformCodes0:
     - nifti_header != null
   checks:
     - nifti_header.qform_code != 0 || nifti_header.sform_code != 0
+
+NiftiPEDir:
+  issue:
+    code: NIFTI_PE_DIRECTION_CONSISTENCY
+    message: |
+      Phase encoding direction is inconsistent with `dir-` entity.
```
oesteban, line 57:
```suggestion
      Phase encoding direction may be inconsistent with `dir-` entity.
```

oesteban (commented):
This is great to make the situation better without introducing backward incompatible changes. I made some suggestions to be more explicit about what the problem is and providing examples, that said, the phrasing will likely require improvements.

src/schema/objects/entities.yaml
effigies: This is introducing a new RECOMMENDation that entity values should not indicate the values of the associated metadata. This is not generally stated in BIDS, and runs counter to common usage; if you would like to make that proposal, I would probably start an issue or a mailing list thread to see if there's much support for it.

I don't want to try to introduce it here. Beyond it being out-of-scope for this fairly narrow attempt to warn on inconsistency, it would introduce a warning even when the metadata is consistent, which seems user-hostile.

That said, we can expand the previous "Please note that the `<label>` does not need to match the actual value of the field." into a brief discussion of non-axis-code conventions:

```suggestion
    Labels need not follow any particular convention,
    and JSON metadata MUST take precedence in interpreting an image.
    Generic labels, such as `dir-reference` and `dir-reversed` avoid any possible inconsistency.
    `dir-j0` and `dir-j1` can indicate `"PhaseEncodingDirection": "j"` and `"j-"` (or vice versa),
    identifying the phase-encoding axis while avoiding indicating polarity.
    
    To avoid inconsistency, if the `<label>` is `AP`, `PA`, `LR`, `RL`, `SI`, or `IS`,
    then the `"PhaseEncodingDirection"` SHOULD correspond to the data
    axis that most closely aligns with the cardinal direction and orientation.
    For example, `dir-PA` and `dir-AP` for an image in RAS+ orientation SHOULD have
    `"PhaseEncodingDirection": "j"` and `"j-"`, respectively.
```

src/schema/rules/checks/nifti.yaml
effigies: I'm not opposed to this, but I don't understand the case where this is triggered and the direction isn't inconsistent. We can expand on this text to explain why it's inconsistent in general terms, though we don't have a way to vary the text based on the values of fields.

effigies:
```suggestion
      The PhaseEncodingDirection and affine orientation indicate
      that the entity value (one of AP, PA, RL, LR, SI, IS) may be inaccurate.
```

src/schema/meta/context.yaml
```diff
@@ -405,6 +406,15 @@ properties:
         name: 'sform code'
         description: 'Use of the affine fields.'
         type: integer
+      axis_codes:
+        name: 'axis codes'
+        description: 'Orientation labels indicating primary direction of data axes.'
```
oesteban:
```suggestion
        description: 'Orientation labels indicating primary direction of data axes defined with respect to the object of interest.'
```

I think it is relevant to mention here the object of interest, so that we are generic enough (without taking a stance) on the debate in rodent imaging where it is unclear what is what.

src/schema/objects/entities.yaml
```diff
@@ -83,7 +83,18 @@ direction:
     This entity represents the `"PhaseEncodingDirection"` metadata field.
     Therefore, if the `dir-<label>` entity is present in a filename,
     `"PhaseEncodingDirection"` MUST be defined in the associated metadata.
-    Please note that the `<label>` does not need to match the actual value of the field.
+
+    Labels need not follow any particular convention,
+    and JSON metadata MUST take precedence in interpreting an image.
+    Generic labels, such as `dir-reference` and `dir-reversed` avoid any possible inconsistency.
+    `dir-j0` and `dir-j1` can indicate `"PhaseEncodingDirection": "j"` and `"j-"` (or vice versa),
+    identifying the phase-encoding axis while avoiding indicating polarity.
```
oesteban:
```suggestion
    Labels need not follow any particular convention and in case of conflict,
    JSON metadata MUST take precedence in interpreting an image.
    
    For example, `dir-j0` and `dir-j1` can indicate `"PhaseEncodingDirection": "j"` and `"j-"` (or vice versa),
    identifying the phase-encoding axis while avoiding indicating polarity.
    However, in case of a dataset indicating `dir-j0` with `"PhaseEncodingDirection": "i"`, the
    JSON metadata overrides the entity `dir-` setting.
    
    The use of generic labels, such as `dir-reference` and `dir-reversed`, is RECOMMENDED
    to avoid any possible inconsistency.
```

PR #2098, opened 2025-04-10
[SCHEMA] BEP028 - BIDS-Prov

bclenet:
Work in progress schema modification for the [BEP028 - BIDS-Prov](https://github.com/bids-standard/BEP028_BIDSprov).


src/schema/rules/files/common/modality_agnostic.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Files and directories that can be found at any level of a dataset.
+
+# Provenance json files
+provenance:
+  level: optional
+  path: prov/*
```
effigies: This should be compared to `phenotype/

```suggestion
  datatypes:
    - prov
```

src/schema/rules/files/common/modality_agnostic.yaml
```diff
@@ -0,0 +1,17 @@
...
+  suffixes:
+    - act
+    - base
+    - ent
+    - env
+    - soft
+  extensions:
+    - .json
+  entities:
+    prov: required
```
effigies: So this should be permitting:

```
/prov/prov-<label>_act.json
/prov/prov-<label>_base.json
/prov/prov-<label>_ent.json
/prov/prov-<label>_env.json
/prov/prov-<label>_soft.json
```

This is the intent?

src/schema/rules/directories.yaml
```diff
@@ -72,6 +77,7 @@ raw:
     value: datatype
     level: required
     opaque: false
+    subdirs:
```
effigies:
```suggestion
```

src/schema/rules/files/common/modality_agnostic.yaml
bclenet: We would like to allow:
```
/prov/[<subdirs>/]prov-<label>_act.json
/prov/[<subdirs>/]prov-<label>_ent.json
/prov/[<subdirs>/]prov-<label>_env.json
/prov/[<subdirs>/]prov-<label>_soft.json
```

(And I need to remove the base suffix)

bclenet: Also, is it ok to introduce the new `src/schema/rules/files/common/modality_agnostic.yaml` file in the schema ?
Is there something to do for this to be taken into account ?

bclenet:
Modification were added in #2099

PR #2100, opened 2025-04-14
[ENH] Allow direction entity in MESE files

tsalo:
Closes none.

Changes proposed:
- Split MESE and MEGRE suffix entity sets.
- Add optional direction entity to the raw MESE anatomical suffix.

### DIFF
--- a/src/schema/rules/files/deriv/preprocessed_data.yaml
+++ b/src/schema/rules/files/deriv/preprocessed_data.yaml
@@ -26,12 +26,21 @@ anat_defacemask_common:
     space: optional
     description: optional
 
-anat_multiecho_common:
+anat_megre_common:
   selectors:
     - dataset.dataset_description.DatasetType == 'derivative'
-  $ref: rules.files.raw.anat.multiecho
+  $ref: rules.files.raw.anat.megre
   entities:
-    $ref: rules.files.raw.anat.multiecho.entities
+    $ref: rules.files.raw.anat.megre.entities
+    space: optional
+    description: optional
+
+anat_mese_common:
+  selectors:
+    - dataset.dataset_description.DatasetType == 'derivative'
+  $ref: rules.files.raw.anat.mese
+  entities:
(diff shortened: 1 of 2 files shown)


effigies:
What's the use case?

tsalo:
Apparently we plan to use the AP and PA MESE scans for distortion correction.

oesteban:
In anatomical images, I think this would be an acceptable use of acq-, but I'd be okay with this change in general.



> Apparently we plan to use the AP and PA MESE scans for distortion correction.

Please make sure there's nothing in the B0FieldIdentifier/B0FieldSources interface that disallows it from setting identifiers on these, so distortion  can be estimated.

effigies:
Okay, so you have a use case for `AP/PA` MESE scans, but what about all the other groups? Do you get `AP/PA` T1w scans, or is this more in-passing to avoid doing this piecemeal, if other use cases are found?

At least for parametric volumes, I would assume that directions would be inapplicable by the time the final volumes are generated, similar to how `echo`, `flip` and `part` no longer apply.

I am hesitant to add more to the nonparametric group without a specific use case.

oesteban:
> Okay, so you have a use case for `AP/PA` MESE scans, but what about all the other groups?

If the only use case is fieldmap estimation, then acq- plus ensuring `B0FieldIdentifier` is explicitly allowed in their metadata should be sufficient to encode this.

yarikoptic:
If there is operator ability to specify direction in anatomicals, we should allow for this entity. IMHO, overall, when there is appropriate targeted entity, we better allow for it instead of inspiring to absorb into some more generic one (like _acq).

effigies:
I agree that we shouldn't overload `acq`, and am happy to approve this for MESE. My hesitation is over a lack of use-case for other suffixes.

tsalo:
Okay, I've limited my change to the MESE and MEGRE suffixes. I don't have any multi-PED MEGRE data, but I figured it was easier to just add direction to that one entity set than to split the rule.

PR #2102, opened 2025-04-14
Fix log line %interpolation in validator.py

yarikoptic:
I am surprised that none of the tests picked it up


yarikoptic:
A quick release here would be appreciated!

effigies (approved):

aaronkanzer:
@yarikoptic does this need to get released? Seemed it was merged, but not released

PR #2104, opened 2025-04-18
[ENH] Formalize presence of optional docs/ folder

yarikoptic:
Inspired by the success of the
https://github.com/bids-standard/bids-specification/pull/1962 adding formalization of `logs/` folder, and triggered by the use-case in DANDI https://github.com/dandi/dandi-docs/pull/200 seeking to add more elaborate descriptions to the README (or more specifically README.md), I would like to propose adding any extra documentation and associated artwork (images, videos, etc) under `docs/` folder.  I do not think it is worth breaking down more (as videos/) etc, since it could confuse with data types (videos of behavior or alike).

Not sure if we would want to formalize inner structure anyhow. We could potentially recommend using `docs/images/` subfolder for images.

Then README.md on top level could potentially use those images via references to `docs/` paths.

This would provide further improvement for establishing 
- #1972

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -392,6 +392,7 @@ Derivatives can be stored/distributed in two ways:
     that were used to generate the derivatives.
     Likewise, any code used to generate the derivatives from the source data
     MAY be included in the `code/` subdirectory.
+    Extra documentation (and relevant images) MAY be included in the `docs/` subdirectory.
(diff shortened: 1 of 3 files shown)


effigies (approved):
Yeah, seems reasonable.

kabilar (commented):
Thank you, Yarik.

robertoostenveld (approved):
I like this, thanks!

PR #2105, opened 2025-04-18
[ENH] Clarify definition of space entity

effigies:
While working through BEP38 stuff, we saw the "space" entity definition reads more restrictive than it has generally been validated, and is overly-specific to the first few data types accepted into BIDS.

In particular,

> The `<label>` MUST be taken from one of the modality specific lists in the Coordinate Systems Appendix.

conflicts with the common derivatives specification that `SpatialReference` MUST be provided if `<label>` *is not* found in that list.

This update aims to be a bit more general while noting that additional restrictions may apply to the values of `<label>` in some cases.

> [!Tip]
>
> [**HTML preview of this PR**](insert URL to HTML preview once available)
>

### DIFF
--- a/src/schema/objects/entities.yaml
+++ b/src/schema/objects/entities.yaml
@@ -324,17 +324,20 @@ space:
   name: space
   display_name: Space
   description: |
-    The `space-<label>` entity can be used to indicate the way in which electrode positions are interpreted
-    (for EEG/MEG/iEEG data)
-    or the spatial reference to which a file has been aligned (for MRI data).
-    The `<label>` MUST be taken from one of the modality specific lists in the
-    [Coordinate Systems Appendix](SPEC_ROOT/appendices/coordinate-systems.md).
-    For example, for iEEG data, the restricted keywords listed under
-    [iEEG Specific Coordinate Systems](SPEC_ROOT/appendices/coordinate-systems.md#ieeg-specific-coordinate-systems)
-    are acceptable for `<label>`.
+    The `space-<label>` entity indicates the coordinate system for interpreting
+    spatial coordinates.
+    For sensor-based modalities, such as EEG/MEG/iEEG,
+    the space indicates how sensor positions are reported;
+    for imaging modalities,
+    the space indicates a reference image to which a file has been aligned.
+    A list of common spaces is defined in the
+    [Coordinate Systems Appendix](SPEC_ROOT/appendices/coordinate-systems.md),
+    and these SHOULD be used as `<label>` values for data reported in those
+    spaces.
 
-    For EEG/MEG/iEEG data, this entity can be applied to raw data,
-    but for other data types, it is restricted to derivative data.
+    Note that the space entity is broadly applicable to derivative data,
+    but is only defined for a subset of raw data types.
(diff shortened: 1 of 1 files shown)


oesteban (approved):
Thanks Chris!

src/schema/objects/entities.yaml
```diff
@@ -324,17 +324,20 @@ space:
...
+    for imaging modalities,
+    the space indicates a reference image to which a file has been aligned.
+    A list of common spaces is defined in the
+    [Coordinate Systems Appendix](SPEC_ROOT/appendices/coordinate-systems.md),
+    and these SHOULD be used as `<label>` values for data reported in those
+    spaces.
+
+    Note that the space entity is broadly applicable to derivative data,
+    but is only defined for a subset of raw data types.
+    For particular use cases, `<label>` may be restricted to a set of enumerated values.
```
yarikoptic, line 340: From reading this, I don't quite follow how such restrictions would be specified  and what are those particular use cases.

effigies: In https://bids-specification.readthedocs.io/en/latest/modality-specific-files/intracranial-electroencephalography.html#electrode-description-_electrodestsv:

> The optional [`space-<label>`](https://bids-specification.readthedocs.io/en/latest/appendices/entities.html#space) entity (`*[_space-<label>]_electrodes.tsv`) can be used to indicate the way in which electrode positions are interpreted. The space `<label>` MUST be taken from one of the modality-specific lists in the [Coordinate Systems Appendix](https://bids-specification.readthedocs.io/en/latest/appendices/coordinate-systems.html). For example for iEEG data, the restricted keywords listed under [iEEG Specific Coordinate Systems](https://bids-specification.readthedocs.io/en/latest/appendices/coordinate-systems.html#ieeg-specific-coordinate-systems) are acceptable for `<label>`.
> 
> For examples:
> 
>     `*_space-MNI152Lin` (electrodes are coregistred and scaled to a specific MNI template)
> 
>     `*_space-Talairach` (electrodes are coregistred and scaled to Talairach space)

src/schema/objects/entities.yaml
```diff
@@ -324,17 +324,20 @@ space:
...
+    the space indicates how sensor positions are reported;
+    for imaging modalities,
+    the space indicates a reference image to which a file has been aligned.
+    A list of common spaces is defined in the
+    [Coordinate Systems Appendix](SPEC_ROOT/appendices/coordinate-systems.md),
+    and these SHOULD be used as `<label>` values for data reported in those
+    spaces.
+
+    Note that the space entity is broadly applicable to derivative data,
+    but is only defined for a subset of raw data types.
```
PeerHerholz, line 339: Maybe it would be helpful for users to add a few examples here?

PeerHerholz (approved):
Thanks @effigies!  

I only added one comment re providing/linking to examples to make it easier for users but I don't think it's strictly necessary.

effigies:
I'm going to merge this since there were no specific requests for improvements. @yarikoptic and @PeerHerholz if you have suggestions for clarifications, please open a PR and ping me directly (off GitHub) for prompt review, as I have a *lot* of mentions from bids-standard, and so it's hard to filter things that need attention.

PR #2109, opened 2025-04-29
rf(schemacode): Update packaging, minimum Python, resource loading

effigies:
This PR is some house-keeping before some larger changes I would like to make to the bidsschematools package. Those changes include switching to PDM for writing build hooks to install a compiled JSON schema in the package. To do that, we need to switch to pyproject.toml, at which point we should probably drop Python 3.8 support and use the [acres](https://pypi.org/project/acres/) package to abstract away `importlib_resources`.

### DIFF
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -61,7 +61,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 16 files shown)


effigies (commented):
A couple notes on less obvious bits for a reviewer.

effigies:
I removed an API change (removing `utils.get_bundled_schema_path()`) after seeing that someone is using it (https://github.com/search?q=get_bundled_schema_path&type=code).

Otherwise, this should be neutral from an API, CLI, schema and spec perspective, so I will merge by lazy consensus if I do not hear back by the end of week.

Issue #2055, opened 2025-02-14
new subsection in Appendix


CPernet:
### Your idea

Following a discussion on supporting new file format(s) - we realised that, while mentioned in published papers, we do not have an explicit explanation in the spec.

_**Proposal:**_ add a section in the appendix on supported file format, actual and future if one wants to propose to add a file format

Based on published papers and discussion at the steering group (February 2025), we propose that the guidelines are:

```
1. The format must be free and open with open-access documentation and an open-source implementation for reading and writing.
2. The format must be widely supported in external software.
3. The format must add value to the existing BIDS-supported formats or be widely adopted in existing datasets.
4. An active community should develop the format or have established and demonstrable procedures for proposal, review, and acceptance of fixes and enhancements from the public.
5. The format must meet the technical requirements of neuroscientific workflows, such as saving numerical data with high precision

These criteria will be reviewed by people previously or currently involved in the relevant domain area (e.g., microscopy, electrophysiology) and other formats development.
```

Question: do all criteria need to be met? @robertoostenveld is that the case of MEG for instance 
@effigies @yarikoptic @dorahermes

robertoostenveld:
When we wrote the MEG extension back in the period 2016-2018, there was no well-defined format that could be adopted as shared format, i.e., data converters and code for writing were limited. Hence we decided that to support all MEG systems we would allow their native file format (i.e., "Unprocessed MEG data MUST be stored in the native file format of the MEG instrument with which the data was collected."). For all MEG systems that we considered back then the file formats were sufficiently open, but considering [this appendix](https://bids-specification.readthedocs.io/en/stable/appendices/meg-file-formats.html) I would say that KDF and ITAB are not widely supported (violating item 2 on the list above). Also, I don't know of publicly accessible documentation on those formats.

By now we have seen that new MEG/OPM companies (FieldLine, Cerca) are able to adopt and write the FIF format. I would for MEG propose that the file format list is "closed" from now on and to encourage new vendors to adopt one of the existing formats in their software, rather than them adding new formats to BIDS. The recommended and most open format (regarding code, not documentation) and most widely used is FIF.

Note that reviewing the list above with the MEG file formats, the "open documentation" is also problem. There is open source code, but no open documentation for most (if not all).

robertoostenveld:
> 4. An active community should develop the format or have established and demonstrable procedures for proposal, review, and acceptance of fixes and enhancements from the public.

[...]

Issue #1552, opened 2023-07-12
DICOM "file-format" support


yarikoptic:
### Your idea

ATM this is a "satellite" issue for the #1551 where such support is to be discussed as a potential BEP.
Please direct all comments etc to that PR.

PR #2111, opened 2025-05-06
Add .nwb as supported format for EEG

yarikoptic:
NWB is BRAIN Initiative supported, and open data standard https://neurodatawithoutborders.github.io/ .

The main motivation here is to support joint iEEG and EEG datasets for EMBER archive project.  Quick google can show that there is a good number of studies which have conjoint EEG and iEEG.  Since .nwb is supported for iEEG, it makes only sense to add support for it in EEG modality as well.

There is a good number of .nwb "specific" tools in the ecosystem, with a notable mention of https://neurosift.app which allows "smart" client-side (in browser) access to remote .nwb files for exploration.

It seems that there is also growing number of toolkits in EEG world which support NWB.  E.g., EEGLAB already input support for NWB: https://eeglab.org/plugins/NWB-io/ so data would be accessible in this popular toolkit.

Added listing of NWB into EEG formats table + harmonize (change URL also) in iEEG section

Attn @bids-standard/raw-electrophys-eeg and @dorahermes


effigies (commented):
I support this on the grounds of consistency across modalities and working with other standards.

That said, as someone who wouldn't be tasked with implementing this, I would like to see support from EEG software groups, e.g., MNE-Python (@larsoner), FieldTrip (@robertoostenveld), etc.

robertoostenveld:
In theory FieldTrip has support for NWB as per https://github.com/fieldtrip/fieldtrip/pull/1419, but in practice it does not work. I just tried it (again, I recall having struggled with it before) and I cannot read any of the NWB test files that I have. 

From the FieldTrip pull request:

> It is assumed that the user is familiar with the basics of MatNWB (Github MatNWB) and the user's system is in a state that would also allow loading the data using MatNWB (MatNWB must be in the path, generateCore must have been run, the correct schema must be installed, see NWB schemas).

Those assumptions are apparently not met by me and/or my computer.

dorahermes:
This discussion is related to guidelines for new formats in issue #2055 
Point 5 is important: "The format must meet the technical requirements of neuroscientific workflows". In my interpretation, this means that developers of a format should make sure that common software packages should be able to read and write these data.

On the workflow side for iEEG, data are often collected concurrently with EEG, MEG, ECG etc and everything is collected on one amplifier and stored in one format. IMHO this does not mean that iEEG formats should propagate to such other formats.

yarikoptic:
> > It is assumed that the user is familiar with the basics of MatNWB (Github MatNWB) and the user's system is in a state that would also allow loading the data using MatNWB (MatNWB must be in the path, generateCore must have been run, the correct schema must be installed, see NWB schemas).
> 
> Those assumptions are apparently not met by me and/or my computer.

That PR is from half a decade ago, or are these observations current? I believe such assumptions were addressed or at least relaxed since then, e.g. see
- https://github.com/NeurodataWithoutBorders/matnwb/issues/493

FWIW, as for "one file format to rule them all" insofar I know only two overarching standards of such kind: DICOM and NWB ;-)
My concern here would indeed be NWB's adoption particularly for EEG data (it is **the** standard for microelectrode neurophysiology data  AFAIK, hence would be the main format for BEP032).  May be @rly and @oruebel could comment on, point to prominent examples?

hoechenberger:
I may be a bit naive, but I'm unable to find an official specification of the file format? I can only seem to find implementations, but not an authoritative spec? Can someone help me out here please?

cbrnr:
I 100% agree with @robertoostenveld. As a maintainer of MNE-Python, I don't think anyone should spend time implementing support unless (1) it is absolutely clear what the advantage of NWB is over other formats, and (2) it can be demonstrated that the format has been widely adopted.

effigies:
@hoechenberger https://nwb-schema.readthedocs.io/en/latest/format_description.html

hoechenberger:
Thank you!

hoechenberger:
Thank you!

robertoostenveld:
Interesting! I now see that it is based on HDF5 as the container file format and that NWB basically defines the schema that is used within the HDF5 container. As such it is actually quite similar to SNIRF for NIRS.

yarikoptic:
Correct, it is primarily a schema on top of HDF5. Moreover, due to inherent limitations of a single binary file approach, work is in progress to implement NWB on top of Zarr: this is where it is going on - https://github.com/hdmf-dev/hdmf-zarr (HDMF is core library, agnostic of NWB schema). 
Related to that, used by @magland's mentioned above neurosift  , he implemented https://github.com/NeurodataWithoutBorders/lindi which exposes internal to NWB schema as a json, which is actually a valid zarr archive. See e.g. https://lindi.neurosift.org/dandi/dandisets/000458/assets/6ab37be4-adfe-4bea-a031-eb1a2b0782a8/nwb.lindi.json  which is a lindi view of a nwb with EEG and behavioral data which could also be navigated at https://neurosift.app/nwb?url=https://api.dandiarchive.org/api/assets/6ab37be4-adfe-4bea-a031-eb1a2b0782a8/download/&dandisetId=000458&dandisetVersion=0.230317.0039 .

neuromechanist:
> My concern here would indeed be NWB's adoption particularly for EEG data

When looking at the newly merged #2055 as a [guideline](https://bids.neuroimaging.io/extensions/guidelines.html#use-existing-and-common-practicesformats), it seems all conditions are passing, except for adoption, which, IMO, is a major issue.

Playing with file format selector at NEMAR.org, EEG datasets (currently 303 datasets) are 53% SET files, 27% BVDF, and 20% EDF/BDF.
iEEG datasets (62 datasets) are 56% EDF, 30% BVDF, 11% MEF3, and **0% NWB**.

I wonder what is the incentive for BIDS data providers, curators, and software developers to replace their current method and adopt NWB?

To be fair, there is true value in using NWB, but adoption is not there. Using HDF5/NWB is awesome for data management and visualization (which was a motivation to support it in EEGLAB, AFAIK), but isn't that a task for dataset users/services?

yarikoptic:
Thank you @alexrockhill for sharing your perspective. Here is my condensed comment to the questions you posed + my answer to @effigies of what could be next:

BIDS grew from its initial `.nii.gz` + `.json` (since .nii.gz lacks metadata) + named folders + `subjects.tsv` (since we want to know basic demographics) into a standard which covers a **WIDE** range of domain specific file formats and metadata.  We use original file formats so that existing domain specific tools could use the files in BIDS datasets. IMHO it would be non-productive for us to try forcing underlying efforts like  OME-Zarr/NWB/DICOM/etc to adopt our BIDSy .json + .tsv files as the requirement for the metadata to accompany their already nicely modularized structured data formats. 

As a result, I take BIDS as a "harmonization + organization + summarization layer" on top of domain specific file formats/standards to organize them into datasets across different acquisitions, potentially from different modalities, typically of different file formats.  And somewhat more elaborate and visual on the topic of differences among standards -- see SfN 2025 poster we had https://zenodo.org/uploads/18333008

With that in mind, for this PR, I would

1. similarly to what I think was and will be done in virtually *any other* BEP -- I would look for alignment between metadata in BIDS sidecar .json + .tsv files to what is in NWB files: what metadata in NWB for EEG maps within BIDS EEG already or what is missing in NWB or BIDS? For example, taking ongoing [BEP032](https://bids.neuroimaging.io/bep032) which introduces support for .nwb, @CodyCBakerPhD in ongoing satellite to BEP032 effort [nwb2bids](https://nwb2bids.readthedocs.io) project established similar mappings: e.g.  for microephys `_channels.tsv` see [are table to alignment between BIDS and NWB](https://nwb2bids.readthedocs.io/en/latest/conversion_gallery.html#ecephys-channels). Such approach also loosely relates in what we do in schema ATM manually annotating correspondence to original DICOM fields, which might come handy later if we decide to adopt DICOM "file standard" within BIDS (#1552).

2. do work out a BIDS-aware converter for some less popular file format(s) which were permitted for EEG modality into how it would look like if converted to NWB, and potentially demonstrate advantages (e.g. visualization with neurosift). May be there is in openneuro some dataset with some .bidsignore'd eeg files in custom format? That is where I might look for assistance from @effigies to discover potential use case, and @CodyCBakerPhD , @rly and @oruebel to consult on representation in NWB.

3. potentially even recommend .nwb as the structure for some new data yet to be harmonized to BIDS which is neither of those permitted for EEG file formats.

alexrockhill:
Yeah I'm mixing a bit my opinion/desire for more nii-like formats (less extra stuff on top of the matrix) in BIDS and my opinion on NWB in BIDS. Always the critical tradeoff: accepting more formats and losing specification specificity and accepting fewer and making it harder to integrate more communities. I like the underlying hdf5 format, I hope the traction builds for more of that in BIDS.

yarikoptic:
FWIW, @arnodelorme reported at OHBM 2026 brainhack about https://github.com/sccn/nwbio

<details>
<summary>the slide -- click to expand</summary> 

image

</details>

so there is support in the flagman eeglab for EEG load(import)/save(export) in NWB

PR #2118, opened 2025-05-21
BF(test): reintroduce import of the load

yarikoptic:
This code seems got broken by #4604.  My semi-blind (did not test locally) attempt to fix

otherwise we have multiple CIs red ATM, e.g.

image

image


effigies (approved):
Thanks for catching this. Debugging this was on my todo list.

PR #2125, opened 2025-06-02
[SCHEMA] Allow HED has an optional column on selected files

VisLab:
HED used to be allowed in various `.tsv` files including the `participants.tsv`, `scans.tsv`, `sessions.tsv`, and `samples.tsv` as well as the `.tsv` files in the `phenotypes` directory. At some point this became disallowed, and the HED in these files may not be validated. The [`eeg_ds003645s_hed_demo`](https://github.com/bids-standard/bids-examples/tree/master/eeg_ds003645s_hed_demo) dataset in the bids examples illustrates this usage.

This PR modifies the schema to allow them again.  It does not address the `phenotypes` directory, as I am not sure how this should be done.

### DIFF
--- a/src/schema/rules/tabular_data/modality_agnostic.yaml
+++ b/src/schema/rules/tabular_data/modality_agnostic.yaml
@@ -15,6 +15,7 @@ Participants:
     handedness: recommended
     strain: recommended
     strain_rrid: recommended
+    HED: optional
   index_columns: [participant_id]
   additional_columns: allowed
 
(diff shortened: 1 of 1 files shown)


effigies (approved):
I'm not sure this was disallowed so much as an oversight in schema development, that turned into dropped support when we switched from the legacy validator.

I think this patch could probably be made against the `maint/1.10.0` branch as a fixup to the schema.

Issue #2120, opened 2025-05-23
[BUG] unable to get schema path in bidsschematools


nbeliy:
### Describe your problem in detail.

Starting from version 1.0.7, I'm unable to get the path to the schema using `get_bundled_schema_path` function. I got the following error:

```python
from bidsschematools.utils import get_bundled_schema_path

get_bundled_schema_path()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 get_bundled_schema_path()

File ~/.virtualenvs/test.3.10/lib/python3.10/site-packages/bidsschematools/utils.py:18, in get_bundled_schema_path()
     10 def get_bundled_schema_path():
     11     """Get the path to the schema directory.
     12 
     13     Returns
   (...)
     16         Absolute path to the directory containing schema-related files.
     17     """
---> 18     return str(data.load_resource("schema"))

AttributeError: module 'bidsschematools.data' has no attribute 'load_resource'
```

This error appears in versions 1.07, 1.0.8, 1.0,9

### Describe what you expected.

In version 1.0.5 schema path is returned normally:

```python
from bidsschematools.utils import get_bundled_schema_path

print(get_bundled_schema_path())
/home/beliy/.virtualenvs/test.3.10/lib/python3.10/site-packages/bidsschematools/data/schema
```

### BIDS specification section

https://bids-specification.readthedocs.io/en/latest/...

effigies:
Ah, that's my bad. It should be just `data.load()`. Incidentally, what are you using this function for? It might be easier to just stop using it.

nbeliy:
I use bidsschema tools for my bidsification tool, and want to report the version and path to schema, for transparency.

effigies:
Would this be sufficient?

```python
    schema = bst.schema.load_schema()
    return schema.bids_version
```

nbeliy:
I would like to be able to get schema version without loading full schema in memory, that's why I was using the hidden `_get_bids_version`.

effigies:
So what happened here is https://github.com/bids-standard/bids-specification/pull/2115, where we started compiling the `schema.json` file during the package build, so we now include that instead of the `schema/` directory. Loading JSON is *dramatically* faster than the YAML directory, and it's half a megabyte of memory.

Probably the simplest way to get the versions is now:

```
import bidsschematools as bst
schema_version = bst.__version__
bids_version = bst.__bids_version__
```

[...]

PR #2129, opened 2025-06-05
[FIX] schemapath resolution in dev mode install of bidsschematools

CodyCBakerPhD:
I ran into some issues when trying to achieve a local dev install to thoroughly test the `bep032` branch on some real datasets

Investigation revealed that this file: https://github.com/bids-standard/bids-specification/blob/9540a823ed70fa65dcf914e992adabd11c7285fc/tools/schemacode/src/bidsschematools/data/schema, which effectively seems to be a symlink, was being passed diectly into the `Namespace.from_json` as a raw string, instead of being recognized as a path to a directory

This small fix resolved my problems and allowed local installation to occur smoothly as expected

Speculation: possibly related to #2120? A collaborator (@yarikoptic) who showed me how to do this weeks ago was able to do it themselves without any issues

### TODO
- [ ] Add self to contributing
- [ ] Increment patch version

### DIFF
--- a/tools/schemacode/src/bidsschematools/schema.py
+++ b/tools/schemacode/src/bidsschematools/schema.py
@@ -224,7 +224,13 @@ def load_schema(schema_path=None):
 
     # JSON file: just load it
     if schema_path.is_file():
-        return Namespace.from_json(schema_path.read_text())
+        content = schema_path.read_text()
+
+        # Bug in Windows dev mode; 'symlink' style to schema directory does not automatically resolve
+        if content[:3] == "../":
+            schema_path = (Path(__file__) / content).resolve()
+        else:
+            return Namespace.from_json(jsonstr=content)
 
     # YAML directory: load, dereference and set versions
     schema = Namespace.from_directory(schema_path)


CodyCBakerPhD:
Before doing so, I wanted to confirm that https://github.com/bids-standard/bids-specification/blob/fix_dev_install/src/schema/BIDS_VERSION is the right place to change to bump the version for this type of fix?

effigies:
Is this a Windows issue? It would be good to test this.

effigies:
No worries. Just want to understand. Before merging this, I want to add CI tests for Linux/Mac/Windows to verify that loading the schema from full/editable installs works as expected. If you're up to exploring that, feel free. I'm currently working on a different PR on this repo, so will fiddle around with this later.

effigies:
I think I would add an `install_type: [wheel]` line to the build matrix:

https://github.com/bids-standard/bids-specification/blob/11611bd09862cc441986d87e7366ca73defb8470/.github/workflows/schemacode_ci.yml#L57-L62

```diff
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        install-type: [wheel]
         include:
           - os: macos-latest
             python-version: 3
+            install-type: wheel
           - os: windows-latest
             python-version: 3
+            install-type: wheel
```

Then add some inclusions for `install-type: dev`. Then you could duplicate this step, making one conditional on wheel and one on dev.

https://github.com/bids-standard/bids-specification/blob/11611bd09862cc441986d87e7366ca73defb8470/.github/workflows/schemacode_ci.yml#L82-L84

effigies:
One last thing: What is the actual command / Python code you run to induce the error?

CodyCBakerPhD:
So far as I can tell, it is simply that one file which git clones as a text file containing the symlink target, and is not established as an actual symlink

Speculation: using a more magical built-in `pathlib` logic to generate the symlink target (or other form of navigation) in-place might be a more robust solution overall than using a static file (symlink or no)

But I feel like I am only one with a very rare particular problem 🤷‍♂️ so maybe no need for over-engineering...

CodyCBakerPhD:
@effigies Sure thing - re-implemented that and it works fine

Last question is how you want to handle version bump?

effigies:
There's no need. It'll be in the next release. I'll backport this onto `maint/1.10.0` so that it will be in any future releases we make there, also.

Also, can you add yourself to https://github.com/bids-standard/bids-specification/wiki/Recent-Contributors? The change to `contributors.md` would get wiped out at the next release, otherwise.

PR #2132, opened 2025-06-05
[FIX] BEP032 - Correcting validations against real data example

CodyCBakerPhD:
With respect to BEP032, I am preparing a real data example to be hosted on the DANDI archive featuring multi-probe NeuroPixels 2.0

This branch contains essential fixes to the main `bep032` branch to get the `deno`-based BIDS validator to pass without (major) errors

Also with respect to allowing `nwb2bids` package to work as expected in conjunction with the BEP


CodyCBakerPhD:
There may be a little bit more work needed so I will keep in draft until ready to ping @yarikoptic on

CodyCBakerPhD:
@yarikoptic This small change is probably good to go - I'll add any more (including `rules`) in small follow-ups

PR #2133, opened 2025-06-07
feat: Generate context types from schema

effigies:
This PR follows-up #2115 by procedurally generating Python types from the `schema.meta.context` object. This uses [structural subtyping](https://typing.python.org/en/latest/spec/protocol.html#protocols) to allow for a variety of implementations to satisfy type-checkers, while also providing default standard library dataclass implementations.

The dataclass fields are annotated with protocols instead of one another, so that you can flexibly mix-and-match the provided dataclasses and your own classes, for example:

```python
from functools import cached_property
import json

import attrs
from bidsschematools.types import context as ctx

@attrs.define
class Dataset:
    tree: FileTree
    ignored: list[str]

    @cached_property
    def dataset_description(self) -> dict[str, Any]:
        with open(tree['dataset_description.json']) as fobj:
            return json.load(fobj)

    @cached_property
    def modalities(self) -> list[str]:
        ...

    @cached_property
    def datatypes(self) -> list[str]:
        ...

    @cached_property
    def subjects(self) -> ctx.Subjects:
        ...
```

`Dataset` is assignable to a variable annotated as `bidsschematools.types.protocols.Dataset`, but not `bidsschematools.types.context.Dataset`.

Protocols doc: https://bidsschematools--2133.org.readthedocs.build/en/2133/api/bidsschematools.types.protocols.html
Dataclasses doc: https://bidsschematools--2133.org.readthedocs.build/en/2133/api/bidsschematools.types.context.html

As with #2115, the goal here is to remain dynamic in editable installation mode, constructing and `exec`ing the module on-the-fly, while freezing a copy for optimal loading when performing full installations or building packages.

### DIFF
--- a/.github/workflows/schemacode_ci.yml
+++ b/.github/workflows/schemacode_ci.yml
@@ -57,18 +57,18 @@ jobs:
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         include:
           - os: macos-latest
-            python-version: 3
+            python-version: 3.13
           - os: windows-latest
-            python-version: 3
+            python-version: 3.13
     name: ${{ matrix.os }} with Python ${{ matrix.python-version }}
     steps:
       - uses: actions/checkout@v4
 
-      - name: "Set up Python"
-        uses: actions/setup-python@v5
+      - name: Set up Python ${{ matrix.python-version }} (uv)
+        uses: astral-sh/setup-uv@v6
         with:
           python-version: ${{ matrix.python-version }}
-          allow-prereleases: true
+          activate-environment: true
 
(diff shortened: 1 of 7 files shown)


effigies:
Documentation here: https://bidsschematools--2133.org.readthedocs.build/en/2133/api/bidsschematools.types.context.html

image

effigies:
@ubdbra001 Just a heads up on this. This is going to replace much of https://github.com/bids-standard/python-validator/pull/24, which will instead do something like what I have above.

bendhouseart (approved):
Very, very, cool 👍

Issue #2108, opened 2025-04-29
Defining Criteria for Data Types and Modalities in BIDS


neuromechanist:
## Background

The Brain Imaging Data Structure (BIDS) specification currently distinguishes between data types (represented as subdirectories under each subject) and modalities (represented as file suffixes). However, there appears to be inconsistency in how these distinctions are made across different kinds of data.

Some examples of the current state:
- `anat` is a data type for MRI anatomical recordings, with `t1w`, `t2w`, etc, as modalities
- `eeg`, `ieeg`, and `meg` are separate data types and modalities for different neural recording methods
- `motion` exists as its own data type and modality
- `physio` exists as a modality but not yet as a data type (being addressed in BEP045, #1675)
- Eye tracking is being added as a recording type under the `physio` modality (BEP020, #1128), with standalone eye tracking placed under the `beh` data type
- `emg` data type and modality is being proposed (BEP042, #1998)

## Current Discussion

There is an ongoing discussion under #1998 about whether certain data should:
1. Have their own dedicated data type and modality
2. Be incorporated under an existing umbrella data type
3. Be embedded within other modalities when appropriate

The discussion initiated as to whether EMG should:
- Have its own data type and modality (similar to EEG/MEG/iEEG)
- Be incorporated under the `physio` modality (similar to eye tracking in BEP020)

However, I believe that the scope of the issue is larger than EMG, and appreciate the community to provide their inputs.

## Key Considerations

When determining whether a data category deserves its own data type/modality or should be incorporated under an existing umbrella, the following factors have been raised:

1. **Signal source and nature**: Is the signal brain-derived vs. peripheral? Neural vs. non-neural?

2. **Data dimensionality and complexity**: Does the data have unique requirements in terms of channel count, sampling rate, or format that make existing structures insufficient?

3. **Research usage patterns**: Is the data commonly used as a standalone dataset, or primarily as an auxiliary measurement to other data types?

4. **Technical requirements**: Does the data require specific metadata fields, coordinate systems, or other specifications that don't align with existing structures?

5. **Community needs**: Is there sufficient research activity and community interest to warrant a dedicated structure?

6. **Fragmentation concerns**: Does creating a new data type/modality risk fragmenting the BIDS ecosystem unnecessarily?

7. **Consistency with existing structures**: How would the decision align with precedents set by other data types?

## Examples from Current Discussions

For EMG data (BEP042), arguments for a dedicated data type include:
[...]

PR #2135, opened 2025-06-10
[ENH] data-type recommendation based on simplicity.

neuromechanist:
Following a discussion in the BIDS maintainer meeting on June 10 on #2108 (defining criteria for data types and modalities), @yarikoptic suggested creating proposals to be considered by the community.

This proposal is compatible with the decisions made so far, and might help lay a path forward.

I formed my response to the issue as a dark PR for better visibility.

Please make suggestion, changes and edits regarding this proposal here. Please keep the general discussion under #2108. 

> [!Note]
> The rendered version is available here: https://bids-specification--2135.org.readthedocs.build/en/2135/common-principles.html#data-type-and-modality-selection

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -863,7 +863,74 @@ A guide for using macros can be found at
             }
         }
     }
-) }}
+  ) }}
+
+## Data type and modality selection
+
+BIDS is primarily for brain-related data, which should be interpreted as a research perspective rather than an inherent property of the data.
+For example, motion is a data type in BIDS because it informs brain activity research,
+while eye-tracking data in BIDS serves brain research perspectives.
+As long as data serves as a window to the brain or supports brain-related research objectives,
+it falls within BIDS' scope.
+
+When developing BIDS Extension Proposals (BEPs) that introduce new types of data,
+it is important to determine whether the data should:
+
+1.  Have its own dedicated data type and modality
+2.  Be incorporated under an existing data type
+3.  Be embedded within other modalities when appropriate
+
+### Simplicity principle
+
+Decisions about data organization SHOULD be guided by a simplicity principle
+that considers how easy it is to curate, manage, and reuse data.
+This principle can be quantified through the following criteria:
+
+1.  **Minimize file requirements**: Reduce the number of files needed (excluding sidecars) to describe new data
+
+1.  **Maximize reuse**: Keep data types, modalities, and entities minimal while reusing existing structures for new data
+    (metadata fields may change when new data is added to the specification)
+
+1.  **Avoid data type congestion**: The majority of data files in a host data type directory
+    should be directly related to that data type's primary purpose
+
+1.  **Balance reuse and congestion**: Balance the benefits of reusing existing structures against
+    the risk of overcrowding a data type with unrelated files
+
+1.  **Maintain data coherence**: Keep data with shared characteristics
+    (common acquisition clock, recording instrument, coordinate system) in one file
+    unless specific features (such as task, run, or data source) warrant separation
+
+### Application guidelines
+
+The simplicity principle SHOULD be applied early in BEP development based on:
+
+-   **File count analysis**: Consider the total number of files required to represent the data
+    and how this affects dataset organization
+
+-   **Usage patterns**: Evaluate whether the data is commonly used standalone
+    or primarily as auxiliary measurements alongside other data types
+
+-   **Technical requirements**: Assess whether the data requires specific metadata fields,
+    coordinate systems, or other specifications that align with existing structures
+
+-   **Community perspective**: Consider both data curator and end-user perspectives
+    when evaluating organizational approaches
+
+### Examples
+
+**Motion data**: Uses three core files (data, channels, events) plus sidecars,
+introduces one data type and essentially one entity while reusing important resources,
+avoiding congestion in other data types.
+
+**Electrocardiogram (EKG) data**: Can serve as an example of flexible placement options:
+
+-   As a channel in EEG recordings when recorded with the same instrument
+-   As a physio modality in a host EEG data type for standalone EKG with different instruments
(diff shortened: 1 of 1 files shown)


src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
-) }}
+  ) }}
+
+## Data type and modality selection
+
+BIDS is primarily for brain-related data, which should be interpreted as a research perspective rather than an inherent property of the data.
+For example, motion is a data type in BIDS because it informs brain activity research,
+while eye-tracking data in BIDS serves brain research perspectives.
+As long as data serves as a window to the brain or supports brain-related research objectives,
+it falls within BIDS' scope.
```
oesteban, line 874: Defining the project scope as a core principle is a great idea—well spotted! To keep everything consistent, could we move this definition into [`common-principles.yaml`](https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/common_principles.yaml)? Once it’s in place, we can fine-tune the exact wording alongside the other principles.

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
+
+1.  **Avoid data type congestion**: The majority of data files in a host data type directory
+    should be directly related to that data type's primary purpose
+
+1.  **Balance reuse and congestion**: Balance the benefits of reusing existing structures against
+    the risk of overcrowding a data type with unrelated files
+
+1.  **Maintain data coherence**: Keep data with shared characteristics
+    (common acquisition clock, recording instrument, coordinate system) in one file
+    unless specific features (such as task, run, or data source) warrant separation
```
oesteban, line 902: Similar to the previous point, this section reads more like governance guidance than a spec requirement. Perhaps we could move it into the governance documentation, keeping the spec text lean and focused on mandatory details.

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
+it falls within BIDS' scope.
+
+When developing BIDS Extension Proposals (BEPs) that introduce new types of data,
+it is important to determine whether the data should:
+
+1.  Have its own dedicated data type and modality
+2.  Be incorporated under an existing data type
+3.  Be embedded within other modalities when appropriate
+
+### Simplicity principle
```
oesteban, line 883: If we want “Simplicity” as a common principle, I suggest adding it to [`common-principles.yaml`](https://github.com/bids-standard/bids-specification/blob/master/src/schema/objects/common_principles.yaml) with a concise, broad definition. In adding new common principles, we must make sure we have evaluated the overlap/collision/compatibility with the existing principles (e.g., this one could possibly be overlaping the 80/20 rule)

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
+    and how this affects dataset organization
+
+-   **Usage patterns**: Evaluate whether the data is commonly used standalone
+    or primarily as auxiliary measurements alongside other data types
+
+-   **Technical requirements**: Assess whether the data requires specific metadata fields,
+    coordinate systems, or other specifications that align with existing structures
+
+-   **Community perspective**: Consider both data curator and end-user perspectives
+    when evaluating organizational approaches
```
oesteban, line 918: This is clearly guidelines. Should not be in the spec but in the guidelines.

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
+avoiding congestion in other data types.
+
+**Electrocardiogram (EKG) data**: Can serve as an example of flexible placement options:
+
+-   As a channel in EEG recordings when recorded with the same instrument
+-   As a physio modality in a host EEG data type for standalone EKG with different instruments
+-   In a dedicated physio data type (once physio-BEP is merged) when multiple physiological recordings
+    (eye-tracking, galvanic skin response, metabolics, SpO2, EKG) are present
+
+The simplicity principle guides the recommendation based on the specific dataset context and usage requirements.
```
oesteban, line 933: For maximum clarity, spec examples should show actual file structures, dataset layouts, or encoding snippets. Since this is primarily illustrative guidance, it may fit better within the BEP guidelines document rather than in the core spec.

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
             }
         }
     }
-) }}
+  ) }}
+
+## Data type and modality selection
```
oesteban, line 868: Thanks for clarifying the intent here! To align with your proposal to introduce a new common principle, could we retitle this heading so it directly reflects the scope? For example:

```suggestion
## BIDS' scope
```

src/common-principles.md
```diff
@@ -863,7 +863,74 @@ A guide for using macros can be found at
...
+while eye-tracking data in BIDS serves brain research perspectives.
+As long as data serves as a window to the brain or supports brain-related research objectives,
+it falls within BIDS' scope.
+
+When developing BIDS Extension Proposals (BEPs) that introduce new types of data,
+it is important to determine whether the data should:
+
+1.  Have its own dedicated data type and modality
+2.  Be incorporated under an existing data type
+3.  Be embedded within other modalities when appropriate
```
oesteban, line 881: It looks like this content serves more as a BEP guideline on how to implement the scope principle adequately. To maintain clear boundaries, let’s relocate it to the [bids-standard/bids-website](https://github.com/bids-standard/bids-website) once the new scope definition is accepted. We can then adapt the wording to fit the BEP guidelines format.

src/common-principles.md
neuromechanist: This is a good point. Ideally, any standard should define its scope early on. But, I predict that this would be a long discussion as well.
I do think BIDS needs a **Scope** statement, but probably under the `introduction.md`

PR #1714, opened 2024-02-28
BEP-038: Atlases

effigies:
This PR proposes the inclusion of BEP-038 into the BIDS specification.

The BEP is hosted on its own branch in the main repository, and BEP leads have been given write permissions to the branch. If you are a BEP lead and having issues writing to the branch, please let a maintainer know and we can make sure you have access.

At present this is a starting point, and I anticipate significant rewriting and reorganization before final merge. There is extensive discussion in #1281 that needs to be turned into concrete proposals. I would suggest keeping the discussion on this PR minimal so that newcomers can follow the thread. When opening PRs against this branch, please include a reference to this PR so that it shows up.

----

Related discussions

* #1281
* ... (to be added to)

PR #2140, opened 2025-06-18
feat(schema): Add meta.templates and allow for composing multiple references

effigies:
This is pulled out of #1714, which needs a way of constructing file rules by composing raw rules with rules adapted for atlas datasets. Since atlas datasets are just derivatives datasets, we can do much of the work in advance.


effigies:
@oesteban Pinging you because this is adapted from your work in BEP38. I'll also make a PR against BEP38 with the relevant changes.

rwblair (approved):
Looks good. Was debating if `null` needs a shout out in the documentation. Its meaning to schema writers is clear enough, and schema consumers don't need to know about what doesn't exist, so probably fine.

PR #2152, opened 2025-07-02
[BEP032] [FIX] Validation rules for events

CodyCBakerPhD:
This resolved some issues with my events file validation (under an ecephys directory following BEP032)

Included the other new suffixes for good measure

However, I get the following invalidation remaining on https://dandiarchive.org/dandiset/001470/draft/files?location=sub-M541/ses-M54120240831/ecephys

```bash
        [ERROR] REQUIRED_COORDSYSTEM If an electrodes.tsv file is provided, an associated coordsystem.json must also be present.

                /sub-M541/ses-M54120240831/ecephys/sub-M541_ses-M54120240831_electrodes.tsv - rules.checks.channels.RequiredCoordsystem

        Please visit https://neurostars.org/search?q=REQUIRED_COORDSYSTEM for existing conversations about this issue.

        [ERROR] SIDECAR_WITHOUT_DATAFILE A json sidecar file was found without a corresponding data file
                /sub-M541/ses-M54120240831/ecephys/sub-M541_ses-M54120240831_coordsystem.json
```
 which as you can see, has a `coordsystem.json` present (granted it's empty - still working on how the content should actually appear... but that wouldn't matter, right? Or at least give a different error like 'coordsystem content is empty'?)

Any ideas on resolving that issue @effigies?


src/schema/rules/files/raw/task.yaml
```diff
@@ -60,6 +60,18 @@ timeseries__mri_no_task:
...
 
+events__microephys:
+  $ref: rules.files.raw.task.events
+  datatypes:
+    - ecephys
+    - icephys
+  entities:
+    subject: required
+    session: optional
+    task: optional
```
yarikoptic, line 71: is that what we want?  (have task optional)

I think we better
- align with what most of other modalities mandate for entities
- do not add this section here, but rather add our two `datatypes` to the top record in this file on `events` which already covers most of other datatypes

CodyCBakerPhD: > is that what we want? (have task optional)

Yes, decided in meeting earlier. There might not always be a prescribed 'task' that can be labeled, nor multiple tasks to be distinguished



> do not add this section here, but rather add our two datatypes to the top record in this file on events which already covers most of other datatypes

That is how the PR started but then the other location was discovered, and that seems to be where all the other subspecialty types are grouped so we decided to follow established convention

PR #2158, opened 2025-07-19
[ENH] Add "Keywords" to `dataset_description.json`

markmikkelsen:
Keywords are standard in publications. Including them in BIDS as optional metadata will hopefully increase the visibility of datasets (perhaps this is something that can also be added to OpenNeuro).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
       - id: codespell
         args: ["--config=.codespellrc", "--dictionary=-", "--dictionary=.codespell_dict"]
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.17.0
+    rev: v1.17.1
     hooks:
       - id: mypy
         # Sync with project.optional-dependencies.typing
@@ -68,6 +68,6 @@ repos:
         args: ["tools/schemacode/src"]
         pass_filenames: false
   - repo: https://github.com/koalaman/shellcheck-precommit
-    rev: v0.10.0
+    rev: v0.11.0
     hooks:
       - id: shellcheck


src/modality-agnostic-files/dataset-description.md
```diff
@@ -33,6 +33,10 @@ and a guide for using macros can be found at
       "DatasetType": "RECOMMENDED",
       "License": "RECOMMENDED",
       "Authors": "RECOMMENDED if CITATION.cff is not present",
+      "Keywords": "OPTIONAL",
+      "Acknowledgements": "OPTIONAL",
+      "HowToAcknowledge": "OPTIONAL",
+      "Funding": "OPTIONAL",
```
effigies: Duplicate.

```suggestion
```

src/schema/objects/metadata.yaml
```diff
@@ -1728,6 +1728,16 @@ JNTANGChannelCount:
     Number of joint angle channels.
   type: integer
   minimum: 0
+Keywords:
+  name: Keywords
+  display_name: Keywords
+  description: |
+    List of three to six keywords that describe the content or subject matter of the dataset.
+    These are useful for enhancing the findability of the dataset through Search Engine
+    Optimization (SEO).
```
effigies: I would not mention numbers or SEO. Feels a bit out-of-place for just defining the meaning of the field.

```suggestion
    A list of keywords that describe the content or subject matter of the dataset.
    These may be used to enhance the findability of the dataset on public archives.
```

neuromechanist (approved):
This is a very nice addition.
I wonder if at a later date and PR, we'd want to consider a list of suggested keywords, which can be quite easily derived from semantic analysis of the current datasets?

PR #2161, opened 2025-08-04
[FIX][SCHEMA] make schema.rules.sidecars.anat consistent entitiy -> entities

AugustijnVrolijk:
My interpreter has currently got a hardcoded line to convert "entity" into "entities" as used everywhere else....


effigies:
Can you rebase against the `maint/1.10.0` branch (assuming this bug is present there)? Fixes in the schema that do not change the rules of BIDS should be able to be released without a new release of BIDS.

AugustijnVrolijk:
I think it should be rebased to maint/1.10.0 now

PR #1775, opened 2024-04-12
BIDS-2.0

yarikoptic:
References
- https://github.com/bids-standard/bids-2-devel/issues/57
- Full list of issues to consider: https://github.com/bids-standard/bids-2-devel/issues
- BIDS 2.0 Project of selected: https://github.com/orgs/bids-standard/projects/10

Plan
- [x] Initial "setup" of needed support tools/tests/changes
   - [x] Closes https://github.com/bids-standard/bids-2-devel/issues/67
   - [x] migrate all examples
   - [x] validate all examples **after** migration (some are excluded since not passing even in BIDS 1.x using deno validator)
- [x] Establish development procedures
   - **See [CONTRIBUTING-BIDS2.0.md](https://github.com/bids-standard/bids-specification/pull/1775/files#diff-11381827c2d95e8cdab4b87229fed3e742e25a41707e167d98d34ab12da44ef7)**
   - [ ] WiP -- being re-implemented/moved into https://github.com/bids-standard/bids-utils
  - diff between this branch and with all the "functional patches" applied: https://github.com/bids-standard/bids-specification/compare/bids-2.0...bids-2.0-patched
- Add patch sets to solve specific issues staged for BIDS 2.0 as outlined in https://github.com/orgs/bids-standard/projects/10
  - **Major** 
    - [ ] Closes https://github.com/bids-standard/bids-2-devel/issues/65
  - **Addressing quirks/inconsistencies of 1.x**. [Full list](https://github.com/bids-standard/bids-2-devel/issues?q=is%3Aopen%20is%3Aissue%20label%3Ainconsistency%20project%3Abids-standard%2F10)
    - [x] Closes https://github.com/bids-standard/bids-2-devel/issues/14
       - [ ] ?Introduce deprecation cycle within 2.0 to allow for coexistence of `participants.tsv` and `subjects.tsv` . 
    - [ ] Closes https://github.com/bids-standard/bids-2-devel/issues/69
    - [ ] Closes https://github.com/bids-standard/bids-2-devel/issues/78, which also generalization over
         - https://github.com/bids-standard/bids-2-devel/issues/15
         - WIP: https://github.com/bids-standard/bids-specification/pull/1821
   - **Making schema stricter** (incompatible within 1.x)
     - [ ] when https://github.com/bids-standard/bids-specification/pull/2400 merged -- add changing that 'warning' into 'error' 
     - [ ] go from warning to error for inheritance principle detected overwrites
[...]

PR #2162, opened 2025-08-04
[ENH] Deprecate 89+ string for default age column, increase expressiveness of column definitions in sidecar files

### LINKED ISSUE #2045: columns.yml: unclear difference in schema in defining within record or within "definition" sub-record

### LINKED ISSUE #2048: Provide clear(er) and visible to users distinction between columns which could be "overridden" in JSON sidecars and which could not
This issue inspired by 
- https://github.com/bids-standard/bids-specification/pull/2046/files#discussion_r1945589499

which is to address
- #2045 

to clarify the current state of schema/specification . Thus I do not think this issue is not pertinent to that PR which is not intended to change anything in the schema.

- #1838 

introduces notions of "fully-defined" and "conventional" columns, where "fully-defined" are defined through "BIDS schema" directly (`type`, `unit`, ...) and "conventional" defined through "JSON Sidecar schema" (in `definition`).

I think it does indeed make sense to "restrict" some column "properties" to indeed be 
- "fully-defined" (e.g. to disallow assignment of "Units" and "Levels" for `_id` fields);  but still allow to assign "TermURL" or even "Description"; or
- "conventional" - as allow to overload any (?) of the properties per our definitions in [common principles: tabular files: columns](https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files)
  - "user defined" are pretty much those "conventional".

The rule in #1838 on handling different types of columns says

> 1. If a column has a schema definition (type 1), validate using that schema. **Warn on attempted override.**
 
which makes it very user-visible and thus should be depicted in the specification.

But ATM

- in https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files I do not see any notion of different "kinds" of columns, and they all stated to allow e.g. `Units` specification, although for some (e.g. all the `{entity}_id`s makes no sense).
- In https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files.html#participants-file I see no statement that somehow `participant_id` is "fully defined" and the others in the list are "conventional"
- word "conventional" is nowhere in the spec

Related to this, at the level of schema: 
- seeing "JSON-schema" confused me into thinking we are somehow talking about [json-schema](https://json-schema.org/) but it is about "definitions allowed in JSON sidecar files per our [spec for .tsv files](https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files). I will call it "JSON Sidecar Schema"
- "explicit better than implicit" so I think reliance on seeing "JSON" record in YAML as indicator is quite suboptimal.  I think it would be better to be explicit in the schema and announce what "definitions" (Unit? TermURL? Levels? Description? or all of those) are not "overridable" for a given column
   - partially because of such "hidden semantic" there are entries with just Description overloaded with some shorter one thus causing duplication and ambiguity (which description to use where?)... e.g.

                species:
                  name: species
                  display_name: Species
                  description: |
                    The `species` column SHOULD be a binomial species name from the
                    [NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi)
                    (for example, `homo sapiens`, `mus musculus`, `rattus norvegicus`).
                    For backwards compatibility, if `species` is absent, the participant is assumed to be
                    `homo sapiens`.
                  definition: {
                    "Description":
                      "binomial species name from the NCBI Taxonomy (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi)"
                  }
                
- Related: [metadata fields (in .json files)](http://github.com/bids-standard/bids-specification/blob/HEAD/src/schema/objects/metadata.yaml) are very alike to tsv columns and already have many characteristics of the 'definition' **but** only using "schema way" and more flexible at times (`anyOf`s) and "Levels" defined via Enums... [ref bids-2-devel/issues/85 on metadata "life-cycle"](https://github.com/bids-standard/bids-2-devel/issues/85)

- It remains unclear why we need both ( as original "YAML schema" vs this "JSON Sidecar Schema"!? since it kinda happened historically -- may be should be harmonized somehow? e.g. go back to "YAML schema" but with explicit listing of what is disallowed and known mapping of what in "JSON sidecar schema" overloads "YAML schema" fields 

Before trying to propose solution, I have a few questions on @effigies statements in https://github.com/bids-standard/bids-specification/pull/1838

- > ... our column description object is not as powerful as JSON schema

on what aspects are missing from our  "BIDS YAML schema" which are present in "JSON Sidecar Schema"?  IMHO we should strive to equate them one way (make YAML schema more expressive) or another (just adopt "JSON Sidecar" uniformly).

- what "fully defined" really means, as whether any field is allowed to be overwritten?

effigies:
This is a minimal change to column JSON definitions to permit them to be as expressive as JSON schema for defining TSV columns. This adds:

1) `Format`, which covers both `type` and `format`, given that all TSV values are strings until cast by a parser.
2) `Minimum` and `Maximum`, which are used for some values. Here I permit them also to be strings, on the grounds that someone might want a minimum/maximum datetime.

This also deprecates the string "89+" for ages above 88, instead setting a `Maximum` of `89`. It is expected that a validator implementing this will carve out an exception for `age` for 89+. The alternative would be something along the lines of:

```
anyOf:
  - type: number
  - type: string
    enum:
      - "89+"
```

Adding the equivalent of `anyOf` to JSON sidecar column descriptions would be excessive complexity. Looking at the existing uses of `anyOf`, two are degenerate (`anyOf: {type: string}`) and one allowed strings or numbers, which is equivalent to `type: string` for TSV values.

At this point, this PR does *not* attempt to adopt a single convention for schema-defined and sidecar-defined columns. I think we probably should, and posted some options in https://github.com/bids-standard/bids-specification/issues/2048#issuecomment-3148785347.

- Closes #1791
- Closes #2045
- Closes #2048

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -2191,9 +2191,7 @@ Maximum:
   display_name: Maximum value
   description: |
     Maximum value a column entry is permitted to have.
-  anyOf:
-    - type: number
-    - type: string
+  type: number
 MeasurementToolMetadata:
   name: MeasurementToolMetadata
   display_name: Measurement Tool Metadata
@@ -2237,9 +2235,7 @@ Minimum:
   display_name: Minimum value
   description: |
     Minimum value a column entry is permitted to have.
-  anyOf:
-    - type: number
-    - type: string
+  type: number
 MiscChannelCount:
   name: MiscChannelCount
   display_name: Misc Channel Count


src/schema/objects/columns.yaml
```diff
@@ -550,7 +553,8 @@ species:
     `homo sapiens`.
   definition: {
     "Description":
-      "binomial species name from the NCBI Taxonomy (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi)"
+      "binomial species name from the NCBI Taxonomy (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi)",
```
CodyCBakerPhD:
```suggestion
      "Latin binomial species name from the NCBI Taxonomy (https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi).",
```

src/schema/objects/columns.yaml
```diff
@@ -592,7 +596,8 @@ strain:
     For species different from `homo sapiens`, string value indicating
     the strain of the species, for example: `C57BL/6J`.
   definition: {
-    "Description": "name of the strain of the species"
+    "Description": "name of the strain of the species",
```
CodyCBakerPhD:
```suggestion
    "Description": "Name of the strain of the species.",
```
Standardizing the punctuation among some of the changed descriptions (not holistic, just as I see them changed here)

CodyCBakerPhD:
@effigies Mostly LGTM as for what we had discussed - I will take a second look at src/schema/objects/metadata.yaml later to ensure I understand

Were you planning to submit a separate pre-PR to fix the mypy issues on `master`?

src/schema/objects/metadata.yaml
```diff
@@ -1199,6 +1199,39 @@ FlipAngle:
         unit: degree
         exclusiveMinimum: 0
         maximum: 360
+Format:
+  name: Format
+  display_name: Value format
+  description: |
+    Permitted formats for values in the described column.
+  type: string
+  # Keep synced with objects.formats
```
yarikoptic, line 1208: IMHO worth a unittest in bidsschematools to guarantee that, otherwise it is nearly guaranteed to diverge at some point (e.g. when objects.formats gets extended)

effigies: Done.

src/schema/objects/metadata.yaml
```diff
@@ -2144,6 +2177,14 @@ MaxMovement:
...
   type: number
   unit: mm
+Maximum:
+  name: Maximum
+  display_name: Maximum value
+  description: |
+    Maximum value a column entry is permitted to have.
+  anyOf:
+    - type: number
+    - type: string
```
yarikoptic: do you see (have) usecases for min/max of strings?  so far I see only numeric values used

```
❯ git grep -E '(min|max)imum: [0-9]' | wc -l
57
❯ git grep -E '(min|max)imum: [^0-9]' | wc -l
0
```

if not -- may be we should keep it restricted to numbers until use-case(s) come up?

src/schema/objects/columns.yaml
```diff
@@ -592,7 +596,8 @@ strain:
     For species different from `homo sapiens`, string value indicating
     the strain of the species, for example: `C57BL/6J`.
   definition: {
-    "Description": "name of the strain of the species"
+    "Description": "Name of the strain of the species.",
+    "Format": "string",
```
yarikoptic, line 600: I didn't know yaml is tolerant to those trailing `,` (like Python is), but I know that JSON is not ok with them, so may be we should avoid them?

although I see that we do have more already

```
❯ sed -n '/,$/ { N; /\n[[:space:]]*}/p; }' src/**/*.yaml
    "Units": "year",
  }
      "Ambidextrous": "Ambidextrous",
    }
      "other": "Other",
    }
```

src/schema/objects/columns.yaml
```diff
@@ -38,12 +38,14 @@ age:
...
-    It is recommended to tag participant ages that are 89 or higher as 89+,
-    for privacy purposes.
+    For privacy purposes, participant ages should be capped at 89.
+    Using "89+" for ages above 88 is DEPRECATED.
   definition: {
     "LongName": "Subject age",
     "Description": "Subject age in postnatal years",
+    "Format": "number",
     "Units": "year",
+    "Maximum": 89,
```
yarikoptic, line 48: this is a wonderful motivation for me to extract the `bst migrate-datasets` from https://github.com/bids-standard/bids-specification/pull/1775/files#diff-24b8a97543a1dc43190b8d5b070df26e1b88023f2b468b4483255af049788788R39 , likely to call `upgrade-datasets` instead.  And add migration for this so while going to newer version we could get `participants.tsv` fixed up automagically?

FTR -- there is a one bids-examples with 89+

```
❯ git grep 89\+ **/*.tsv
genetics_ukbb/participants.tsv:sub-05       89+     M       control 623589  yes     b002    SNV/CNV hg19/hg38       WES/SNP_intensity
genetics_ukbb/participants.tsv:sub-06       89+     F       patient 458712  no      b003    SNV     hg19    WES
genetics_ukbb/participants.tsv:sub-07       89+     M       patient 985632  yes     b004    SNV     hg19    WES
genetics_ukbb/participants.tsv:sub-13       89+     M       patient 225587  no      b001    SNV/CNV hg19/hg38       WES/SNP_intensity
```

src/schema/objects/metadata.yaml
effigies: datetimes

src/schema/objects/columns.yaml
CodyCBakerPhD: I'd say if this YAML file is meant to contain JSON strings, those strings should be loadable

As long as there is a simple automated test for that, it resolves any concerns I'd have. Preferably a pre-commit of some kind that can safely auto-push such changes

effigies: YAML is a superset of JSON. I just intended to make this bit look like JSON, but I wanted to take advantage of trailing commas to reduce diffs. If this is serialized to JSON, then the trailing commas will go away.

src/schema/objects/metadata.yaml
CodyCBakerPhD: This also threw me for a loop a bit

Can you provide documentation + examples of a string valued min/max syntax here? (only used to `minLength` and such to refer with repect to length of string)

src/schema/objects/columns.yaml
yarikoptic: found for myself in https://yaml.org/spec/1.2.2/#flow-collection-styles 

> Flow collection entries are terminated by the “,” indicator. The final “,” **may** be omitted. This does not cause ambiguity because flow collection entries can never be [completely empty](https://yaml.org/spec/1.2.2/#example-empty-content).

with emphasis on **may** mine. So you are right that YAML does support those!  I was just afraid that it might be some python yaml library "robustification" and thus could still lead to some problems with other yaml libraries.

src/schema/objects/metadata.yaml
yarikoptic: > datetimes

*cute* ! some folks could then setup min/max to e.g. have some "operationPerformed" to be within the duration of the study etc. Kinda opens possibilities but I do not see an immediate demand.

What worries me more is that it would likely mean that there should be custom format-specific comparator implemented to ensure that comparisons are done correctly if string representation allows for some "freedom" (e.g. leading 0s removed... `time` format regex seems to allow) which could alter the actual value/comparison. 

Thus I would just stay away for it until a specific case comes up to avoid implementation of comparisons within all of those known formats.

tools/schemacode/src/bidsschematools/tests/test_schema.py
```diff
@@ -182,6 +182,11 @@ def test_formats(schema_obj):
             )
 
 
+def test_format_consistency(schema_obj):
+    """Test that the "Format" field is consistent with objects.formats."""
+    assert set(schema_obj.objects.metadata.Format.enum) == schema_obj.objects.formats.keys()
```
CodyCBakerPhD, line 187: Do you think any additional testing would be worth it (e.g., example files asserting their column values to be of a particular type, and ensuring validation errors if not that particular non-string type, etc.) or is the assumption that the current testing suite would cover that?

effigies: I think it's out-of-scope for bidsschematools, but will definitely be in-scope for the validator.

src/schema/objects/metadata.yaml
effigies: Non-number Maximum/Minimum removed.

tools/schemacode/src/bidsschematools/tests/test_schema.py
CodyCBakerPhD: Did you want to start a sister PR over on the validator to ensure this works as expected or simply proceed and patch?

yarikoptic (approved):
It is great to see for specifications to converge!  You should come over to North more often!

PR #2172, opened 2025-08-17
[FIX] Update ResonantFrequency and SpectrometerFrequency to support both string/number and arrays

markmikkelsen:
Noticed that ResonantFrequency and SpectrometerFrequency required arrays. Should allow either a string or number, respectively, (most common case) or arrays of strings or numbers, respectively.


markmikkelsen:
I don't believe RF coil hardware has ever been designed to tune for more than two isotopes. So a maximum length of two would be reasonable.

src/schema/objects/metadata.yaml
```diff
@@ -3605,10 +3607,13 @@ SpectrometerFrequency:
...
-  items:
-    type: number
-    unit: MHz
+  anyOf:
+    - type: number
+      unit: MHz
+    - type: array
+      items:
+        type: number
+        unit: MHz
```
effigies, line 3619:
```suggestion
        unit: MHz
        minItems: 1
        maxItems: 2
```

src/schema/objects/metadata.yaml
```diff
@@ -3045,9 +3045,11 @@ ResonantNucleus:
...
     an array can be used: `["1H", "13C"]`.
     Corresponds to [DICOM Tag 0018, 9100](https://dicomlookup.com/dicomtags/(0018,9100)) `Resonant Nucleus`.
-  type: array
-  items:
-    type: string
+  anyOf:
+    - type: string
+    - type: array
+      items:
+        type: string
```
effigies, line 3053:
```suggestion
        type: string
        minItems: 1
        maxItems: 2
```

effigies (commented):
If minimum/maximum lengths are worth validating...

neuromechanist (approved):

effigies (approved):

PR #2173, opened 2025-08-18
Add 1PE to complement 2PE (microscopy)

yarikoptic:
https://blog.biodock.ai/one-vs-two-photon-microscopy/ seems to provide a nice comparison between one- and two-photon microscopy. Also
https://www.gophotonics.com/community/what-are-single-photon-and-two-photon-excitation provides another comparison and uses "excitation" consistently between them.

Note that there is also already "FLUO" suffix for "Fluorescence microscopy" in general, of which both 1PE and 2PE and likely MPE (multi-photon) mechanisms are, so unclear when FLUO to be used, but I guess 2PE, MPE and now 1PE should be RECOMMENDED to be used in favor of FLUO as more specific

Some related references:
- https://github.com/bids-standard/bids-specification/pull/1632

ping @bids-standard/bep031 @bendichter

### DIFF
--- a/src/modality-specific-files/microscopy.md
+++ b/src/modality-specific-files/microscopy.md
@@ -28,7 +28,7 @@ and a guide for using macros can be found at
  https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
 -->
 {{ MACROS___make_filename_template("raw", datatypes=["micr"], suffixes=["TEM", "SEM", "uCT", "BF", "DF",
-"PC", "DIC", "FLUO", "CONF", "PLI", "CARS", "2PE", "MPE", "SR", "NLO", "OCT", "SPIM", "XPCT"], n_dupes_to_combine=4) }}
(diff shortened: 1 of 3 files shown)


effigies:
cc @bids-standard/bep031

yarikoptic:
@CodyCBakerPhD @bendichter , may be you could assist
- how much of one-photon imaging data do we have in dandi?
- do you see anything critical missing from https://bids-specification.readthedocs.io/en/stable/modality-specific-files/microscopy.html#microscopy-metadata-sidecar-json which would needs to be there for 1PE in addition to 2PE or FLUO?

yarikoptic:
> @yarikoptic BIDS microscopy only applies to static images to my understanding.

as I have mentioned in 
- https://github.com/bids-standard/bids-specification/pull/1632#issuecomment-1769248031

there is nothing in BIDS microscopy restricting to anatomical, although indeed lacking desired temporal data formalized.

But overall questions mostly relevant to this particular PR: 
-given current listed support of 2PE and FLUO, would addition of 1PE be warranted for 1PE datasets (temporal data is not forbidden, thus IMHO permitted)? 
- if not - would FLUO be appropriate for 1PE (but at the cost of loosing expresiveness)? 
- Could there be FLUO which is not 1PE or 2PE?

PR #2179, opened 2025-08-22
[ENH] Clarify that task label is not required to derive from TaskName

### LINKED ISSUE #2081: Clarify relation of task entity to TaskName metadata using RFC 2119 MUST/SHOULD/MAY language
e.g., This file: https://github.com/bids-standard/bids-examples/blob/7dc73456fe41bf204c1c71f6512293d2f33c6686/eeg_matchingpennies/task-matchingpennies_eeg.json#L2

has as task name `"Matching Pennies"` although the label in the files is `task-matchingpennies`

I think the validator should ensure that these two "fields" match

effigies:
Following up on decision made in Copenhagen. This "is" was interpreted by some to be a requirement, but this has never been validated, would be breaking to turn into an error, and is impractical to encode in the schema. Rather than contort ourselves to devise a check, it makes more sense to use RFC 2119 language to make clear that this is an optional convention, not a requirement.

Closes #2081.

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -3753,7 +3753,7 @@ TaskName:
   description: |
     Name of the task.
     No two tasks should have the same name.
-    The task label included in the filename is derived from this `"TaskName"` field
+    The task label included in the filename MAY be derived from this `"TaskName"` field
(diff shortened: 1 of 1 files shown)


neuromechanist (approved):
LGTM.
Actually, I had someone reach out about this very sentence separately, agreed that *May* works best

PR #2184, opened 2025-08-25
[ENH] Formalize presence of optional scratch/ folder

michellewang:
As per @yarikoptic's suggestion, this PR proposes formalizing an optional `scratch/` subdirectory for storing temporary/working files. [Nipoppy](https://github.com/nipoppy/nipoppy/blob/main/README.md) projects have a `scratch/` directory, and it would be nice if it were added to the BIDS specification as well since we are moving towards aligning the Nipoppy project layout with the BIDS "study" dataset type.

I based this PR off of the changes made in #2104 for adding the `docs/` directory, though I notice that #1962 has slightly different changes for the `logs/` directory. I'm not sure what the differences in the schema changes are and if it would make sense to add [the additional changes from #1962](https://github.com/bids-standard/bids-specification/pull/1962/files#diff-978630391ad0c59128c99bf80532c444f7d075be5763e3f2b43839e078234ce2). @yarikoptic what do you think?

### DIFF
--- a/src/schema/rules/directories.yaml
+++ b/src/schema/rules/directories.yaml
@@ -23,6 +23,7 @@ study:
       - docs
       - derivatives
       - logs
+      - scratch
       - sourcedata
   code:
     name: code
@@ -38,6 +39,10 @@ study:
     opaque: true
   logs:
     name: logs
+    level: optional
+    opaque: true
+  scratch:
+    name: scratch
     level: optional
     opaque: true
   sourcedata:
@@ -76,10 +81,6 @@ raw:
     name: phenotype
     level: optional
     opaque: false
-  scratch:
-    name: scratch
-    level: optional
-    opaque: true
   sourcedata:
     name: sourcedata
     level: optional


yarikoptic:
heh heh -- just a curse of duplication! I, frankly, missed the fact that in https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/directories.yaml we describe directories separately for each of the dataset types, so I need to complement it now with the one for 
- #1972 

uff -- filed

- https://github.com/bids-standard/bids-specification/pull/2185
 
IMHO `scratch/` is good for all dataset types, so just list it for all.

effigies:
I oppose adding scratch directories to raw and derivative datasets. Those dataset types should be archive-ready, and a dumping ground of intermediate files does not belong in either, IMO.

yarikoptic:
Hi @effigies . Indeed I might have not thought it fully through!
The intent was more of an instruction to user *where* uniformly to have such a temporary location. But since it is **not** intended to be shipped with the dataset, it should not contain anything there during upload!  So, oddly, I tried to standardize the location which **must not** contain any files,  but for BIDS anything which is not prescribed by BIDS, is not legit BIDS... unless people do add it to `.bidsignore`, when it becomes ignored and still not legit.

So, even though I would have loved BIDS to prescribe as many locations as possible in a generic fashion, may be this "intermediate results" folder then should be "workflow specific" and instructed outside of the BIDS specification context.

michellewang:
@yarikoptic are you saying that `scratch` should not be in the schema even for the `study` dataset type? In that case in Nipoppy we will probably end up putting it in `.bidsignore`

PR #2185, opened 2025-08-26
Add "directories" description for "study" DatasetType and add "docs" where was missing

yarikoptic:
Apparently I have missed this file entirely whenever I was preparing

-  https://github.com/bids-standard/bids-specification/pull/1972 (study DatasetType)

and also due to all the duplication we (I) missed that docs was not listed among "root.subdirs" or not listed at all.

This would provide a fix, but I wonder if we could/should make it so we could avoid duplication altogether.  As I have argued in #1972 I feel that "study" is the base dataset type and next ones just add potentially more to them.  So may be we could come up with some more compact representation here... but not in this PR

### DIFF
--- a/src/schema/rules/directories.yaml
+++ b/src/schema/rules/directories.yaml
@@ -1,5 +1,5 @@
 ---
-# This file defines layouts of directories.
+# This file defines layouts of directories for each DatasetType
 #
 # A layout defines a collection of directory specifiers.
 # Each specifier has a naming convention, requirement level, opacity, and subdirectories.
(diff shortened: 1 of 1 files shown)


src/schema/rules/directories.yaml
```diff
@@ -16,10 +16,50 @@
...
+    level: optional
+    opaque: false
+  sourcedata:
+    name: sourcedata
+    level: optional
+    opaque: true
+  stimuli:
+    name: stimuli
+    level: optional
+    opaque: true
```
effigies: Are `phenotype/` and `stimuli/` expected to occur at the top level or inside a `sourcedata/` or `derivatives/` subdirectory?

yarikoptic: just like in "raw" - on the top hence we have them listed in `root:` above.  `sourcedata/` and `derivatives/` are `opaque` so we do not really describe what appears there, right?

or what am I missing?

effigies: I'm asking whether phenotype and stimuli should be removed from the top level of the study dataset, since they would presumably appear in the raw dataset.

yarikoptic: ah.  yeah, good point -- they would either be sources or in corresponding "raw" (or "derivative") datasets

Looking at discussions with @asmacdo and @michellewang  on
- https://github.com/nipoppy/nipoppy/pull/687

and the [diff](https://github.com/nipoppy/nipoppy/pull/687/files) there not using top level `stimuli/` or `phenotype/` , let's indeed kick them out from the `study` level description!

yarikoptic: pushed

julia-pfarr (approved):

jbpoline:
Not sure that this is the right place (still learning... apologies if that comment is to be moved) but it would be worth having the  "bids study" directory layout revisited.  Other place that might be impacted :
https://github.com/bids-standard/bids-specification/pull/1861)
https://github.com/bids-standard/bids-specification/pull/1972

PR #2203, opened 2025-09-05
[ENH] Clarify that "source datasets" do not have to be BIDS, and provide definitions for DOI and URL

yarikoptic:
came up while reviewing... but could be wrong and we demand only BIDS datasets there?

### DIFF
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -901,7 +901,7 @@ DOI:
   name: DOI
   display_name: DOI
   description: |
-    Used to specify the locations of all source datasets.
+    Used to specify the locations of all source datasets (BIDS or not).
     (see [URI](SPEC_ROOT/common-principles.md#uniform-resource-indicator))
   type: string
 DoseCalibrationFactor:
@@ -3543,7 +3543,7 @@ SourceDatasets:
   name: SourceDatasets
   display_name: Source Datasets
   description: |
-    Used to specify the locations and relevant attributes of all source datasets.
+    Used to specify the locations and relevant attributes of all source datasets (BIDS or not).
     Valid keys in each object include `"URL"`, `"DOI"` (see
     [URI](SPEC_ROOT/common-principles.md#uniform-resource-indicator)), and
     `"Version"` with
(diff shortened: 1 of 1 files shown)


src/schema/objects/metadata.yaml
```diff
@@ -3543,7 +3543,7 @@ SourceDatasets:
   name: SourceDatasets
   display_name: Source Datasets
   description: |
-    Used to specify the locations and relevant attributes of all source datasets.
+    Used to specify the locations and relevant attributes of all source datasets (BIDS or not).
```
effigies, line 3546: This seems fine. Duplicating in the DOI and URL fields seems excessive.

src/schema/objects/metadata.yaml
```diff
@@ -901,7 +901,7 @@ DOI:
   name: DOI
   display_name: DOI
   description: |
-    Used to specify the locations of all source datasets.
+    Used to specify the locations of all source datasets (BIDS or not).
```
yarikoptic: I think such wording is not even correct, 

```suggestion
    Used to specify the Digital Object Identifier (DOI).
```

as a generic thing, I do not think we should associate it here in documentation with source datasets... and it is not about location per se!

src/schema/objects/metadata.yaml
yarikoptic: I think it is just a result of excessive (if not wrong to start with) specification in both of those. I do not see a need to make them "source dataset specific" and clearly should not relate to "all".

src/schema/objects/metadata.yaml
```diff
@@ -3985,7 +3985,7 @@ URL:
   name: URL
   display_name: URL
   description: |
-    Used to specify the locations of all source datasets.
+    Used to specify the locations of all source datasets (BIDS or not).
```
yarikoptic:
```suggestion
    Used to specify the location (local or on Internet).
```
?

src/schema/objects/metadata.yaml
effigies: If it's a term, we might as well define it.

```suggestion
    A Digital Object Identifier (DOI) is a persistent, resolvable name for a digital resource.
```

src/schema/objects/metadata.yaml
effigies:
```suggestion
    A Uniform Resource Locator (URL) is a structured reference to an object that
    includes a resource transfer scheme and a location.
```

julia-pfarr (approved):

PR #2207, opened 2025-09-07
Update current BIDS maintainers list

sappelhoff:
Hi all,

with new priorities that came up for me, I have over the past year not been having enough time to make meaningful contributions to BIDS as a maintainer. Therefore, I would like to move my name from active to past BIDS maintainers.

That is the only change though -- as before I will still be available and involved in the project when needed, and/or when I find capacity to pick up tasks.

xref https://github.com/bids-standard/bids-website/commit/a84fa60634a441f41c3edade5a021bc4de749350


yarikoptic:
:hugs:

PR #2209, opened 2025-09-09
[ENH] Add motion entity table

JuliusWelzel:


effigies:
Thanks for catching this! Could you add nirs and mrs as well?

JuliusWelzel:
Sorry, that is true. I just revisited the definition of recording "[This entity is commonly applied when continuous recordings are from different acquisition instruments, or have different sampling frequencies or start times.]". I will remove it again.

tsalo (approved):
I checked the rendered docs and it looks good to me.

PR #2211, opened 2025-09-15
[ENH] BEP 16: Diffusion modeling derivatives

arokem:
This transfers work that so far has been done in a separate repository: https://github.com/bids-standard/bids-bep016. 

I believe this is not too far from ready to be reviewed by the community, but also seeking input from @Lestropie, @francopestilli, @oesteban, @PeerHerholz.


arokem:
Forgot to tag @mattcieslak, whose input here also useful. And probably others, please tag and add to this discussion.

PeerHerholz:
Hey @arokem and everyone,

thx @arokem! Yeah, I think once it's decided that the current `markdown` and respective information/proposals are in a good status we would create a couple of example datasets (I can update the [conversion software](https://github.com/PeerHerholz/bids_bep16_conv) respectively). I do agree that the examples should be relatively easy, especially considering that we would provide empty imaging files and focus on the `metadata`. The `schema` appears to be a different beast and having someone from the maintainers/developers involved would be great.

effigies:
> One general question: does the validator currently look into the content of data files to validate that their content is sensible?

It validates tsv contents and NIfTI headers.

Validating NIfTI data blocks is likely to be significantly slower, and would probably make sense for a separate tool like [miniQC](https://github.com/openneuroorg/miniqc).

What kind of check are you thinking?

arokem:
One thing that comes to mind is to validate the shape of the data. For example, the tensor should have six elements on the last dimension; CSD coefficients would also have a certain dimensionality that can be checked. This can be done without reading the data. So, if there is precedent for this kind of check, that's good.

effigies:
Yes, checking NIfTI shapes is doable:

https://github.com/bids-standard/bids-specification/blob/b9a66732e67de7bc51402b7ed0a73a1e58ce85d6/src/schema/rules/checks/func.yaml#L57-L67

You'll want to use `make_sidecar_table` if it corresponds to a sidecar schema rule, but for describing deeper models, you'll probably want `make_metadata_table`, e.g.,

https://github.com/bids-standard/bids-specification/blob/b9a66732e67de7bc51402b7ed0a73a1e58ce85d6/src/common-principles.md?plain=1#L539-L561

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,946 @@
...
+```JSON
+{
+    "Description": "Estimated signal intensity with no diffusion weighting, ie. S0",
+    "Model": {
+        "Description": "Diffusion Tensor",
+        "Parameters": {
+            "FitMethod": "ols",
+            "OutlierRejectionmethod": "None"
+        }
+    }
```
arokem, line 396: Maybe add a `Sources` field here, that points to the BIDS URI for the `param-diffusivity` file?

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,946 @@
...
+            "sub-01_model-csd_param-csf_dwimap.nii.gz": "",
+        },
+        },
+    },
+    }
+) }}
+
+Dimensions of NIfTI image "`sub-01_model-csd_param-wm_model.nii.gz`": *I*x*J*x*K*x45 ([spherical harmonics](#encoding-sh))
+Dimensions of NIfTI image "`sub-01_model-csd_param-gm_model.nii.gz`": *I*x*J*x*K*x1 ([spherical harmonics](#encoding-sh))
+Dimensions of NIfTI image "`sub-01_model-csd_param-csf_model.nii.gz`": *I*x*J*x*K*x1 ([spherical harmonics](#encoding-sh))
```
arokem: Change `model` to `dwimap` in these three lines above.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,946 @@
...
+
+Contents of JSON file "`sub-01_model-csd_param-wm_model.json`":
```
arokem: Here too.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,1094 @@
...
+-   As per [common data types](common-data-types.md) for derivative data, a
+    JSON sidecar file is REQUIRED due to the REQUIRED `SkullStripped` field.
+
+```Text
+<pipeline_name>/
+    sub-<participant_label>/
+        dwi/
+            <source_keywords>[_space-<space>]_desc-preproc_dwi.nii[.gz]
+            <source_keywords>[_space-<space>]_desc-preproc_dwi.bvals
+            <source_keywords>[_space-<space>]_desc-preproc_dwi.bvecs
```
gagnonanthony: Sorry to jump in here, but would it be better for consistency to keep the extension identical to the raw data files (e.g., `.bval` and `.bvec`, [here for reference](https://bids-specification.readthedocs.io/en/stable/glossary.html#bval-extensions))?

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,1125 @@
...
+        "LongName": "URL",
+        "Description": (
+        "OPTIONAL",
+        "String. URL to the specific implementation of the model utilized.",
+        )
+    }
+   }
+) }}
+
+Dictionary `"Model["Parameters"]"` has the following reserved keywords that may be applicable to a broad range of models:
```
tsalo, line 250: For `"Model["Parameters"]"`, do we want to restrict keys to the reserved keywords or do we want it to be like column names in some TSVs (i.e., some columns are described in the specification and are automatically validated against that, but if you add your own columns you need to define what they are)?

The tough part is that, since this is already in the JSON, there's no separate place to define the key if you're adding a new one.

As an example, QSIRecon has three parameters for its NODDI interface: "ParallelDiffusivity", "PerpendicularDiffusivity", and "IsExVivo". The first two are proposed, with specific units, in this BEP, but the third one is just a parameter for the AMICO NODDI implementation, and my bet is that it tunes a bunch of knobs internally and QSIRecon doesn't have access to those knobs' values. Therefore, if we wanted to include "IsExVivo", would we need a nested dictionary like the following?

```json
{
    "Model":
        "Parameters": {
            "ParallelDiffusivity": 0.017,
            "PerpendicularDiffusivity": 0.025,
            "IsExVivo": {
                "Value": true,
                "Description": "A boolean indicating if the data are ex vivo (true) or in vivo (false)."
                "Type": "boolean"
            }
        }
}
```

tsalo: Pinging @arokem @effigies

arokem: Is there any precedent to this situation (i.e., passing parameters to software to create derivatives)? For example, what does fitlins do with its input parameters in metadata?

tsalo: The only thing I can think of is the provenance BEP, but that uses json.ld files instead of BIDS sidecars.

effigies: Sorry for missing this. No, there's no way to describe JSON values in JSON. We could propose a JSON-schema (or JSON Type Definition) format that allows a dataset to declare the schema that its values should be validated against. Something like `model-<label>_param-<label>_dwimap.schema.json`, that validates sidecars matching those entities.

I don't think it would be worth trying to extend the column definitions in sidecars for TSV files to meet this use case. If you're at the point of describing JSON, using an existing validation technology is probably less surprising to anybody producing or consuming these values.

arokem: OK - let's keep this simple, by using only reserved keywords described in the model parameters table, and specified explicitly in the schema. We can always add more models/parameters through subsequent additions.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,1025 @@
...
+        | ...     | ...                |
+
+    -   Relationship between maximal spherical harmonic degree *l<sub>max</sub>*
+        and number of image volumes *N*:
+
+        *N* = ((*l<sub>max</sub>*+1) x (*l<sub>max</sub>*+2)) / 2
+
+        | ***l<sub>max</sub>*** | 0 | 2 | 4 | 6 | 8 | 10 | ... |
+        | --------------------- | --:| --:| --: | --: | --: | --: | :--: |
+        | ***N***               | 1 | 6 | 15 | 28 | 45 | 66 | ... |
```
arokem, line 1014: Could we consider basis sets that also have odd harmonics? We've been using that to encode asymmetric ODFs. It seems like it would be a relatively straightforward addition here, but I'll concede that it's a bit iffy on the 80/20 rule.

mattcieslak: I'd like to be able to represent asymmetric odfs in BIDS

src/schema/objects/metadata.yaml
```diff
@@ -4197,3 +4197,273 @@ iEEGReference:
...
+    SphericalHarmonicBasis:
+      name: SphericalHarmonicBasis
+      display_name: Spherical Harmonic Basis
+      description: |
+        String. Options are: { `mrtrix3`, `descoteaux` }. Details are provided in the
+        [spherical harmonics bases](#spherical-harmonics-bases) section.
+      type: string
+      enum:
+        - mrtrix3
+        - descoteaux
```
arokem, line 4366: For asymmetric ODFs, would it be sufficient if we add `mrtrix3_asym` and `descoteaux_asym` here, to imply that the coefficients include both even and odd orders?

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,1025 @@
...
+<!-- This block generates a file tree.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_filetree_example(
+    {
+    "msmtcsd_pipeline": {
+        "sub-01": {
+        "dwi": {
+            "sub-01_model-csd_param-wm_dwimap.nii.gz": "",
```
arokem, line 446: @36000 points out that if we fit both a "standard" csd and a msmt csd model within the same pipeline, we would end up with the same file names here. Should we use the full "msmtcsd" here? Or use "ssstcsd" for the "standard" (single-shell, single-tissue) version of CSD?

PR #2216, opened 2025-09-21
[INFRA] Use full page width

effigies:
In many cases, our content is naturally wider than our theme (mkdocs-material) handles comfortably. Filename templates and metadata tables are frequently a bit squashed feeling.

This PR sets the max-width of the content pane to `fit-content` instead of `61rem`. I believe this will always be the same as using the full screen width, but at least allows the possibility that a page with narrow contents would be better centered.

Relates somewhat to #1784, which addressed the impact of the narrow content frame on the entity tables.

References:

* https://github.com/squidfunk/mkdocs-material/discussions/6404
* https://developer.mozilla.org/en-US/docs/Web/CSS/max-width

## Before

image

image

## After

image

image

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -107,6 +107,7 @@ extra_javascript:
   - js/jquery-3.6.0.min.js
 extra_css:
   - css/tsv.css
+  - css/dynamic-width.css
 markdown_extensions:
   - toc:
       anchorlink: true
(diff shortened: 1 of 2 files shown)


julia-pfarr (approved):
makes sense

nellh (approved):
I think this is a good change.

PR #1525, opened 2023-06-24
[ENH] Permit CITATION.cff as strict alternative to Authors field

effigies:
Closes #901.

The short-term fix for the legacy validator would be adding `CITATION.cff` to the file regexes, but I do not currently plan to write code to enforce the mutual exclusion. Unless there's strong feeling that all aspects must be implemented by the legacy validator until we're ready to flip the switch?

PR #2221, opened 2025-09-29
[ENH] Allow datacite.yml in dataset root

### LINKED ISSUE #1955: Support some datacite.yml in addition/instead of CITATION.cff
### Your idea

ATM there is an idea to relay more of information for citations to CITATION.cff and not strive to make `dataset_description.json` to contain all needed information (e.g. affiliations of the authors etc).

BUT, as far as I see, CITATION.cff development stalled, while it is still lacking some rudimentary needed features such as supporting multiple affiliations for authors (https://github.com/citation-file-format/citation-file-format/issues/268) and unclear how to reference funders (https://github.com/citation-file-format/citation-file-format/issues/491).

While that is happening, there is https://schema.datacite.org/ with a working group, continuous progress, and used widely by anyone minting DOIs via datacite (e.g. we do for DANDI and IIRC OpenNeuro also uses datacite).

A well aged example of a datacite record from @mih for a BIDS dataset could be found  at 
https://github.com/psychoinformatics-de/studyforrest-data-phase2/blob/master/datacite.yml . It lacks affiliations but otherwise a good example .

Ideally there should be a tool to convert from datacite to citation.cff format, but ATM I didn't google up any. may be someone could add some support into https://github.com/con/tributors or alike.

effigies:
The DataCite Metadata Schema (https://schema.datacite.org/meta/kernel-4/) is a fairly comprehensive structure for describing citable datasets, including [contributors](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/) (including [organizations](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/#a-nametype) and contributor [roles](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/contributor/#a-contributortype)), [funders](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/fundingreference/), [licenses](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/rights/) and [abstracts](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/description/).

The DataCite Metadata Working Group publish [XSD](https://schema.datacite.org/meta/kernel-4.6/metadata.xsd) and Invenio hosts JSON-Schema translations at https://github.com/inveniosoftware/datacite/tree/master/datacite/schemas.

Following the lead of CITATION.cff, we would use YAML, a superset of JSON that is interpreted as a JSON-compatible object.

I would propose that we add the Invenio JSON-Schema documents to the [jsr:@bids/schema](https://jsr.io/@bids/schema/) package, similar to https://jsr.io/@bids/schema/1.1.0/citation/schema.json.

Closes #1955.

TODO:

* [x] Add schema checks for MUST/SHOULD statements.
* [ ] Recommend against having both CITATION.cff and datacite.yml? They could get out-of-sync, but if someone was syncing them, it might make the dataset legible to more tools.

### DIFF
--- a/src/modality-agnostic-files/dataset-description.md
+++ b/src/modality-agnostic-files/dataset-description.md
@@ -5,6 +5,7 @@ Templates:
 -   `dataset_description.json`
 -   `README[.md|.rst|.txt]`
 -   `CITATION.cff`
+-   `datacite.yml`
 -   `CHANGES`
 -   `LICENSE[.md|.rst|.txt]`
 
@@ -155,7 +156,24 @@ A guide for using macros can be found at
 -->
 {{ MACROS___render_text("objects.files.README.description") }}
 
-## `CITATION.cff`
+## Structured citation files
+
+BIDS permits structured citation files that may improve interoperability with
+dataset indexing utilities or afford higher precision than fields in
+`dataset_description.json`.
+
+If these files are used, fields that provide information that is redundant with
+`dataset_description.json` fields SHOULD be preferred to those in `dataset_description.json`.
+
+To avoid inconsistency, metadata present in one of the following files SHOULD NOT be
+be included in `dataset_description.json`, with the exception of `Name` and
+`DatasetDOI`, to ensure that tools that are only aware of `dataset_description.json`
+can generate references to the dataset.
+
+In particular, if a structured citation file is present,
+the `"Authors"` field of `dataset_description.json` MUST be omitted.
+
+### `CITATION.cff`
 
 <!-- This block generates a file tree.
 A guide for using macros can be found at
@@ -163,17 +181,21 @@ A guide for using macros can be found at
 -->
 {{ MACROS___render_text("objects.files.CITATION.description") }}
 
-For most redundant fields between `CITATION.cff` and `dataset_description.json`,
-the `CITATION.cff` SHOULD take precedence.
-To avoid inconsistency, metadata present in `CITATION.cff` SHOULD NOT be
-be included in `dataset_description.json`, with the exception of `Name` and
-`DatasetDOI`, to ensure that `CITATION.cff`-unaware tools can generate
-references to the dataset.
-In particular, if `CITATION.cff` is present,
-the `"Authors"` field of `dataset_description.json` MUST be omitted,
-and the `"HowToAcknowledge"`, `"License"` and `"ReferencesAndLinks"` SHOULD be omitted
+If `CITATION.cff` is present,
+the `"HowToAcknowledge"`, `"License"` and `"ReferencesAndLinks"` SHOULD be omitted
 in favor of the `CITATION.cff` fields `message`/`preferred-citation`, `license` and
 `references`.
+
+### `datacite.yml`
+
+<!-- This block generates a file tree.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___render_text("objects.files.datacite.description") }}
+
+If `datacite.yml` is present, the `"License"` SHOULD be omitted
+in favor of the `datacite.yml` field `rightsList`.
 
 ## `CHANGES`
 
--- a/src/schema/objects/files.yaml
+++ b/src/schema/objects/files.yaml
@@ -49,6 +49,16 @@ README:
     even if the used format is not rendered.
     A guideline for creating a good `README` file can be found in the
     [bids-starter-kit](https://github.com/bids-standard/bids-starter-kit/tree/main/templates/).
+datacite:
+  display_name: datacite.yml
+  file_type: regular
+  description: |
+    A description of the citation information for the dataset, following the
+    [DataCite Metadata Schema v4](https://schema.datacite.org/meta/kernel-4/)
+    (or higher) schema.
+    This file permits more detailed and structured descriptions than
+    [dataset_description.json](SPEC_ROOT/glossary.md#dataset_description-files),
+    suitable for submission to DataCite to generate a DOI.
 dataset_description:
   display_name: Dataset Description
   file_type: regular
--- a/src/schema/rules/checks/dataset.yaml
+++ b/src/schema/rules/checks/dataset.yaml
@@ -89,12 +89,24 @@ SingleSourceAuthors:
     code: AUTHORS_AND_CITATION_FILE_MUTUALLY_EXCLUSIVE
     level: error
     message: |
-      'CITATION.cff' file found. The "Authors" field of 'dataset_description.json'
+      {path} file found. The "Authors" field of 'dataset_description.json'
       must be removed to avoid inconsistency.
   selectors:
-    - path == '/CITATION.cff'
+    - intersects([path], ['/CITATION.cff', '/datacite.yml'])
   checks:
     - '!("Authors" in dataset.dataset_description)'
+
+SingleSourceDataciteFields:
+  issue:
+    code: SINGLE_SOURCE_CITATION_FIELDS
+    level: warning
+    message: |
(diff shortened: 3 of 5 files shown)


effigies:
Rebased on #2315.

Anybody have thoughts on this?

> * [ ] Recommend against having both CITATION.cff and datacite.yml? They could get out-of-sync, but if someone was syncing them, it might make the dataset legible to more tools.

julia-pfarr:
Who would be the "someone" syncing them? The authors of the dataset?

src/modality-agnostic-files/dataset-description.md
```diff
@@ -172,26 +154,47 @@ A guide for using macros can be found at
...
+### `datacite.yml`
+
+<!-- This block generates a file tree.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___render_text("objects.files.datacite.description") }}
+
+If `datacite.yml` is present, the `"License"` SHOULD be omitted
+in favor of the `datacite.yml` field `rightsList`.
```
effigies, line 198:
```suggestion
the `"HowToAcknowledge"`, `"License"` and `"ReferencesAndLinks"` fields SHOULD be omitted
from `dataset_description.json` in favor of the `CITATION.cff` fields
`message`/`preferred-citation`, `license` and `references`.

### `datacite.yml`

<!-- This block generates a file tree.
A guide for using macros can be found at
 https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
-->
{{ MACROS___render_text("objects.files.datacite.description") }}

If `datacite.yml` is present, the `"License"` field SHOULD be omitted
from `dataset_description.json` in favor of the `datacite.yml` field `rightsList`.
```

effigies:
> > I'm not sure where that leaves us. Do we need crosswalks and validation before allowing datacite.yml? Or do we disallow duplication until we have crosswalks?
> 
> I would say that we kinda follow 'summarization' principle of IP that we do allow for duplication but consistent one (thus I think echoing @ericearl ); and thus
> 
>     * tune up PR text to keep those fields required in dataset_description.json in particular if defined elsewhere.

`Authors` is currently forbidden if `CITATION.cff` is present. At most we can make it recommended again, but I don't like the idea of someone fixing an error from one version of BIDS and then suddenly getting warned about the fix in another version. I would suggest removing the error at most.

We can stop warning on duplication of metadata, but without consistency checks, I think we're inviting pain.

>     * file an issue in bids-validator so we do not forget about adding at least basic cross-walk checks

We should have a paired PR, not just an issue. The question is what should be in it. I am of the opinion that we should not allow duplication at all, but I would appreciate if those who want duplication and consistency would make specific proposals.

>     * research more into a generic (non BIDS) cross-walk between CITATION.cff and datacite.cff -- I guess in https://codemeta.github.io/crosswalk/ .  But oddly CITATION.cff is not even there among crosswalks: https://github.com/search?q=org%3Acodemeta+CITATION.cff&type=pullrequests  I am confused a little why not even seems discussed????

<details><summary>Expand for Claude's breakdown of the three name representations:</summary>

| | BIDS `Authors` | CITATION.cff person | DataCite creator/contributor |
|---|---|---|---|
| Shape | ordered list of **opaque strings** | ordered list of person **or** entity | ordered list; `creators` (required) + `contributors` (roled) |
| Given/family split | none | `given-names`, `family-names` (both *optional*) | `givenName`, `familyName` (both optional) |
| Combined name | the string *is* the name, free-form | **none** for persons | `creatorName` **required**, format "Family, Given" |
| Particles / suffixes | inline, unmarked | `name-particle`, `name-suffix` (separate fields) | **no** separate fields — folded into name/family |
| Org vs person | indistinguishable | `entity{name}` vs `person{...}` | `nameType: Personal\|Organizational` |
| Identifier | none | `orcid` (URL form) | `nameIdentifier` + `nameIdentifierScheme` |
| Affiliation | none | single string | **0–n**, with identifiers |

</details>

I'm going to present two new proposals. The first attempts to unify all three representations, and the second aims to permit duplication with `datacite.yml`/`dataset_description.json` while keeping the rules for `CITATION.cff` as-is:

### Proposal 1: Unify on `datacite:creatorName` with specific serialization of CITATION.cff authors

[`datacite:creatorName`](https://datacite-metadata-schema.readthedocs.io/en/4.6/properties/creator/#creatorname) is technically free-form but should (they do not use the RFC2119 "SHOULD", so normative status is unclear) be `Last, First`. Extending that to particles and suffixes, that would be `{particle} {family-names}, {given-names}, {suffix}` (with commas and spaces stripped). So:

1) If `datacite.yml` is present, then `dataset_description.Authors == [creator.creatorName for creator in datacite.creators]`. This is a required field and defined for both people and entities.
2) If `CITATION.cff` is present, then `dataset_description.Authors` needs to be checked against a constructed list:
   ```python
   authors = []
   for author in citation_cff.authors:
       if name := author.get("name"):  # Entity
           authors.append(name)
       else:  # Person
           given = author.get("given-names", "")
           particle = author.get("name-particle", "")
           family = author.get("family-names", "")
           suffix = author.get("name-suffix", "")
           # Simplified, as "Seneca, , the Younger" is possible
           authors.append(f"{particle} {family}, {given}, {suffix}".strip(", "))
   ```
3) If both are present and `dataset_description.Authors` is absent, then these two lists must also be compared.

This follows @yarikoptic's preference for verified duplication over deduplication, and does not prevent the simultaneous use of CITATION.cff and datacite.yml, which people may want.

I do not think this can be reasonably represented in the schema. This would need to be a directive to validator and probably an appendix with a reference implementation.

#### Proposal 1 problems

BIDS would have to include a name serialization standard, which will be culturally fraught. If we do want to go down that road, I think that would be worth a working group or even BEP to make sure that we get it right, where this PR aims to be a practical solution to allow structured metadata that includes contributor roles.

### Proposal 2: Allow `datacite:creatorName`/`dataset_description.Authors` equality, disallow `CITATION.cff`+`datacite.yml`

Because `creatorName` and `Authors` are both free-form text, we can reduce this to:

1) If `datacite.yml` is present and `Authors` is present in `dataset_description.json`, then `dataset_description.Authors == [creator.creatorName for creator in datacite.creators]`.
2) If `datacite.yml` is present, `CITATION.cff` is forbidden.

In the case where `CITATION.cff` is absent, this is identical to proposal 1, since no serialization is required in either case.

These two rules could be represented in schema, provided that `[creator.creatorName for creator in datacite.creators]` is made available in the dataset-level context. Rule 2 could later be relaxed if a datacite/CITATION.cff cross-walk is made available as an outside standard (or BIDS does decide to maintain one). This avoids/defers the problem of maintaining our own standard, and requires no change to the existing rules for `CITATION.cff`.

#### Proposal 2 problems

This introduces an asymmetry between datacite.yml and CITATION.cff. This reflects both an evolution of consensus on whether duplication+consistency checks are preferable to deduplication since #1525 and that consistency checks between CITATION.cff name structures and free-form text remains a difficult problem.

### Proposal 3: Single source

Just to write up the current state of the PR, the alternative proposal is to always prefer structured metadata to the exclusion of unstructured. If a structured metadata file (CITATION.cff or datacite.yml) are present, then fields SHOULD (and in the case of Authors, MUST) be deduplicated from dataset_description.json.

The PR does not currently disallow having both files, but for it to be truly single source, I think it should.

#### Proposal 3 problems

There is an expressed desire not to make `dataset_description.json` fields change requirement status based on the presence/absence of a structured metadata file. Multiple structured metadata files may be desirable to some.

----

Of these, I have a strong preference for 3, due to simplicity. I am willing to put in the work to make 2 happen. I strongly think 1 would be a mistake. I'm open to hearing other proposals.

PR #2225, opened 2025-10-01
[MISC]: Fix placement of picture for Benjamin Dichter

bendichter:


bendichter:
I don't know how it happened

effigies (approved):
Bad source data in `.all-contributorsrc`. Fixed and confirmed that running `make update_contributors` doesn't revert your change.

Issue #2057, opened 2025-02-18
Motion trajectories & pose extracted data for animal beh research


yarikoptic:
### Your idea

Popular relevant toolkits used

- [SLEAP](https://sleap.ai/)  (attn @talmo) : Social LEAP Estimates Animal Poses  - storing into `.pkg.slp` (HDF5). Some more pointers: https://github.com/talmolab/sleap/discussions/2093#discussioncomment-11908290
- [DeepLabCut](https://github.com/DeepLabCut/DeepLabCut) (attn @MMathisLab): yet to discover / add here ...

Even though that section was primarily aiming for hardware-based motion tracking of humans (AFAIK; e.g. like point-light and other systems), I think it would be great to review the [Motion](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/motion.html) BIDS data type which was and see how it could/should be extended to be able to capture results of the processing from SLEAP, DeepLabCut, or any other similar tool.  We might need to workout a BEP and also see if some common format could be established for tool-agnostic format.

Also attn @bids-standard/bep032 

Related issues
- #1771

talmo:
Relevant:
- Unifying library by @niksirbi and colleagues: https://github.com/neuroinformatics-unit/movement
- NWB extension by @rly: https://github.com/rly/ndx-pose

MMathisLab:
Thanks @yarikoptic! [DeepLabCut](https://github.com/DeepLabCut/DeepLabCut) uses `HDF5`, and has an export for `NWB` for saving analyzed data. 
- function to [convert HDF5 to NWB](https://github.com/DeepLabCut/DeepLabCut/blob/fa2ebee4bf9b9723cf32e0a1ffe664de95b9b2e1/deeplabcut/utils/conversioncode.py#L261)

yarikoptic:
Thank you @talmo and @MMathisLab .  So both SLEAP and DeepLabCut has export to ndx-pose.  I didn't check though if such representation is generalized that e.g. they could potentially read it back (I see there is "load" functionality within https://github.com/DeepLabCut/DLC2NWB)?

@niksirbi 's movement has open issue
- https://github.com/neuroinformatics-unit/movement/issues/23
[...]

PR #2022, opened 2024-12-22
[ENH] BEP044 - Stim-BIDS

neuromechanist:
This PR addresses #153 by introducing specifications for handling and standardizing stimulus files and their annotations within the BIDS specifications. The changes focus on improving the organization, referencing, and metadata of stimulus files to enhance consistency, reusability, and efficiency.

> [!IMPORTANT]
> HTML render of the stimuli page : https://bids-specification--2022.org.readthedocs.build/en/2022/modality-agnostic-files/stimuli.html

> [!NOTE]
> The issue: #153
> [Google Doc](https://docs.google.com/document/d/1DoQghbJlQksCHEFs0boT816p3ejX2Hd9l-OWDTXFtV8/edit?pli=1&tab=t.0#heading=h.ab3cjzoaom6d) 
> Example PR: https://github.com/bids-standard/bids-examples/pull/433

Known issues:
- The validator fails because extensions in `/src/schema/rules/files/raw` require `datatype`. Stimuli might be a special data type that can only be present at the root of the dataset. So, the datatype field is missing for now.
- There are some style errors by the remark validators

cc: @bids-standard/bep044 and @monique2208

PR #2231, opened 2025-10-25
[ENH] BEP047 - Add audio/video recordings to behavioral experiments

### LINKED ISSUE #1771: BEP for audio/video capture of behaving subjects

bendichter:
fix #1771 

edits by @yarikoptic:
- preview - https://bids-specification--2231.org.readthedocs.build/en/2231/modality-specific-files/behavioral-experiments.html
- bids-examples PR: https://github.com/bids-standard/bids-examples/pull/523


src/schema/objects/metadata.yaml
```diff
@@ -2090,6 +2097,54 @@ MEGREFChannelCount:
...
+    - mm
+    - cm
+    - pixels
+MicroephysCoordinateSystemPhoto:
+  name: MicroephysCoordinateSystemPhoto
+  display_name: Microephys Coordinate System Photo
+  description: |
+    A link to a photo or drawing of the microelectrode probe system.
+  type: string
+  format: uri
```
bendichter:
```suggestion
```

src/schema/objects/metadata.yaml
```diff
@@ -2208,6 +2215,12 @@ ManufacturersModelName:
   description: |
     Manufacturer's model name of the equipment that produced the measurements.
   type: string
+ManufacturersModelVersion:
+  name: ManufacturersModelVersion
+  display_name: Manufacturers Model Version
+  description: |
+    Manufacturer's model version of the equipment that produced the measurements.
+  type: string
```
bendichter:
```suggestion
```

src/schema/objects/metadata.yaml
```diff
@@ -3789,6 +3802,13 @@ SubjectArtefactDescription:
     If this field is set to `"n/a"`, it will be interpreted as absence of major
     source of artifacts except cardiac and blinks.
   type: string
+SupplementarySignals:
+  name: SupplementarySignals
+  display_name: Supplementary Signals
+  description: |
+    Description of the supplementary signal (additional modalities) recorded
+    in parallel and are also stored in the data file.
+  type: string
```
bendichter:
```suggestion
```

src/schema/objects/metadata.yaml
```diff
@@ -2990,6 +2997,12 @@ RecordingDuration:
     Length of the recording in seconds (for example, `3600`).
   type: number
   unit: s
+RecordingSetupName:
+  name: RecordingSetupName
+  display_name: Recording Setup Name
+  description: |
+    Custom name of the recording setup.
+  type: string
```
bendichter:
```suggestion
```

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -1,4 +1,4 @@
-# Behavioral experiments (with no neural recordings)
+# Behavioral experiments
```
yarikoptic:
```suggestion
# Behavioral experiments (large or no neural recordings)
```

since otherwise would "conflict" with the current situation that typically behavioral data (`_events.tsv`) is with data (`func`, `eeg`, `ecephys`, ...) and not in a separate folder. So we need to somehow tell here that it is in those other cases when to consider this section

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +72,141 @@ A guide for using macros can be found at
...
+
+-   `task` - OPTIONAL for audio and video recordings
+-   `acquisition` - OPTIONAL, can distinguish different recording setups
+-   `run` - OPTIONAL, for multiple recordings with identical parameters
+-   `recording` - OPTIONAL, to differentiate simultaneous recordings from different angles, locations, or devices
+-   `split` - OPTIONAL, for continuous recordings split into multiple files
+
+### Examples
+
+```Text
```
yarikoptic: see/use MACROS___make_filetree_example here or provide comprehension when such trees are encountered in `Text` blocks, parsed into hierarchical structury,  and then proxied to that MACROS for actual uniform rendering

src/modality-specific-files/behavioral-experiments.md
bendichter: How about "Behavioral recordings"? I'm not a big fan of "large" because it is very subjective

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +68,167 @@ A guide for using macros can be found at
...
+
+    These recordings are generally more suitable for internal use or for sharing
+    non-human subject data, unless appropriate privacy protections are implemented.
+
+### File formats
+
+Audio recordings MUST use one of the following extensions:
+
+-   `.mp3` - MPEG Audio Layer III
+-   `.wav` - Waveform Audio File Format
```
effigies, line 106: Would it be worth including `.flac` as an alternative to `.wav`, since it provides lossless compression? If I were working with these data, I would almost certainly use FLAC for my own archives and then export to MP3 if that were necessary to satisfy BIDS.

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +68,167 @@ A guide for using macros can be found at
...
+{{ MACROS___make_filetree_example(
+   {
+   "sub-01": {
+      "ses-01": {
+         "beh": {
+            "sub-01_ses-01_task-freeplay_run-01_video": {
+               "split-001.mp4": "",
+               "split-002.mp4": "",
+               "split-003.mp4": "",
+               },
```
effigies: This is a substantial change in how much extensions, entities and suffixes can be reordered. I talked with @yarikoptic yesterday and I think this is an interesting future direction, but something that does not require any changes at the moment would be:

```suggestion
            "sub-01_ses-01_task-freeplay_run-01_split-001_video.mp4": "",
            "sub-01_ses-01_task-freeplay_run-01_split-002_video.mp4": "",
            "sub-01_ses-01_task-freeplay_run-01_split-003_video.mp4": "",
```

src/modality-specific-files/behavioral-experiments.md
yarikoptic: +1 from me on adding .flac for lossless option (worth stating that in description) and even state preference over old uncompressed .wav (`<dandi-archive-hat>`should we even forbid it to not waste space?!`</dandi-archive-hat>` )

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -15,19 +15,11 @@ and a guide for using macros can be found at
...
-The results of those experiments MAY be stored in the `beh` directory
-using the same formats for event timing (`_events.tsv`),
-metadata (`_events.json`),
-physiological (`_physio.tsv.gz`, `_physio.json`)
-and other continuous recordings (`_stim.tsv.gz`, `_stim.json`)
-as for tasks performed during MRI, electrophysiological or other neural recordings.
-Additionally, events files
-that do not include the mandatory `onset` and `duration` columns
-MAY be included,
+The `beh` directory MAY store behavioral recordings such as audio (`_audio.*`) and video (`_video.*`) recordings, physiological (`_physio.*`) recordings, and other continuous recordings (`_stim.tsv.gz`, `_stim.json`).
```
effigies: Just to be clear, there is no proposal to support this for experiments where there are non-physiological recordings, such as fMRI or EEG? Or is this a one-step-at-a-time strategy, get agreement on behavior and then make applicable where stim and physio are supported?

src/schema/rules/sidecars/beh.yaml
```diff
@@ -25,3 +25,24 @@ BEHInstitutionInformation:
...
+  selectors:
+    - datatype == "beh"
+    - intersects([suffix], ["audio", "video"])
+  fields:
+    AudioChannelCount: optional
+    AudioSampleRate: optional
+    FrameRate: optional
+    Height: optional
+    Width: optional
+    Duration: optional
```
effigies: I think you should probably have a `StartTime`, as with `physio` and `stim` files.

effigies (commented):
Overall this makes sense to me. It would be good to get some feedback from contributors to related BEPs, such as eye-tracking (20), motion (29), stimuli (44) and physio (45). Even if this PR doesn't propose adding this as an associated file to those data types, the potential is there and it's worth getting opinions and identifying potential conflicts.

cc @bids-standard/bep029 @bids-standard/bep044
cc @mszinte @julia-pfarr @oesteban (BEP020)
cc @m-miedema @smoia @SouravKulkarni (?) (BEP045)

src/modality-specific-files/behavioral-experiments.md
effigies: I would personally be fine with just dropping wav, since FLAC is unencumbered by patents and has implementations in:

* Python: https://pyflac.readthedocs.io/en/latest/
* MATLAB: https://www.mathworks.com/help/matlab/import_export/supported-video-file-formats.html
* R: https://cran.r-universe.dev/seewave
* Julia: https://www.juliapackages.com/packages?search=flac
* C/C++: https://github.com/xiph/flac
* Rust: https://crates.io/search?q=flac&sort=downloads

The advantage of wav is the sheer simplicity of the format, but I feel like the size of .wav files of any length is sufficient justification to use a more complicated format.

bendichter: Also 

* Javacript: https://github.com/mmig/libflac.js/

Yes, supporting .flac sounds like a good idea, but I don't know about forbidding .wav. It's just so ubiquitous. What if we allow for .wav but recommend using .flac instead?

src/modality-specific-files/behavioral-experiments.md
bendichter: @yarikoptic 's [concern](https://github.com/bids-standard/bids-specification/issues/1771#issuecomment-2654835205) here was that if we have hundreds or even thousands of video files this would get cumbersome. @talmo [also preferred](https://github.com/bids-standard/bids-specification/issues/1771#issuecomment-2658029897) the subdirectory approach.

From my perspective, my philosophy is to only propose changes to the standard for *expressibility* or *performance* and this a large proposed change that seems merely for convenience, not expressibility or performance (correct me if I am wrong), so I am fine with reverting to the standard split organization (i.e. no subdirectory).

src/modality-specific-files/behavioral-experiments.md
effigies: If we recommend it, then that implies making a warning for wav. Also fine with me, and an archive that doesn't want to support wav can promote the warning to an error.

The only weird thing is permitting something we want to warn on immediately. If it hasn't been previously allowed, making it an error gives the opportunity to convert before submitting to an archive, since many people just ignore warnings.

src/schema/rules/sidecars/beh.yaml
bendichter: What would StartTime indicate? Shouldn't that information be in the scans.tsv file?

src/modality-specific-files/behavioral-experiments.md
bendichter: This is just a statement about where data goes. If that data exists, it should go in `/func`.

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +68,165 @@ A guide for using macros can be found at
...
+    local data protection laws) when handling these recordings.
+
+    These recordings are generally more suitable for internal use or for sharing
+    non-human subject data, unless appropriate privacy protections are implemented.
+
+### File formats
+
+Audio recordings MUST use one of the following extensions:
+
+-   `.mp3` - MPEG Audio Layer III
```
bendichter, line 104:
```suggestion
-   `.flac` - Free Lossless Audio Codec
-   `.mp3` - MPEG Audio Layer III
```

src/schema/rules/sidecars/beh.yaml
effigies: If you're going to make `scans.tsv` and the `acq_time` column required, then that would be an alternative, but we haven't done that before.

`StartTime` indicates the start of the current file with reference to the start of the experiment/observation, which is also onset 0 in any associated events.tsv. It can just be 0, if there is no need to align to another file, or if you want it to be the reference to which other files are aligned.

src/modality-specific-files/behavioral-experiments.md
bendichter: I could see users preferring .wav for performance reasons - I assume it is faster to read from wav than flac. Also, for compatibility reasons. You listed a bunch of existing libraries but I could still imagine situations where this would come up. Really though I just get a bad feeling about forbidding what is by far the most popular approach for storing sounds. It creates a conversion burden that feels unnecessary.

If BIDS is *just* for depositing data in OpenNeuro, I think it makes sense to enforce lossless data compression. If BIDS is for other applications too, e.g. working with data internally in a lab, multi-lab collaborations, etc. it may make sense to be more flexible.

In NWB and DANDI, we have a concept of requirements for NWB files that are specific to DANDI and go beyond the minimal requirements for valid NWB. For example, in NWB specifying the age of a subject is optional, but it is required for NWB files being uploaded to DANDI. Is there a mechanism like that we could use here?

src/schema/rules/sidecars/beh.yaml
bendichter: I may be confused here, but I thought `scans.tsv` was the standard way to specify starting time of each stream?

src/modality-specific-files/behavioral-experiments.md
effigies: I see I missed this discussion, apologies for jumping in late. I unresolved this so others can see this conversation, but reading through the previous discussion I think there's an 80/20 split here where it's better to work within the current BIDS constraints and then separately push for a generalization to `_<suffix>/split-<label>.<ext>` if that becomes a practical problem for people.

I would say it's a BEP-level change that would require design for the schema and validator and would significantly slow down this proposal to include it here.

bendichter: Agreed

src/modality-specific-files/behavioral-experiments.md
effigies: Okay, we'll need to treat these similarly to `physio` and `stim`. See `rules.files.raw.task`.

src/schema/rules/files/raw/beh.yaml
```diff
@@ -14,3 +14,42 @@ noncontinuous:
...
     acquisition: optional
     run: optional
+
+# Audio recordings
+audio:
+  suffixes:
+    - audio
+  extensions:
+    - .mp3
+    - .wav
```
bendichter, line 27:
```suggestion
    - .wav
    - .flac
```

src/modality-specific-files/behavioral-experiments.md
bendichter: What's wrong with the current rules in https://github.com/bids-standard/bids-specification/pull/2231/files#diff-163780b72c1b903fa0debc351d7c04f083773598cb0b4ecb5e787acaa63d4d0d?

src/schema/rules/sidecars/beh.yaml
effigies: `scans.tsv` indicates the start of recording of each file, and it can be used to align streams, but it is an optional file, and `acq_time` is an optional column within it. `StartTime` is currently required metadata for `physio` and `stim` and intended to be used to align them to other recordings, identified by matching entities.

I see these options:

1) Treat these as basically equivalent to `stim` and `physio`, and require `StartTime`. It's simple if a bit unsatisfying if we want to encourage universal use of `scans.tsv` to identify simultaneous recordings across datatypes.
2) Require the file be included in `scans.tsv` with an `acq_time`. This will require some design for the schema and validator, as `scans.tsv` is an optional file and `acq_time` is an optional column, but we could probably figure something out.
3) Treat these as qualitative data that do not specifically need to be aligned to be useful. In the absence of `StartTime` or a non-`n/a` `scans.tsv:acq_time` metadata, alignment will simply be undefined. No change is needed if this is considered optional, but a warning that the files cannot be aligned would have the same design requirements as (2), just emitting a warning instead of an error.

src/modality-specific-files/behavioral-experiments.md
effigies: BEP36 has a mechanism that could be used to enable suites of checks that are disabled by default. Without looking, I think it's called "AdditionalValidations" in `dataset_description.json`. If that's the path we want to go, then no change is needed here, and we should not make a specific recommendation against WAV, and we can add those validations when it becomes convenient.

src/modality-specific-files/behavioral-experiments.md
effigies: The `datatypes` is limited to `beh`. You'll need to add every directory these will need to be found in, and you will need to adjust for the differences in permitted entities across datatypes.

src/schema/rules/sidecars/beh.yaml
bendichter: OK, I think it makes sense to be consistent with `stim` and `physio` and include a required `StartTime`.

While ideally videos would be aligned, sometimes you do have video that is during the same session but cannot be precisely aligned. Is it possible to have a n/a-type value for `StartTime`?

effigies: Yes, it would require duplicating the `StartTime` metadata to something like `StartTime__audiovideo` and using `anyOf` to allow `"n/a"`.

Given that there hasn't previously been a call to allow for this in other contexts, I probably wouldn't just update `objects.metadata.StartTime` to allow `"n/a"`, but that is an option as well.

src/modality-specific-files/behavioral-experiments.md
bendichter: The intention here is that video and audio are always in `beh`

effigies: Oh, I see, I think we were talking past each other. Your proposal is `beh/*_{audio,video}.*` and any alignment must be done with `scans.tsv:acq_time`. I would probably explicitly state that in the spec. Something like "Audio/video recordings MAY occur simultaneously with other recordings, such as BOLD or EEG. Relative timing between  files may be determined by consulting the `scans.tsv` file. If no `scans.tsv` file is present, the alignment is undefined."

bendichter: That text is fine with me. So would you recommend I revert the StartTime addition?

effigies: The only thing it can be relative to now is other things in `beh`, and if the goal is for it to be alignable with any other recording, having a relative start time is probably an attractive nuisance. So I guess I'd probably revert.

src/modality-specific-files/behavioral-experiments.md
robertoostenveld: I am fine including `.flac`, but would like `.wav` to remain as an option for uncompressed audio as well, because it is so widely supported by (hardware) audio recorders.

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +71,166 @@ A guide for using macros can be found at
...
+-   `.mp4` - MPEG-4 Part 14
+-   `.mkv` - Matroska video container
+-   `.avi` - Audio Video Interleave
+
+### Entities
+
+Audio and video files MAY use the following entities:
+
+-   `task` - OPTIONAL for audio and video recordings
+-   `acquisition` - OPTIONAL, can distinguish different recording setups
```
robertoostenveld: should this not be `acq` ?

bendichter:
```suggestion
-   `acq` - OPTIONAL, can distinguish different recording setups
```

bendichter: yes, thanks for catching this.

src/modality-specific-files/behavioral-experiments.md
gdevenyi: If we're going to recommend an unencumbered (lossy) compressed audio, it should probably be `.ogg`

bendichter: Based on this conversation, I think it would be best to include flac and ogg. I'll resolve.

src/schema/objects/extensions.yaml
```diff
@@ -153,6 +165,24 @@ md:
...
+  display_name: Matroska Video
+  description: |
+    A [Matroska](https://www.matroska.org/) video container file.
+    This format is commonly used for behavioral video recordings.
+mp3:
+  value: .mp3
+  display_name: MPEG Audio Layer III
+  description: |
+    An [MP3](https://en.wikipedia.org/wiki/MP3) audio file.
+    This format is commonly used for behavioral audio recordings.
```
bendichter:
```suggestion
    This format is commonly used for behavioral audio recordings or auditory stimuli.
```

src/schema/rules/sidecars/beh.yaml
robertoostenveld: For clarification: how would this work if there is both a `StartTime` in the json and an `acq_time` in the scans.tsv?  

Am I correct with the interpretation that `StartTime=0` with `acq_time=1900-01-01T12:00:00` and `StartTime=1` with `acq_time=1900-01-01T11:59:59` would mean the same?

effigies: I don't think anybody has put semantics on `acq_time` and `StartTime` coexisting. My expectation would be that `acq_time` reliably describes the acquisition time of the current file, while `StartTime` indicates an offset relative to a collection of simultaneously collected files.

The `StartTime` proposal has been withdrawn, so I don't think this needs to be settled here.

src/schema/objects/metadata.yaml
```diff
@@ -4197,3 +4204,63 @@ iEEGReference:
...
+    Frame rate of the video recording in frames per second (for example, `30.0`).
+  type: number
+  exclusiveMinimum: 0
+  unit: Hz
+
+Height:
+  name: Height
+  display_name: Video Height
+  description: |
+    Height of the video in pixels (for example, `1920`).
```
robertoostenveld: I suggest for the example to swap height and width to 1080 and 1920, respectively, to align with a landscape HDMI screen and camcorder. Now the example is in portrait orientation.

robertoostenveld (approved):
looks good to me, I look forward to having this as part of the specification

satra:
a colleague asked why all recording sources are being put into a single folder when other parts of bids are separated by modality/technique. perhaps a sentence or two for this rationale would be helpful to include in the specification.

src/schema/objects/metadata.yaml
bendichter:
```suggestion
    Height of the video in pixels (for example, `1080`).
```

src/schema/objects/metadata.yaml
```diff
@@ -4197,3 +4204,63 @@ iEEGReference:
...
+  type: integer
+  minimum: 1
+
+Width:
+  name: Width
+  display_name: Video Width
+  description: |
+    Width of the video in pixels (for example, `1080`).
```
bendichter:
```suggestion
    Width of the video in pixels (for example, `1920`).
```

bendichter:
@satra The rationale here is that the organizing principle for the `beh/` directory is the *type of data being captured* (behavior) rather than the *recording technique* (audio, video, events, etc.). 

In the context of behavioral experiments, audio recordings (capturing vocalizations, speech), video recordings (capturing facial expressions, body movements), and event files all serve the same fundamental purpose: documenting subject behavior. They are different modalities of the same underlying phenomenon.

This is analogous to how `func/` contains both BOLD and ASL data - different acquisition techniques measuring the same underlying phenomenon (functional brain activity).

Additionally, keeping these files together in `beh/` makes it easier to:
1. Find all behavioral data for a session in one place
2. Maintain temporal alignment metadata between synchronized recordings
3. Keep behavioral data organized when collected alongside neural recordings (where the neural data lives in its respective modality folder like `ieeg/` or `eeg/`)

I can add a sentence or two to the specification explaining this rationale. Would something like the following work?

> "Audio and video recordings are stored together in the `beh/` directory because they represent different modalities for capturing the same underlying phenomenon: subject behavior. This organization mirrors other BIDS directories where the organizing principle is the type of data being measured rather than the specific recording technique."

@satra did I understand the question? Do you think this would satisfy the colleague in the context of your conversation with them?

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +71,167 @@ A guide for using macros can be found at
...
+
+```JSON
+{
+  "TaskName": "RestingState",
+  "Device": "Sony FDR-AX53",
+  "AudioChannelCount": 2,
+  "AudioSampleRate": 48000,
+  "FrameRate": 30.0,
+  "Height": 1920,
+  "Width": 1080,
```
bendichter:
```suggestion
  "Height": 1080,
  "Width": 1920,
```

neuromechanist:
My first suggestion (🤓) would be to request a BEP number for this PR from maintainers, as this is quite a change, and arguably/conceptually adds a new modality (although audio and video are being added in BEP044), which would require a BEP. I'll take it up to this week's maintainer meeting.

bendichter:
@neuromechanist 

OK, I have submitted a PR to the website to make this an official BEP here: https://github.com/bids-standard/bids-website/pull/759

bendichter:
## Comparison: vs PR #2022

### Summary
- This PR: Adds audio/video recordings of **subjects behaving** to the `beh/` datatype
- **PR #2022** (BEP044 - Stim-BIDS): Adds organized **stimulus files** to the root-level `/stimuli` directory

### Key Differences

| Aspect | This PR | PR #2022 (Stim-BIDS) |
|--------|---------------------------|----------------------|
| **Location** | `sub-XX/beh/` (subject-scoped) | `/stimuli/` (root-level, shared) |
| **Suffixes** | `_audio`, `_video` | `_audio`, `_video`, `_audiovideo`, `_image` |
| **Audio formats** | `.flac`, `.mp3`, `.ogg`, `.wav` | `.wav`, `.mp3`, `.aac`, `.ogg` |
| **Video formats** | `.mp4`, `.mkv`, `.avi` | `.mp4`, `.avi`, `.mkv`, `.webm` |
| **Unique formats** | `.flac` (audio) | `.aac` (audio), `.webm` (video), image formats (`.jpg`, `.png`, `.svg`, `.webp`) |
| **Key entity** | `recording-<label>` (for multiple angles) | `stim-<label>` (stimulus identifier) |
| **Catalog files** | None (uses `scans.tsv` for timing) | `stimuli.tsv`, `annotations.tsv` |
| **Events linking** | `_events.tsv` alongside recordings | `stim_id` column in events.tsv |
| **Metadata focus** | Technical (AudioSampleRate, FrameRate, Height, Width, Duration, CameraPosition, AudioBitDepth) | Attribution (License, Copyright, URL, Description) |
| **Privacy concern** | Explicit PII warning for human subjects | Not emphasized (stimuli, not subjects) |
| **Part splitting** | `split-<index>` for continuous recordings | `part-<label>` for stimulus segments |
| **Annotation system** | Standard `_events.tsv` | Dedicated `_annot-<label>_events.tsv` |

### Notable Distinctions

1. **`_audiovideo` suffix**: Only in PR #2022 (BEP044), explicitly distinguishes files with both audio and video streams from video-only files

2. **`_image` suffix**: Only in PR #2022 (BEP044), adds support for static visual stimuli

3. **Reusability**: PR #2022 emphasizes stimulus reuse across subjects/studies (centralized in `/stimuli`), while PR #2231 ties recordings to specific subjects

4. **Annotation richness**: PR #2022 has a more elaborate annotation system with `annotations.tsv` and `annot-<label>` entity for multiple annotation sets per stimulus

5. **New columns**: PR #2022 adds `stim_id` column to events.tsv; PR #2231 doesn't add new event columns

6. **Timing alignment**: PR #2231 references `scans.tsv` for synchronization with other modalities; PR #2022 focuses on stimulus onset/duration in events files

These PRs are complementary—one captures what the subject does (behavioral recordings), the other captures what's shown to the subject (stimuli).

@neuromechanist , thoughts?

src/schema/objects/metadata.yaml
```diff
@@ -4366,3 +4373,82 @@ iEEGReference:
...
+  type: integer
+  minimum: 1
+
+Width:
+  name: Width
+  display_name: Video Width
+  description: |
+    Width of the video in pixels (for example, `1920`).
+  type: integer
+  minimum: 1
```
yarikoptic: I think it would be quite valuable to expose at least some details on the underlying codec(s) used for audio/video within the files, so e.g. we could assess if e.g. browser would play it etc?

oesteban:
> Here are the issues relating to potentially changing the folder organization within BIDS, I think it is better to discuss it among those

IMHO, this is a problem of today. It'd be great that BIDS 2.0 had an elegant/more consistent response. However, BIDS needs to address this for future BEPs (cc/ @ericearl)

bendichter:
@neuromechanist 

1. On the audio video vs. audiovideo labeling, we went back and forth a bit in the issue. For us it doesn't make a huge difference, since you would be able to parse that from the metadata about the streams in the json sidecar anyway. If you feel strongly about audiovideo I would be fine with changing it in interest of consistency.

2. _image. I suppose one could take a picture of a subject performing a task task. I don't know if that's recording behavior per se, but I'd be fine with adding it if you think we should.

3. I think one of our biggest differences is with the metadata in the sidecar files. Yours is attribution (License, Copyright, URL, Description) and this one is technical (AudioSampleRate, FrameRate, Height, Width, Duration, CameraPosition, AudioBitDepth). I don't think adding attribution to ours makes much sense. It will generally share the license of the rest of the dataset. However, I do think it might make sense for you to adopt our technical attributes. Maybe not CameraPosition, but it might be nice to be able to get AudioSampleRate, FrameRate, Height, Width, Duration without reading the data file.

4. Our splitting is different. I think split is more consistent with existing usage. The only mention I see of part in the [existing schema](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#part) is:

> [part](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#part)
>Full name: Part
>
>Format: part-<label>
>
>Allowed values: mag, phase, real, imag
>
> Definition: This entity is used to indicate which component of the complex representation of the MRI signal is represented in voxel data. The part-<label> entity is associated with the DICOM Tag 0008, 9208. Allowed label values for this entity are phase, mag, real and imag, which are typically used in part-mag/part-phase or part-real/part-imag pairs of files.
> 
> Phase images MAY be in radians or in arbitrary units. The sidecar JSON file MUST include the "Units" of the phase image. The possible options are "rad" or "arbitrary".
> 
> When there is only a magnitude image of a given type, the part entity MAY be omitted.

whereas [split already has to do with splitting large files](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#split):

> [split](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#split)
> Full name: Split
> 
> Format: split-<index>
> 
> Definition: In the case of long data recordings that exceed a file size of 2Gb, .fif files are conventionally split into multiple parts. Each of these files has an internal pointer to the next file. This is important when renaming these split recordings to the BIDS convention.
> 
> Instead of a simple renaming, files should be read in and saved under their new names with dedicated tools like [MNE-Python](https://mne.tools/), which will ensure that not only the filenames, but also the internal file pointers, will be updated.
> 
> It is RECOMMENDED that .fif files with multiple parts use the split-<index> entity to indicate each part. If there are multiple parts of a recording and the optional scans.tsv is provided, all files MUST be listed separately in scans.tsv and the entries for the acq_time column in scans.tsv MUST all be identical, as described in [Scans file](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#scans-file).
> 

though I can see in your case why you might want to use part, if you are splitting the stimulus up into logical components, like chapters of an audiobook. I don't mind terribly if we use different approaches for this.

src/schema/objects/metadata.yaml
bendichter: OK but let's try to make it as easy as possible to acquire and standardize this data. Can we just use something that is output by FFMPEG? Do we want to make codec information required?

CodyCBakerPhD: > Can we just use something that is output by FFMPEG

Exactly what was proposed in https://github.com/bids-standard/bids-specification/pull/2367#discussion_r2989971312, I would start with that PR maybe

>  Do we want to make codec information required?

I believe we have reached a positive consensus; this is critical and highly valuable information

src/modality-specific-files/behavioral-experiments.md
```diff
@@ -76,6 +71,234 @@ A guide for using macros can be found at
...
+
+Audio and video recordings of behaving subjects MAY be stored in the `beh` directory
+using the `_audio`, `_video`, and `_audiovideo` suffixes.
+The `_audio` suffix is for audio-only recordings, `_video` for video-only recordings,
+and `_audiovideo` for recordings that contain both audio and video streams.
+These recordings are typically used to capture vocalizations, speech, facial expressions,
+body movements, or other behavioral aspects during experimental tasks or rest periods.
+
+Still images captured during behavioral experiments MAY be stored in the `beh` directory
+using the `_image` suffix.
```
bendichter, line 84: I'm on the fence about image. I mostly want a way to store audio/video data and I don't care as much about images/photos. I don't know whether to include it, and if we include it I don't know whether to call it image or photo

CodyCBakerPhD: An image can be thought of (and indeed sometimes is extracted and saved) as a single frame of a video, so I think we should include it

"Image" is good with me, "photo" implies more about how the image was captured and stored

yarikoptic:
A great usecase came from @xiaonansun is video recording of epileptic patients along with EEG or iEEG or  other (single units) recording.

yarikoptic:
and @mvdm 's lab records rodent faces for eye tracking + potential whiskers analysis etc!

aimeme-ag:
I wanted to ask whether is considering support for subtitle or caption files, such as .srt or .vtt, associated with the recordings. In our MoBI arts datasets, especially a Balinese performance project, subtitles could be useful for storing time-aligned transcriptions in the original language, as well as English translations. 

Would subtitle files be better represented as external files linked to the video, as _events.tsv annotations, or should subtitle/caption formats be considered as part of the audio/video extension?

bendichter:
## Consolidation update

This PR now carries the **complete BEP047 proposal** in one place. It combines what were previously three separate PRs:

- #2367 (Yarik's common media file definitions for BEP044/BEP047) — now closed
- yarikoptic#2 (BEP047 behavioral layer on top of those definitions) — now closed
- the original BEP047 work on this branch

Both upstream PRs have been closed with pointers here, so **#2231 is the single canonical PR to review.** The branch has been merged onto current `master`.

### What's included (12 files)

- **`src/appendices/media-files.md`** — shared appendix defining media file naming, accepted formats/codecs, and common stream metadata (duration, audio/video/image properties). Applies to both BEP044 (stimuli) and BEP047 (recordings).
- **`src/modality-specific-files/behavioral-experiments.md`** — the BEP047 behavioral layer: `_audio`, `_video`, `_audiovideo`, and `_image` recordings in `beh/`, with the `recording`/`run`/`split` entities and behavioral-specific device metadata.
- Schema objects/rules (`extensions`, `suffixes`, `metadata`, `rules/files/raw/beh`, `rules/sidecars/{beh,media}`) plus the mkdocs/macro plumbing for the metadata tables.

bendichter:
https://github.com/bids-standard/bids-examples/pull/523 is up do date with this branch

tools/mkdocs_macros_bids/macros.py
```diff
@@ -203,6 +203,52 @@ def make_suffix_table(suffixes, src_path=None):
     return table
 
 
+def make_extension_table(extensions, src_path=None):
```
yarikoptic, line 206: we already have where such macro could be used seems to me

Image

could you submit a separate PR with commit(s) adding this macro as in this PR but also making it used in those other places?  if anyhow different formatting - we better harmonize/arrive at common reusable macro one way or another

bendichter: Good callout on reuse. Digging in, I think these are actually two different jobs:

- The **EEG/EMG/iEEG tables** are *format*-level: they group multiple extensions per row (BrainVision `.vhdr/.vmrk/.eeg`, EEGLAB `.set/.fdt`), link the format to its homepage, and carry format-specific rules (`edf+`/`bdf+`, "capital `.EDF` MUST NOT be used"). The schema's `objects.extensions` doesn't model any of that, and these formats are very stable.
- The **media-files macro** is a simple *per-extension* list linked to the glossary.

To make one macro cover both, we'd have to add homepage URLs + format grouping to the schema purely to serve 3 rarely-changing tables — or lose content to fit the current shape. IMO that's more complexity than the duplication it removes, so I'd lean toward leaving those three as manual tables and keeping `make_extension_table` scoped to plain per-extension lists. If a shared "format" model comes up again elsewhere, happy to revisit it as its own thing. WDYT?

yarikoptic (approved):
I think this is in a good shape and would allow also BEP042 to start making use of the metadata fields and macros

tools/mkdocs_macros_bids/macros.py
yarikoptic: ok, filed a dedicated issue for htat
-  https://github.com/bids-standard/bids-specification/issues/2461

yarikoptic:
Since this would relate to the software which would use such recordings, I would love to invite @talmo @MMathisLab @niksirbi @roaldarbol as relating to 
- https://github.com/bids-standard/bids-specification/issues/2057

Please have a look at this PR changes and/or rendered version at https://bids-specification--2231.org.readthedocs.build/en/2231/modality-specific-files/behavioral-experiments.html . It would be great to hear your opinion about the suggested BIDS file naming conventions for audiovideo recordings and metadata to come along in the sidecars.

edit: note -- you are welcome to submit an "official" review of the PR

niksirbi:
Thanks for tagging me @yarikoptic! I'm quite busy at the moment, but would definitely want to review this. Is there a time horizon / deadline for the review period?

roaldarbol:
Currently on holidays, I'll have a look in early August. :-)

bendichter:
@roaldarbol checking in about a review

src/modality-specific-files/behavioral-experiments.md
roaldarbol: I think it's absolutely valuable to be able to store images. Especially since they're often used for training the pose estimation models, as pointed out by @talmo in https://github.com/bids-standard/bids-specification/issues/2057#issuecomment-2670031635. So in order to enable full reproducibility end-to-end, not only do we need the raw videos, but also the specific images that were used for training. I would note though, that it's mostly useful if the training data - where the ground-truth points were placed - is also available, and I don't know if it currently could be somehow.

roaldarbol:
@bendichter Finally found some time to look through it. Generally I think it looks very thorough and well thought out. I mentioned in a comment that if the idea with images (or one of the things that it should help with) is regarding reproducibility of pose estimation or other models, then the annotations on those images also need to be stored somewhere. Is that already possible, like is there place those could meaningfully belong? And if so, it may be worth adding that extra bit of context in the section that discusses images and explicitly mentions pose estimation. 

Great work! I'll be looking more closely at your metadata sidecars for adoption in animovement.

CodyCBakerPhD:
> I mentioned in a comment that if the idea with images (or one of the things that it should help with) is regarding reproducibility of pose estimation or other models, then the annotations on those images also need to be stored somewhere.

Such annotations would count as 'derivative' data since they could be tagged or interpreted in multiple ways, possibly even automatically by software

BTW this little conveience app exports in BEP047-compliant format (and supports both image type as well as associated text descriptions): https://clip-extractor.brain-bbqs.org/

You can try it out with pose overlay on the SLEAP tutorial data: https://github.com/talmolab/sleap-tutorial-data/tree/main/new_data

roaldarbol:
@CodyCBakerPhD You could also see https://roald-arboel.com/croppy/ which is also for extracting clips (and command lab video operations). If there's anything I could do to make it better fit BEP047, please let me know on the repo! :-)

CodyCBakerPhD:
Thought just occurred - given the recent conversations about alignment event information on BEP032 into the `_events.tsv` file with relevant tags, what then should the recommendation be for BEP047 when we have multiple cameras running at roughly (but maybe not exactly) the same time?

> @CodyCBakerPhD You could also see https://roald-arboel.com/croppy/ which is also for extracting clips (and command lab video operations). If there's anything I could do to make it better fit BEP047, please let me know on the repo! :-)

Thanks will do - local apps certainly have some advantages in terms of hardware performance (web apps have a notoriously hard time leveraging GPU for things like this)

yarikoptic:
> Thought just occurred - given the recent conversations about alignment event information on BEP032 into the `_events.tsv` file with relevant tags, what then should the recommendation be for BEP047 when we have multiple cameras running at roughly (but maybe not exactly) the same time?

there is `_scans.tsv` where per each data file (related rename issue: https://github.com/bids-standard/bids-2-devel/issues/29) we have already [spec'ed](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#scans-file) `acq_time` . So if various videos started to be acquired at different times -- there it will be. @vmdocua  is to inject such metadata for "stimuli video" for

- https://github.com/ReproNim/reprostim/issues/276

src/modality-specific-files/behavioral-experiments.md
bendichter: Agreed on the images themselves — that's why `_image` is in the PR, and lines 83-86 call out training frames for pose estimation and frames extracted from video as the motivating cases. The ground-truth markings are a different category though: keypoint placements are derived data, not acquired data, and I'm keeping BEP047 narrowly scoped to what comes off the instrument. Annotations are out of scope for this PR. They'd belong in a derivatives spec, where provenance back to the source images can be expressed properly.

PR #2232, opened 2025-10-26
DOCS: Fix HED resources link in hierarchical event descriptors documentation

bendichter:
image

https://github.com/hed-standard/hed-resources/commit/3b3461009d2db6de8d45f40a97d51abad9b51eba#diff-e91be36adf148d7583082e3b2d81a1e8075677b1c1080ed03ed9fd3a1646326d


yarikoptic:
I think it is a no-brainer since just a url and prior one is indeed broken, so I will just merge it

PR #2235, opened 2025-10-27
fix: Use max of float-or-array fields for numerical metadata checks

### LINKED ISSUE #2234: [BUG] Issues with `rules.checks.asl`

Some current ASL rules appear to assume - or only define rule for - cases where these metadata fields contain numbers rather than arrays.

We have multi-delay ASL data where PostLabelingDelay is a 90-element array with values ranging from 0 to 2.2 (i.e., none are above the threshold of 10 in the check), and the check is being triggered.

LABELING_DURATION_GREATER https://github.com/bids-standard/bids-specification/blob/d62a5d2deb1a3ce6d53e99f65d083f758d153dfb/src/schema/rules/checks/asl.yaml#L241-L250

POST_LABELING_DELAY_GREATER
https://github.com/bids-standard/bids-specification/blob/d62a5d2deb1a3ce6d53e99f65d083f758d153dfb/src/schema/rules/checks/asl.yaml#L216-L226

_No response_

https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#_asljson-file

tsalo:
Closes #2234.

Changes proposed:

- Instead of directly comparing fields that can either be floats or arrays to a single value, take the max of the field. This assumes that `max(float)` will return the float. The checks I modified are `PostLabelingDelayGreater`, `BolusCutOffDelayTimeGreater`, `LabelingDurationGreater`, and `EchoTimeGreaterThan`.

### DIFF
--- a/src/schema/rules/checks/asl.yaml
+++ b/src/schema/rules/checks/asl.yaml
@@ -201,16 +201,18 @@ ASLTotalAcquiredPairsASLContextLength:
     code: TOTAL_ACQUIRED_VOLUMES_NOT_CONSISTENT
     message: |
       The number of values for 'TotalAcquiredPairs' for this file does not match number of
-      volumes in the associated 'aslcontext.tsv'.
-      'TotalAcquiredPairs' is the original number of 3D volumes acquired for each volume defined in the
-      associated 'aslcontext.tsv'.
+      control-label pairs in the associated 'aslcontext.tsv'.
+      'TotalAcquiredPairs' is the original number of control-label pairs
+      defined in the associated 'aslcontext.tsv'.
     level: warning
   selectors:
     - suffix == "asl"
     - type(associations.aslcontext) != "null"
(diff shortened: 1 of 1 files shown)


src/schema/rules/checks/func.yaml
```diff
@@ -27,7 +27,7 @@ RepetitionTimeGreaterThan:
     - suffix == "bold"
     - type(sidecar.RepetitionTime) != "null"
   checks:
-    - sidecar.RepetitionTime <= 100
+    - max(sidecar.RepetitionTime) <= 100
```
effigies: I believe RepetitionTime is always a single number, so this is unnecessary.

tsalo: Oh you're right. I was thinking of RepetitionTimePreparation. I'll fix that.

effigies:
@tsalo is this ready for review?

effigies (approved):

PR #2237, opened 2025-10-27
[ENH] Updated HED Version to 8.4.0 in examples

VisLab:
Updated specification to use latest HED version in the examples.

### DIFF
--- a/src/appendices/hed.md
+++ b/src/appendices/hed.md
@@ -250,7 +250,7 @@ The following `dataset_description.json` indicates that only the HED-SCORE libra
 ```JSON
 {
   "Name": "A great experiment",
-  "BIDSVersion": "1.10.0",
+  "BIDSVersion": "1.10.1",
   "HEDVersion": "score_1.0.0"
 }
 ```


src/appendices/hed.md
```diff
@@ -250,7 +250,7 @@ The following `dataset_description.json` indicates that only the HED-SCORE libra
 ```JSON
 {
   "Name": "A great experiment",
-  "BIDSVersion": "1.7.0",
+  "BIDSVersion": "1.10.0",
```
neuromechanist:
```suggestion
  "BIDSVersion": "1.10.1",
```

I wonder why this one is not 1.10.1 like others.

VisLab: should be....thanks

yarikoptic:
2 approvals, I will merge. but I am curious 

- was that necessary for anything @VisLab ?
- are you planing to perform similar upgrades in the future just to point to more recent version?
- should we strive to make it some form of a "convention" or do the opposite and point to minimal compatible version or ...
?

VisLab:
@yarikoptic  @effigies The reason that it is important to revise the examples to the latest version of HED is that these are EXAMPLES.  This means that people hopefully use them as models when putting their dataset into BIDS. These examples should be continually updated to show best practices and the latest versions. Some of the examples both in the Examples and the Specification have ancient values of BIDS also. 

> 2 approvals, I will merge. but I am curious
> 
>     * was that necessary for anything @VisLab ?
> 
>     * are you planing to perform similar upgrades in the future just to point to more recent version?
> 
>     * should we strive to make it some form of a "convention" or do the opposite and point to minimal compatible version or ...
>       ?

Issue #2010, opened 2024-12-16
extra warning during build with recent panda


Remi-Gau:
Getting a lot of those.

Should fix probably soon before it becomes too annoying.

```
FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly
           call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
             File "/home/remi/github/bids/website/env/lib/python3.12/site-packages/pandas/core/internals/blocks.py", line 904, in replace
               warnings.warn(
             File "/home/remi/github/bids/website/env/lib/python3.12/site-packages/bidsschematools/render/utils.py", line 132, in
               df = df.replace("", math.nan).dropna(axis=1, how="all").fillna("")
```

PR #2240, opened 2025-10-29
fix: Address warnings, error on malformed TSV contents

### LINKED ISSUE #2238: "ParserWarning: Length of header or names does not match length of data"
`mkdocs serve` has some of such warnings.  Am I alone or you see them too @bids-standard/maintainers ???

```
INFO    -  ParserWarning: Length of header or names does not match length of data. This leads to a loss of
           data with index_col=False.
             File
           "/home/yoh/proj/bids/bids-specification/.venv/lib/python3.11/site-packages/pandas/io/parsers/base_parser.py",
           line 925, in _check_data_length
               warnings.warn(
             File
           "/home/yoh/proj/bids/bids-specification/tools/schemacode/src/bidsschematools/render/tsv.py", line
           31, in
               df = pd.read_csv(
INFO    -  ParserWarning: Length of header or names does not match length of data. This leads to a loss of
           data with index_col=False.
             File
           "/home/yoh/proj/bids/bids-specification/.venv/lib/python3.11/site-packages/pandas/io/parsers/base_parser.py",
           line 925, in _check_data_length
               warnings.warn(
             File
           "/home/yoh/proj/bids/bids-specification/tools/schemacode/src/bidsschematools/render/tsv.py", line
           31, in
               df = pd.read_csv(

```

those suggest some misformatted TSV and I think we should make them into ERRORs so PRs do not allow for such to appear. WDYT?

effigies:
Closes https://github.com/bids-standard/bids-specification/issues/2010 following pattern used in https://github.com/bids-standard/pybids/pull/1136.

Closes #2238.

### DIFF
--- a/src/derivatives/imaging.md
+++ b/src/derivatives/imaging.md
@@ -548,9 +548,9 @@ standard BIDS labels:
 
 ```tsv
 index	name	abbreviation
-137	pars	opercularis	IFGop
-138	pars	triangularis	IFGtr
-139	pars	orbitalis	IFGor
+137	pars opercularis	IFGop
+138	pars triangularis	IFGtr
+139	pars orbitalis	IFGor
 ```
 
 <!-- Link Definitions -->
--- a/src/modality-specific-files/microscopy.md
+++ b/src/modality-specific-files/microscopy.md
@@ -420,8 +420,8 @@ Additional optional columns MAY be used to describe other subjects' attributes.
 `participants.tsv` example:
 ```tsv
 participant_id	species	strain	strain_rrid
-sub-01	mus	musculus	C57BL/6J	RRID:IMSR_JAX:000664
-sub-02	mus	musculus	C57BL/6J	RRID:IMSR_JAX:000664
+sub-01	mus musculus	C57BL/6J	RRID:IMSR_JAX:000664
+sub-02	mus musculus	C57BL/6J	RRID:IMSR_JAX:000664
 ```
 
 `participants.json` example:


tools/schemacode/src/bidsschematools/render/tsv.py
```diff
@@ -28,31 +30,30 @@ def fence(
...
-
-        # Remove newlines from HTML to prevent copy-paste from inserting spaces
-        return html.replace("\n", "")
-    except Exception:
-        import traceback
-
-        exc = traceback.format_exc()
-        return f"<pre>{exc}</pre>"
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
```
yarikoptic: sorry -- wouldn't that ignore warnings entirely and we wouldn't know when something is off or about to be removed?

tools/schemacode/src/bidsschematools/render/utils.py
```diff
@@ -3,6 +3,45 @@
...
+
+    This also opts-in to copy-on-write, which previously required `copy=False`
+    to be set.
+    """
+    if args := _pandas_3_0_options():
+        return pd.option_context(*args)
+
+    import contextlib
+
+    return contextlib.nullcontext()
```
yarikoptic, line 29: such a complexity with a context manager only so we do not affect potential outside libraries functioning and scope it for our code here? (just that `mkdocs` runs for me already so long that I start weighing in when adding additional cpu-cycle-usages like that ;) )

tools/schemacode/src/bidsschematools/render/tsv.py
effigies: It promotes all warnings to errors in this block.

tools/schemacode/src/bidsschematools/render/utils.py
effigies: Correct. I do not want to modify global state.

yarikoptic: FWIW I would've gone with a function decorator for smaller diffs

tools/schemacode/src/bidsschematools/render/tsv.py
yarikoptic: ignorant me misread such specification (I thought it would "filter" them out), but docs at https://docs.python.org/3/library/warnings.html#warning-filter set me straight:

> "error"   turn matching warnings into exceptions

tools/schemacode/src/bidsschematools/render/utils.py
effigies: Reworked as decorators.

yarikoptic (approved):
Thank you @effigies . I test driven it (both context handler and decorator versions) and have not see warnings popping up any longer -- life will be so much better now!!!!

PR #2248, opened 2025-11-06
fix: Get TSV tables rendering again

effigies:
#2240 broke TSV builds due to an issue with `@contextlib.contextmanager`-built context managers. This replaces it with a normal context manager.


effigies:
Before:

image


After:

image

PR #2255, opened 2025-11-12
[ENH] provide RECOMMENDATION of the "Last, First" form for the `Authors` names

yarikoptic:
ATM there is no consistency across e.g. OpenNeuro datasets, e.g.

    ds006267/dataset_description.json:
      Authors=['Katherine M. Cole', 'Shau-Ming Wei', 'Pedro E. Martinez', 'Tuong-Vi Nguyen', 'Michael D. Gregory', 'J. Shane Kippenhan', 'Philip D. Kohn', 'Steven J. Soldin', 'Lynnette K. Nieman', 'Jack A. Yanovski', 'Peter J. Schmidt', 'Karen F. Berman']
    ds006269/dataset_description.json:
      Authors=['Lucy Pritchard', 'Ingrid Buller-Peralta', 'Sally M Till', 'Peter C Kind', 'Alfredo Gonzalez-Sulser']
    ds006303/dataset_description.json:
      Authors=['Linke, Julia', 'Naim, Reut', 'Haller, Simone', 'Khosravi, Parmis', 'Scheinberg, Beck', 'Byrne, Meghan', 'Harrewijn, Anita', 'Leibenluft, Ellen', 'Brotman, Melissa', 'Winkler, Anderson', 'Pine, Daniel']

and that is why some are left ambigous like

    ds003834/dataset_description.json:
      Authors=['Matteo Visconti di Oleggio Castello', 'James V. Haxby', 'M. Ida Gobbini']

where for Matteo I believe there is a composite last name of "Visconti di Oleggio Castello" per e.g.

	❯ curl --silent https://raw.githubusercontent.com/bids-standard/pybids/refs/heads/main/.zenodo.json | grep Matteo
		  "name": "Visconti di Oleggio Castello, Matteo",

but for the other 2 authors, the only last word is the Family name.

TODOs
- [ ] validation: add a check for validator to WARN about using `First Last`, in particular if any of the names has more than 2 components?  @effigies do you see an easy way to do that?
- [ ] anywhere else in the text to add information about this?

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -547,6 +547,7 @@
           "datetime",
           "file_relative",
           "participant_relative",
+          "person_name",
           "rrid",
           "stimuli_relative",
           "time",
--- a/src/schema/objects/formats.yaml
+++ b/src/schema/objects/formats.yaml
@@ -112,6 +112,12 @@ participant_relative:
     without starting with "/" (an absolute path) or "sub/"
     (a relative path starting with the participant directory, rather than relative to that directory).
   pattern: '(?!/)(?!sub-)[0-9a-zA-Z+/_\-.]+'
+person_name:
+  display_name: Name of a person
+  description: |
+    Name for a person in a single string with clear disambiguation of First (Given) and Last (Family)
+    name(s).  Should follow `Last Names, First Names` convention.
+  pattern: '.+, .+'
 rrid:
   display_name: Research resource identifier
   description: |
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -213,6 +213,7 @@ Authors:
   type: array
   items:
     type: string
+    format: person_name
 B0FieldIdentifier:
   name: B0FieldIdentifier
   display_name: B0 Field Identifier
@@ -1327,6 +1328,7 @@ Format:
     - stimuli_relative
     # Miscellaneous
     - hed_version
+    - person_name
 FrameAcquisitionDuration:
   name: FrameAcquisitionDuration
   display_name: Frame Acquisition Duration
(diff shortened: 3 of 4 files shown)


effigies:
> * add a check for validator to WARN about using `First Last`, in particular if any of the names has more than 2 components?  @effigies do you see an easy way to do that?

Regex?

CodyCBakerPhD:
I would personally prefer a list of objects for the `Authors` field. That could resolve all types of name regex issues that might exist by splitting out first/last/middle/family/maiden/etc fields as well as pair with the lovely metadata as shown in https://github.com/bids-standard/bids-specification/pull/2397#issue-4263584415

But I suppose this counts as a 'back compatability' break?

yarikoptic:
ping on this, and I filed another related PR which might complicate the situation here:
- #2397 

since for institutions there would be no regex to check, and without explicit marker (having records per author) there would be no way.  So - it is kinda one or another.  

And now that CITATION.cff is "supported" (as disabling check for Authors to be defined), I guess we could shift into offloading that all into CITATION.cff and recommend it in favor over Authors here.

yarikoptic:
potentially, we could attempt at migrating  the names from prior versions of BIDS . People would be advised to review anyways, thus potentially fixing up automigrated names.  But then in the light of 

- #2397

similarly to how CFF does it pretty much, then starting from that BIDS version we would be able to tell persons from organizations by following the `Last, First` pattern in 'Authors'.

yarikoptic:
I think I (?) prematurely marked that PR as closing this one in https://github.com/bids-standard/bids-specification/pull/2397 -- that one just allows for institutions but gives no recommendation for human entries. So this one is still open, although now would need to be adjusted as to just add RECOMMENDATION text likely without validation.  Ideally, indeed, we should eventually converge on unambigous schema of records.

PR #2258, opened 2025-11-13
[ENH] BEP 16: Diffusion modeling derivatives

arokem:
Introduces specifications for diffusion modeling derivatives. 

This PR replaces #2211 so that maintainers can push changes into the PR during its development.

> [!Note]
>
> At this point, the best way to communicate about this BEP is through this GitHub pull request.
>
>
> [**HTML preview of this BEP**](XXX - will add when available)
>

### DIFF
--- a/src/schema/rules/sidecars/derivatives/diffusion.yaml
+++ b/src/schema/rules/sidecars/derivatives/diffusion.yaml
@@ -1,0 +1,46 @@
+---
+OrientationEncodingTypeAmplitudes:
+  selectors:
+    - dataset.dataset_description.DatasetType == "derivative"
+    - suffix == "dwimap"
+    - sidecar.OrientationEncoding.Type == "amplitudes"
+  fields:
+    AmplitudesDirections__OrientationEncoding:
+      level: required
+      level_addendum: if `OrientationEncoding.Type` is `"amplitudes"`
+
+OrientationEncodingTypeSphericalHarmonics:
+  selectors:
+    - dataset.dataset_description.DatasetType == "derivative"
+    - suffix == "dwimap"
+    - sidecar.OrientationEncoding.Type == "sh"
+  fields:
+    SphericalHarmonicsBasis__OrientationEncoding:
+      level: required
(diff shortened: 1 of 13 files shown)


arokem:
@PeerHerholz : there was some discussion of examples on #2211. Have you made any progress on that?

src/schema/rules/sidecars/derivatives/diffusion.yaml
```diff
@@ -0,0 +1,46 @@
+---
+OrientationEncodingTypeAmplitudes:
+  selectors:
+    - dataset.dataset_description.DatasetType == "derivative"
+    - suffix == "dwimap"
+    - sidecar.OrientationEncoding.Type == "amplitudes"
+  fields:
+    AmplitudesDirections__OrientationEncoding:
+      level: required
+      level_addendum: if `OrientationEncoding.Type` is `"amplitudes"`
```
tsalo, line 10: I am having trouble figuring out how to validate subobjects' field values. In this case, if "OrientationEncoding"["Type"] is "amplitudes", "OrientationEncoding"["AmplitudesDirections"] MUST be defined.

arokem:
Bringing over two questions/issues that I accidentally recorded in #2211 but now belong here: 

1. @36000 points out that if we fit both a "standard" csd and a msmt csd model within the same pipeline, we would end up with the same file names here. Should we use the full "msmtcsd" here? Or use "ssstcsd" for the "standard" (single-shell, single-tissue) version of CSD?

2. For asymmetric ODFs, would it be sufficient if we add `mrtrix3_asym` and `descoteaux_asym` to the list of possible basis sets, to imply that the coefficients include both even and odd orders?

arnaudbore:
I was just looking at the way derivated images are named from NODDI model: [here](https://github.com/bids-standard/bids-specification/pull/2258/changes#diff-b68220a64b2c9520ee1fe82d980dcbd38e2eeea89f07b9dd3b328822a311ba16R647).

 What happened to ISOVF/FWF ? Thank you

yarikoptic:
This PR is read in CI and has conflicts. Needs TLC. ping @arokem

PR #2262, opened 2025-11-17
[ENH] [Derivatives] Add MImap suffix for molecular imaging maps

melanieganz:
In order to enable the addition of PET examples to the current [atlas BEP](https://github.com/bids-standard/bids-specification/discussions/2250) proposal, we need to define what a PET derivative will be called. The PET derivatives are still under development (see [here](https://bids.neuroimaging.io/extensions/beps/bep_023.html)), but we have a pretty good idea how they should look like. 

Hence we added a MImap suffix with details on units and descriptions to the schema.

Derived with help of @mnoergaard @CPernet

### DIFF
--- a/src/schema/objects/suffixes.yaml
+++ b/src/schema/objects/suffixes.yaml
@@ -112,6 +112,16 @@ MESE:
     times and is primarily used for T2 mapping.
     Please note that this suffix is not intended for the logical grouping of
     images acquired using an Echo Planar Imaging (EPI) readout.
+MImap:
+  value: MImap
+  display_name: Molecular Imaging map
+  description: |
+    MImaps can have different units, e.g. unitless, ml/ml or pmol/ml. See 
+    [Innis et al. 2007](https://journals.sagepub.com/doi/full/10.1038/sj.jcbfm.9600493) for unit descriptions.
+    MImaps are derivatives of molecular imaging (PET or SPECT) raw data. They can 
+    either be unitless, e.g. a BP_ND or SUVR, or have the units ml/cm^3 for a V_T or pmol/ml 
+    for average density maps (see e.g. [Beliveau et al. 2017](https://pmc.ncbi.nlm.nih.gov/articles/PMC5214625/)).     
+  unit: arbitrary
 MP2RAGE:
   value: MP2RAGE
   display_name: Magnetization Prepared Two Gradient Echoes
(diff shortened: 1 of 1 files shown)


src/schema/objects/suffixes.yaml
```diff
@@ -112,6 +112,16 @@ MESE:
...
     images acquired using an Echo Planar Imaging (EPI) readout.
+MImap:
+  value: MImap
+  display_name: Molecular Imaging map
+  description: |
+    MImaps can have different units, e.g. unitless, ml/ml or pmol/ml. See
+    [Innis et al. 2007](https://journals.sagepub.com/doi/full/10.1038/sj.jcbfm.9600493) for unit descriptions.
+    MImaps are derivatives of molecular imaging (PET or SPECT) raw data. They can
+    either be unitless, e.g. a BP_ND or SUVR, or have the units ml/cm^3 for a V_T or pmol/ml
+    for average density maps (see e.g. [Beliveau et al. 2017](https://pmc.ncbi.nlm.nih.gov/articles/PMC5214625/)).
```
bendhouseart: Fix for bad Latin check.

```suggestion
    for average density maps (see [Beliveau et al. 2017](https://pmc.ncbi.nlm.nih.gov/articles/PMC5214625/)).
```

bendhouseart (changes requested):
Approval pending latin check fix.

src/schema/objects/suffixes.yaml
```diff
@@ -112,6 +112,16 @@ MESE:
...
     images acquired using an Echo Planar Imaging (EPI) readout.
+MImap:
+  value: MImap
+  display_name: Molecular Imaging map
+  description: |
+    MImaps can have different units, e.g. unitless, ml/ml or pmol/ml. See
+    [Innis et al. 2007](https://journals.sagepub.com/doi/full/10.1038/sj.jcbfm.9600493) for unit descriptions.
+    MImaps are derivatives of molecular imaging (PET or SPECT) raw data. They can
+    either be unitless, e.g. a BP_ND or SUVR, or have the units ml/cm^3 for a V_T or pmol/ml
+    for average density maps (see [Beliveau et al. 2017](https://pmc.ncbi.nlm.nih.gov/articles/PMC5214625/)).
```
mathesong, line 123:
```suggestion
    MImaps are derivatives of molecular imaging (PET or SPECT) raw data which 
    can represent different quantities with different units. As defined by 
    [Innis et al. 2007]
    (https://journals.sagepub.com/doi/full/10.1038/sj.jcbfm.9600493), these 
    maps can be unitless (e.g. BP_ND or SUVR), or have units of ml/cm^3 (e.g. 
    V_T) or pmol/ml (average density maps, e.g. 
    [Beliveau et al. 2017](https://pmc.ncbi.nlm.nih.gov/articles/PMC5214625/)),
    among others.
```

mathesong (commented):
I would suggest changing the phrasing a little bit, at least to clean up the structure (and avoid being about units, then what mimaps are and then units again); but also to open up the possibility of mimaps representing other quantities too (e.g. Kd maps as estimated by the Lassen plot filter)

PR #2271, opened 2025-12-03
chore: Add "serve" rule to Makefile and provide description to what actions available

yarikoptic:
by default make does nothing!  I kept it that way but made that "nothing" explicit. I also added serve target so it is straightforward to trigger build&serve

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -267,7 +267,8 @@ such as when you build the documentation locally.
 
 ### 4. Ready to build!
 
-Using the terminal (command line) please enter `uv run mkdocs serve`. Alternatively, if you would like to use `make`, run `make serve` which will ensure the node and `uv` environments are present and trigger the `mkdocs` build.
+Using the terminal (command line) please enter `uv run mkdocs serve`.
+Alternatively, if you would like to use `make`, run `make serve`, which will run `uv sync` and then `mkdocs`.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,15 @@
...
+
+.PHONY: all serve install update_contributors formatschema
+
 all:
+	@echo "Nothing is done by default. Consider following targets:"
+	@echo "  install -- prep environment"
+	@echo "  serve   -- prep environment and build and serve docs"
+	@echo "  formatschema -- format and commit(!) schema"
+
+serve: install
```
effigies: No javascript is needed.

```suggestion
serve: .venv
```

CONTRIBUTING.md
```diff
@@ -267,12 +267,12 @@ such as when you build the documentation locally.
 
 ### 4. Ready to build!
 
-Using the terminal (command line) please enter `uv run mkdocs serve`.
+Using the terminal (command line) please enter `uv run mkdocs serve`. Alternatively, if you would like to use `make`, run `make serve` which will ensure the node and `uv` environments are present and trigger the `mkdocs` build.
```
effigies:
```suggestion
Using the terminal (command line) please enter `uv run mkdocs serve`.
Alternatively, if you would like to use `make`, run `make serve`, which will run `uv sync` and then `mkdocs`.
```

PR #2286, opened 2025-12-18
Replace `atlas-<label>_description.json` with `atlases.tsv` pattern and also add `templates.tsv` and `cohorts.tsv`

yarikoptic:
This is a follow-up to PR #1714 (BEP038: Templates and atlases) to make atlas metadata approach consistent with existing BIDS conventions for entity-specific TSV files (like participants.tsv, sessions.tsv, descriptions.tsv, etc).

It was edited to be split into 2 commits, first one is submitted independently for a dedicated, and less radical discussion:

- https://github.com/bids-standard/bids-specification/pull/2287

Changes of the 2nd commit here:
- Replace atlas-<label>_description.json with atlases.tsv

The atlases.tsv includes columns derived from the original JSON fields:
- atlas_id, name, license (REQUIRED)
- description, authors (RECOMMENDED)
- curators, funding, references_and_links, species, sample_size, derived_from, level_type, rrid (OPTIONAL)

Array fields (authors, curators, funding, references_and_links) use JSON-encoded strings in TSV cells.

Schema changes:
- Added column definitions for all atlas/template/cohort TSV columns
- Added suffix definitions for atlases, templates, cohorts
- Added tabular_data rules for column validation
- Added file rules supporting both root-level and template-level naming
- Removed old atlas_description JSON-based rules and associations

This addresses the inconsistency noted in:

- Issue #2285: atlas-<label>_description.json pattern introduced by BEP038

via already prevalent in BIDS

- Issue #2283: Proposal to formalize {entities_plural}.tsv pattern

Issue #2285, opened 2025-12-18
Formalize concept of '{entity}-<label>_description.json'


yarikoptic:
To overall provide a mechanism for per an `{entity}_id` metadata, alternative to `{entity_plural}.tsv` which we used so far and for which I filed a separate

- #2283 

It was born to provide a similar (if not inspired) to `dataset_description.json` nested (includes lists of Authors etc) metadata record per each value of the `atlas-<label>` in 

- https://github.com/bids-standard/bids-specification/pull/1714/ by @bep038 group

I argued that we already have a conceptual solution for such aggregated metadata in the form of `{entity_plural}.tsv` (such as `participants.tsv`, `sessions.tsv`) and that introduction of such files really introduces a new conceptual new location for metadata specification as for no other entity we have such a file ATM, thus this relates heavily to

-  https://github.com/bids-standard/bids-2-devel/issues/85

In principle we can have both this and `{entity_plural}.tsv`  "interchangeable" besides forbidding aspects:

- names are `PascalCased` in json and `snake_cased` in tsv, and we do not have "crosswalks"
- optional columns in .tsv could still get their schema described whenever we do not have anything similar for .json so if anything custom (e.g. `Version` as was already provided in feedback via email) added to such a file -- would not be validatable.

thus making us to choose one or another for any particular entity, with `atlas` being the only one for `_description.json` ATM and mixing it with non-entity `dataset_description.json`.  Also it might potentially "interfer" with Inheritance principle as suggesting that it applies to all `_description.json` files undeneath not really for all instances of that `atlas` label being used. 

Without formalization, we would just make it more complicated for people to 

- create datasets (do I use .tsv or .json for this entity?)
- operate on datasets programatically (need to code for both principles in general but only for specific entities)

Issue #2283, opened 2025-12-17
Formalize the concept of `[{leading entities}_]{entity_plural}.{tsv,json}`  file(s)


yarikoptic:
This is precursor to

- https://github.com/bids-standard/bids-specification/issues/2273

and mentioned through out the existing issues and PRs, e.g.

- https://github.com/bids-standard/bids-specification/pull/1328#issuecomment-2610160946

and would be required to facilitate

- #1809 (for https://github.com/bids-standard/bids-2-devel/issues/54)
- https://github.com/bids-standard/bids-2-devel/issues/94
- https://github.com/bids-standard/bids-2-devel/issues/85
- https://github.com/bids-standard/bids-specification/issues/2013

but I failed to find a dedicated issue.

### Current situation/issue

<details>
<summary>
We have an expanding list of already defined .tsv files summarized below (including `_scans.tsv` which is not per-entity per se but close in spirit):
</summary>

####  Table 1: Entity-level TSV files ({entity-plural}.tsv pattern)

  | File Pattern | Entity | Location | Column 1 | Column 2 | Column 3 | Column 4 | Issues etc |
  |---|---|---|---|---|---|---|---|
  | [`participants.tsv`](https://github.com/bids-standard/bids-specification/blob/master/src/modality-agnostic-files/data-summary-files.md?plain=1#L3) | subject (`sub-<label>`) | Dataset root | `participant_id` | `species` (R) | `age` (R) | `sex` (R) |  [bids-2-devel#14](https://github.com/bids-standard/bids-2-devel/issues/14) |
  | [`samples.tsv`](https://github.com/bids-standard/bids-specification/blob/master/src/modality-agnostic-files/data-summary-files.md?plain=1#L84) | sample (`sample-<label>`) | Dataset root | `sample_id` | `participant_id` | `sample_type` | `pathology`  (R) | Composite index (`sample_id` + `participant_id`) |
  | [`sub-<label>_sessions.tsv`](https://github.com/bids-standard/bids-specification/blob/master/src/modality-agnostic-files/data-summary-files.md?plain=1#L200) | session (`ses-<label>`) | Subject folder | `session_id` | `acq_time` (O) | `pathology` (R)  | `HED` (O) | |
[...]

PR #2287, opened 2025-12-18
ENH: Add templates.tsv and cohorts.tsv for template/cohort entity metadata

yarikoptic:
Introduce templates.tsv and cohorts.tsv files consistent with existing BIDS entity-specific TSV files (participants.tsv, sessions.tsv).

- templates.tsv: placed at derivative dataset root, describes tpl-<label> entities
- cohorts.tsv: placed within tpl-<label>/ directory, describes cohort-<label> entities

This enables consistent documentation of template and cohort entities in derivative datasets. Part of BEP038 atlas metadata improvements.

Schema changes:
- Add template_id and cohort_id columns to columns.yaml
- Add templates and cohorts suffixes to suffixes.yaml
- Add Templates and Cohorts rules to common_derivatives.yaml
- Add templates and cohorts file rules to tables.yaml
- Update atlas.md with templates.tsv and cohorts.tsv documentation

Related to:
- https://github.com/bids-standard/bids-specification/issues/2285
- https://github.com/bids-standard/bids-specification/issues/2283

Part of the larger
- #2286 

for independent consideration. Attn @bids-standard/bep038

### DIFF
--- a/src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
+++ b/src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
@@ -26,7 +26,7 @@ Templates:
       description_addendum: |
         The corresponding label column is `template_id`.
   index_columns: [template_id]
-  additional_columns: allowed
+  additional_columns: allowed_if_defined
 
 Cohorts:
   selectors:
@@ -42,7 +42,7 @@ Cohorts:
       description_addendum: |
         The corresponding label column is `cohort_id`.
   index_columns: [cohort_id]
-  additional_columns: allowed
+  additional_columns: allowed_if_defined
 
 Descriptions:
   selectors:


src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
```diff
@@ -12,6 +12,38 @@ SegmentationLookup:
...
+    - cohort_id
+    - description__entities
+  columns:
+    cohort_id: required
+    description__entities:
+      level: required
+      description_addendum: |
+        The corresponding label column is `cohort_id`.
+  index_columns: [cohort_id]
+  additional_columns: allowed
```
effigies: We have an opportunity here to set `additional_columns` to `allowed_if_defined`, which would require curators to define columns they are going to use. There's a philosophical question of whether we should aim for consistency with the laxness of existing rules or to push towards strictness in new rules.

The validator will check types if definitions are provided, but will only produce a warning if they are not.

What do people think?

```suggestion
  additional_columns: allowed_if_defined

Cohorts:
  selectors:
    - suffix == "cohorts"
    - extension == ".tsv"
  initial_columns:
    - cohort_id
    - description__entities
  columns:
    cohort_id: required
    description__entities:
      level: required
      description_addendum: |
        The corresponding label column is `cohort_id`.
  index_columns: [cohort_id]
  additional_columns: allowed_if_defined
```

effigies (dismissed):
Overall this seems reasonable. We should update at least one example to use these features.

src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
yarikoptic: FWIW I like "explicit better than implicit" zen, thus to restrict nearly all new introduced .tsv to have that setting!

oesteban:
> do we actually have some stats on that?

It is my subjective perception that licenses of atlases/templates follow the 80/20 rule (only 20 use standard licenses, and often with no-derivs restrictions, which is nuts for that kind of resource).

src/derivatives/atlas.md
```diff
@@ -76,6 +76,79 @@ as described in the [Derived atlases](#derived-atlases) section, below.
...
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_columns_table("derivatives.common_derivatives.Templates") }}
+
+This file MUST be located at the root of the derivative dataset.
+
+`templates.tsv` example:
+
+```tsv
+template_id	description
```
oesteban, line 108:
```suggestion
template_id	long_name
```

oesteban: I would strongly suggest RRID here as a second column before "long_name"

src/derivatives/atlas.md
```diff
@@ -76,6 +76,79 @@ as described in the [Derived atlases](#derived-atlases) section, below.
...
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_columns_table("derivatives.common_derivatives.Cohorts") }}
+
+This file MUST be located within the `tpl-<label>/` directory for which it describes cohorts.
+
+`tpl-MNIPediatricAsym_cohorts.tsv` example:
+
+```tsv
+cohort_id	description
```
oesteban, line 143: This is great :+1:

src/schema/objects/columns.yaml
```diff
@@ -120,6 +120,22 @@ derived_from:
     for example a slice of tissue (`sample-02`) derived from a block of tissue (`sample-01`).
   type: string
   pattern: ^sample-[0-9a-zA-Z+]+$
+cohort_id:
+  name: cohort_id
+  display_name: Cohort ID
+  description: |
+    A cohort identifier of the form `cohort-<label>`,
```
oesteban, line 127: I would recommend using indices instead of labels, but I know that will be contentious

oesteban (approved):
This is great, thanks!

src/schema/objects/columns.yaml
oesteban: Actually the BEP is already merged as label, disregard my comment above.

src/derivatives/atlas.md
CPernet: +1 for RRID, as always columns are optional but we know that when we put things in example they are more adopted than when we do not  -  might also consider DOI

src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
```diff
@@ -12,6 +12,38 @@ SegmentationLookup:
   index_columns: [index]
   additional_columns: allowed
 
+Templates:
+  selectors:
+    - suffix == "templates"
+    - extension == ".tsv"
+  initial_columns:
+    - template_id
+    - description__entities
```
effigies, line 21:
```suggestion
    - template_id
```

`description` being free-text, I would expect many people to choose to put it last for better alignment. I would drop the initial column requirement.

src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
```diff
@@ -12,6 +12,38 @@ SegmentationLookup:
...
+  columns:
+    template_id: required
+    description__entities:
+      level: required
+      description_addendum: |
+        The corresponding label column is `template_id`.
```
effigies, line 27: We can make `rrid` show up in the table and also validate that it matches the RRID schema:

```suggestion
    description__entities:
      level: required
      description_addendum: |
        The corresponding label column is `template_id`.
    rrid: optional
```

This also requires adding `rrid` to `objects.columns`.

cc @oesteban @CPernet

src/schema/rules/tabular_data/derivatives/common_derivatives.yaml
```diff
@@ -12,6 +12,38 @@ SegmentationLookup:
...
+  index_columns: [template_id]
+  additional_columns: allowed_if_defined
+
+Cohorts:
+  selectors:
+    - suffix == "cohorts"
+    - extension == ".tsv"
+  initial_columns:
+    - cohort_id
+    - description__entities
```
effigies, line 37: Again, I think this is overly prescriptive with column ordering.

```suggestion
```

src/derivatives/atlas.md
effigies: Right now `description` is REQUIRED by the table definition, so either that needs to stay in addition to `long_name` or it needs to be demoted from REQUIRED. `long_name` would need to be defined in `objects.columns` and added to the table definition.

src/schema/objects/suffixes.yaml
```diff
@@ -566,6 +566,15 @@ colFA:
     image color is determined by the principal eigenvector of the diffusion tensor,
     where red encodes the magnitude of the left-right component of that orientation,
     with green and blue similarly encoding the anterior-posterior and inferior-superior components respectively.
+cohorts:
+  value: cohorts
+  display_name: Cohort Entity Definitions
+  description: |
+    A TSV file describing labels found for the `cohort` entity within a template.
+
+    This file MUST be located within a `tpl-<label>/` directory.
```
effigies, line 575: This would need an explicit check to validate. Any TSV or JSON file can show up at the root without, even if it doesn't make much sense to do it.

src/schema/objects/suffixes.yaml
```diff
@@ -870,6 +879,15 @@ stim:
   display_name: Continuous recording
   description: |
     Continuous measures, such as parameters of a film or audio stimulus.
+templates:
+  value: templates
+  display_name: Template Entity Definitions
+  description: |
+    A TSV file describing labels found for the `tpl` entity in a Derivatives dataset.
```
effigies, line 886: This is a weird way of talking about it. I would say `participants.tsv` describes the participants, not the participant labels. Similarly with templates. Cohorts feels more reasonable to say labels.

effigies (commented):
Ongoing discussion. Cleared my approval until these are wrapped up.

PR #2294, opened 2026-01-07
[ENH] Make Pharmaceutical metadata terms less PET-specific

yarikoptic:
We are to introduce use of Pharma* fields into BEP032 (#1705) where it has no notion of tracers. I am moving this extra description to become PET specific, but also @bids-standard/raw-pet  people might want to double check if that is what was actually intended (or it could be administered separate from the tracer

### DIFF
--- a/src/schema/rules/sidecars/pet.yaml
+++ b/src/schema/rules/sidecars/pet.yaml
@@ -123,6 +123,7 @@ PETPharmaceuticals:
     PharmaceuticalName:
       level: recommended
       description_addendum: |
+        For pharmaceutical co-administered with tracer.
         Corresponds to [DICOM Tag 0018, 0034](https://dicomlookup.com/dicomtags/(0018,0034)) `Intervention Drug Name`.
     PharmaceuticalDoseAmount:
       level: recommended
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -2839,7 +2839,7 @@ PharmaceuticalName:
   name: PharmaceuticalName
   display_name: Pharmaceutical Name
   description: |
-    Name of pharmaceutical coadministered with tracer.
+    Name of pharmaceutical.
   type: string
 PhaseEncodingDirection:
   name: PhaseEncodingDirection


src/schema/rules/sidecars/pet.yaml
```diff
@@ -127,6 +127,7 @@ PETPharmaceuticals:
     PharmaceuticalDoseAmount:
       level: recommended
       description_addendum: |
+        For pharmaceutical co-administered with tracer.
         Corresponds to [DICOM Tag 0008, 0028](https://dicomlookup.com/dicomtags/(0008,0028)) `Intervention Drug Dose`.
```
effigies: DICOM lookup is dead, but we should still use https://dicom.innolitics.com/ciods/nuclear-medicine-image/nm-isotope/00180026/00180028

```suggestion
        Corresponds to [DICOM Tag 0018, 0028](https://dicomlookup.com/dicomtags/(0018,0028)) `Intervention Drug Dose`.
```

And 0008 needs to be updated to 0018 in PharmaceuticalName and PharmaceuticalDoseTime as well.

CPernet: I have 2 comments - questions
1) why would that be microscopy or PET specific? we do drug studies in fmri too, seems like it can be a generic addition
2) regarding PET, I am not sure of the wording 'co-administered with tracer' -- to clarify, while a drug can be tagged by a tracer, in PET/humans it unlikely as the concentration would be too low to detect, so usually it's more like take a drug that do X and, not necessarily the same time, inject a tracer that targets some receptors or vesicules -- co-admisnister suggest as the same time / simultaneously IMO

otherwise, DICOM wise make sense (but then fMRI too)

effigies: I'm okay moving the DICOM tag addendum into the term definitions.

The "co-administered" term comes from the original definition. I think the point is to indicate that the pharmaceutical dose is present at the same time as, but distinct from, the tracer. If that's correct, we could just make the term definition:

```yaml
PharmaceuticalDoseAmount:
  name: PharmaceuticalDoseAmount
  display_name: Pharmaceutical Dose Amount
  description: |
    Dose amount of administered pharmaceutical.
    Note that this is distinct from any radiotracer or contrast agent used in the scan protocol.
    Corresponds to [DICOM Tag 0018, 0028](https://dicomlookup.com/dicomtags/(0018,0028)) `Intervention Drug Dose`.
```

yarikoptic: I really like this suggestion if that does not change original intention but what do you mean by "original definition"? I see 

image

effigies: I mean that the word was taken from the current `objects.metadata` definition:

image

yarikoptic: gotcha, thanks! Unfortunately due to squashed history can't see potential reasoning in the 04c2d282cc58645c164b56781df204549f8d40eb which added it.  If there is no such "coadministration", I would vote for @effigies 's approach. Any :+1: :-1: from others?

mnoergaard: Sounds good to me!

mnoergaard (approved):

effigies:
Okay, let's merge this and I'll post a broader PR to move PET DICOM mappings into the `objects.metadata` terms.

PR #1021, opened 2022-03-03
[FIX] PET Spec; added known DICOM tags, fixed tag error, updated citation, clarified scale factor.

bendhouseart:
Small updates to the PET spec:

- Added references/correspondences between all known dicom tags relating to PET entities
- Clarified scale factor to, hopefully, avoid double scaling of images converted to nifti. 
- Fixed typo on injection start 
- Updated citation to include PET BIDS Paper

PR #2298, opened 2026-01-09
[ENH] Move PET DICOM correspondences into term definitions

effigies:
Follow-up to #2294.

This PR moves the `Corresponds to [DICOM Tag ...` section into `objects.metadata` terms for PET. This excludes terms like `InstitutionName`, which are recommended for many data types that are not converted from DICOM.

Minor cleanups here:

1) Remove a duplicate `BodyPart` entry. This was probably just an error in schematization.
2) For `InjectionStart` and `InjectionStop`, change "with respect to `"TimeZero"` in the default unit seconds." to "with respect to `"TimeZero"`, in seconds." BIDS does not support alternative units, so we can use simple language here.
3) Added "Note that this is distinct from any radiotracer or contrast agent used in the scan protocol." to `PharmaceuticalName`. While worth pointing out, it doesn't need to be pointed out multiple times, and it seems reasonable to include this in the term definition itself.
4) Some DICOM notices were removed from `rules.sidecars.pet` without adding to `objects.metadata` because they were already present in the `objects.metadata` term.


src/schema/objects/metadata.yaml
```diff
@@ -4130,6 +4155,8 @@ TracerName:
   display_name: Tracer Name
   description: |
     Name of the tracer compound used (for example, `"CIMBI-36"`)
+    Corresponds to DICOM Tags (0008,0105) `Mapping Resource` and
```
mnoergaard: This does not seem correct to me.

src/schema/objects/metadata.yaml
```diff
@@ -4142,6 +4169,7 @@ TracerRadionuclide:
   display_name: Tracer Radionuclide
   description: |
     Radioisotope labeling tracer (for example, `"C11"`).
+    Corresponds to DICOM Tags (0008,0104) `CodeValue` and (0008,0104) `CodeMeaning`.
```
mnoergaard: This does not seem correct to me.

src/schema/rules/sidecars/pet.yaml
```diff
@@ -61,13 +54,8 @@ PETRadioChemistry:
   fields:
     TracerName:
       level: required
-      description_addendum: |
-        Corresponds to DICOM Tags (0008,0105) `Mapping Resource` and
-        (0008,0122) `Mapping Resource Name`.
```
mnoergaard, line 73: I am not sure how these ended up here. I know tracer information can go into anonymous dicom tags, but this mapping resource seems confusing to me from a dicom lookup point of view.

mnoergaard (changes requested):
Many thanks for these changes, @effigies. It mostly looks good to me, except for the tracer dicom parts. But would like to loop @CPernet into this.

src/schema/objects/metadata.yaml
```diff
@@ -2885,7 +2893,9 @@ PharmaceuticalName:
   name: PharmaceuticalName
   display_name: Pharmaceutical Name
   description: |
-    Name of pharmaceutical.
+    Name of administered pharmaceutical.
+    Note that this is distinct from any radiotracer or contrast agent used in the scan protocol.
+    Corresponds to [DICOM Tag 0018, 0034](https://dicomlookup.com/dicomtags/(0018,0034)) `Intervention Drug Name`.
```
CPernet: This partly matches

CPernet (dismissed):

src/schema/objects/metadata.yaml
CPernet: the correct code is Tag (0054,0300) Radionuclide Code Sequence -- see https://dicom.nema.org/medical/Dicom/2024b/output/chtml/part03/sect_C.8.4.10.html

src/schema/objects/metadata.yaml
effigies: What should this say?

effigies (commented):
Tracer DICOM tags were added by @bendhouseart in #1021, and signed off by @mnoergaard, @sappelhoff and @tsalo. The question of whether to put them in terms or in the table as addenda was addressed:

> I'm not confident enough to say that these tags should belong in the `metadata.yml` as many of these terms are often more explicitly defined (or known....) by the experiment designer/performer or simply not present in the dicom headers. Adding them to the tables provides a bread crumb for a user to extract this info if they're not omnipotent concerning the experiment or the metadata.

_Originally posted by @bendhouseart in https://github.com/bids-standard/bids-specification/pull/1021#discussion_r821878990_

I'm okay removing them altogether, if that's the updated consensus, or putting them back (with comments) in `metadata.yaml`. From the perspective of someone reading the PET page, there is no difference; the difference is in the glossary or looking the term up in the schema directly.

Any updates to the text for clarity would be appreciated.

src/schema/rules/sidecars/pet.yaml
mnoergaard: I take this back @CPernet - apologies. After having looked at the dicom standard this does make sense e.g. https://dicom.innolitics.com/ciods/nuclear-medicine-image/nm-isotope/00540016/00540304/00080104

mnoergaard (approved):

src/schema/objects/metadata.yaml
```diff
@@ -4130,6 +4155,8 @@ TracerName:
   display_name: Tracer Name
   description: |
     Name of the tracer compound used (for example, `"CIMBI-36"`)
+    Corresponds to DICOM Tags (0008,0105) `Mapping Resource` and
+    (0008,0122) `Mapping Resource Name`.
```
effigies: @mnoergaard I think you meant [Radionuclide Code Sequence](https://dicom.innolitics.com/ciods/nuclear-medicine-image/nm-isotope/00540016/00540300) and not [Radiopharmaceutical Code Sequence](https://dicom.innolitics.com/ciods/nuclear-medicine-image/nm-isotope/00540016/00540304).

It looks like these are the fields that we want to point people to most prominently, rather than the generic terms inside the sequence. What do you think about language like this?

```suggestion
    Corresponds to [DICOM Tag 0054, 0300](https://dicomlookup.com/dicomtags/(0054,0300)) `Radionuclide Code Sequence`,
    extracted from the fields [0008, 0122](https://dicomlookup.com/dicomtags/(0008,0122)) `Mapping Resource Name`
     and/or [0008, 0105](https://dicomlookup.com/dicomtags/(0008,0105)) `Mapping Resource`.
```

bendhouseart:
Can't say that I'm not pleased to see [src/schema/rules/sidecars/pet.yaml](https://github.com/bids-standard/bids-specification/pull/2298/files#diff-a2b7f646b63bd9903b98925b2abd6ba01d83d95393534fec7f9a1cb039da8ef9) slimmed down nearly 100 lines.

Cheers 👍

src/schema/objects/metadata.yaml
mnoergaard: Think this should be 'This partly matches the [DICOM Tag 0018, 0034](https://dicomlookup.com/dicomtags/(0018,0034)) `Intervention Drug Name`.

effigies: Done.

mnoergaard: Many thanks, @effigies !

bendhouseart (approved):

PR #2304, opened 2026-01-15
[FIX] Correct `bval` and clarify `bvec` descriptions

kabilar:
Part of the current `bval` description seems to be duplicated from the `bvec` description.

cc @satra @ayendiki

### DIFF
--- a/src/schema/objects/extensions.yaml
+++ b/src/schema/objects/extensions.yaml
@@ -20,7 +20,7 @@ bval:
   value: .bval
   display_name: FSL-Format Gradient Amplitudes
   description: |
-    A space-delimited file containing gradient directions (b-vectors) of diffusion measurement.
+    A space-delimited file containing gradient amplitudes (b-values) of diffusion measurement.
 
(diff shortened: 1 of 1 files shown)


src/schema/objects/extensions.yaml
```diff
@@ -20,7 +20,7 @@ bval:
   value: .bval
   display_name: FSL-Format Gradient Amplitudes
   description: |
-    A space-delimited file containing gradient directions (b-vectors) of diffusion measurement.
+    A space-delimited file containing gradient amplitudes (b-values) of diffusion measurement.
```
arokem: Good catch! I think that technically this is not the gradient amplitude but a combination of different factors, including the amplitude of the gradient. So how about: 
```suggestion
    A space-delimited file containing gradient weightings (b-values) of diffusion measurement.
```

ayendiki:
I suggest "diffusion-encoding gradients" for bvecs and "diffusion-encoding weightings" for bvals.

arokem (approved):

effigies (approved):

PR #2307, opened 2026-01-15
[ENH]: BEP032 formal review

CodyCBakerPhD:
# BEP 032: Under review

This BEP is currently under review: https://github.com/bids-standard/bids-specification/discussions/2406

Please post significant discussion to that thread. Minor issues, such as typos or clarifications, may be posted as suggestions to this PR. These suggestions may be accepted or rejected without discussion, so do not make substantial proposals in this way.

---

# Original post

To all @bids-standard/maintainers 

In line with the penultimate steps outlined in the [Proposed BEP Review](https://bids.neuroimaging.io//extensions/process.html#proposed-bep-review) process, the BEP032 team would like to submit our pull request for formal review by the maintainers and community!

This PR adds a specification for microelectrode electrophysiology datasets based on the [BEP032 proposal](https://bids.neuroimaging.io/bep032).

We happily welcome all enhancements, fixes, suggestions, requests, and critiques by the broader community.

This work is the result of some enormous efforts over the last year by @ree-gupta @bendichter @yarikoptic @twachtler @robertoostenveld @lzehl @apdavison and many others whose full work history has been squashed as an extended description on this commit and can be seen in the [original draft PR](https://github.com/bids-standard/bids-specification/pull/1705), a copy of which will be kept on a branch in perpetuity on my personal fork (https://github.com/CodyCBakerPhD/bids-specification/tree/backup-bep032-work).

### Quick links

- [**HTML preview of this BEP**](https://bids-specification--2307.org.readthedocs.build/en/2307/modality-specific-files/microelectrode-electrophysiology.html)
- Working [overleaf manuscript project](https://www.overleaf.com/6614877716cftfyhffrfvp#25764f) with source and git history  [**here**](https://github.com/INCF/bids-microephys-manuscript)
- [**Example dataset**](https://github.com/bids-standard/bids-examples/pull/491)
- [**Another example dataset**](https://github.com/bids-standard/bids-examples/pull/527)


robertoostenveld:
@bids-maintenance can we get this announced to the community through the mailing list for formal review, and get a few experts explicitly assigned/invited as potential reviewers?

julia-pfarr:
Hi, we have two scheduled BEP community review periods per year, the next one is in April. We put you on the list for this review period! Before that, some maintainers would like to meet with all of you and go through the BEP, to make sure everything is ready for the community review. @effigies and @rwblair, can you post a when2meet with your availabilities here? Thank you!

effigies:
Another approach could be to add a required boolean field like `"PharmacologicalIntervention"`. If `False`, then fields are OPTIONAL; if `True`, they are REQUIRED/RECOMMENDED (as appropriate).

effigies:
We can push off if you prefer, but if we're aiming for a review period in April, there's only so much time to push. I'll leave it to you. Otherwise, it looks like Monday at 2pm EDT is going to be the best time. LMK.

ree-gupta:
> Hi, we have two scheduled BEP community review periods per year, the next one is in April. We put you on the list for this review period

Hi BIDS-maintainers :) do we have an update in which review period would this proposal fall in? Are there any more open questions or changes pending on our end?
Thanks!

Tagging to notify: @julia-pfarr @effigies

effigies:
Was just going to send out emails. We're planning to open review on Monday.

effigies:
This PR is under review: https://github.com/bids-standard/bids-specification/discussions/2406

Please post significant discussion to that thread. Minor issues, such as typos or clarifications, may be posted as suggestions to this PR. These suggestions may be accepted or rejected without discussion, so do not make substantial proposals in this way.

src/schema/objects/metadata.yaml
```diff
@@ -2384,6 +2384,55 @@ MEGREFChannelCount:
...
+  description: |
+    Units of the coordinates of `"MicroephysCoordinateSystem"`.
+  type: string
+  # TODO: Clarify if units need to be used or enums, if units, define pixels
+  enum:
+    - m
+    - mm
+    - cm
+    - um
+    - pixels
```
effigies, line 2443: I don't understand this TODO. Apologies if I missed a discussion.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -0,0 +1,915 @@
...
+In such a case, this file is REQUIRED.
+This allows benefiting from the capability of the supported data formats (NIX and NWB) to store multiple
+recordings in a single file, which can be convenient when these recordings share numerous characteristics
+(for example, for subsequent recordings obtained on a single cell in intracellular electrophysiology).
+In such case, the information about these recordings should be stored in columns added in the
+`*_events.tsv` file, which are listed now.
+
+Optional column names in `events.tsv` to support multiple recordings in a single data file:
+
+<!-- TODO: Macro for events -->
```
effigies, line 715: Missing table here. Looks like the following, rendered:

Image

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -0,0 +1,915 @@
...
+
+## Examples of Real Datasets
+
+Several real-world datasets have been formatted using this specification and can be used for practical guidance when curating a new dataset.
+<!-- TODO: Update with current real datasets. A current version of these datasets [can be found on GIN](https://gin.g-node.org/NeuralEnsemble/BEP032-examples) .
+
+For a complete dataset including all data samples the extracellular microelectrode dataset published in [Brochier (2018)](https://doi.org/10.1038/sdata.2018.55) has been reorganized according to the current version of this BEP, using the NIX data format.
+The up-to-date version of the dataset [can be found on GIN](https://gin.g-node.org/sprenger/multielectrode_grasp/src/bep_animalephys) .
+
+We will also publish another dataset using the NWB data format in the near future, and a dataset acquired -->
```
effigies, line 985: This TODO appears undone. If it is not ready, I would remove this section and open an issue to add it later.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -0,0 +1,915 @@
...
+
+For the first subject only a single sample (a cell for patch-clamp terminology) was extracted (sample-cell001), on which three different protocol recordings were performed: two runs of current injection to characterize intrinsic properties, and one run of synaptic stimulation. The `scans.tsv` file stores information such as the starting recording times. The detailed information on the recording channel (such as the recording mode used) is stored in the `channels.tsv` which, in this case, is common to all available recordings. The probes and electrodes files provide information on the pipette and solutions used for the recordings and are also shared across data files.
```
effigies, line 852: The `sample-<label>` entity is first introduced here and only described in the photos section. As a reader it came out of nowhere.

This appears to have been taken from microscopy. I would reuse text from there to the extent that it's applicable.

Is `samples.tsv` required if the `sample-<label>` entity is used? Is it permitted?

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -0,0 +1,915 @@
...
+{{ MACROS___make_sidecar_table("microephys.microephysProcessing") }}
+
+### Additional Procedure Information
+
+Furthermore, additional information can be stored about the recording procedure.
+We RECOMMEND to use a dedicated `Procedure` node with the following keys:
+
+-   `Pharmaceuticals`
+-   `Sample`
+-   `Supplementary`
```
effigies, line 117: This appears to be vestigial, as there is no `Procedure` metadata defined in the schema, and `PharmaceuticalName` is shown directly in the top-level metadata. (I'm assuming "node" means a field in a JSON structure.)

src/schema/rules/files/raw/channels.yaml
```diff
@@ -110,6 +134,25 @@ electrodes__meg:
...
+    - .tsv
+  datatypes:
+    - ecephys
+    - icephys
+  entities:
+    subject: required
+    session: optional
+    sample: optional
+    acquisition: optional
+    processing: optional
```
effigies: This is not used in any other files, and is not obviously related to electrodes specifically.

```suggestion
```

src/schema/rules/files/raw/channels.yaml
```diff
@@ -110,6 +134,25 @@ electrodes__meg:
...
     processing: optional
 
+# microephys electrodes exclude task/run entities
+electrodes__microephys:
+  $ref: rules.files.raw.channels.electrodes
+  suffixes:
+    - electrodes
+  extensions:
+    - .json
+    - .tsv
```
effigies: These are implied by the `$ref: rules.files.raw.channels.electrodes`.

```suggestion
```

src/schema/rules/files/raw/channels.yaml
```diff
@@ -11,6 +11,22 @@ channels:
...
     - nirs
 
+# microephys has an additional 'sample' entity but excludes task/run
+channels__microephys:
+  $ref: rules.files.raw.channels.channels
+  suffixes:
+    - channels
+  extensions:
+    - .json
+    - .tsv
```
effigies:
```suggestion
```

effigies:
```suggestion
```

effigies (commented):
Still working on a final review. These are my comments from reading the text. I haven't yet gone through the schema closely.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Hey BEP032 team, let's discuss @effigies 's point about `sample-<label>` coming out of nowhere. He's right that it's under-explained, and digging in I found the requirement level is inconsistent between the prose and the examples, plus a schema gap. Here's what I'd suggest, but curious what you all think:

**Introduce the entity up front.** Right now its only real definition is buried in the Photos section, framed as just for histological photos, but the intracellular examples use it on every file (`sample-cell001`, etc.). I'd propose adding a definition near the top in "Terminology: Modality and Datatypes," adapted from the microscopy text.

**Split the requirement level by datatype** rather than making it uniformly optional or copying microscopy's blanket REQUIRED:

- `ecephys` (in-vivo extracellular): OPTIONAL, and typically absent, since there's no tissue sample. Requiring it here would be wrong.
- `icephys` (patch): REQUIRED. A patched cell effectively *is* a sample, and the datatype inherently deals in discrete cells/slices. This matches what our intracellular example already does (every file is `sample-cellNNN`) and mirrors how `micr` treats it.

In the schema this is just leaving `sample: optional` for ecephys and changing it to `sample: required` in the icephys block of `microephys.yaml`.

One sub-question: `micr` does *not* exempt the single-sample case (one sample per subject still gets `sample-01`). I'd suggest we follow that, since a single-sample exemption causes filename churn if a second cell is added later and forces tools to special-case it. So: required for icephys even when there's only one sample.

**`samples.tsv`.** Chris also asked whether it's required when the entity is used. It currently isn't enforced at all for microephys, since `SamplesTSVMissing` only triggers on `"micr" in dataset.modalities`. I'd suggest requiring `samples.tsv` whenever the `sample-` entity is present (which, with the split above, means always for icephys and only-if-used for ecephys), and adding a check keyed on entity-presence rather than modality.

**`samples.tsv` columns we inherit.** These are modality-agnostic, so microephys gets them for free: `sample_id` (required), `participant_id` (required), `sample_type` (required), `pathology` (recommended), `derived_from` (recommended). They all fit microephys fine, so I don't think we need to define anything new. Two things worth a quick look:

- `sample_type` is a **required** enum, and its full value list is broader than just cell types: `tissue`, `whole organisms`, `organoid`, `technical sample`, plus `cell line` / `in vitro differentiated cells` / `primary cell` / `cell-free sample` / `cloning host`. So an acute slice maps cleanly to `tissue` and organoid recordings to `organoid` — no missing values. The only thing to settle is a **granularity convention**: for patch clamp, is the "sample" the slice or the individual cell? If it's the cell, `tissue` is still the closest `sample_type`. We could also register both and use `derived_from` to link `sample-cell001` back to `sample-slice01`. I'd suggest we pick one convention and document it.
- `sample_id` labels are restricted to `^sample-[0-9a-zA-Z+]+$` (no underscores/hyphens), so `sample-cell001` is fine but keep that in mind for examples.

**Cleanup.** Trim the Photos-section mention down to a back-reference to the new definition.

If the per-datatype split (optional ecephys / required icephys) and the entity-presence trigger for `samples.tsv` sound right, I'm happy to put up the edits. The main thing I'd want to settle first is the sample granularity convention (cell vs. slice) for patch data.

src/schema/rules/files/raw/channels.yaml
bendichter: Agreed, let's remove it. You're right that it doesn't belong here as written: `processing` was only on the microephys electrodes block and on none of the microephys data, channels, probes, or coordsystem files, so a `proc-` electrodes file would have no `proc-`-tagged recording to pair with. It looks like it got copied over from the `electrodes__meg` template (which does carry `processing: optional`, for MaxFilter-style variants), rather than being a deliberate choice.

I'm fine to drop it for now. If we later decide microephys needs to distinguish device-side processing variants (e.g. on-device filtering/referencing on Open Ephys / Intan / SpikeGLX), we can add `proc` properly to the data and channels files at that point rather than leaving it stranded on electrodes. Accepting the suggestion.

src/modality-specific-files/microelectrode-electrophysiology.md
yarikoptic: nice dig!  I do not know much about icephys, but is it really that there would be multiple samples per each subject?  ATM we do mandate `sub-` entity already, so if it is actually typical to have multiple samples - people would just use it if/where needed to disambiguate within subject, so we do not even need to make it RECOMMENDED.

And yes, I would recommend having `samples.tsv` if `_sample-` entity is used.

NB/backref -- this will add one more "consistent" case for `samples.tsv` to be inconsistent' in having dual-index for a single plural entity filename as mentioned in #2283

src/modality-specific-files/microelectrode-electrophysiology.md
yarikoptic: I think it came from good old times when also @JuliaSprenger was involved in this BEP, so that https://gin.g-node.org/sprenger/multielectrode_grasp/src/bep_animalephys was created. But overall it was left as TODO since it is a chicken-egg issue. We can't have this until we finalize BEP since otherwise those datasets might not be valid ones. I wonder also if we could/should just point to https://github.com/bids-dandisets/ which would have many of such examples.  I would not bother with https://gin.g-node.org/NeuralEnsemble/BEP032-examples though since those are largely replaced with bids-examples. But I would leave in smth like

```suggestion
Several real-world datasets have been formatted using this specification and can be used for practical guidance when curating a new dataset.
<!-- TODO: Upon finalization of the BEP032, update with current real datasets if they are updated to the spec and pass bids-validator.

For a complete dataset including all data samples the extracellular microelectrode dataset published in [Brochier (2018)](https://doi.org/10.1038/sdata.2018.55) has been reorganized according to the current version of this BEP, using the NIX data format and [hosted on GIN](https://gin.g-node.org/sprenger/multielectrode_grasp/src/bep_animalephys) .
Also numerous datasets from the [DANDI Archive](https://dandiarchive.org) have been converted to BIDS using [nwb2bids]() and shared [on GitHub](https://github.com/bids-dandisets/).
>
```

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Yes, multiple samples per subject is normal for icephys. A single animal commonly yields several slices, and several cells are patched per slice, so >1 sample per subject is the rule, not the exception.

On the requirement level: `sub-` alone can't disambiguate samples *within* a subject, and as you note people would reach for `sample-` to do that anyway. So the real choice is whether `sample-` is present inconsistently (only when >1) or always. The problem with "only when needed" is churn: collect one cell → `sub-X_...`, collect a second later → you now have to rename the first file to add `sample-`, and tools have to special-case the single-sample form. `micr` avoids this by requiring `sample-` even for a single sample, and I'd lean toward icephys following it for the same reason.

Agreed on requiring `samples.tsv` whenever `sample-` is used.

That said, the required-vs-optional call (and the related cell-vs-slice granularity convention) has real trade-offs in both directions, so I think it warrants input from the larger BEP032 team before we settle it.

src/schema/meta/associations.yaml
```diff
@@ -101,6 +103,14 @@ electrodes:
       - space
   inherit: true
 
+probes:
+  selectors:
+    - intersects([suffix], ['ecephys', 'icephys'])
+    - extension != '.json'
+  target:
+    suffix: probes
+    extension: .tsv
```
effigies, line 112: Since `probes.tsv` is a TSV file, it presumptively follows the inheritance principle, and I haven't seen anything that suggests otherwise. Let me know if I missed something.

```suggestion
probes:
  selectors:
    - intersects([suffix], ['ecephys', 'icephys'])
    - extension != '.json'
  target:
    suffix: probes
    extension: .tsv
  inherit: true
```

src/schema/objects/columns.yaml
```diff
@@ -49,6 +49,45 @@ age:
...
   }
+AP_angle:
+  name: AP_angle
+  display_name: AP angle
+  description: |
+    Anterior-Posterior rotation angle measured as rotation from the vertical axis in the sagittal plane.
+    0° represents vertical along DV axis. Positive values indicate anterior rotation.
+  unit: degree
+  type: number
+  minimum: -180
```
effigies: Probably want a half-open interval, otherwise you have two representations of the same value.

```suggestion
  exclusiveMinimum: -180
```

src/schema/objects/columns.yaml
```diff
@@ -345,6 +495,36 @@ material:
...
+  type: string
+ML_angle:
+  name: ML_angle
+  display_name: ML angle
+  description: |
+    Medial-Lateral rotation angle measured as rotation from the vertical axis in the coronal plane.
+    0° represents vertical along DV axis. Positive values indicate rightward/clockwise rotation (as seen from behind).
+  unit: degree
+  type: number
+  minimum: -180
```
effigies:
```suggestion
  exclusiveMinimum: -180
```

src/schema/objects/columns.yaml
```diff
@@ -533,6 +761,16 @@ response_time:
...
   unit: s
+rotation_angle:
+  name: rotation_angle
+  display_name: Rotation angle
+  description: |
+    Rotation angle around the probe axis. 0° when probe features align with the coronal plane.
+    Positive rotation is clockwise when viewed from above.
+  unit: degree
+  type: number
+  minimum: -180
```
effigies:
```suggestion
  exclusiveMinimum: -180
```

src/schema/objects/columns.yaml
```diff
@@ -49,6 +49,45 @@ age:
...
+  description: |
+    An identifier of the associated brain region based on the Uberon ontology
+    for anatomical structures in animals, for example "UBERON:0010415"
+  type: string
+associated_brain_region_quality_type:
+  name: associated_brain_region_quality_type
+  display_name: Associated brain region quality type
+  description: |
+    The method used to identify the associated brain region (estimated|proof)
+    depending on anatomical pictures proofing the location or indirect estimation of the location.
```
effigies: This `(estimated|proof)` is causing truncation:

Image

I don't understand this sentence in any case. Could it be rewritten not to use `(x|y)` notation and edited for clarity?

src/schema/objects/columns.yaml
```diff
@@ -345,6 +495,36 @@ material:
...
+  unit: degree
+  type: number
+  minimum: -180
+  maximum: 180
+ML__probes:
+  name: ML
+  display_name: Probe ML position
+  description: |
+    Probe position along the Medial-Lateral axis.
+    Positive values are to the right (as seen from behind).
```
effigies:
```suggestion
    Positive values are to the right, as seen from behind (neurological convention).
```

src/schema/objects/columns.yaml
```diff
@@ -94,6 +141,30 @@ component:
...
+    Point of the probe that is described by the probe coordinates and
+    on which the yaw, pitch, and roll rotations are applied.
+  type: string
+anatomical_reference_point:
+  name: anatomical_reference_point
+  display_name: Anatomical reference point
+  description: |
+    Anatomical reference point for stereotaxic coordinates (for example, `Bregma`, `Lambda`).
+    If not specified, `Bregma` is assumed for rodents.
+    MUST be defined for species other than rodents.
```
effigies: This "MUST" comes with no corresponding check, meaning it will not be enforced. What would it look like to enforce this? I assume `species` should be defined in `participants.tsv`, but how should rodents vs non-rodents be identified?

Would it be easier just to make this required?

src/schema/objects/columns.yaml
```diff
@@ -506,6 +727,13 @@ reference__ieeg:
     Specification of the reference (for example, `mastoid`, `ElectrodeName01`, `intracranial`, `CAR`, `other`, `n/a`).
     If the channel is not an electrode channel (for example, a microphone channel) use `n/a`.
   type: string
+reference_atlas:
+  name: reference_atlas
+  display_name: Reference atlas
+  description: |
+    Name of reference atlas used for associated brain region identification,
+    preferably an [ebrains supported atlas](https://www.ebrains.eu/brain-atlases/reference-atlases/#services).
```
effigies: IIUC, "EBRAINS" is the typical spelling.

```suggestion
    preferably an [EBRAINS-supported atlas](https://www.ebrains.eu/brain-atlases/reference-atlases/#services).
```

This URL no longer seems valid, by the way. It redirects to https://ebrains.eu/data-tools-services/brain-atlases#services. What should it point to?

src/schema/objects/columns.yaml
```diff
@@ -49,6 +49,45 @@ age:
...
+    A textual indication on the location of the probe,
+    preferably species-independent terms as obtained, for example from Uberon.
+  type: string
+associated_brain_region_id:
+  name: associated_brain_region_id
+  display_name: Associated brain region identifier
+  description: |
+    An identifier of the associated brain region based on the Uberon ontology
+    for anatomical structures in animals, for example "UBERON:0010415"
+  type: string
```
effigies: One of these seems to require Uberon, one uses Uberon as an example. If I were to read this as a curator, I would assume Uberon is required, and `associated_brain_region` is just vaguely worded.

What is the intent here?

src/schema/objects/columns.yaml
```diff
@@ -71,6 +110,14 @@ channel:
...
     of a channel name.
   type: string
+channel_label:
+  name: channel_label
+  display_name: Channel label
+  description: |
+    Human readable identifier. Use this name to specify the
+    content of signals not generated by electrodes. For example,
+    `DAQ internal synchronization signals`, `behavioral signals`, `behavioral cues`.
+  type: string
```
effigies: The examples seem to match more with what I would think of as a description than an identifier. How does this differ from description?

src/schema/objects/columns.yaml
```diff
@@ -637,6 +889,19 @@ software_filters:
...
+  description: |
+    The types of software filters applied to this channel.
+    The `Levels` for this column SHOULD be defined in the accompanying
+    `*_channels.json` file, mapping each filter type key to its description.
+    Use `n/a` if no software filters were applied to this channel.
+  anyOf:
+    - type: string
+    - type: string
+      enum:
+        - n/a
```
effigies: Redundant. and `n/a` is always permissible.

```suggestion
  type: string
```

src/schema/objects/columns.yaml
```diff
@@ -928,7 +1236,13 @@ z:
...
   description: |
     Recorded position along the z-axis.
-  type: number
+
+    For 2D electrode localizations, this SHOULD be a column of `n/a` values.
+  anyOf:
+    - type: number
+    - type: string
+      enum:
+        - n/a
```
effigies: n/a is always permissible

```suggestion
  type: number
```

src/schema/objects/columns.yaml
```diff
@@ -948,8 +1262,12 @@ z__optodes:
...
   description: |
     Recorded position along the z-axis.
-    `"n/a"` if not available.
-  type: number
+    "`n/a`" if not available.
+  anyOf:
+    - type: number
+    - type: string
+      enum:
+        - n/a
```
effigies:
```suggestion
    `n/a` if not available.
  type: number
```

effigies (commented):
Got through columns.yaml. Working my way in lexical order.

src/schema/meta/associations.yaml
yarikoptic: with this in mind we found another missing annotation and necessity/desire to adjust metaschema to make it "explicit": competing human-vs-claude+human:
- #2478 
- #2479

bendichter:
I was inspired by @effigies's questions about the brain region columns to look more closely at how anatomical location is represented in this BEP, and I think there is a structural issue worth discussing as a group.

## Current Situation

Anatomical location is annotated in two separate places. In `probes.tsv` there are four recommended columns: `associated_brain_region`, a textual label (for example, `primary visual cortex`), `associated_brain_region_id`, an ontology identifier (for example, `UBERON:0010415`), `associated_brain_region_quality_type`, indicating whether the region was established by anatomical evidence or estimated, and `reference_atlas`, naming the atlas used. In `electrodes.tsv` there is a single recommended column, `location`, described as "An indication on the location of the electrode (for example, `cortical layer 3`, `CA1`)", with no accompanying identifier column and no ontology mapping.

## Why This Is a Problem

There are three issues, and they compound.

First, a probe frequently spans multiple regions. A single shank can pass through cortex, hippocampus, and thalamus in one insertion, so one brain region per probe cannot describe where the recording actually happened.

Second, localization really happens at the electrode level. Once you have histology or an atlas registration, what you recover is a region assignment per contact. The probe-level annotation is better understood as the intended target of the insertion than as the result of localization.

Third, `electrodes.tsv` has no ontology support. So the one table where per-contact localization can be recorded is also the place where the annotation is least machine readable, while the ontology columns sit on the coarser of the two tables.

The third point is the one I would emphasize, because it is not only a matter of granularity. If the ontology term lives on the electrode, we can support different ontologies for different localizations, since different rows can carry identifiers from whichever atlas was actually used for that assignment. That flexibility is not really available at the probe level, where a single value has to stand for the whole insertion.

## Proposed Changes

For `probes.tsv`, redefine the existing region columns as the *target* brain region rather than the observed one. This can be done in the column descriptions without renaming anything, so it is a low-cost change: `associated_brain_region` and `associated_brain_region_id` become the region the probe was aimed at, which is genuinely useful information and is what a probe-level annotation can honestly express.

For `electrodes.tsv`, rename `location` to `brain_region` and add a companion `brain_region_id` column holding an ontology identifier, allowing UBERON as well as species-specific atlases such as the Allen Mouse Brain Atlas. The naming would then parallel the probes table, and the distinction between the two would be target versus observed rather than two unrelated concepts.

Once the columns are split this way, `associated_brain_region_quality_type` should move to `electrodes.tsv` as well. A quality or evidence qualifier on a stated target is not meaningful, since the target is declared in advance by the experimenter and there is nothing about it to have been estimated or verified. The qualifier only makes sense attached to an observed localization, so it should follow `brain_region` to the electrodes table rather than stay behind on the probes table.

I anticipate the objection that the evidence method is usually constant across a probe's contacts, since there is typically one histological reconstruction per track, so the value would repeat down the whole column. I do not find that persuasive. Constant columns are routine in BIDS tabular files, and the method does vary in real data: contacts near a region boundary versus contacts deep inside a structure, contacts whose slice was damaged or lost, and assignments that were directly observed versus interpolated between two observed points. Per-contact seems like the safer granularity to me. If the group would rather avoid the repetition, the alternative worth considering is a sidecar JSON field, not leaving the column on `probes.tsv`.

I would also take the opportunity to fix two things about that column, which is the subject of @effigies's open thread on it. The name suggests a quality grade, but the concept is really how the region assignment was established, so something like `brain_region_localization_method` or `brain_region_evidence` describes it more directly. And the permitted values are currently written into the prose as `(estimated|proof)` rather than declared as an enum, which is what is causing the rendered table to truncate. Making it a real enum would fix the rendering and the ambiguity together.

I think this also resolves the question @effigies raised about whether Uberon is required or merely an example, since we would need to state the intent explicitly for the new electrode columns in any case.

A few things I do not have a strong opinion on and would like input on:

- Which ontologies do we allow, and do we express that as prefixed identifiers (`UBERON:0010415`, `MBA:385`) so a single column can carry more than one source?
- Should `reference_atlas` also become an `electrodes.tsv` column, or is a prefix on the identifier sufficient to say where a term came from?
- What should the enum values for the localization method be? I read the current "proof" as meaning histologically verified, so something like `histology`, `imaging`, `estimated`, and possibly `interpolated` seems closer to the real distinctions, but this is a domain question and I would rather the group set the list.

bendichter:
Following up on @effigies's question about `anatomical_reference_point`, I looked into what it would take to enforce the current language, and I do not think the conditional as written can be implemented.

## Current Situation

The column is `optional` in `microephysProbes`, but its description says "If not specified, `Bregma` is assumed for rodents" and "MUST be defined for species other than rodents". So an optional column carries a conditional MUST, and both the requirement and the default depend on knowing the subject's species.

## Why the Species Condition Cannot Be Enforced

The validator context does not expose species at all: `context.dataset.subjects` provides only `sub_dirs` and `participant_id`, so a rule evaluating `probes.tsv` cannot read the `species` column of `participants.tsv` without first extending the context. `species` is itself only RECOMMENDED and is assumed to be `homo sapiens` when absent, so the condition rests on a frequently missing column whose fallback would make the reference point required by default. And deciding rodent versus non-rodent needs taxonomic reasoning rather than string comparison, since Rodentia contains thousands of species and the expression language has no notion of taxonomy. None of this seems worth solving for one column.

## Proposed Change

@effigies asked whether it would be easier to just make the column required. That is close, but `microephysProbes` covers `icephys` as well, and a patch clamp recording in an acute slice has no stereotaxic reference point, so a flat requirement would force those datasets to write `n/a`.

Instead, I propose requiring `anatomical_reference_point` when stereotaxic coordinates are present, meaning when any of `AP`, `ML`, or `DV` are given in the same file. That is the condition that actually matters, since the reference point exists to anchor those coordinates, and it handles the slice case correctly. Unlike the species condition it is expressible today: checks can reference `columns.<name>` of the file under validation, as `EMGCoordSysMatch` in `rules/checks/emg.yaml` already does with `columns.coordinate_system`.

I would also drop "If not specified, `Bregma` is assumed for rodents". It has the same species-detection problem, and if coordinates are present the column is required, so no default is needed.

bendichter:
@effigies asked how `channel_label` differs from `description`, and looking at the table I do not think it does.

## Current Situation

`microephysChannels` already carries columns that cover this ground. `name` is required and serves as the index, defined as "Label of the channel". `type` is required and takes an enum. `description` is optional, defined as "Brief free-text description of the channel, or other information of interest". `channel_label` sits alongside them as optional, defined as "Human readable identifier" and intended to specify the content of signals not generated by electrodes.

The examples in its description are the clearest sign of the overlap: `DAQ internal synchronization signals`, `behavioral signals`, and `behavioral cues`. Each of those is already a `type` value. This BEP itself added `SYNC` and `BEH` to the channel type enum, and `SYSCLOCK`, `TRIG`, `STIM`, and `MISC` were there before it. So the categorical content of those examples belongs in `type`, and whatever is left over is free text, which is `description`.

Calling it an identifier makes the overlap worse rather than better, since the identifier for a channel is `name`, which is required and indexed.

## Proposed Change

Remove `channel_label`. A channel not generated by an electrode sets `electrode_name` to `n/a`, takes an appropriate `type` such as `SYNC` or `BEH`, and uses `description` for anything the type does not capture. That path is already fully specified and needs no additional column.

The one reading that might justify a dedicated column is an alternative identifier that does not fit `name`. That does not hold up either. `name` is an unconstrained string with no pattern or character restrictions, so there is nothing to escape from, and the spec already recommends carrying alternative identifier sets as custom columns just below the example tables. That approach also handles datasets with more than one alternative scheme, which a single `channel_label` column cannot.

src/modality-specific-files/microelectrode-electrophysiology.md
lzehl: Hey. Here my feedback.

**Introduce the entity up front.** I agree with your suggestion.

**Split the requirement level by datatype** It's good to make it OPTIONAL for ecephys (although "since there's no tissue sample" is not correct, just not the typical case). It should indeed be required for icephys.

**requiring samples.tsv whenever sample- is used** I would agree to that (I think this is the case already no?). From a metadata management point of view this is very annoying that this is not the case for subjects. Even if there is only one subject (or sample) you want to have structured information on that subject (or sample).

From my point of view all those cases can be true and should be possible to model in BIDS: one sample per subject, multiple samples per subject, one sample from multiple subjects, sample with no subject.

**sample_type** the enum list is indeed insufficient and should be extended/revised. We encountered this problem already for the atlas/microscopy people that started to use BIDS. However I'm not sure if they ever raised that issue officially. As inspiration, here is the list we use in [openMINDS](https://openminds.docs.om-i.org/en/latest/instance_libraries/terminologies/tissueSampleType.html) (which was usage driven so most likely also not complete).

src/modality-specific-files/microelectrode-electrophysiology.md
lzehl: Hey BEP032 team, @effigies is right that this it is very unclear what "node" refers to and I honestly don't remember. Prior to this we say `"The *_ecephys.json or *_icephys.json file can be used to store any microephys-specific metadata for the dataset. All setup-related metadata should be stored in a dedicated node of the JSON file called Setup."`. For me this reads as if we intended the `*_ecephys.json` or `*_icephys.json` to be nested jsons, that provide information of setup and procedure under highlevel keys`"setup"` and `"procedure"`, but I think this is not true. Except for the filter metadata `*_ecephys.json` or `*_icephys.json` are supposed to be flat. 

My suggestion: The node reference is confusing I would just delete it. Setup, Procedure (with Pharmaceuticals, Sample, Supplementary) are just subheaders to structure all possible keys of the `*_ecephys.json` and `*_icephys.json` sidecars.

This brings up though another general issue I see: for me the sample metadata in the `*_ecephys.json` and `*_icephys.json` sidecars (and actually in the microscopy sidecar JSON files) compared to what is/should be provided in the samples.tsv and samples.json is a bit unclear. I would have assumed that all in samples.tsv and samples.json refers to the unprocessed sample (so all that is true always for sample X, such as sample type, slice thickness, extraction protocol, etc) while the sample keys in the `*_ecephys.json`, `*_icephys.json` and microscopy JSON sidecars cover additional processes applied sequentially (such as staining, pharmaceuticals, etc) or sample related acquisition variables (such as deformation, etc) or information on tissue structures when there is no sample (such as examined body part in a typical ecephys case,or intravital microscopy settings). This seems to be not the logic though. Sorry to raise this now... I guess we cannot do much about this now, but maybe we can deliver good examples of what to define where and when for samples?

src/schema/objects/columns.yaml
lzehl: @bendichter I think we just expect an enum here "estimated" or "proof". I would suggest to reformulate to:

```
    The quality type of the method used to identify the associated brain region.
    MUST be 'estimated' or 'proof'.
```

In openMINDS we call this [criteria quality types](https://openminds.docs.om-i.org/en/latest/instance_libraries/terminologies/criteriaQualityType.html) and our terms here are 'asserted' (claim/subjective) vs 'processive' (repeatable process/objective).

src/schema/objects/columns.yaml
lzehl: I think `associated_brain_region` and the corresponding `associated_brain_region_id` should not be limited to UBERON (which is only a generic cross species ontology). There are species specific anatomical atlas brain regions and their corresponding IDs out there which should be used if fitting (I don't understand why we say cross-species is preferred... not really a good advise). UBERON should just be an example. So the text needs revision. My sugestion:

```
associated_brain_region:
  name: associated_brain_region
  display_name: Associated brain region
  description: |
    A textual indication on the location of the probe,
    preferably ontology-based, for example from UBERON.
  type: string
associated_brain_region_id:
  name: associated_brain_region_id
  display_name: Associated brain region identifier
  description: |
    The corresponding persistent identifier of the associated brain region, 
    for example "UBERON:0010415".
  type: string
```

src/schema/objects/columns.yaml
lzehl: @effigies human readable identifiers or maybe use better the term label here (?) differ from descriptions in multiple ways: shorter (best actually one word, max two words) and not necessarily explanatory. Descriptions should be fully formulated sentences actually explaining the thing or the content. 

I think channel label rather corresponds to channel type. So I would remove channel label.

src/modality-specific-files/microelectrode-electrophysiology.md
SylvainTakerkart: > **sample_type** the enum list is indeed insufficient and should be extended/revised. We encountered this problem already for the atlas/microscopy people that started to use BIDS. However I'm not sure if they ever raised that issue officially. As inspiration, here is the list we use in [openMINDS](https://openminds.docs.om-i.org/en/latest/instance_libraries/terminologies/tissueSampleType.html) (which was usage driven so most likely also not complete).

Hi, I jump in a little bit (yes, I'm still around and following ;) ). Wanted to attend the meeting yesterday but finally couldn't...

I've had the same experience with the definition and use of sample for some microscopy data recently... With two issues:
- the enum list (I concurr with @lzehl , it should be extended/revised)
- the granularity question, which is in fact, if I'm not mistaken, not formally settled in the microscopy specs, leading to potential different interpretations of the micr specs (a case I encountered)

So I agree and support what you're doing here, in particular the addition of the text to define the granularity!

src/modality-specific-files/microelectrode-electrophysiology.md
SylvainTakerkart: > I think it came from good old times when also @JuliaSprenger was involved in this BEP, so that https://gin.g-node.org/sprenger/multielectrode_grasp/src/bep_animalephys was created. But overall it was left as TODO since it is a chicken-egg issue. We can't have this until we finalize BEP since otherwise those datasets might not be valid ones. I wonder also if we could/should just point to https://github.com/bids-dandisets/ which would have many of such examples. I would not bother with https://gin.g-node.org/NeuralEnsemble/BEP032-examples though since those are largely replaced with bids-examples. But I would leave in smth like

I confirm this historical point! The use of GIN to host such example datasets is probably obsolete now and any reference to it, at least the ones created by Julia Sprenger, should  be removed... The TODO list is also outdated (same timeframe / investigator)...

SylvainTakerkart: from what I know, the data available at https://gin.g-node.org/sprenger/multielectrode_grasp/src/bep_animalephys is old (last change 4 years ago), and therefore not compliant with the current state of the specs... I think @ree-gupta has produced a version of the same dataset (from Brochier 2018) that should be compliant with the current specs... we could simply replace the old link with the one that's up-to-date...

src/schema/objects/metadata.yaml
bendichter: Addressed in #2514, which drops the TODO and makes the field follow the same pattern as `iEEGCoordinateUnits`.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Addressed in #2504 (merged). The section no longer promises a column table; it now describes one `events.tsv` row per recording with `onset`, `duration`, `stream_id`, and a HED `Recording` tag.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Addressed in #2504 (merged). The GIN links are gone and the section now points to the two bids-examples datasets and to bids-dandisets.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: The `sample-<label>` entity is now introduced in the terminology section and `sample_type` was extended in #2484 (merged). Requiring the entity for icephys and requiring `samples.tsv` whenever it is used is in #2480, which is waiting on Sylvain's review.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Addressed in #2504 (merged). The references to `Setup` and `Procedure` nodes are removed and the prose now says the sidecar is a flat object whose tables group keys by topic only.

src/schema/rules/files/raw/channels.yaml
bendichter: Applied in 2aa6e305; the microephys rules now only override `datatypes` and `entities`.

src/schema/objects/columns.yaml
bendichter: Addressed in #2481 (merged). The quality type column was dropped along with the other `associated_brain_region*` columns; anatomical location now uses a single `anatomical_location` column with `Levels` in the sidecar.

src/schema/objects/columns.yaml
bendichter: Addressed in #2482 (merged). `anatomical_reference_point` is now required for every species whenever `AP`, `ML`, or `DV` is present, with a validator check.

src/schema/objects/columns.yaml
bendichter: Addressed in #2481 (merged). The `associated_brain_region*` columns are replaced by `anatomical_location`, and ontology terms are given through `Levels` and `TermURL` in the sidecar rather than tied to Uberon.

src/schema/objects/columns.yaml
bendichter: Addressed in #2483 (merged), which removed the `channel_label` column.

bendichter:
@effigies, since your July pass we have merged a set of follow-up PRs into the review branch that address all of the threads you opened, and I have replied on each thread with a pointer to the PR that handled it.

- #2481: anatomical location moved from the probe to the electrode level, with a single `anatomical_location` column and ontology terms given through `Levels` in the sidecar. This also removed the `(estimated|proof)` quality-type column and the Uberon wording.
- #2482: `anatomical_reference_point` is required whenever stereotaxic coordinates are given, with a validator check.
- #2483: `channel_label` removed.
- #2484: `sample_type` extended and an anatomical location column added to `samples.tsv`.
- #2480: the `sample-<label>` entity is introduced up front, required for `icephys`, and `samples.tsv` is enforced whenever the entity is used.
- #2504 and #2513: the vestigial `Setup` and `Procedure` node text is gone, the multiple-recordings section describes `events.tsv` rows with `stream_identifier` and a HED `Recording` tag instead of promising a column table, and the real-dataset links point to bids-examples.
- #2514: the remaining schema TODO on `MicroephysCoordinateUnits` now follows the `iEEGCoordinateUnits` pattern.
- The channels.yaml suggestions were applied. CI is green.

On the review discussion, I followed up with Saskia by email and she is fine with the proposal moving forward. I left a note on the thread.

You mentioned you had gotten through `columns.yaml`. Would you be able to finish the pass over the rest of the schema? The working group meets on the 23rd to sign off and Cody presents at the BEP meeting on October 1, so having your remaining comments before then would let us fold them in.

xgui3783:
Not too sure if this is the right place to sought some clarification with regard to the proposed specification (if not, please point to the right forum)

In the specification[1][2], it appears that the column *must* be in the following order: `name`, `probe_name`, `x`, `y`, `z` ... (others may be in any other order)

But in the examples given[3][4], the tsv column is in the following order `name`, `probe_name`, `hemisphere`, `x`, `y`, `z` ... (other columns)

Is this because of some spec which I misinterpreted? Or hemisphere column should come after x y z?

[1] https://bids-specification--2307.org.readthedocs.build/en/2307/modality-specific-files/microelectrode-electrophysiology.html#electrodes-description-_electrodestsv
[2] https://github.com/bids-standard/bids-specification/pull/2307/changes#diff-80ed9627be097bf1b19f2cba255d3dc345f267f8530bdc5fa6a09d54233aee79R46-R51
[3] https://bids-specification--2307.org.readthedocs.build/en/2307/modality-specific-files/microelectrode-electrophysiology.html#example-_electrodestsv
[4] https://github.com/bids-standard/bids-specification/pull/2307/changes#diff-207c85ee9805015015facf6485c0caffc3eb189fef11b89533e0902e7f0fb9a8R426-R445

lzehl:
@xgui3783 thanks for noticing this. I actually do not know if the columns must come in a fixed order. @effigies? or @bendichter if you know?

bendichter:
@xgui3783 you read it correctly, and this is the right place to ask. The schema lists `name`, `probe_name`, `x`, `y`, `z` as the `initial_columns` of `*_electrodes.tsv`, which means they must be the first five columns in that order, and the validator will flag a file where `hemisphere` comes before `x`. Columns after those five can appear in any order. The examples in the text were wrong; the example datasets in bids-standard/bids-examples#527 and bids-standard/bids-examples#491 already put `hemisphere` after `z`. I opened #2533 to fix both examples in the text.

@lzehl the fixed order is a schema-level property (`initial_columns`). iEEG has the same rule for `name`, `x`, `y`, `z`, `size`, and the `*_probes.tsv` examples already follow it.

bendichter:
@effigies, following up on my comment from the 11th. Would you be able to finish your pass over the schema (metadata.yaml, the rules, and the sidecars)? Cody presents at the BEP meeting on October 1, and we would like to have your remaining comments addressed before then.

effigies:
I'll try to finish this this week.

src/schema/objects/columns.yaml
```diff
@@ -49,6 +49,66 @@ age:
...
+  maximum: 180
+AP__probes:
+  name: AP
+  display_name: Probe AP position
+  description: |
+    Probe position along the Anterior-Posterior axis.
+    Positive values are anterior to the reference point.
+  type: number
+  unit: mm
+anatomical_location:
```
effigies:
```suggestion
anatomical_location__electrodes:
```

src/schema/rules/tabular_data/microephys.yaml
```diff
@@ -0,0 +1,147 @@
...
+        Units are specified by `MicroephysCoordinateUnits` in the corresponding `*_coordsystem.json` file.
+        For 2D electrode localizations (for example, when the coordinate system is
+        `Pixels`), this SHOULD be a column of `n/a` values.
+    hemisphere: recommended
+    impedance: optional
+    shank_id: optional
+    size__microephys: optional
+    electrode_shape: optional
+    material: optional
+    anatomical_location: recommended
```
effigies:
```suggestion
    anatomical_location__electrodes: recommended
```

PR #2309, opened 2026-01-16
[ENH] Recommend reporting of BigDelta and SmallDelta for DWI images

### LINKED ISSUE #2302: Combined DWI scheme with different acquisition parameters
Hi BIDS team, we have a situation that is different than the [Multipart (split) DWI scheme](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#multipart-split-dwi-schemes), whereby we have acquired a single concatenated DWI run but parts of the DWI run have different acquisition parameters (i.e. echo time [TE], time between the leading edges of two successive diffusion-encoding gradient pulses [Δ], duration of a diffusion-encoding gradient pulse [δ]).  It is ideal for us to keep this data as a single volume since downstream analysis is performed on the entire DWI run.

Right now we are thinking of saving sidecar files similar to the `.bvals` and `.bvecs` with the TE, Δ, and δ as `.techo`, `.tbigdelta`, `.tsmalldelta`, respectively.  What do you think?  Happy to submit a pull request to the spec if this approach is agreed upon.  Thanks.

cc @satra @tinggong @ayendiki

kabilar:
Closes https://github.com/bids-standard/bids-specification/issues/2302

cc @satra @ayendiki @tinggong


src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -912,6 +912,8 @@ JSON example:
 
 ```JSON
 {
+  "DiffGradientDuration": 11.0,
+  "DiffGradientSeparation": 15.2,
```
effigies: We don't generally use abbreviations. I would say either "DiffusionGradientDuration/Separation" or "GradientDuration/Separation". I would probably prefer the latter if the terms are unambiguous within a diffusion context and wouldn't refer to something else in another context BIDS is likely to touch on.

kabilar: Thanks @effigies.  Just avoid any potential future confusion, I would prefer "DiffusionGradientDuration/Separation".

@ayendiki @tinggong What do you think?

arokem: I don't necessarily like the word "separation", which to me implies the interval between the end of one pulse and the beginning of the other pulse. Other words to potentially use here are "interval", or "offset". I do wonder if given their prevalence in the literature, `BigDelta` and `SmallDelta` wouldn't be clearer. Also, I do believe that this is a redefinition of something that could already be captured by `MixingTime`: https://bids-specification.readthedocs.io/en/stable/glossary.html#mixingtime-metadata?

arokem: I see this as related to #2311

mharms: I'm not familiar with how diffusion sensitization plays out in a "double spin echo" diffusion weighted sequence, but are `MixingTime` and `BigDelta` really the same thing?

arokem: The timing parameters are defined quite differently for "double" or "twice refocused" spin echo sequences, see https://onlinelibrary.wiley.com/doi/10.1002/mrm.10308. IIUC (see Fig 1 therein), there is no "BigDelta" in these sequences, and the intensity of the gradients is somehow integrated across all those small deltas, so I think that's to be handled separately, maybe in the advanced dMRI BEP.

mharms: So then, the existing `MixingTime` is not appropriate here, right?

kabilar: I am not too familiar with `MixingTime` but this [paper](https://pmc.ncbi.nlm.nih.gov/articles/PMC4231024/) declares the mixing time, gradient separation, and gradient duration as separate values:

`...mixing time TM=173.0 ms, gradient separation Δ=185.8 ms and a gradient duration of δ=5.4ms.`

src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -922,6 +922,8 @@ JSON example:
 
 ```JSON
 {
+  "BigDelta": 15.2,
+  "SmallDelta": 11.0,
```
mharms:
```suggestion
  "BigDelta": 0.040,
  "SmallDelta": 0.015,
```
Shouldn't the example have something reflective of seconds?

src/schema/objects/metadata.yaml
```diff
@@ -350,6 +350,15 @@ BasedOn:
       items:
         type: string
         format: participant_relative
+BigDelta:
+  name: BigDelta
+  display_name: Diffusion Gradient Separation
```
mharms: I'm not sure where/how these "display_name" fields show up, but I wonder if it should just be "Diffusion BigDelta"?

src/modality-specific-files/magnetic-resonance-imaging-data.md
kabilar: Good catch. I was moving too fast.  I have updated it.

src/schema/objects/metadata.yaml
kabilar: One place it shows up is in the [Glossary](https://bids-specification.readthedocs.io/en/stable/glossary.html) under the `Name` field.  How about the following?
```suggestion
  display_name: Diffusion Big Delta (Δ)
```

src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -922,8 +922,8 @@ JSON example:
 
 ```JSON
 {
-  "BigDelta": 15.2,
-  "SmallDelta": 11.0,
+  "BigDelta": 0.0152,
+  "SmallDelta": 0.011,
```
mharms: Are these "example" values from a low b-value acquisition?  While they are just an example, I'm trying to figure out why they are so small.

src/schema/objects/metadata.yaml
mharms: Fine by me, although the entry for SmallDelta should be changed accordingly as well.

src/modality-specific-files/magnetic-resonance-imaging-data.md
kabilar: Hi @mharms, these values are from an upcoming publication for macaque dMRI protocols on a Bruker 4.7T system.  Similar values are used for human dMRI protocols on the Connectome 2.0 3T system.

src/schema/objects/metadata.yaml
```diff
@@ -3889,6 +3898,15 @@ SliceTiming:
     type: number
     minimum: 0
     unit: s
+SmallDelta:
+  name: SmallDelta
+  display_name: Diffusion Gradient Duration
```
kabilar:
```suggestion
  display_name: Diffusion Small Delta (δ)
```

src/modality-specific-files/magnetic-resonance-imaging-data.md
mharms: Ok, although as a more typical "example", values of 0.040 and 0.015 would be more in the ballpark of what others would likely be using.

src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -922,6 +922,8 @@ JSON example:
 
 ```JSON
 {
+  "BigDelta": 0.0152,
+  "SmallDelta": 0.011,
```
kabilar:
```suggestion
  "BigDelta": 0.040,
  "SmallDelta": 0.015,
```

src/modality-specific-files/magnetic-resonance-imaging-data.md
kabilar: Sounds good, updated, thanks.

effigies (dismissed):
LGTM:

image

mharms:
> Are there DICOM tags that should be used? This will be important for converters like dcm2niix.

If this info exists in Siemens DICOMs, I don't know where it is.  The fact that this is not easy to know/extract is probably why it hasn't been added previously.

src/schema/rules/sidecars/dwi.yaml
```diff
@@ -16,5 +16,7 @@ MRIDiffusionOtherMetadata:
     - suffix == "dwi"
     - match(extension, "^\.nii(\.gz)?$")
   fields:
+    BigDelta: recommended
+    SmallDelta: recommended
```
mharms: What are the other options beyond "recommended"?  i.e., I have a hard time giving `BigDelta/SmallDelta` the same level of emphasis as something like `PhaseEncodingDirection` and `TotalReadoutTime`.

mharms: If everything simply becomes "recommended", the distinction becomes meaningless.

tsalo: The levels are OPTIONAL (no warning if missing), RECOMMENDED (warning, but no error, if missing), and REQUIRED (error if missing).

mharms: To me, Big/SmallDelta should be considered OPTIONAL then (the level below RECOMMENDED)

eauerbach:
> > > On siemens you can find the deltas:
> > > ```
> > > sSpecPara.dSpecLipidSupprBandwidth     =     XX.XX        // delta
> > > sSpecPara.dSpecLipidSupprDeltaPos     =     XX.XX        // DELTA
> > > sSpecPara.dSpecWaterSupprBandwidth     =     0.XXX    // spoiler b-value
> > > ```
> > 
> > 
> > Are you sure? Do you have a reference for that? It seems odd to me that variables with names referring to `LipidSuppr` and `WaterSuppr` would be where Siemens is storing information related to diffusion gradient timing.
> 
> > I got this reference from Mark Elliott at Penn. I'm not sure if he's on github
> 
> Well, color me surprised. Those do indeed contain the delta/DELTA that I'd expect in two Siemens sessions that I checked -- one on XA30, and the other on VE11C. Both were using the CMRR MB diff sequence though. If someone has convenient access to diffusion data collected with the Siemens product sequence, it would be good to check those as well. e.g., I wonder if @eauerbach might be using those variables to store information in a way that is not part of the Siemens product diffusion sequence (esp. considering that those 3 values -- delta, DELTA, and spoiler b-value -- are all values that you can check via a tool-tip in his sequence).

Yes, this is a non-standard feature in my sequences; Siemens does not provide this information in the product (unless it has changed in the newest versions).

src/schema/rules/sidecars/dwi.yaml
kabilar: Hi @mharms, I noted these down here as RECOMMENDED since some downstream packages require these values for processing (e.g. https://github.com/Tinggong/Microstructure.jl).  I opted to not make it REQUIRED since these values are sometimes not exposed to the user.

mattcieslak: These values are in the hbcd 1.0 data release sidecars as LargeDelta and SmallDelta, but I bet we could get that changed. It looks like a couple ad-hoc sidecar conventions have popped up across software packages in the absence of official names

mharms: These do not rise to the level of RECOMMENDED for me personally -- i.e., I don't think warnings need to generated by compliance checkers if those values are absent.

kabilar:
```suggestion
    BigDelta: optional
    SmallDelta: optional
```

kabilar: Sounds good, I have updated them to optional, thanks.

src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -900,6 +900,35 @@ A guide for using macros can be found at
...
+
+The `EchoTime`, `SmallDelta`, and `BigDelta` fields MAY be
+saved in the sidecar JSON file on any level of the directory structure and thus define those values for
+all sessions and/or subjects in one place (see
+[the inheritance principle](../common-principles.md#the-inheritance-principle)).
+
+As an exception to the [common principles](../common-principles.md#definitions)
+that parameters are constant across runs, the `EchoTime`, `SmallDelta`, and `BigDelta` values MAY change across the DWI run.
+
 ### Other RECOMMENDED metadata
```
kabilar: Updating the heading since the table in this section would list RECOMMENDED and OPTIONAL fields.

```suggestion
### Other metadata fields
```

src/modality-specific-files/magnetic-resonance-imaging-data.md
```diff
@@ -900,7 +900,36 @@ A guide for using macros can be found at
    }
 ) }}
 
-### Other RECOMMENDED metadata
+### Combined DWI schemes
+
+Some DWI schemes are acquired by varying parameters including the
+[`EchoTime (TE)`](../glossary.html#objects.metadata.EchoTime),
```
effigies:
```suggestion
[`EchoTime (TE)`](../glossary.mdl#objects.metadata.EchoTime),
```

src/schema/objects/metadata.yaml
```diff
@@ -350,6 +350,22 @@ BasedOn:
       items:
         type: string
         format: participant_relative
+BigDelta:
+  name: BigDelta
+  display_name: Diffusion Big Delta (Δ)
+  description: |
+    In the context of a diffusion-weighted pulsed gradient spin echo sequence,
+    corresponds to the time between the leading edges of two successive
+    diffusion-encoding gradient pulses (Δ), specified in seconds.
```
effigies, line 362:
```suggestion
    diffusion-encoding gradient pulses (Δ), specified in seconds.
    The value may be an array, if Δ varies by volume, in which case the number
    of values MUST correspond to the number of volumes in the data file.
```

src/schema/objects/metadata.yaml
```diff
@@ -3889,6 +3905,22 @@ SliceTiming:
     type: number
     minimum: 0
     unit: s
+SmallDelta:
+  name: SmallDelta
+  display_name: Diffusion Small Delta (δ)
+  description: |
+    In the context of a diffusion-weighted pulsed gradient spin echo sequence,
+    corresponds to the duration of a diffusion-encoding gradient pulse (δ),
+    specified in seconds.
```
effigies, line 3974:
```suggestion
    specified in seconds.
    The value may be an array, if δ varies by volume, in which case the number
    of values MUST correspond to the number of volumes in the data file.
```

effigies (commented):
Please also add checks, similar to

https://github.com/bids-standard/bids-specification/blob/1fd10bb21b9c3b261c18d8d674286dd956b497e5/src/schema/rules/checks/mri.yaml#L134-L147

src/modality-specific-files/magnetic-resonance-imaging-data.md
kabilar: I think you meant to add the `.md` extension so I will commit that.

```suggestion
[`EchoTime (TE)`](../glossary.md#objects.metadata.EchoTime),
```

src/schema/rules/checks/mri.yaml
```diff
@@ -145,3 +145,31 @@ RepetitionTimePreparationNotConsistent:
...
+
+SmallDeltaNotConsistent:
+  issue:
+    code: SMALLDELTA_NOT_CONSISTENT
+    message: |
+      The number of values for 'SmallDelta' for this file does
+      not match the 4th dimension of the NIfTI header.
+    level: error
+  selectors:
+    - modality == "mri"
```
effigies: I believe this only applies to DWI data? Would it ever appear for an associated `sbref` or `epi` (reverse-PE fieldmap)?

```suggestion
    - suffix == "dwi"
    - type(sidecar.BigDelta) == "array"
    - type(nifti_header) != "null"
  checks:
    - length(sidecar.BigDelta) == nifti_header.dim[4]

SmallDeltaNotConsistent:
  issue:
    code: SMALLDELTA_NOT_CONSISTENT
    message: |
      The number of values for 'SmallDelta' for this file does
      not match the 4th dimension of the NIfTI header.
    level: error
  selectors:
    - suffix == "dwi"
```

kabilar: Thanks.  I am not sure if it applies to other modalities, but I have committed your suggestion.

kabilar:
Hi @effigies et al., I just wanted to check in to see if there was any additional input needed from me for this pull request.  No rush on my end if it is just in the queue for being reviewed.  Thank you.

effigies (approved):
Makes sense from a BIDS/schema perspective. At least one positive review from a discussant and no objections would be appreciated.

My understanding of the state of the discussion is:

1) There is no standard way to encode these data in DICOM, so this PR is giving a name to these fields that will require manual (or automatic with specialized tooling) entry for the time being. The main concern here would be that we don't make life difficult if/when this does become standardized by DICOM. cc @mharms @neurolabusc for this
2) The field names follow common conventions in the DWI community, and the predominant (but not universal) BIDS convention of using seconds instead of ms.

arokem (approved):

kabilar:
Hi @effigies, thanks for the review and summary.  Since this has now been approved by @arokem, could we merge this?  Thanks.

src/schema/objects/metadata.yaml
```diff
@@ -939,6 +957,38 @@ DewarPosition:
...
+    diffusion-encoding gradient pulses (Δ), specified in milliseconds.
+  anyOf:
+    - type: number
+      unit: ms
+      exclusiveMinimum: 0
+    - type: array
+      items:
+        type: number
+        unit: ms
+        exclusiveMinimum: 0
```
kabilar: It looks like I forgot to remove the previous names for these variables.
```suggestion
```

mharms: I thought the "unit" was supposed to be in seconds...?

kabilar:
Preview: https://bids-specification--2309.org.readthedocs.build/en/2309/modality-specific-files/magnetic-resonance-imaging-data.html#combined-dwi-schemes

kabilar:
The failing CI appears to be unrelated.

src/schema/objects/metadata.yaml
kabilar: Hi @mharms, sorry for the confusion.  I have removed this text altogether.  It was an outdated, duplicate.

kabilar: Please see the latest file changes.

PR #2310, opened 2026-01-16
[Schema][Fix] add checks for multiple frame consistency issues 

bendhouseart:
This was discovered by @mnoergaard during the processing of a dataset recently uploaded to openneuro -> https://github.com/nipreps/petprep/issues/223

(Not a validator issue 👍 )

This adds checks for the following cases:

- Number of frames differ between nifti, FrameDuration, and FrameTimesStart
- Number of frames mismatch between FrameDuration and nifti
- Number of frames mismatch between FramesTimeStart and nifti
- Number of frames mismatch between FramesTimesStart and FrameDuration when nifti header is null (`--ignoreNiftiHeader` flag use case)

### DIFF
--- a/src/schema/rules/checks/pet.yaml
+++ b/src/schema/rules/checks/pet.yaml
@@ -5,20 +5,15 @@ PETFrameConsistency:
   issue:
     code: PET_FRAME_CONSISTENCY
     message: |
-      The number of frames in this scan does not match the number of frames
-      (as defined by FrameDuration and FrameTimesStart) in the
+      The number of frames as defined by FrameDuration and FrameTimesStart do not match in the
       associated '.json' file.
     level: error
   selectors:
     - suffix == 'pet'
-    - type(nifti_header) != "null"
     - sidecar.FrameDuration
     - sidecar.FrameTimesStart
   checks:
-    - |
-      length(sidecar.FrameDuration) == nifti_header.dim[4] &&
-      length(sidecar.FrameTimesStart) == nifti_header.dim[4] &&
-      length(sidecar.FrameDuration) == length(sidecar.FrameTimesStart)
+    - length(sidecar.FrameDuration) == length(sidecar.FrameTimesStart)
 
 PETFrameConsistencyFrameDuration:
   issue:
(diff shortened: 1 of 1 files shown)


src/schema/rules/checks/pet.yaml
```diff
@@ -4,6 +4,25 @@
...
+    - sidecar.FrameTimesStart
+  checks:
+    - |
+      length(sidecar.FrameDuration) == nifti_header.dim[4] &&
+      length(sidecar.FrameTimesStart) == nifti_header.dim[4] &&
+      length(sidecar.FrameDuration) == length(sidecar.FrameTimesStart)
+
+PETFrameConsistencyFrameDuration:
+  issue:
+    code: PET_FRAME_CONSISTENCY_FRAME_DURATION
```
effigies, line 20: This just duplicates the other three. And we should keep the name and code the same, as people may look up the rule in the schema or filter the issue in the validator output.

```suggestion
```

src/schema/rules/checks/pet.yaml
```diff
@@ -4,6 +4,25 @@
...
+  selectors:
+    - suffix == 'pet'
+    - type(nifti_header) != "null"
+    - sidecar.FrameDuration
+    - sidecar.FrameTimesStart
+  checks:
+    - |
+      length(sidecar.FrameDuration) == nifti_header.dim[4] &&
+      length(sidecar.FrameTimesStart) == nifti_header.dim[4] &&
+      length(sidecar.FrameDuration) == length(sidecar.FrameTimesStart)
```
bendhouseart:
```suggestion
```

I think we should keep the name/code specific though.

src/schema/rules/checks/pet.yaml
bendhouseart: Adding a suggestion to a suggestion is tricky, see my reply below.

src/schema/rules/checks/pet.yaml
effigies: I'll summon @rwblair to call it on consistency. If we are going to change the name and code of an existing rule, then we definitely can't do this as a bug-fix. The schema minor version needs to be bumped for that.

mnoergaard:
I would define this as a bug fix, as the BEP assumes that FrameDuration and FrameTimesStart have the same number of entries. We added the validation that FrameDuration should be checked against the number of frames in the imaging data to verify consistency, but didn't think of also adding this check to FrameTimeStart. Again, assuming they should have similar size.

bendhouseart:
Woohoo

PR #2311, opened 2026-01-16
[FIX] Add `MixingTime` to the DWI section

kabilar:
I am not too familiar with the `MixingTime` field, but the [MixingTime](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.metadata.MixingTime) description says that certain diffusion-weighted sequences should include this field.  The `MixingTime` is currently listed in the [Spatial Encoding](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#in-and-out-of-plane-spatial-encoding) section, and here I have added it to the DWI section so that it is easier to discover.

cc @satra @ayendiki

### DIFF
--- a/src/modality-specific-files/magnetic-resonance-imaging-data.md
+++ b/src/modality-specific-files/magnetic-resonance-imaging-data.md
@@ -900,7 +900,7 @@ A guide for using macros can be found at
    }
 ) }}
 
-### Other RECOMMENDED metadata
+### Other RECOMMENDED and OPTIONAL metadata
 
 The `PhaseEncodingDirection` and `TotalReadoutTime` metadata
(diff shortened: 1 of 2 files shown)


effigies:
I'm not sure I agree with this. The first half of the page consists of fields that apply to many MRI data types; duplicating them for each applicable type of data when there is no modification (such as an increased requirement level, or updated description) does not serve a clear purpose. If anything, it might imply that fields that are not explicitly listed in the DWI section do not apply to DWI.

effigies:
What would be the conditions under which it's recommended? Can you tell from other metadata, or would it be for all DWI?

kabilar:
Closing since this is not critical and is not my area of expertise.

PR #2315, opened 2026-01-20
[FIX] Update and render schema definition of dataset_description.json

effigies:
The metadata table displayed in https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/dataset-description.html and the one in the schema that is validated got out of sync.

This updates the schema table to match the one on the spec, and renders the table from the schema.

* [Updated table](https://bids-specification--2315.org.readthedocs.build/en/2315/modality-agnostic-files/dataset-description.html#dataset_descriptionjson)
* [Original](https://bids-specification.readthedocs.io/en/latest/modality-agnostic-files/dataset-description.html#dataset_descriptionjson)

### DIFF
--- a/src/modality-agnostic-files/dataset-description.md
+++ b/src/modality-agnostic-files/dataset-description.md
@@ -24,26 +24,7 @@ The definitions of these fields can be found in
 and a guide for using macros can be found at
  https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
 -->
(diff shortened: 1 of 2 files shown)


yarikoptic (approved):
Duplication is evil -- we should get rid from manually duplicated specifications in .md files -- all should be in the schema... then we should hunt some AI to ensure any free text is also in compliance!

PR #2322, opened 2026-01-23
fix(schema): Check SpatialReference for space, template entities

effigies:
The rule changes in #1714 were a bit off. If the `space-<label>` entity is set, then the file is in that space, regardless of whether the file is `tpl-<label>-*` or `sub-<label>-*`, so check for a spatial reference if that value is unknown.

If space is not absent, *then* we check for whether `tpl-<label>` is defined and unknown.


bendhouseart (approved):
Much clearer logic; tested w/ bep023 petprep examples no more false errors 👍

PR #2324, opened 2026-01-26
[ENH] Render phenotype sidecar table from schema

effigies:
Follow up to #2315. Should be pretty uncontroversial, I think.

Updated table:

image


bendhouseart (approved):
lgtm, no sematic change waiving second review.

Issue #2327, opened 2026-01-27
Inconsistency/duplication: SoftwareVersion and SoftwareVersions


yarikoptic:
Came up while reviewing paper for BEP020 (atten @bids-standard/bep020), claude picked up inconsistency of having SoftwareVersion whenever BIDS has SoftwareVersions.  Well - the surprise is that we have both! :-/

```
❯ git grep '\<SoftwareVersions\>'
src/CHANGES.md:-   Added `SoftwareVersions`.
src/modality-specific-files/magnetoencephalography.md:   "SoftwareVersions": "Acq 5.4.2-linux-20070507",
src/pregh-changes.md:-   Added `SoftwareVersions`.
src/schema/objects/metadata.yaml:SoftwareVersions:
src/schema/objects/metadata.yaml:  name: SoftwareVersions
src/schema/rules/sidecars/continuous.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/eeg.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/ieeg.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/meg.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/micr.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/motion.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/mri.yaml:    SoftwareVersions:
src/schema/rules/sidecars/mrs.yaml:    SoftwareVersions: recommended
src/schema/rules/sidecars/nirs.yaml:    SoftwareVersions: recommended
```

```
❯ git grep '\<SoftwareVersion\>'
src/modality-agnostic-files/events.md:        "SoftwareVersion": "3.0.14",
src/schema/objects/metadata.yaml:SoftwareVersion:
src/schema/objects/metadata.yaml:  name: SoftwareVersion
src/schema/objects/metadata.yaml:    - SoftwareVersion
src/schema/objects/metadata.yaml:    SoftwareVersion:
src/schema/objects/metadata.yaml:      $ref: objects.metadata.SoftwareVersion
```

with `master` gaining 1 more hit

```
src/modality-specific-files/physiological-recordings.md:    "SoftwareVersion": "SREB1.10.1630 WIN32 LID:F2AE011 Mod:2017.04.21 15:19 CEST",
```

with plural form used more and singular, even as of current master (hence having bep020 merged) having singular form described as

```yaml
[...]

PR #2328, opened 2026-01-27
fix: should be plural SoftwareVersions for _physio.md (eye tracking)

yarikoptic:
As already prescribed

```
❯ grep -B7 SoftwareVer ./src/schema/rules/sidecars/continuous.yaml
PhysioHardware:
  selectors:
    - suffix == "physio"
  fields:
    DeviceSerialNumber: optional
    Manufacturer: optional
    ManufacturersModelName: optional
    SoftwareVersions: optional
```

More on this point:

-  https://github.com/bids-standard/bids-specification/issues/2327

attn @bids-standard/bep020

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,4 @@
 {
-  "name": "bids-specification-master",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {


package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "specification",
+  "name": "bids-specification-master",
```
effigies:
```suggestion
```

package-lock.json
```diff
@@ -1,5 +1,4 @@
 {
-  "name": "specification",
```
yarikoptic, line 2: dang, sorry -- missed that change introduced while building it :-/

effigies: Resolved future issues of this sort in 675ef95d.

PR #2333, opened 2026-01-31
[ENH]: BEP046, diffusion tractography

arokem:
Starting to port over information from https://docs.google.com/document/d/1ubDQ2RhgjnfGqoeukzEkPV9YEHhfYMERrj7-3b0c2HI/edit?tab=t.0

> [!Note]
>
> **We meet regularly to discuss this BEP at the IST Standardization Unit**
> Please reach out to @arokem (arokem@gmail.com) if you would like to join these meetings, which usually occur on the last Friday of each month at 7:30 AM PT.
>

### DIFF
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -1,0 +1,31 @@
+# Diffusion derivatives
+
+## Diffusion-based tractography
+
+Tractography based on diffusion-weighted MRI data is stored as streamlines in
+the TRX file format:
+
+```Text
+<pipeline_name>/
+    sub-<label>/
+        [ses-<label>/]
+            <modality/
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.trx
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.json
+```
+
+Where "tract” is the anatomical/structural entity that is being imaged, and
+“track” uses as its value one of the items in a controlled vocabulary.
+
+For example:
+
+```Text
+
+AFQ/
+    sub–01/
+        dwi/
+            sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.trx
+            sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.json
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.trx
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.json
+```
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -19,9 +19,6 @@ Where `tract` is the anatomical/structural entity that is being imaged, and
 
 For example:
 
-```Text
-
-AFQ/
     sub–01/
         dwi/
             sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.trx
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -9,7 +9,7 @@ the [TRX file format](https://tee-ar-ex.github.io):
 <pipeline_name>/
     sub-<label>/
         [ses-<label>/]
-            <modality/
+            <modality>/
                 <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.trx
                 <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.json
 ```
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -26,6 +26,6 @@ AFQ/
         dwi/
             sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.trx
             sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.json
-            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.trx
-            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.json
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.trx
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.json
 ```
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -3,7 +3,7 @@
 ## Diffusion-based tractography
 
 Tractography based on diffusion-weighted MRI data is stored as streamlines in
-the TRX file format:
+the [TRX file format](https://tee-ar-ex.github.io):
 
 ```Text
 <pipeline_name>/
--- a/src/derivatives/diffusion-derivatives.md
+++ b/src/derivatives/diffusion-derivatives.md
@@ -15,7 +15,11 @@ the [TRX file format](https://tee-ar-ex.github.io):
 ```
 
 Where `tract` is the anatomical/structural entity that is being imaged, and
-`track` uses as its value one of the items in a controlled vocabulary.
+`track` is fully specified in the sidecar as described below.
+
+For the most common case, a tractogram will be generated from some DWI data, so
+<modality> will typically be “dwi”. In any case, the <modality> allows to
+unequivocally identify the data type from which the tractogram originated.
 
 For example:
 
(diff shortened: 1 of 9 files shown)


ayendiki:
Thanks for the update, Ariel. It sounds like support for developers is growing. What about support for users? If one had a tractogram in .trx format and wanted to interact with it and do a manual dissection for instance, what would they use?

________________________________

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,31 @@
...
+
+```Text
+
+AFQ/
+    sub–01/
+        dwi/
+            sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.trx
+            sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.json
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.trx
+            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-left_track-eudx_tractogram.json
```
kabilar: Hi Ariel, it looks like the label values for `hemi` have to be `L` or `R`.

https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#hemi

```suggestion
            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.trx
            sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.json
```

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,31 @@
...
+<pipeline_name>/
+    sub-<label>/
+        [ses-<label>/]
+            <modality/
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.trx
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.json
+```
+
+Where `tract` is the anatomical/structural entity that is being imaged, and
+`track` uses as its value one of the items in a controlled vocabulary.
```
kabilar, line 18: Could you please provide a list of the controlled vocabulary (or a link)?  Thanks.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,31 @@
...
+
+For example:
+
+```Text
+
+AFQ/
```
kabilar: Removing extra line for consistency with above.

```suggestion
```Text
AFQ/
```

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,31 @@
...
+## Diffusion-based tractography
+
+Tractography based on diffusion-weighted MRI data is stored as streamlines in
+the TRX file format:
+
+```Text
+<pipeline_name>/
+    sub-<label>/
+        [ses-<label>/]
+            <modality/
```
kabilar:
```suggestion
            <modality>/
```

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,31 @@
...
+<pipeline_name>/
+    sub-<label>/
+        [ses-<label>/]
+            <modality>/
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.trx
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.json
+```
+
+Where `tract` is the anatomical/structural entity that is being imaged, and
+`track` uses as its value one of the items in a controlled vocabulary.
```
kabilar, line 18:
```suggestion
Where [`tract-<label>`](../appendices/entities.md#tract) is the anatomical/structural entity that is being imaged, and
[`track-<label>`](../appendices/entities.md#track) uses as its value one of the items in a controlled vocabulary.
```

src/schema/objects/entities.yaml
```diff
@@ -526,3 +534,14 @@ tracksys:
     may be longer and more human readable.
   type: string
   format: label
```
kabilar, line 536: Since there is already an entity that uses `track` as part of `tracksys` to describe motion tracking, should we use a more distinct entity?  @ayendiki had suggested `trace`, which I also like.

src/derivatives/diffusion-derivatives.md
arokem: It's table 10 in the Google doc: https://docs.google.com/document/d/1ubDQ2RhgjnfGqoeukzEkPV9YEHhfYMERrj7-3b0c2HI/edit?tab=t.0

src/schema/objects/entities.yaml
arokem: Wouldn't that be confused with the trace of a tensor? 

Would "tracking" make this less ambiguous? Or "tracto"? We could also go more abstract with "method" or "algo", but that might also ultimately conflict with other corners of the spec.

francopestilli:
@effigies @PeerHerholz @kimberlylray can we merge this into the spec please?

francopestilli (approved):
@arokem and @francopestilli worked together on this.

francopestilli (approved):

francopestilli:
@jdtournier @kabilar @erdalkaraca @oesteban can one of you please review and approve?

arokem:
The last series of commits (starting with [b32ea9f](https://github.com/bids-standard/bids-specification/pull/2333/commits/b32ea9fbcaa71278aa485a0d3e0b7729001ecbd1)) completes the transition of content that was previously in the google doc, so I am also going to mark the google doc as "deprecated". 

Disclosure: these last few commits were generated using the "Big Pickle" model on opencode, but I carefully reviewed every line and adjusted where the model had introduced things I would deem inaccurate. 

There are a few files (in the schema in particular) where I have tried to emulate existing patterns, but might be missing important knowledge, so I might need specifically review of these files by someone who better understands how the schema works.

arokem:
Now also rebased on master.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,166 @@
...
+    {
+        "Metric": ("REQUIRED", "String", "Text description of the anisotropy metric mediating whether a streamline was permitted to propagate from the seed point."),
+        "Source": ("REQUIRED", "String", "Filesystem path or BIDS URI of image data from which the corresponding anisotropy metric can be extracted/computed."),
+        "Threshold": ("REQUIRED", "Float", "Numerical threshold applied to anisotropy metric."),
+    }
+) }}
+
+Table 2d (contents of elements of dict `StreamlineSeeding["InitialDirection"]`):
+
+TODO
```
arokem, line 86: This was absent in the google doc, so is still absent here as well. Not 100% sure what this was supposed to describe originally, a random guess is that @Lestropie might have some memory of this (even though it's been a while).

src/schema/objects/entities.yaml
```diff
@@ -502,16 +502,16 @@ tracer:
...
   description: |
-    The `voi-<label>` entity can be used to distinguish acquisitions localized to different regions.
-    The label SHOULD be the name of the body region or part scanned.
-    If used, the fields `"BodyPart"` and `"BodyPartDetails"` MUST be defined in the JSON file.
-    `BodyPartDetailsOntology` is OPTIONAL to also include.
+    The `track-<label>` entity identifies the tractography method
+    employed to reconstruct white matter fiber streamlines.
   type: string
   format: label
+  enum:
```
arokem: I believe that something about this added section is raising the test errors currently apparent on the CI, but not sure what is offending part. Is it that the top-level key and the value associated with the "name" key are identical? Or something to do with the enum?

arokem: I'm testing one of these hypotheses with cee4f742

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,166 @@
...
+```Text
+<pipeline_name>/
+    sub-<label>/
+        [ses-<label>/]
+            <modality>/
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.trx
+                <source-entities>[space-<space>]_[tract-<tract name>]_[track-<tracking method>]_tractogram.json
+```
+
+Where `tract` is the anatomical/structural entity that is being imaged, and
```
arokem, line 17: Alternatives proposed for "tract" are "fasc" (as in "fascicle"), "subset", "bndl" (as in "bundle"), "part", "str" (for "structure", maybe "strc" or somesuch instead?).

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,166 @@
...
+For example:
+
+```Text
+    pyAFQ/
+        sub–01/
+            dwi/
+                sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.trx
+                sub-01_space-MNI152NLin2009cAsym_tract-wholebrain_track-eudx_tractogram.json
+                sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.trx
+                sub-01_space-MNI152NLin2009cAsym_tract-ArcuateFasciculus_hemi-L_track-eudx_tractogram.json
```
arokem, line 33: Should values associated with "tract" be a controlled vocabulary? Having to define all possible structures that someone might want to describe would be a real bear. On the other hand, not defining this opens up the possibility that different people would name the Arcuate "ArcuateFasciculus", "ARC", "Arcuate", "FrontoTemporal", and so forth., making it hard to share informatively, integrate across datasets, etc.

src/derivatives/diffusion-derivatives.md
```diff
@@ -0,0 +1,166 @@
...
+TODO
+
+Table 3a (contents of the value for key `StreamlinePropagation`):
+
+{{ MACROS___make_metadata_table(
+    {
+        "Algorithm": ("REQUIRED", "Dict", "The contents of this dictionary are described in Table 3b."),
+        "AnglePerStep": ("OPTIONAL", "Float", "The maximal change in degrees in the streamline tangent between successive steps. Note that exactly how this parameter is interpreted may depend on the value of IntegrationOrder."),
+        "Interpolation": ("REQUIRED", "String", "Selection from the following: ['Nearest', 'Linear', 'Cubic', 'Spline']. The mechanism by which sub-voxel information is drawn from the Source based on the precise streamline location."),
+        "InterpolationOrder": ("REQUIRED", "Int", "The order of interpolation used (or 0 if linear)"),
```
arokem: Added here to address comments made in the google doc by @francopestilli and @jhlegarreta

src/schema/objects/entities.yaml
jhlegarreta: How about "pathway", "fascicle" or "fasciculus"? I think that would resolve all concerns. Related to https://github.com/bids-standard/bids-specification/pull/2333/changes#r3464492486

src/derivatives/diffusion-derivatives.md
jhlegarreta: Following to the google doc comment, I am not sure whether at the time I commented on this there was a proposal for the vocabulary (i.e. `linear`, `cubic`, etc.). I am fine using that vocabulary. In that case, the order is redundant. The only concern here are splines, as they would require additional information to be reproducible. Does anybody know, out of the top of their heads, how this is handled in other BEPs where splines are used?

src/derivatives/diffusion-derivatives.md
jhlegarreta: I think we should agree on a minimal set. We can ask the IST subunit to help us with the terms. I know defining them anatomically is not an easy task, but here our job is not to define them anatomically, just to provide a controlled vocabulary, and maybe provide a very broad explanation about them. I believe this is doable for a reasonable amount of tracts.

src/schema/objects/entities.yaml
jhlegarreta: Still learning about the errors, but I think you are right in here:
> Is it that the top-level key and the value associated with the "name" key are identical

src/schema/objects/entities.yaml
arokem: I think this demonstrates the confusion! Here, we are referring to the method used to propagate streamlines ("track") and not to the anatomical structure ("tract"). Confusing.

src/schema/objects/enums.yaml
```diff
@@ -79,6 +79,18 @@ _iEEGCoordSys:
...
     - $ref: objects.enums.Other.value
+_TrackMethod:
+  type: string
+  enum:
+    - $ref: objects.enums.act.value
+    - $ref: objects.enums.eudx.value
+    - $ref: objects.enums.fact.value
+    - $ref: objects.enums.ifod.value
+    - $ref: objects.enums.sdstream.value
+    - $ref: objects.enums.tensor.value
```
arokem, line 90: This seems a little under-constrained. What did you do with this tensor?

src/schema/objects/entities.yaml
jhlegarreta: ~Still learning about the errors, but I think you are right in here:~ Maybe not; other similar cases exist in the document.

So, maybe the issue is that https://github.com/bids-standard/bids-specification/pull/2333/changes#diff-b2ff216a7b94882e7d1d45af322955cc874f5176a75cf0cab41245b7bd75e595R514 cannot be resolved and we need to specify
```
    - $ref: objects.enums.act.value
    - $ref: objects.enums.eudx.value
    - $ref: objects.enums.fact.value
    - $ref: objects.enums.ifod.value
    - $ref: objects.enums.sdstream.value
    - $ref: objects.enums.tensor.value
    - $ref: objects.enums.pft.value
    - $ref: objects.enums.ptt.value
    - $ref: objects.enums.set.value
```

Another possibility is that https://github.com/arokem/bids-specification/blob/75ceda1b00ac66add59721fd8d3185ea269fd37b/tools/schemacode/src/bidsschematools/render/text.py#L76
should be
```
allowed_values.append(value["value"])
```

src/derivatives/diffusion-derivatives.md
arokem: One related idea (from @kimberlylray) as expressed just now in the IST Standarization Unit meeting is to allow for flexibility to use one of several “atlases” and then the names are defined by that atlas (for example, the [HCP-1065 Young Adult Fiber Templates](https://brain.labsolver.org/hcp_template.html). Metadata would report which atlas was used.

src/schema/objects/entities.yaml
jhlegarreta: 🙃 Another reason to avoid "tract" when referring to the anatomical structure.

Lestropie: Recommend forbidding "`track`" throughout. It causes far too many problems. "`tract`" is fine as long as there is diligence around use of "streamline" rather than "track".

src/derivatives/diffusion-derivatives.md
Lestropie: I wouldn't even shoot for a controlled vocabulary. Even for something as well-known as the arcuate, there is disagreement about exactly what breadth of connections that name refers to---not just in the tractography community, in the neuroanatomy community---so controlling the name doesn't standardise the data.

One could make the argument that anything relating to exactly how that tract is defined, ie. the set of criteria used to either reconstruct a tract or extract it from a whole-brain tractogram, is actually a BIDS Provenance issue, and a BEP like this shouldn't attempt to plug that gap. Because even the esoterics of how a particular software package applies that particular set of rules to the data may be consequential, so by the time you embed enough information about not only the set of rules defining the tract but also the software and the processing by which they were applied, hey presto, you've duplicated provenance. If a particular pipeline can provide some information about how the tract was extracted, e.g. a WMQL string or a filename from a piece of software containing a set of rules or a process, maybe there's no harm in it being there as long as it's not to be treated as facilitating reproduction.

src/derivatives/diffusion-derivatives.md
Lestropie: Did not finish translating the set of streamline reconstruction mechanisms to the document. But as per prior discussion, I would be stripping ***all*** of that content out of the BEP. It's not the right environment for that deconstruction, it's not the right resolution of information required for BIDS Derivatives data, some of it is too fine-grained for even tractography experts to be able to fill in, and it draws too much attention and effort away from what should actually be in the BEP.

arokem:
Since this came up earlier in this thread, just a heads up that the new release of FSLeyes now supports the TRX format: 

https://open.oxcin.ox.ac.uk/pages/fsl/fsleyes/fsleyes/userdoc/changelog.html

Hopefully that is helpful for end users of data compliant with the proposed standard, and let me know if there are other places where such support would be helpful.

src/schema/objects/entities.yaml
jhlegarreta: Addressed in 1a6af2a5ec1a77c05578a91886aa166105082c88. Resolving.

francopestilli (commented):
This comment was addressed above

@arokem @kimberlylray

src/derivatives/diffusion-derivatives.md
francopestilli: I just made a related comment below. I will delete that. I agree that we should implement this, given that not all Arcuate Fasciculi are created equal, we need to know how your arcuate was created, meaning which segmentation or atlas you used.

I was thinking to NOT allow produce different tracts by different segmentation methods, but asumen that there will be oalwaysone altas for an entire dataset, if that is a good assumption we can add the atlas / segmentation in the JSON side-car, no? @arokem @kimberlylray

Issue #1566, opened 2023-07-28
Support for susceptibility imaging including QSM, SWI and SWI-MIP


astewartau:
### Your idea

Support for Quantitative Susceptibility Mapping (QSM) currently exists as it is considered a [quantitative](https://bids-specification.readthedocs.io/en/stable/appendices/qmri.html#quantitative-mri), [derivative](https://bids-specification.readthedocs.io/en/stable/appendices/qmri.html#quantitative-maps-are-derivatives) and [anatomical](https://bids-specification.readthedocs.io/en/stable/appendices/entity-table.html#magnetic-resonance-imaging) MR image and uses the [`_Chimap`](https://bids-specification.readthedocs.io/en/stable/glossary.html#chimap-suffixes) suffix. It also has a [BIDS example](https://github.com/bids-standard/bids-examples/tree/master/qmri_qsm). I believe this was introduced by the [qMRI extension](https://bids-specification.readthedocs.io/en/stable/appendices/qmri.html) integrated in 2022 based on Karakuzu et al.'s Scientific Data [publication](https://www.nature.com/articles/s41597-022-01571-4).

However, similar support for Susceptibility Weighted Imaging (SWI) and associated Minimum Intensity Projections (SWI-MIP) is to be integrated. The inactive [BEP004](https://docs.google.com/document/d/1kyw9mGgacNqeMbp4xZet3RnDhcMmf4_BmRgKaOkO2Sc/edit) proposal suggests a `swi/` datatype and proposes storing the input `_GRE` images there with the resultant files using proposed `_swi` and `_minIP` suffixes.

Currently, `_MEGRE` and `_T2starw` suffixes already exist in the `anat/` datatype, so the remaining changes to fully support SWI may be smaller and consist only of adding the `_swi` and `_minIP` suffixes.

Myself, [@stebo85](https://github.com/stebo85), and [@neurolabusc](https://github.com/neurolabusc) propose to finalise support for susceptibility imaging by making some minor pull requests. We also have [open examples on OSF](https://osf.io/ru43c/) of DICOM input data and its mapping to BIDS, which are suitable for QSM and SWI reconstruction that we propose to adapt into similar BIDS examples.

We are opening this issue to begin discussions with the wider susceptibility imaging and software developer community so we can share ideas before changes are made and integrated.

tsalo:
I believe you will also need a coil or channel entity. There's an open PR for that (#425), but it stalled a while back (mostly my fault).

[...]

PR #2343, opened 2026-02-12
Add swi and minIP suffixes for susceptibility weighted imaging

astewartau:
## Summary

Adds support for susceptibility weighted images (`swi`) and minimum intensity projections (`minIP`) as non-parametric anatomical MRI suffixes, addressing the remaining gaps identified in #1566.

- Adds `swi` and `minIP` suffix definitions to `src/schema/objects/suffixes.yaml`
- Adds a file rule group in `src/schema/rules/files/raw/anat.yaml` with standard MRI entities (no `echo` or `part`, since multi-echo combination occurs before SWI generation)
- Adds `swi` and `minIP` to the non-parametric suffix table and filename template in the MRI documentation

These suffixes are classified as non-parametric (arbitrary intensity scale), consistent with how `T2starw` and other contrast-weighted images are handled. Like `Chimap` and other parametric maps, they can appear in raw datasets when pre-generated by the scanner, or in derivative datasets when computed offline.

Ref: https://github.com/bids-standard/bids-specification/issues/1566#issuecomment-3878452407

## Test plan

- [ ] Schema validation passes
- [ ] Rendered docs show `swi` and `minIP` in the non-parametric suffix table and filename templates
- [ ] Example dataset validates (see companion PR in bids-examples)

### DIFF
--- a/src/schema/rules/files/raw/anat.yaml
+++ b/src/schema/rules/files/raw/anat.yaml
@@ -53,6 +53,20 @@ parametric:
     $ref: meta.templates.raw.mri.entities
     task: optional
 
+swi:
+  suffixes:
+    - swi
+    - minIP
+  extensions:
+    - .nii.gz
+    - .nii
+    - .json
+  datatypes:
+    - anat
+  entities:
+    $ref: meta.templates.raw.mri.entities
+    task: optional
+
 defacemask:
   suffixes:
     - defacemask
--- a/src/modality-specific-files/magnetic-resonance-imaging-data.md
+++ b/src/modality-specific-files/magnetic-resonance-imaging-data.md
@@ -237,6 +237,8 @@ and a guide for using macros can be found at
          "PDT2",
          "UNIT1",
          "angio",
+         "swi",
+         "minIP",
       ])
 }}
 
@@ -261,6 +263,8 @@ and a guide for using macros can be found at
          "PDT2",
          "UNIT1",
          "angio",
+         "swi",
+         "minIP",
       ]
    )
 }}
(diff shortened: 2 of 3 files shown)


src/schema/rules/files/raw/anat.yaml
```diff
@@ -53,6 +53,20 @@ parametric:
     $ref: meta.templates.raw.mri.entities
     task: optional
 
+swi:
+  suffixes:
+    - swi
+    - minIP
```
effigies, line 59: Are these both raw images (i.e., you get them off a scanner) or should minIP be a derivative?

Also, is minIP ever sensibly used in a non-SWI context?

tsalo: I think the scanner will produce minIP images. Also, I got minIP images from an angio scan in [ds002278](https://openneuro.org/datasets/ds002278/versions/2.0.0) (see [the JSON](https://openneuro.org/datasets/ds002278/versions/2.0.0/file-display/sub-Bubbles:ses-01:anat:sub-Bubbles_ses-01_acq-axial_run-1_mod-angio_minIP.json)).

effigies: How are minIP used? Are they for looking at or processing? If just looking at, minIP as a suffix seems fine, but if not, maybe it's more a `stat-minIP_swi` than a `mod-swi_minIP`?

tsalo: @astewartau would be better able to answer that, but I _think_ they're for looking at.

fabriciocravo (commented):
Given the ongoing minIP discussion, would it make sense to split this PR? My thinking is that the swi suffix could move forward on its own, as it is needed to share SWI data, while minIP and how it should be represented get addressed separately. The SWI addition appears to be agreed upon, so pushing it alone right now seems reasonable.

Part of the reasoning is that minIP is applied to modalities beyond SWI, so it may warrant its own scope rather than being resolved as part of this one.

PR #2344, opened 2026-02-12
[FIX] Use http://www.dicomlookup.com for DICOM tags

effigies:
http://www.dicomlookup.com is back up. It no longer accepts HTTPS connections or dropping the `www`.

Based on http://www.dicomlookup.com/about:

> This site is a hobby project. Your support to cover hosting costs is greatly appreciated.

I think it's not ideal to slam their server thousands of times a day, so I'm still excluding them from the link checker.


rwblair (approved):

bendhouseart (approved):
👍

yarikoptic:
PR released in [`v1.11.1`](https://github.com/bids-standard/bids-specification/releases/tag/v1.11.1)

PR #2346, opened 2026-02-14
[FIX] Add emg to timeseries rule for physio/stim files

neuromechanist:
## Summary
- Add `emg` to the base `timeseries` rule datatypes in `src/schema/rules/files/raw/task.yaml`
- This allows `_physio`, `_physioevents`, and `_stim` suffixes to be valid within `emg/` directories
- EMG was the only datatype missing from the timeseries rule, despite being a first-class BIDS datatype since v1.11.0 (BEP 042)

## Context
When converting the [HySER dataset](https://physionet.org/content/hd-semg/1.0.0/) to BIDS-EMG format, finger force sensor recordings (`_physio.tsv.gz`) placed alongside EMG data cause `ALL_FILENAME_RULES_HAVE_ISSUES` validation errors because the validator schema does not include `emg` in the timeseries rule.

All other electrophysiology datatypes (`eeg`, `ieeg`, `nirs`) as well as `beh`, `func`, `meg`, `motion`, `pet`, `anat`, `dwi`, and `perf` already support physio/stim files.

## Test plan
- [x] Verify `emg` is a defined datatype in `src/schema/objects/datatypes.yaml`
- [x] Run schema tests
- [x] Validate a BIDS-EMG dataset with `_physio.tsv.gz` files passes after this change

### DIFF
--- a/src/modality-specific-files/electromyography.md
+++ b/src/modality-specific-files/electromyography.md
@@ -13,10 +13,17 @@ this extension when referring to it in the context of the academic literature.
 
 ## EMG data
 
+<!--
+This block generates a filename templates.
+The inputs for this macro can be found in the directory
+  src/schema/rules/files/raw
+and a guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
 {{ MACROS___make_filename_template(
-"raw",
-datatypes=["emg"],
-suffixes=["emg", "events"])
+   "raw",
+   datatypes=["emg"],
+   suffixes=["emg", "events", "physio", "stim"])
 }}
 
 EMG device manufacturers use a variety of formats for storing raw data, and there is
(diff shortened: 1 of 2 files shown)


effigies:
I agree with this. A question is whether this should wait on (or justify) another BIDS release (`1.11.1`) or can be a schema patch (`schema-1.2.1`) on v1.11.0.

For it to be a schema patch, it cannot change the semantic meaning of the BIDS specification version 1.11.0. While I think this was obviously intended in v1.11.0, it does change the literal text of the document, so I think a stickler could argue for a new spec release.

There has been very little activity since the 1.11.0 release, so we could simply make a 1.11.1 release with this fix, and avoid the whole question.

Opinions? @bids-standard/maintainers @bids-standard/steering?

effigies:
@neuromechanist Could you add a physio file to an example dataset to validate this?

effigies (approved):

bendhouseart (approved):
👍 lgtm

yarikoptic:
PR released in [`v1.11.1`](https://github.com/bids-standard/bids-specification/releases/tag/v1.11.1)

PR #2348, opened 2026-02-19
rf(schema): Consolidate `rules.dataset_metadata` under `rules.json`

effigies:
This is a slight cleanup. `rules.dataset_metadata` and `rules.json` developed in parallel, but are ultimately the same thing. This makes it easier on schema users to find the rules that apply to any given JSON file by consolidating all under `rules.json`.

I did move the `dataset_description_with_genetics` rule to `rules.json.genetics` instead of `rules.json.dataset`, since those rules were introduced in the genetics BEP. It could also be argued that it should go under `rules.json.dataset`, since it directly impacts `dataset_description.json`. The location has no impact on validation, it's just a question of what's most intuitive.


bendhouseart (approved):
Agree, it's slightly more intuitive.

effigies:
Tempting @yarikoptic into reviewing with a promise of a more consistent schema.

yarikoptic:
doesn't this change/break metaschema and thus requiring boost of https://github.com/bids-standard/bids-specification/blob/master/src/schema/SCHEMA_VERSION to 2.0.0-dev?

effigies:
It seemed like a relatively small schema change, so I figured the minor version bump would be fine, but I have no objection to calling it major.

rwblair (approved):

Issue #965, opened 2022-01-06
[SCHEMA] make schemacode into bids-schema (bids_schema pkg)


yarikoptic:
#892 made the main work to detach schema handling code into an installable python package.

https://github.com/bids-standard/bids-specification/pull/892/files#diff-fca7304d24a5898ab105cce4ccd5a21761e7673dbeccb692b65bc6cddef9eee7R3 lists https://github.com/bids-standard/schemacode and setup.py https://github.com/bids-standard/bids-specification/pull/892/files#diff-a2a4329d57cbccf54036da2be45bf77cf3059065fb0e856244f86d05394e3082R17 lists it as `"schemacode" .

I think eventually, whenever this code matures to leave the mothership of the bids-specification, its name should become globally unique. And even before the detachment from this repo, it would be nice to have a copy deposited to PyPI while building/releasing straight from this repo (with its own versioning in 0.x.y series while it matures, independently from bids-specification, so versioneer might need to be gone to replaced).  Upload to PyPI would require the name to be unique.  So I am proposing to rename the beast into `bids_schema` package (but for PyPI to make it prettier, could be `bids-schema` name).  Then whoever wants to use that as a library would be able to do so.

Attn @TheChymera

TheChymera:
@yarikoptic 
Set up this repo to try out full separation and allow easier management of schema PRs without having to keep track of full bids-specification repository history and merge conflicts:
https://github.com/TheChymera/bids-schemadata

TheChymera:
Following up on this:

With the addition of the validator XS code which checks for readability and may also be loaded by external tools needing a simple officially working backend, probably simply splitting off the schema in and of itself is not such a great idea as I had originally thought.

I have experimented a bit and the schemacode package works well on its own, and given access to a linux package manager (in my example, Gentoo) it can easily if not very elegantly be distributed as such with zero changes to the current codebase: https://github.com/TheChymera/overlay/commit/ae712de08ac3af94627ea895e6bd6f2ff0779760

The follow-up issues I am thinking of are:

1. since this is schema with testcode, perhaps changing the name to `bids_schemacode` (@yarikoptic 's suggestion) is a good idea. This should be easy to sort out, just let me know if you agree and I can make a PR.
[...]

PR #2351, opened 2026-02-27
[ENH][SCHEMA] Use Universal Path Library for Schema Retrieval

bendhouseart:
I had initially added this to a PR for pybids but it seems more appropriate here. I think this also plays a bit into the discussion at issue #965 in so far as python users are concerned.

Conversely, a user could make a request first to `https://bids-specification.readthedocs.io/en/stable/schema.json` then use `schema.load_schema` following that retrieval. 

I don't know, maybe this is a solution in search of a problem and/or should be fleshed out to retrieve specific versions of the schema from bids-standard/bids-schema.

### DIFF
--- a/tools/schemacode/pyproject.toml
+++ b/tools/schemacode/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["pdm-backend", "acres", "pyyaml"]
+requires = ["pdm-backend", "acres", "pyyaml", "universal-pathlib", "requests", "aiohttp"]
 build-backend = "pdm.backend"
 
 [project]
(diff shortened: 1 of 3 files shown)


effigies (commented):
I really don't want to do this. If anything, I want fewer dependencies for tools that want to use the schema. We allow users to use `UPath`s.

What problems are you having with pdm?

bendhouseart:
> Interesting. I only use `uv` (mainly `uv sync`), and not the pdm frontend. It's just a quite easy backend to write plugins for. If you want to rewrite in a different backend and get the same result, I'll have no objection.

Yeah I think it's a me uv/pdm/brew thing, but I kept getting invalid/unavailable backend when I was running uv run pytest... will make a smaller PR when I get the chance next week.

PR #2363, opened 2026-03-16
BEP032: Relax many levels to optional

CodyCBakerPhD:
@ree-gupta @bendichter These are my suggestions for what can/should be relaxed in terms of 'how easy' things are to extract from NWB

Note that for many of these fields, there might be a generic place in NWB to write some information but it is nearly always free-text strings that cannot be reliably parsed into individual BIDS fields. A fully automated conversion tool could not set those fields, but an interactive agent such as the meditor could likely parse the text and provide guesses to matching fields, so I would suggest that being a post-organization enhancement for data contributors

An extension or core modification would be needed to completely 'harmonize' the two (not that I am convinced that would entirely be worth it, given we can just treat the BIDS metadata as a ground truth)

cc @rwblair @effigies

### DIFF
--- a/src/schema/rules/tabular_data/microephys.yaml
+++ b/src/schema/rules/tabular_data/microephys.yaml
@@ -121,7 +121,7 @@ microephysChannels:
         For channels not associated with an electrode, use `n/a`.
     type__channels: required
     units: required
-    sampling_frequency: optional
+    sampling_frequency: required
     low_cutoff: optional
     high_cutoff: optional
     reference__microephys: optional


bendichter (approved):
looks good, thanks for these changes @CodyCBakerPhD !

src/schema/rules/tabular_data/microephys.yaml
ree-gupta:
```suggestion
    sampling_frequency: required
```

CodyCBakerPhD: required makes the most sense to me to match it across states - some mild duplication, but not terrible

PR #2366, opened 2026-03-17
Add channels -> electrodes -> probes consistency checks

ree-gupta:
Thanks @effigies and @rwblair for the meeting and feedback yesterday. I have tried drafting a rule for the channels -> electrodes -> probes consistency checks.

### What this does

Validates that references from child files resolve in parent files:
- `electrode_name` in `channels.tsv` must exist in `electrodes.tsv`
- `probe_name` in `electrodes.tsv` must exist in `probes.tsv`

Both checks allow `n/a` (for aux channels without electrodes, or standalone electrodes without probes). We intentionally do not check the reverse direction (parent -> child) since not all electrodes need channels (broken contacts, reference-only, run-specific subsets).

### Changes

- **`context.yaml`**: Added `name` column to `electrodes` association and new `probes` association with `probe_name` column
- **`associations.yaml`**: Extended selectors so `electrodes` is resolved when visiting `channels.tsv` and `probes` is resolved when visiting `electrodes.tsv` (follows the `coordsystem` precedent which already includes `electrodes` in its selector). Added `inherit: true` to `probes`.
- **`checks/microephys.yaml`**: Two warning-level child -> parent reference checks

### Looking for feedback on

I was not able to get the validator to fire these checks when I broke a dataset by replacing `probe_name` values in `electrodes.tsv` with nonexistent names. The schema compiles fine, valid datasets still pass, but the broken dataset did not trigger the new warnings.

Also, the check expression uses `length(intersects(...))`. When `intersects()` finds zero overlap it returns `false` rather than an empty array -- how does `length(false)` behave? If this is problematic, would a simpler check like the following be preferred?

```yaml
count(unique(columns.probe_name), "n/a") == length(unique(columns.probe_name))
|| intersects(unique(columns.probe_name), associations.probes.probe_name)
```

Related #2307

### DIFF
--- a/src/schema/meta/associations.yaml
+++ b/src/schema/meta/associations.yaml
@@ -94,7 +94,7 @@ coordsystem:
 
 electrodes:
   selectors:
-    - intersects([suffix], ['eeg', 'emg', 'ieeg', 'meg', 'ecephys', 'icephys'])
+    - intersects([suffix], ['eeg', 'emg', 'ieeg', 'meg', 'ecephys', 'icephys', 'channels'])
     - extension != '.json'
   target:
     suffix: electrodes
@@ -105,11 +105,12 @@ electrodes:
 
 probes:
   selectors:
-    - intersects([suffix], ['ecephys', 'icephys'])
+    - intersects([suffix], ['ecephys', 'icephys', 'electrodes'])
     - extension != '.json'
   target:
     suffix: probes
     extension: .tsv
+  inherit: true
 
 physio:
   selectors:
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -271,6 +271,25 @@ properties:
           path:
             description: 'Path to associated electrodes.tsv file'
             type: string
+          name:
+            description: 'Contents of the name column'
+            type: array
+            items:
+              type: string
+      probes:
+        description: 'Probes file'
+        type: object
+        required: [path]
+        additionalProperties: false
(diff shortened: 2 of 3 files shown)


bendichter:
@ree-gupta I found why the checks were not firing. The schema side of this PR is correct. The validator only reads extra columns from an associated file when it has a loader for that association, and `electrodes` and `probes` use the default loader, which records only the path. So `associations.probes.probe_name` and `associations.electrodes.name` were always null, the selectors failed, and the checks were skipped without any message. I opened bids-standard/bids-validator#453 to add the two loaders.

With that change, I ran the validator against the microephys_toy example from bids-standard/bids-examples#527, using a schema built from this branch. The unmodified dataset reports neither issue. A copy with one `probe_name` in `electrodes.tsv` and one `electrode_name` in `channels.tsv` pointing at nothing reports both. On your question about `length(false)`: when no values match, `length` returns null and the check fails as it should, and a column that is all `n/a` passes. I tested both cases, so the expression can stay as it is.

I pushed two commits to your branch. The first merges `bep032-review`, which resolves the conflict in `checks/microephys.yaml` by keeping both sets of checks. The second raises both checks from warning to error, since the column descriptions say the value MUST match. Let me know if you intended them as warnings.

ree-gupta:
Hi, @bendichter, thank you for continuing the work on this! 
I see about the associations, makes sense.
Feel free to modify this PR as needed :)

PR #2374, opened 2026-03-26
[FIX] Correct BIDS Starter Kit link in src/index.md

ericearl:
Updated link for the BIDS Starter Kit to the correct URL.


bendhouseart (approved):
Thanks Eric!

PR #2392, opened 2026-04-13
[ENH] Add OME-Zarr as a supported format for imaging data

### LINKED ISSUE #1704: Allow OME-Zarr for non-microscopy data
BIDS accepts ome-zarr data in its microscopy extension, but not in its main MRI-related specification.

In most cases, MRI volumes are small enough that chunked formats do not make sense, but MRIs acquired in ex vivo human brains can become quite large. For example, [these 120 um isotropic MRIs](https://dandiarchive.org/dandiset/000026/draft/files?location=sub-I48%2Fses-MRI%2Fanat&page=1) are about 5GB each. This is particularly problematic for web-based viewing, as most (all?) viewers load the entire volume in GPU memory and have a hard memory limit. The MRI linked above is 1600x1400x640 and I am not sure that niiview would be able to display it. Even if it were, having to download the entire file before showing it makes this impractical.

Could we allow OME-Zarr files in the main spec? 

The main problem I see is that, in its current form, the OME-NGFF specification does not allow storing most of the metadata that live in the nifti header -- most improtantly the qform and sform. As an alternative, we have drafted a very lightweight supplement to OME-Zarr, namely [NIfTI-Zarr](https://github.com/neuroscales/nifti-zarr/tree/draft/0.1), which only requires dumping the nifti header under `.zattrs["nifti"]["base64"]` (using base64 encoding). This makes it very easy to decode the nifti metadata for any library that has access to a base64 decoder, and a nifti parser. We have reference implementations in [python](https://github.com/neuroscales/nifti-zarr/pull/2) and [julia](https://github.com/neuroscales/nifti-zarr/pull/3).

I guess this might be somewhat related to #197

@satra @martindurant @yarikoptic

kabilar:
Fix #1704

cc @satra @balbasty @ayendiki

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -520,9 +520,6 @@ Spatial metadata (such as the axis names and units, and coordinate transformatio
 be stored within the OME-Zarr metadata following the
 [OME-Zarr version 0.5 specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html)
 (the latest released version).
-Extended support for coordinate systems and transformations is described in
-[RFC-5: Coordinate Systems and Transformations](https://ngff.openmicroscopy.org/rfc/5/index.html)
-(draft, expected to be released as part of OME-Zarr version 0.6).
 
 ### Tabular files
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -506,6 +506,24 @@ The [BIDS-validator](https://github.com/bids-standard/bids-validator)
 will check for conflicts between the JSON file and the data recorded in the
 NIfTI header.
 
+#### OME-Zarr
+
+[Zarr](https://zarr-specs.readthedocs.io/) is a chunked, cloud-optimized format that provides efficient access to
+large multidimensional datasets without requiring a full download.
+[OME-Zarr](https://ngff.openmicroscopy.org/), developed by the Open Microscopy Environment (OME),
+extends Zarr with bioimaging-specific metadata.
+OME-Zarr is particularly suitable for very large imaging volumes (for example, high-resolution
+ex vivo MRI) where NIfTI would be impractical for streaming or web-based visualization.
+
+OME-Zarr filesets are stored with the `.ome.zarr` extension.
+Spatial metadata (such as the axis names and units, and coordinate transformations) SHOULD
+be stored within the OME-Zarr metadata following the
+[OME-Zarr version 0.5 specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html)
+(the latest released version).
+Extended support for coordinate systems and transformations is described in
+[RFC-5: Coordinate Systems and Transformations](https://ngff.openmicroscopy.org/rfc/5/index.html)
+(draft, expected to be released as part of OME-Zarr version 0.6).
+
 ### Tabular files
 
 Tabular data MUST be saved as plain-text, tab-delimited values (TSV) files
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -480,8 +480,13 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 
 ### Imaging files
 
-All imaging data MUST be stored using the NIfTI file format. We RECOMMEND using
-compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. If using compressed files,
+Imaging data SHOULD be stored using the NIfTI file format.
+Large imaging data MAY instead be stored using the
+[OME-Zarr (OME-NGFF)](https://ngff.openmicroscopy.org/) file format.
+
+#### NIfTI
+
+We RECOMMEND using compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. If using compressed files,
 the gzip header SHOULD lack source filenames and timestamps. Imaging data SHOULD
 be converted to the NIfTI format using a tool that provides as much of the NIfTI
 header information (such as orientation and slice timing information) as
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -520,6 +520,9 @@ Spatial metadata (such as the axis names and units, and coordinate transformatio
 be stored within the OME-Zarr metadata following the
 [OME-Zarr version 0.5 specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html)
 (the latest released version).
+An example dataset containing an OME-Zarr can be found in the
+[BIDS examples repository](https://github.com/bids-standard/bids-examples/tree/master/micr_XPCTzarr)
+and can be used as helpful guidance when curating new datasets.
 
 ### Tabular files
 
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -522,7 +522,7 @@ extends Zarr with bioimaging-specific metadata.
 OME-Zarr is particularly suitable for very large imaging volumes (for example, high-resolution
 ex vivo MRI) where NIfTI would be impractical for streaming or web-based visualization.
 
-OME-Zarr filesets are stored with the `.ome.zarr` extension.
+OME-Zarr filesets are stored in BIDS with the `.ome.zarr` extension.
 Spatial metadata (such as the axis names and units, and coordinate transformations) SHOULD
 be stored within the OME-Zarr metadata following the
 [OME-Zarr version 0.5 specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html)
--- a/src/schema/rules/checks/nifti.yaml
+++ b/src/schema/rules/checks/nifti.yaml
@@ -50,6 +50,20 @@ NiftiPixdimPET:
   checks:
     - min([nifti_header.pixdim[1], nifti_header.pixdim[2], nifti_header.pixdim[3]]) > 0
 
+NiftiLargeVolume:
+  issue:
+    code: NIFTI_LARGE_VOLUME
+    message: |
+      NIfTI file has a spatial dimension exceeding 1024 voxels.
+      Consider using the OME-Zarr file format, which is better suited
+      for large imaging data.
+    level: warning
+  selectors:
+    - type(nifti_header) != "null"
+    - length(nifti_header.shape) >= 3
+  checks:
+    - max([nifti_header.shape[0], nifti_header.shape[1], nifti_header.shape[2]]) <= 1024
+
 # 60
 XformCodes0:
   issue:
--- a/src/schema/rules/checks/nifti.yaml
+++ b/src/schema/rules/checks/nifti.yaml
@@ -59,6 +59,7 @@ NiftiLargeVolume:
       for large imaging data.
     level: warning
   selectors:
+    - match(extension, "^\.nii(\.gz)?$")
     - type(nifti_header) != "null"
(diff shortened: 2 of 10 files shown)


kabilar:
I fixed most of the CI failures.  The remaining 2 failures appear unrelated.

effigies (approved):
I support this. It is in line with the decision made at the Copenhagen 2023 meeting which was: HDF5 and Zarr are preferred formats for new multi-dimensional data.

It looks like we dropped the ball on the wording in #1614, but I do think the decision should hold, nonetheless.

src/common-principles.md
```diff
@@ -501,6 +506,24 @@ The [BIDS-validator](https://github.com/bids-standard/bids-validator)
...
+[OME-Zarr](https://ngff.openmicroscopy.org/), developed by the Open Microscopy Environment (OME),
+extends Zarr with bioimaging-specific metadata.
+OME-Zarr is particularly suitable for very large imaging volumes (for example, high-resolution
+ex vivo MRI) where NIfTI would be impractical for streaming or web-based visualization.
+
+OME-Zarr filesets are stored with the `.ome.zarr` extension.
+Spatial metadata (such as the axis names and units, and coordinate transformations) SHOULD
+be stored within the OME-Zarr metadata following the
+[OME-Zarr version 0.5 specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html)
+(the latest released version).
```
yarikoptic, line 522: I thought to scream "duplication is evil!" but apparently we do not really disclose much in microscopy section about OME or NGFF

```
❯ git grep -i -e OME.Zarr -e ngff -- src/modality-specific-files/microscopy.md
src/modality-specific-files/microscopy.md:file's header. Further, OME-ZARR (sometimes referred to as OME-NGFF or NGFF) has been developed to provide improved
src/modality-specific-files/microscopy.md:-   [OME-ZARR/NGFF](https://ngff.openmicroscopy.org/latest/) (`.ome.zarr` directories)
```

but then we should at least take an opportunity to point to this section here from within that microscopy.md section.  attn @bids-standard/bep031

src/common-principles.md
```diff
@@ -501,6 +506,24 @@ The [BIDS-validator](https://github.com/bids-standard/bids-validator)
...
+Extended support for coordinate systems and transformations is described in
+[RFC-5: Coordinate Systems and Transformations](https://ngff.openmicroscopy.org/rfc/5/index.html)
+(draft, expected to be released as part of OME-Zarr version 0.6).
```
yarikoptic, line 525: oh what a lovely pandora's box to open here.  AFAIK we do not point to WiP within the spec, not even our own BEPs, so I do not see why to mention WiP OME-Zarr work here.

Also ping to freshly created team @bids-standard/bep014  for retrospection on this aspect and PR overall

src/common-principles.md
```diff
@@ -480,8 +480,13 @@ field in `dataset_description.json` of each subdirectory of `derivatives` to:
 
 ### Imaging files
 
-All imaging data MUST be stored using the NIfTI file format. We RECOMMEND using
-compressed NIfTI files (.nii.gz), either version 1.0 or 2.0. If using compressed files,
+Imaging data SHOULD be stored using the NIfTI file format.
+Large imaging data MAY instead be stored using the
+[OME-Zarr (OME-NGFF)](https://ngff.openmicroscopy.org/) file format.
```
yarikoptic, line 485: I wonder if some schema-level check /could/should be added here based on some characteristics of the volume(s) -- e.g. any dimension having over 1k elements?

effigies: Could: yes

```yaml
checkname:
  issue: ...
  selectors:
    - type(nifti_header) != 'null'
  checks:
    - max(nifti_header.shape) <= 1000
```

Should: 🤷🏻

src/common-principles.md
kabilar:
```suggestion
```

kabilar: Sounds good.  I have removed the sentence about RFC-5.

src/common-principles.md
kabilar: Thanks. I don't feel too strongly here, but if we do add a check let's use `level: warning`.

src/common-principles.md
kabilar: Good idea.  I have just pushed a commit.

yarikoptic:
That's shaping up nicely!  Please complement with a PR against bids-examples which would demo some fake dataset with .ome.zarr . FWIW -- [here](https://github.com/bids-standard/bids-examples/pull/451/changes#diff-e24297ce70dde24311a763f6c08816cd5ca82acb2aae135de0359256da748d18) I am creating one but scripting so I could easily recreate etc

kabilar:
> That's shaping up nicely! Please complement with a PR against bids-examples which would demo some fake dataset with .ome.zarr . FWIW -- [here](https://github.com/bids-standard/bids-examples/pull/451/changes#diff-e24297ce70dde24311a763f6c08816cd5ca82acb2aae135de0359256da748d18) I am creating one but scripting so I could easily recreate etc

Thanks @yarikoptic.  There was already a `bids-examples` dataset with an `.ome.zarr` so I pushed a commit that references it.

src/common-principles.md
yarikoptic: yeap, I would warn one that it should be zarr if going e.g. over 1000 for any spatial dimension (I don't know if `.shape` includes time or DTI directions etc), but I guess we would need to also verbalize that in text somehow.

kabilar: Thanks.  I have pushed a commit with a schema-level check.

kabilar:
Hi @yarikoptic, just following up on my recent commits to see if you have any feedback, thanks.

src/common-principles.md
```diff
@@ -508,6 +513,24 @@ The [BIDS-validator](https://github.com/bids-standard/bids-validator)
...
 NIfTI header.
 
+#### OME-Zarr
+
+[Zarr](https://zarr-specs.readthedocs.io/) is a chunked, cloud-optimized format that provides efficient access to
+large multidimensional datasets without requiring a full download.
+[OME-Zarr](https://ngff.openmicroscopy.org/), developed by the Open Microscopy Environment (OME),
+extends Zarr with bioimaging-specific metadata.
+OME-Zarr is particularly suitable for very large imaging volumes (for example, high-resolution
+ex vivo MRI) where NIfTI would be impractical for streaming or web-based visualization.
```
yarikoptic, line 523: echoing today's demo by @neurolabusc and seeing https://github.com/niivue/niivue/blob/53d450fe2e43c40c0d4ad6a2d8dc08ee37575f8c/packages/niivue/src/nvimage/zarr/ZarrChunkClient.ts#L74 -- I guess niivue would be ready to support visualization of those. What about pipelining @neurolabusc and @francopestilli  ?

yarikoptic (approved):
From what I see -- looks good to me

src/common-principles.md
```diff
@@ -508,6 +513,24 @@ The [BIDS-validator](https://github.com/bids-standard/bids-validator)
...
+
+OME-Zarr filesets are stored with the `.ome.zarr` extension.
```
joshmoore: Two quick heads up on the extension:
 - This extension is a SHOULD and not a MUST.
 - RFC-9 (not yet accepted) is proposing an additional (zipped) extension: .ozx

yarikoptic: Thank you @joshmoore for the feedback!  Here, IMHO, we should enforce the ones we consider in BIDS. May be worth making it more specific (not making a generic statement).

```suggestion
OME-Zarr filesets are stored in BIDS with the `.ome.zarr` extension.
```

Good to know about rfc-9 -- ref: https://ngff.openmicroscopy.org/rfc/9/ .

src/common-principles.md
kabilar: Thanks @yarikoptic.  From what I understand, niivue is still in the early phases of adding ome-zarr support.  See demo here: https://niivue.com/demos/features/ome-zarr.html.  I'll resolve this conversation and perhaps we can file an issue in the niivue repo to discuss further?

src/common-principles.md
kabilar: Thanks @joshmoore and @yarikoptic.  I have committed Yarik's suggestion.  Once RFC-9 is accepted, we can propose to include `.ozx` in BIDS.

yarikoptic:
@effigies - we have 2 approvals.  This is a tiny PR with HUGE potential consequences.  I would love us to proceed on it regardless since I feel it is the right thing to do!

kabilar:
@ayendiki @satra @balbasty @tinggong @chiaramff @kimberlylray @francopestilli @fcollman @srheilbronner

Hi all, this topic came up in today's BRAIN Initiative pre-meeting discussion. Given your work in large-scale data generation, visualization, and analysis, I'd value your feedback on this addition. A comment, review, or 👍 would help us gauge consensus before moving forward. Thank you.

src/schema/rules/checks/nifti.yaml
```diff
@@ -50,6 +50,20 @@ NiftiPixdimPET:
...
+    message: |
+      NIfTI file has a spatial dimension exceeding 1024 voxels.
+      Consider using the OME-Zarr file format, which is better suited
+      for large imaging data.
+    level: warning
+  selectors:
+    - type(nifti_header) != "null"
+    - length(nifti_header.shape) >= 3
+  checks:
+    - max([nifti_header.shape[0], nifti_header.shape[1], nifti_header.shape[2]]) <= 1024
```
effigies, line 66: I'm a bit concerned that this will catch up cifti files. That said, refining a warning can be done post release.

effigies (commented):
This has been open some time with no objections. I'm +1 to merge.

@nellh any concerns from the OpenNeuro side? There is the potential for an explosion of inodes, but I would assume these datasets will tend to be small numbers of subjects.

kabilar:
> There is the potential for an explosion of inodes, but I would assume these datasets will tend to be small numbers of subjects.

And with sharding in Zarr v3, the number of files will hopefully be manageable.

src/schema/rules/checks/nifti.yaml
```diff
@@ -50,6 +50,20 @@ NiftiPixdimPET:
...
+  issue:
+    code: NIFTI_LARGE_VOLUME
+    message: |
+      NIfTI file has a spatial dimension exceeding 1024 voxels.
+      Consider using the OME-Zarr file format, which is better suited
+      for large imaging data.
+    level: warning
+  selectors:
+    - type(nifti_header) != "null"
+    - length(nifti_header.shape) >= 3
```
kabilar, line 64: Updating the selector pattern to exclude cifti files.

```suggestion
  selectors:
    - match(extension, "^\.nii(\.gz)?$")
    - type(nifti_header) != "null"
    - length(nifti_header.shape) >= 3
```

src/schema/rules/checks/nifti.yaml
kabilar: Thanks.  This fix should hopefully resolve this issue: https://github.com/bids-standard/bids-specification/pull/2392#discussion_r3874611667

effigies:
Alright, no concerns raised. Let's move forward.

PR #2394, opened 2026-04-13
[FIX] Accept MiscChannelCount in EEG/Motion sidecars; deprecate MISCChannelCount alias

### LINKED ISSUE #2393: Inconsistent EEG misc channel count key: schema rule says `MISCChannelCount`, validator check and docs use `MiscChannelCount`
## Description

The BIDS schema is internally inconsistent for the misc-channel-count key in EEG sidecars. The Recommended-fields rule uses `MISCChannelCount` (all-caps), but the validator check rule, the spec example in the documentation, the MEG/iEEG sidecar rules, and the legacy validator's JSON schema all use `MiscChannelCount` (mixed-case).

Because **datasets in the wild already use both spellings** depending on which writer/version produced them, the resolution should preserve backward compatibility rather than force a single canonical key.

## Where the inconsistency lives

In this repository:

| File | Key used |
| --- | --- |
| `src/schema/rules/sidecars/eeg.yaml:73` (`EEGRecommended`) | **`MISCChannelCount`** |
| `src/schema/rules/sidecars/motion.yaml:64` (`MotionRecommended`) | **`MISCChannelCount`** |
| `src/schema/rules/checks/eeg.yaml:54-66` (`MiscChannelCountReq`) | `MiscChannelCount` |
| `src/schema/rules/sidecars/meg.yaml:101` | `MiscChannelCount` |
| `src/schema/rules/sidecars/ieeg.yaml:79` | `MiscChannelCount` |
| `src/modality-specific-files/electroencephalography.md:179` (example) | `MiscChannelCount` |
| `src/modality-specific-files/intracranial-electroencephalography.md:200` (example) | `MiscChannelCount` |
| `src/schema/objects/metadata.yaml:2387 / :2596` | both keys defined as separate metadata objects |

In `bids-validator@1.15.0` (legacy):
- `validators/json/schemas/eeg.json` lists only `MiscChannelCount` and sets `additionalProperties: false`, so writing `MISCChannelCount` triggers `JSON_SCHEMA_VALIDATION_ERROR (code: 55)`.
- The bundled new BIDS-schema rules also include both names, so different warnings surface depending on which spelling the sidecar contains.

## Reproducer (downstream impact)

mne-tools/mne-bids#1546 reports this on the writing side: `write_raw_bids()` writes `MiscChannelCount`, which validates against the legacy validator and the schema check rule, but the schema-based validator's *Recommended* rule complains that `MISCChannelCount` is missing. Switching to `MISCChannelCount` flips the failures to the legacy validator and the consistency check. There is no key the writer can choose that satisfies every rule.

## Proposed resolution (backward-compatible)

Treat both spellings as valid in the EEG sidecar — pick `MiscChannelCount` as the canonical (it matches the validator check, MEG/iEEG sidecars, both spec examples, and the legacy validator schema), and mark `MISCChannelCount` as a `deprecated` alias so existing datasets keep validating without warnings turning into errors.

Concretely in `src/schema/rules/sidecars/eeg.yaml`:

```yaml
EEGRecommended:
  selectors:
    - datatype == "eeg"
    - suffix == "eeg"
  fields:
    ...
    MiscChannelCount: recommended
    MISCChannelCount:
      level: deprecated
      description_addendum: |
        Deprecated alias of `MiscChannelCount`. Existing datasets remain valid;
        new datasets SHOULD use `MiscChannelCount` (matching MEG and iEEG sidecars,
        the consistency check rule, and the documentation examples).
    ...
```

This way:
- Existing datasets with either spelling continue to validate.
- New writers (mne-bids and others) converge on `MiscChannelCount`.
- The contradiction between the recommended-fields rule and the consistency check disappears.

I'll send a small PR with that change. The `motion.yaml:64` case is a separate question (motion is a newer modality and may have intentionally adopted a different convention) — happy to handle it in a follow-up if maintainers want it aligned.

cc @effigies (since #1319 introduced the EEG check rules)

bruAristimunha:
Closes #2393.

## Problem

The EEG sidecar's `EEGRecommended` rule names the misc-channel-count key as `MISCChannelCount`, but the rest of the schema and the docs use `MiscChannelCount`:

| Source | Key |
| --- | --- |
| `src/schema/rules/sidecars/eeg.yaml:73` (`EEGRecommended`) | **`MISCChannelCount`** |
| `src/schema/rules/checks/eeg.yaml` (`MiscChannelCountReq`) | `MiscChannelCount` |
| `src/schema/rules/sidecars/meg.yaml:101` | `MiscChannelCount` |
| `src/schema/rules/sidecars/ieeg.yaml:79` | `MiscChannelCount` |
| `src/schema/rules/sidecars/motion.yaml:64` | **`MISCChannelCount`** |
| `src/modality-specific-files/electroencephalography.md:179` (example) | `MiscChannelCount` |
| `bids-validator@1.x` legacy `eeg.json` schema (`additionalProperties: false`) | `MiscChannelCount` |

The contradiction means writers cannot satisfy both rules at once: writing `MiscChannelCount` triggers a missing-recommended warning, writing `MISCChannelCount` triggers `JSON_SCHEMA_VALIDATION_ERROR (code 55)` against the legacy validator and silences the consistency check. Reproducer / impact: mne-tools/mne-bids#1546.

## Fix

Make `MiscChannelCount` the canonical recommended key in both EEG and motion sidecars (matching the rest of the schema, the docs example, and the legacy validator), and keep `MISCChannelCount` as a `deprecated` alias in both so datasets in the wild that already use that spelling keep validating.

```yaml
EEGRecommended / motionRecommended:
  fields:
    ...
    MiscChannelCount: recommended
    MISCChannelCount: deprecated
```

The deprecation note lives on the field definition itself (`src/schema/objects/metadata.yaml`), so the glossary entry for `MISCChannelCount` points readers at the canonical `MiscChannelCount` wherever the field is referenced.

## Validator behaviour

Checked with `bids-validator@2.4.1` against the updated schema:

- Datasets using `MiscChannelCount` validate cleanly.
- Datasets using the deprecated `MISCChannelCount` are accepted without error (existing datasets keep passing). bids-validator v2 does not currently emit a dedicated `DEPRECATED_SIDECAR_KEY` warning for any `deprecated`-level sidecar field (e.g. `DCOffsetCorrection`, `HardcopyDeviceSoftwareVersion`) — that gap is pre-existing and belongs upstream in `bids-standard/bids-validator`, not in this spec PR.

cc @effigies (since #1319 introduced the EEG check rules)

### DIFF
--- a/src/schema/rules/sidecars/eeg.yaml
+++ b/src/schema/rules/sidecars/eeg.yaml
@@ -71,13 +71,7 @@ EEGRecommended:
     EMGChannelCount: recommended
     EOGChannelCount: recommended
     MiscChannelCount: recommended
-    MISCChannelCount:
-      level: deprecated
-      description_addendum: |
-        Deprecated alias of `MiscChannelCount`. Existing datasets remain valid;
-        new datasets SHOULD use `MiscChannelCount`, which matches the MEG and
-        iEEG sidecar rules, the EEG channel-count consistency check, and the
-        documentation example in this modality.
+    MISCChannelCount: deprecated
     TriggerChannelCount: recommended
     RecordingDuration: recommended
     RecordingType: recommended
--- a/src/schema/rules/sidecars/eeg.yaml
+++ b/src/schema/rules/sidecars/eeg.yaml
@@ -70,7 +70,14 @@ EEGRecommended:
     ECGChannelCount: recommended
     EMGChannelCount: recommended
     EOGChannelCount: recommended
-    MISCChannelCount: recommended
+    MiscChannelCount: recommended
+    MISCChannelCount:
+      level: deprecated
+      description_addendum: |
+        Deprecated alias of `MiscChannelCount`. Existing datasets remain valid;
+        new datasets SHOULD use `MiscChannelCount`, which matches the MEG and
+        iEEG sidecar rules, the EEG channel-count consistency check, and the
+        documentation example in this modality.
     TriggerChannelCount: recommended
     RecordingDuration: recommended
     RecordingType: recommended
--- a/src/schema/objects/metadata.yaml
+++ b/src/schema/objects/metadata.yaml
@@ -2389,6 +2389,9 @@ MISCChannelCount:
   display_name: Miscellaneous channel count
   description: |
     Number of miscellaneous channels not covered otherwise.
+
+    Deprecated alias of `MiscChannelCount`. New datasets SHOULD use
+    `MiscChannelCount`.
   type: integer
   minimum: 0
 MotionChannelCount:
--- a/src/schema/rules/sidecars/motion.yaml
+++ b/src/schema/rules/sidecars/motion.yaml
@@ -61,7 +61,8 @@ motionRecommended:
     JNTANGChannelCount: recommended
     LATENCYChannelCount: recommended
     MAGNChannelCount: recommended
-    MISCChannelCount: recommended
+    MiscChannelCount: recommended
+    MISCChannelCount: deprecated
     MissingValues: recommended
     MotionChannelCount: recommended
     ORNTChannelCount: recommended


bruAristimunha:
The whole problem emerges from here: https://github.com/bruAristimunha/bids-specification/commit/843f4508018a2a8a3aecef3b26cfee399c3e39c8

bruAristimunha:
@neuromechanist, @sappelhoff and @scott-huberty, can you please review?

effigies (commented):
On the whole, I think unifying the duplicated field makes sense.

```console
$ git grep -i miscchannelcount
src/modality-specific-files/electroencephalography.md:  "MiscChannelCount":0,
src/modality-specific-files/intracranial-electroencephalography.md:  "MiscChannelCount":0,
src/schema/objects/metadata.yaml:MISCChannelCount:
src/schema/objects/metadata.yaml:  name: MISCChannelCount
src/schema/objects/metadata.yaml:MiscChannelCount:
src/schema/objects/metadata.yaml:  name: MiscChannelCount
src/schema/rules/checks/eeg.yaml:MiscChannelCountReq:
src/schema/rules/checks/eeg.yaml:      The MiscChannelCount metadata does not match the number of channels
src/schema/rules/checks/eeg.yaml:    - type(sidecar.MiscChannelCount) != "null"
src/schema/rules/checks/eeg.yaml:    - sidecar.MiscChannelCount == count(associations.channels.type, "MISC")
src/schema/rules/sidecars/eeg.yaml:    MISCChannelCount: recommended
src/schema/rules/sidecars/ieeg.yaml:    MiscChannelCount: recommended
src/schema/rules/sidecars/meg.yaml:    MiscChannelCount: recommended
src/schema/rules/sidecars/motion.yaml:    MISCChannelCount: recommended
```

It looks like we use `MISCChannelCount` in both EEG and motion. I would do the same thing in both cases:

```diff
-    MISCChannelCount: recommended
+    MiscChannelCount: recommended
+    MISCChannelCount: deprecated
```

And update the description for `MISCChannelCount` in `schema/objects/metadata.yml` to indicate that the field is deprecated in favor of `MiscChannelCount`.

bruAristimunha:
hey @effigies!

Many thanks for the quick review. Okay, I simplify, and I update for motion too. I was wondering, can you please review again?

effigies (approved):
LGTM.

bruAristimunha:
Hey @effigies,

I was wondering, can we merge?

effigies:
No, this requires a second review. @bids-standard/bep029 @bids-standard/raw-electrophys-eeg should probably have a look.

src/schema/rules/sidecars/eeg.yaml
```diff
@@ -70,7 +70,8 @@ EEGRecommended:
     ECGChannelCount: recommended
     EMGChannelCount: recommended
     EOGChannelCount: recommended
-    MISCChannelCount: recommended
```
neuromechanist, line 73: I probably see why it was tried to be append `<ChannelType>` with `ChannelCount`, which actually might be a good machine-readable approach to have.

We have `MISC` channel type in many modalities (along with `EEG`, `EMG`, `REF`, `TRIG`, ...) , and the problem might be actually larger. For example, then we should have `TRIGChannelCount` as well, not `TriggerChannelCount` if we want to abide by the above simplicity.

neuromechanist:
Should we make this PR a larger, but focused to ChannelCount vars to make all of them consistent?
@effigies, WDYT?

effigies:
What's the specific proposal?

This PR seems obviously correct on historical grounds (the old validator used, and examples use, "Misc", but we messed up schematization), and I don't want to derail that for something that might need wider discussion. But if what you're proposing is similar in scope and nature, then I'm not inherently opposed to getting it done here.

This is also @bruAristimunha's PR, and he is perfectly welcome to say that he doesn't want to increase the scope.

bruAristimunha:
Hey @neuromechanist!

Super nice suggestion and I would be happy to continue the conversation outside this PR, or even in another PR. 

I think I would prefer to keep this PR really minimal. At the moment, I just would like allow the tools that are writing in bids to be bids compatible again.

robertoostenveld:
I would not object to calling this one `MiscChannelCount` and to leave the wider discussion possibly to BIDS 2.0. I believe we now have certain channel types for which the corresponding `xxxChannelCount` is not mentioned and hence not valid.

neuromechanist:
> What's the specific proposal?

To fix all consistencies all at once, not one at a time!

However, I can also see how it will create problems with tools that have already manually implemented this, and a change would make them incompatible.

Thanks @bruAristimunha for creating the other issue.

neuromechanist (approved):

PR #2400, opened 2026-04-16
[ENH] Recommend controlled vocabulary for age Units, clarify that it can be overloaded

### LINKED ISSUE #1633: Formalize participants' age to actually allow for other than year units/durations

yarikoptic:
Of high relevance to @bids-standard/bep032 effort, but also potentially to @bids-standard/bep036 and others.

- Most recent relevant discussion is in https://github.com/bids-standard/bids-specification/issues/1633#issuecomment-4253350012 
 - Ultimately closes #1633
- Reincarnated  https://github.com/bids-standard/bids-specification/pull/1791 which was partially addressed elsewhere
- and relates to freshly filed https://github.com/bids-standard/bids-validator/issues/386 as we might need to adjust this depending on the solution there. 

TODOs/Notes
- [x] it will make some bids-example datasets invalid since they use "years" for "Units" instead of "year", submitted https://github.com/bids-standard/bids-examples/pull/553
- [x] Review/fix for that test fix.  ~~Unclear why NOW it became applicable/failing~~
   - we ran into this since this is presumably the first rule adding custom checks on the column definition (not sidecar) .json files? (according to claude).  But reworked test selection seems became  too cumbersome
   - no more changes to tests -- validating against .tsv
- [x] For BIDS 2.0 make 'warning' into 'error' there Units not being from allowed values. Added to TODOs of #1775

### DIFF
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -44,7 +44,7 @@ age:
     Using "89+" for ages above 88 is DEPRECATED.
   definition: {
     "LongName": "Subject age",
-    "Description": "Subject postnatal age (by default in years)",
+    "Description": "Subject postnatal age",
     "Format": "number",
     "Units": "year",
     "Maximum": 89,
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -36,13 +36,15 @@ age:
   name: age
   display_name: Subject age
   description: |
-    Numeric value in years (float or integer value).
+    Numeric value (float or integer value).
+    By default expressed using "year" as the unit. Other possible "Units" are described in
+    [Units](SPEC_ROOT/common-principles.md#units).
 
-    For privacy purposes, participant ages should be capped at 89.
+    For privacy purposes, participant ages should be capped at 89 years.
     Using "89+" for ages above 88 is DEPRECATED.
   definition: {
     "LongName": "Subject age",
-    "Description": "Subject age in postnatal years",
+    "Description": "Subject postnatal age (by default in years)",
     "Format": "number",
     "Units": "year",
     "Maximum": 89,
--- a/src/schema/rules/checks/dataset.yaml
+++ b/src/schema/rules/checks/dataset.yaml
@@ -43,6 +43,23 @@ ParticipantIDMismatch:
         sorted(dataset.subjects.sub_dirs)
       )
 
+AgeUnits:
+  issue:
+    code: AGE_UNITS
+    message: |
+      The "Units" value for age in 'participants.json' is not a valid
+      ISO 8601-based duration unit.
+      Allowed values are "year", "month", "week", "day", "hour", "minute",
+      or "second".
+    level: error
+  selectors:
+    - path == '/participants.json'
+    - '"age" in json'
+    - type(json.age) != 'null'
+    - '"Units" in json.age'
+  checks:
+    - intersects([json.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
+
 # 214
 SamplesTSVMissing:
   issue:
--- a/src/schema/rules/checks/dataset.yaml
+++ b/src/schema/rules/checks/dataset.yaml
@@ -51,11 +51,9 @@ AgeUnits:
       ISO 8601-based duration unit.
       Allowed values are "year", "month", "week", "day", "hour", "minute",
       or "second".
-    level: error
+    level: warning
   selectors:
     - path == '/participants.json'
-    - '"age" in json'
-    - type(json.age) != 'null'
     - '"Units" in json.age'
   checks:
     - intersects([json.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
--- a/src/schema/rules/checks/dataset.yaml
+++ b/src/schema/rules/checks/dataset.yaml
@@ -53,10 +53,10 @@ AgeUnits:
       or "second".
     level: warning
   selectors:
-    - path == '/participants.json'
-    - '"Units" in json.age'
+    - path == '/participants.tsv'
+    - '"Units" in sidecar.age'
   checks:
-    - intersects([json.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
+    - intersects([sidecar.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
 
 # 214
 SamplesTSVMissing:
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -1135,9 +1135,16 @@ Describing dates and timestamps:
     for more information.
 
 -   Age SHOULD be given as the number of years since birth at the time of
-    scanning (or first scan in case of multi session datasets). Using higher
-    accuracy (weeks) should in general be avoided due to privacy protection,
-    unless when appropriate given the study goals, for example, when scanning babies.
+    scanning (or first scan in case of multi session datasets).
+    The default unit is `"year"`, but it MAY be overridden in the JSON sidecar
+    (for example, `participants.json`) by setting `"Units"` to one of the
+    following values based on
+    [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) duration
+    designators: `"year"`, `"month"`, `"week"`, `"day"`, `"hour"`, `"minute"`,
+    or `"second"`.
+    Using higher accuracy (for example, weeks or days) should in general be
+    avoided due to privacy protection, unless when appropriate given the study
+    goals, for example, when scanning babies or animals.
 
 ## Directory structure
 


src/schema/rules/checks/dataset.yaml
```diff
@@ -43,6 +43,23 @@ ParticipantIDMismatch:
...
 
+AgeUnits:
+  issue:
+    code: AGE_UNITS
+    message: |
+      The "Units" value for age in 'participants.json' is not a valid
+      ISO 8601-based duration unit.
+      Allowed values are "year", "month", "week", "day", "hour", "minute",
+      or "second".
+    level: error
```
effigies: This is backwards incompatible.


```suggestion
    level: warning
```

src/schema/rules/checks/dataset.yaml
```diff
@@ -43,6 +43,23 @@ ParticipantIDMismatch:
...
+  selectors:
+    - path == '/participants.json'
+    - '"age" in json'
+    - type(json.age) != 'null'
```
effigies: `"Units" in null` is falsey (I believe it should be `null`, but I haven't reread the language spec), so these can be removed. Might be a misguided hope that others won't find the schema intimidating, but I try to keep these simple when possible.

```suggestion
```

src/schema/objects/columns.yaml
```diff
@@ -36,13 +36,15 @@ age:
...
+    By default expressed using "year" as the unit. Other possible "Units" are described in
+    [Units](SPEC_ROOT/common-principles.md#units).
 
-    For privacy purposes, participant ages should be capped at 89.
+    For privacy purposes, participant ages should be capped at 89 years.
     Using "89+" for ages above 88 is DEPRECATED.
   definition: {
     "LongName": "Subject age",
-    "Description": "Subject age in postnatal years",
+    "Description": "Subject postnatal age (by default in years)",
```
effigies: The idea of the `definition` was to be a full column description that you could insert into a `participants.json` file as-is and get a identical meaning and validation. I think updating `objects.columns.age.description` makes a lot of sense, but updating `objects.columns.age.definition.Description` doesn't.

effigies (commented):
Some comments. Overall, I think this is a good push and will help clarity, but needs to be done in a backwards compatible way.

src/schema/rules/checks/dataset.yaml
yarikoptic: rrright. But for BIDS 2.0 we should make it Error. 
I will add label

src/schema/objects/columns.yaml
yarikoptic: I think I follow your interesting point of view and I think I follow it so I am ok to revert "in principle".
But I might actually even suggest  just getting

```suggestion
    "Description": "Subject postnatal age",
```

to avoid "data duplication" since "Units" is the one stating it and we do not encode units in other similar cases AFAIK:

```
❯ git grep -B2 'Units":' -- src/schema/objects/columns.yaml
... removed age here...
--
src/schema/objects/columns.yaml-    "Description": "continuous pulse measurement",
src/schema/objects/columns.yaml-    "Format": "number",
src/schema/objects/columns.yaml:    "Units": "mV"
--
src/schema/objects/columns.yaml-    "Description": "continuous measurements by respiration belt",
src/schema/objects/columns.yaml-    "Format": "number",
src/schema/objects/columns.yaml:    "Units": "mV"
--
src/schema/objects/columns.yaml-    "Description": "continuous measurement of the scanner trigger signal",
src/schema/objects/columns.yaml-    "Format": "number",
src/schema/objects/columns.yaml:    "Units": "arbitrary"
```

and we better then first decide on 

- https://github.com/bids-standard/bids-validator/issues/386

although it seems it is "intended" ATM, at least according to the validator.

Also, is there anything in specification user facing materials which facilitate implementation in users' actions of such an idea?  Because I do not think they would know what to cut/paste really. So far docs just 

- have a 'bad' example in https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html as it does not fully overload the records, e.g. lacking `Format:`

```
    "age": {
        "Description": "age of the participant",
        "Units": "year"
    },
```

where we need to decide on https://github.com/bids-standard/bids-validator/issues/386 first.

- in https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files providing a list of attributes which could be specified, but nothing "copy pasteable"

effigies: Agreed on dropping the mention of years from the description as it duplicates units. And I am pro-fixing the example to include format. At one point we used "Units" to imply numeric types, and it's possible that we still do if there's no `Format`. But explicit is better than implicit.

I reopened https://github.com/bids-standard/bids-specification/issues/2045 to follow up, as I think we should fix this systematically.

yarikoptic:
@effigies I forced pushed rebase on master with tune up the the testing... if you see how to improve it please share (even if an idea). according to claude this is "AgeUnits rule is the first to access json.age in participants.json" and thus we started to hit in this test for sidecar json files.

tools/schemacode/src/bidsschematools/tests/test_expressions.py
```diff
@@ -105,6 +105,24 @@ def walk_schema(schema_obj, predicate):
...
 
 
+def _targets_column_json(selectors):
+    """Check if rule selectors target a column-description JSON file.
+
+    Column-description JSON files (like participants.json, samples.json)
+    have column names as top-level keys, not metadata field names.
+    """
+    # Stems of TSV-companion JSON files defined in rules/files/common/tables.yaml
+    column_json_stems = ("participants", "samples", "scans", "sessions")
```
yarikoptic: yarik dislikes hardcoding like this -- likely we should decide on that based on the schema values if we agree with overall approach here

tools/schemacode/src/bidsschematools/tests/test_expressions.py
```diff
@@ -105,6 +105,24 @@ def walk_schema(schema_obj, predicate):
...
+    for sel in selectors:
+        for stem in column_json_stems:
+            if f"/{stem}.json" in sel:
+                return True
+        # phenotype directory JSONs
+        if "phenotype" in sel:
```
yarikoptic: also feels too broad

tools/schemacode/src/bidsschematools/tests/test_expressions.py
```diff
@@ -118,12 +136,18 @@ def test_valid_sidecar_field(schema_obj):
...
+        # as top-level keys, so json.* may reference column names there.
+        targets_columns = _targets_column_json(selectors)
+
         for expr_type in ("selector", "check"):
             for expr in rule.get(f"{expr_type}s", []):
                 ast = expression.parse_string(expr)[0]
                 for name in find_names(ast):
                     if name.startswith("json."):
-                        assert name.split(".", 2)[1] in field_names, (
+                        allowed = field_or_column_names if targets_columns else field_names
```
effigies: The idea here was that we would never write checks for sidecar files, only data files, and then make queries/assertions on their sidecars. (Root-level and phenotype TSV files are probably the only cases where the inheritance principle doesn't come into play.)

Either way, we're letting `sidecar.*` be loose because of this case, and so I think either you switch to validating the `participants.tsv`'s `sidecar.age`, or we just get rid of this `if/else` and always use `field_or_column_names`. As mentioned in the docstring, typos are the bigger danger than trying to look for column names in a metadata field.

yarikoptic: > or we just get rid of this `if/else` and always use `field_or_column_names`.

that's how I had it originally but felt (and still do) wrong to make test weaker.

> switch to validating the participants.tsv's `sidecar.age`

need more bytes: do you mean somehow in the schema selectors or here in the code?

src/schema/rules/checks/dataset.yaml
```diff
@@ -43,6 +43,21 @@ ParticipantIDMismatch:
...
+      The "Units" value for age in 'participants.json' is not a valid
+      ISO 8601-based duration unit.
+      Allowed values are "year", "month", "week", "day", "hour", "minute",
+      or "second".
+    level: warning
+  selectors:
+    - path == '/participants.json'
+    - '"Units" in json.age'
+  checks:
+    - intersects([json.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
```
effigies: @yarikoptic This is what I mean by validating on participants.tsv:

```suggestion
    - path == '/participants.tsv'
    - '"Units" in sidecar.age'
  checks:
    - intersects([sidecar.age.Units], ["year", "month", "week", "day", "hour", "minute", "second"])
```

yarikoptic:
done, rewritten, force pushed, I think we have converged!

effigies:
Just for the sake of considering alternatives, the ISO 8601 standard for durations is `P47Y1M8D` and such-like. We could allow someone to use `"Format": "ISO8601-Duration"` and define the regex `/P(\d+Y)?(\d+M)?.../` to validate those durations.

That would be using an existing standard instead of creating our own controlled vocabulary inspired by the standard, so I think it's worth considering, if only to record a reason for rejecting it.

My reasons would be:

1) Existing practice is to use decimal durations, and providing a common set of unit names is more practical than suggesting everybody learn to read and write ISO strings.
2) Including it as an optional unit (e.g., `"Units": "ISO8601"`) would be a burden on dataset consumers.

effigies (approved):

yarikoptic:
yeap, search reminded me mentioning it in original
- https://github.com/bids-standard/bids-specification/issues/1633

FWIW, I think adopting standard for durations makes sense. I will extract your comment into an issue

yarikoptic:
@ericearl I would like to invite you to review this PR since somewhat relates to @bids-standard/bep036 and thus could be of interest for feedback or just approval ;) @julia-pfarr also might be of interest to you as a liaison to the  world of all the participants.tsv annotations of neurobagels ;)

src/schema/rules/checks/dataset.yaml
```diff
@@ -43,6 +43,21 @@ ParticipantIDMismatch:
...
+    code: AGE_UNITS
+    message: |
+      The "Units" value for age in 'participants.json' is not a valid
+      ISO 8601-based duration unit.
+      Allowed values are "year", "month", "week", "day", "hour", "minute",
+      or "second".
+    level: warning
+  selectors:
+    - path == '/participants.tsv'
+    - '"Units" in sidecar.age'
```
ericearl, line 57: Should there be a selector here to make sure the sidecar exists first? The spec permits providing a participants.tsv without a participants.json because the participants.json file is only RECOMMENDED.

> https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#participants-file

ericearl (approved):
Find a resolution to my one comment, and it looks good!

src/schema/rules/checks/dataset.yaml
yarikoptic: I would need to ask @effigies for this since I recall he advised to minimize some (may be another) conditioning as it would generally be `falsy` anyways, and might be the case here too.

effigies: If `sidecar` does not exist or is empty, then this results in `False`.

effigies:
Two approvals. This can be merged on/after Thursday, April 23, to ensure there's enough time for people to review since the last significant edit.

julia-pfarr:
Looks good! FYI: Neurobagel uses age in years with different formats (float (31.5), range (30-35), euro (31,5), bounded (30+), iso8601 (31Y6M))

yarikoptic:
> Two approvals. This can be merged on/after Thursday, April 23, to ensure there's enough time for people to review since the last significant edit.

coolio, today is May 07 so let's proceed!

Issue #2402, opened 2026-04-17
Formalize "IndexColumns" for .tsv files


yarikoptic:
> > Eric: I would rather explicitly define the index columns in the JSON header
> 
> Yarik: Please propose it as an alternative proposal to this one then! We might then consider them one by side later.
> 
> Here's an example:
> 
> ```json
> {
>     "IndexColumns": [
>         "participant_id",
>         "session_id",
>         "run_id"
>     ],
>     "participant_id": {
>         ...
>     },
>     ...
>     (all the other usual data dictionary sidecar JSON stuff)
>     ...
> }
> ``` 

 _Originally posted by @ericearl in [#2273](https://github.com/bids-standard/bids-specification/issues/2273#issuecomment-4267997893)_

So proposed as an alternative to 
- https://github.com/bids-standard/bids-specification/issues/2273

yarikoptic:
Overall I like "explicit better than implicit" aspect of such an approach! But immediate cons with my assessment of severity 
 
- (*major*) inability to separate common (e.g. across sessions and runs in the given example) metadata into its own file, and thus encouraging duplication, and **duplication is evil** and avoidance of duplication is inherent to the *foundational* "Inheritance principle" strategy of summarization of common metadata at higher level(s)
- (*medium*) breaks software compatibility which can load current `.tsv` files under assumption of a singular index (and thus unique rows e.g. per `participant_id`)
- (*low*) adds to the mix of columns description entries and now a new "control metadata" (`IndexColumns`, I guess complement to `Columns` of `.tsv.gz`) and that is in addition to other potential metadata we already allow/expect in some cases (ref: https://github.com/bids-standard/bids-2-devel/issues/71; so hopefully would not be an issue in bids 2.0) 
- (*medium*) while we already have `sessions.tsv` etc, *nothing in the name* of `participants.tsv` would suggest that it is a file with content potentially with metadata across participants+sessions+runs.  And the point of BIDS is to provide filenames descriptive to humans! (otherwise we could have just gone into collection of UUID based folders + index table) , and complimentary for machines:
- (*low*) it would require loading .json file to be able to answer if we have a joint index for any operation on .tsv

note that IMHO my proposal in #2273 **does not have any** of those `cons`.

ericearl:
[...]

Issue #2273, opened 2025-12-03
Formalize joint index support in .tsv via {"+".join(entities)}s.tsv files


yarikoptic:
### Your idea

Came up *prominently* in discussion of the BEP036 phenotypic data (attn @bids-standard/bep036) which want to introduce an aggregated demographics table, bringing together information across subjects/sessions and potentially even runs. 

In immediate use case of BEP036 I suggest to formalize allowing for  top level `participant+sessions.tsv` with joint index across `participant_id`, `session_id` pairs.  Then information 
- the same across sessions will go to `participants.tsv` like it does now
- the same across participants but different across sessions - into `sessions.tsv` at the top level to be "taken" per IP principle.
   - @effigies noted complication is that for .tsv we actually do not really define how to inherit .tsv records beyond "take .tsv closest in hierarchy"... I am yet to check.

But I think the issue is common and warrants a more generic solution!  Related desires came up in

- iEEG in BIDS and microephys (BEP032) to need to disambiguate combination of electrodes on probes. iEEG atm just encodes some location in addition to index within "name", thus making it unique 

It is IMHO potentially a general desire to be able to consolidate relevant summary information in the "long form" at higher levels, thus requiring composite indexes. E.g. for an "experimental" composition of OpenNeuro studies I already produced similar [studies_derivatives.tsv]( https://github.com/OpenNeuroStudies/OpenNeuroStudies/blob/master/studies_derivatives.tsv) (to be renamed likely into `study+derivatives.tsv`).

It would "pair" up with summarization inheritance principle that at higher levels, if desired, we can provide common metadata in .json or .tsv files. So ATM `participants.tsv` provide common metadata per participant and `sub-*/sub-*_sessions.tsv` are the ones to provide differing ones across sessions.

## Counter precedents (sneaked in):

- microscopy uses top level [samples.tsv](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#samples-file) which at the top level pairs the two: "The combination of `sample_id` and `participant_id` MUST be unique."

### Flexible joint indexes

By @effigies to do allow multiple columns to serve an index, and define that at the schema level. Details to be provided by @effigies in some other issue/PR? ;)

### Related

- https://github.com/bids-standard/bids-2-devel/issues/69

yarikoptic:
[...]

PR #2123, opened 2025-05-30
[ENH] BEP036 - Phenotypic Data Guidelines

ericearl:
**The BEP leads can meet as-needed to discuss this BEP PR**

Coordinate a meeting by emailing **Eric Earl: [eric.earl@nih.gov](mailto:eric.earl@nih.gov)**.

Communicate on this PR to provide feedback otherwise.

[**HTML preview of this BEP**](https://bids-specification--2123.org.readthedocs.build/en/2123/)

**BEP036** brings guidelines for best tabular phenotypic data to the BIDS specification.

- Includes an appendix called `phenotype.md`
- Includes a new `AdditionalValidation` key for the `dataset_description.json`, for which the usage is described in the modality agnostic files sections
- Includes the new option to use `"IndexColumns"` to define multiple columns as columns to index upon, allowing for aggregation of tabular phenotypic data in the `phenotype/` subfolder.

**Additional Links**

1. [Original Google Doc](https://docs.google.com/document/d/1WTkfES8L0vItZVyyR68fc-9cO03jS-kCnMnw6602pbc/edit?usp=sharing)
2. [Draft BIDS Validator errors and warnings](https://docs.google.com/document/d/1guNgy-E8H7PgvPaaKtD1JLUyiLyfGqJqZaVLvYkfH80/edit?usp=sharing)
3. [BIDS Examples PR](https://github.com/bids-standard/bids-examples/pull/465)
4. [BIDS Validator PR](https://github.com/bids-standard/bids-validator/pull/438)

---------

Co-authored-by: Eric Earl <eric.earl@nih.gov> @ericearl 
Co-authored-by: Samuel Guay <samuel.guay@umontreal.ca> @SamGuay 
Co-authored-by: Sebastian Urchs <sebastian.urchs@mcgill.ca> @surchs 
Co-authored-by: Arshitha Basavaraj <arshitha.basavaraj@iiitb.ac.in> @Arshitha 
Co-authored by: Dustin Moraczewski <dustin.moraczewski@nih.gov> @dmoracze

PR #2403, opened 2026-04-18
[ENH] Add participant+sessions.tsv for session-varying participant metadata

### LINKED ISSUE #1020: Age at session
More of a question first, which may lead to a proposal later.

Is there a consensus on how to represent the concept of an age at session in the BIDS metadata hierarchy?

The context concerns longitudinal studies where imaging visits span a large period of time (several years typically), and the age at session is necessary for modeling disease progression.

On first read, I thought this attribute may go in `sub-<label>_sessions.tsv#age` which would override `participants.tsv#age` as per the inheritance rule. But that contradicts the [rule](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#sessions-file) in the sessions file section mandating the use of distinct attributes between participants and sessions tabular data.

A BIDS-compliant alternative could be to introduce an `age_at_session` attribute in the sessions file.

What are your thoughts?

yarikoptic:
Introduces a single new optional dataset-level file `participant+sessions.tsv` with a composite index `[participant_id, session_id]`.  This provides a single top-level location for metadata that varies across both participants and sessions -- e.g. age at each visit, body weight, clinical scores in longitudinal studies -- complementing the existing `participants.tsv` (participant-constant) and per-subject `*_sessions.tsv` files.

Note that it is already possible to provide such metadata in `sub-*/ses-*_sessions.tsv` file. So such approach just serves the way to provide an "aggregate" collection of metadata.  As such, we might then need to define how it interacts with the inheritance principle, but defining that yet TODO in general for .tsv files.

The `+` in the filename signals a composite index, following the convention proposed in 

- #2273 
- and alternative to freshly proposed #2402 
 
inspired by a work on BEP036 (@bids-standard/bep036):

- https://github.com/bids-standard/bids-specification/pull/2123

Most of the changes are just straightforward interpolation of `participants.tsv` and `sessions.tsv` files definitions.

One of the notable changes is to `meta/context.yaml` where we added `dataset.sessions` (union of all session directories across subjects) to enable session-level validation checks.  I think it is only reasonable given that we did already included dataset level summaries for datatypes and modalities. But it would require bids-validator to support it.  Alternative - is to drop it and that extra check we added.

Ideally though we should figure out how to validate specific combinations of sub/sessions and TODO was left for that.

An example `participant+sessions.tsv` with `body_weight` column for the already `7t_trt` bids-examples dataset is at

- https://github.com/bids-standard/bids-examples/pull/556

where, if you also look into original `participants.tsv`, makes it a little obvious that duplication of all entries across all sessions would be dubious.

- I think overall we can state that it closes #1020 which theoretically could have been closed with original introduction of _sessions.tsv files.

TODOs
- [ ] clarify relationship (summarization AKA formerly inheritance) to the per subject `sub-*/sub-*_sessions.tsv` files. Two possibilities:
   - a. ad-hoc within description of elaborated here "Participant and sessions file" section
   - b. "Common principles/Inheritance" thus pre-empting a solution for more generic #2273 
   - c. both

### DIFF
--- a/src/longitudinal-and-multi-site-studies.md
+++ b/src/longitudinal-and-multi-site-studies.md
@@ -56,6 +56,7 @@ A guide for using macros can be found at
             },
         "sub-control01_sessions.tsv": "",
         },
+    "participant+sessions.tsv": "",
     "participants.tsv": "",
     "dataset_description.json": "",
     "README": "",
@@ -70,6 +71,13 @@ session_id	acq_time	systolic_blood_pressure
 ses-predrug	2009-06-15T13:45:30	120
 ses-postdrug	2009-06-16T13:45:30	100
 ```
+
+In addition to per-subject `_sessions.tsv` files,
+a dataset-level `participant+sessions.tsv` file can aggregate
+metadata that varies across both participants and sessions
+(such as age at each session) into a single file.
+See [Participant and sessions file](modality-agnostic-files/data-summary-files.md#participant-and-sessions-file)
+for details.
 
 See this [example dataset](https://github.com/bids-standard/bids-examples/tree/master/7t_trt)
 that has been formatted
--- a/src/modality-agnostic-files/data-summary-files.md
+++ b/src/modality-agnostic-files/data-summary-files.md
@@ -81,6 +81,63 @@ to date of birth.
 }
 ```
 
+## Participant and sessions file
+
+Template:
+
+```Text
+participant+sessions.tsv
+participant+sessions.json
+```
+
+<!-- This block generates a description.
+A guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___render_text("objects.files.participant_sessions.description") }}
+
+We RECOMMEND to make use of these columns, and
+in case that you do use them, we RECOMMEND to use the following values
+for them:
+
+<!-- This block generates a columns table.
+The definitions of these fields can be found in
+  src/schema/rules/tabular_data/*.yaml
+and a guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_columns_table("modality_agnostic.ParticipantSessions") }}
+
+`participant+sessions.tsv` example:
+
+```tsv
+participant_id	session_id	age	body_weight
+sub-01	ses-predrug	34	81.3
+sub-01	ses-postdrug	35	83.1
+sub-02	ses-predrug	12	45.2
+sub-02	ses-postdrug	13	48.7
+```
+
+It is RECOMMENDED to accompany each `participant+sessions.tsv` file with a sidecar
+`participant+sessions.json` file to describe the TSV column names and properties
+of their values (see also
+the [section on tabular files](../common-principles.md#tabular-files)).
+Such sidecar files are needed to interpret the data, especially so when
+additional columns are defined beyond `age`,
+such as `body_weight` in this example.
+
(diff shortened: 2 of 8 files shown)


dmoracze:
Is the proposed `participants+sessions.tsv` the _only_ place to store data that varies within participant across session?
For example, would a longitudinal dataset that collects demographics and surveys at each timepoint be required to put every item in this file? If so, this file would get out of hand quickly.

Or is `participants+sessions.tsv` a record of joint indices that can be used in other places, say `phenotype/`?

yarikoptic:
Thank you @dmoracze for your question! TL;DR: he full and exhaustive answer is "NO" and  I added TODO to the original description to elaborate on that!

The quick extended answer was alluded to in the original description:

>  complementing the existing `participants.tsv` (participant-constant) and per-subject `*_sessions.tsv` files.

So we **already** have that mechanism and it could be used in cases where more appropriate.

And, although we are yet to actually improve/formalize [inheritance principle](https://bids-specification.readthedocs.io/en/stable/common-principles.html#the-inheritance-principle) for the application to .tsv files, and to this composite indexing in particular , but potentially such metadata could then be present in **both** top level `participants+sessions.tsv` and `sub-.../sub-..._sessions.tsv` file(s) (Hence -- TODO).

dmoracze:
Ah ha, yes that makes sense. I'm both trying to mesh this enhancement with our group's typical usecases and brainstorming a way forward for https://github.com/bids-standard/bids-specification/pull/2123.

We often curate datasets with multiple visits and data types and many times each visit does not contain all data types. Take, for example, two visits where only survey/demographic data is collected, say pre/post drug administration where MR scan sessions take place between the survey sessions. The current spec recommends aggregating these data into `dataset/phenotype`. If we have the same survey given at multiple sessions, I see the following options, none of them satisfactory:

1. Storing all aggregated multi-session survey data within the the proposed `participants+sessions.tsv`. This can make the file very large and unwieldy. 

2. Do not aggregate the data and store within the corresponding subject directory, with the implication that the user will need to aggregate over many tsv files. There are two options here:
    - Store the multi-session survey data in `sessions.tsv`, but then IMO you get the same large and unwieldy files as in 1, except now there is one per subject
    - Store the multi-session survey data in `/dataset/subject/session/phenotype`, but currently the `phenotype` directory is only allowed at the top level

Perhaps this is an edge case in the field as a whole, but we curate many datasets like this. The other option is to allow some type of composite index in TSVs stored in `dataset/phenotype`, as described in #2123. 

That is my current understanding of the situation, let me know if I'm misunderstanding your proposal.

yarikoptic:
There is always a compromise to hit. That is why there is flexibility + "redundancy" (eg. via IP + summarization) available in BIDS to cater a wider range of use cases -- one size fits all might not work.

in "unwieldy", in particular for `sub-.../ses-..._sessions.tsv` file you mean -- too wide (too many columns)?  

> Store the multi-session survey data in /dataset/subject/session/phenotype, but currently the phenotype directory is only allowed at the top level

thanks for bringing it up too!  I recall having some discussions or "brainstorming" toward generalization and allowing for "symmetry" of  datatype folders (could also apply to `stimuli/` as e.g. initiated in good old #751 etc) . In part it relates to #1809 but I would not go there ATM ;)  I am yet to get a fresh look at #2123 before our meetup with @ericearl next week, so I will not comment on that relation yet.

yarikoptic:
re unwildly again, since I am on a mission (I feel like) of promotion -- have you seen/tried https://www.visidata.org ?

yarikoptic:
since I do not think it is something which should be merged without more of extended discussion -- moved to draft "for protection"

PR #2404, opened 2026-04-20
[SCHEMA] Support structured survey data as a BIDS modality

karl-koschutnig:
## Summary

This PR proposes adding schema support for structured instrument-based survey data as a valid BIDS representation, organized in the same subject-, session-, and run-resolved way as other BIDS modalities.

The goal is not to replace aggregated tabular phenotypic data. It is to complement them with a canonical acquisition-facing structure that preserves provenance, timing, and instrument context, while still allowing aggregated tables to be generated later when needed.

A working reference implementation already exists in [PRISM Studio](https://github.com/MRI-Lab-Graz/prism-studio), with documentation at [prism-studio.readthedocs.io](https://prism-studio.readthedocs.io/).

## Rationale

BIDS is strongest when its canonical structures reflect how data are actually acquired. For imaging, physio, events, and other modalities, the normative pattern is a subject-resolved structure first, with higher-level summaries and
derivatives produced later.

Instrument-based phenotypic data fit this same pattern. Treating `phenotype/` as the only primary home for those data flattens acquisition context at the point where BIDS usually preserves it.

A structured survey modality would:

- preserve provenance and administration context,
- support repeated assessments across sessions and runs,
- keep sidecar metadata next to the corresponding data files, and
- still allow aggregated phenotype tables to be generated later.

That direction is structurally stronger because aggregated tables can be written from a structured survey layout with little ambiguity, whereas reconstructing the original structure from only an aggregate table often requires extra
assumptions.

## Proposed Changes

This PR implements a minimal but complete first pass:

| File | What changed |
|---|---|
| `src/modality-specific-files/survey.md` | New documentation page: file naming conventions, sidecar structure, instrument-level and item-level field tables, worked dataset structure and JSON/TSV examples |
| `src/schema/objects/datatypes.yaml` | Added `survey` datatype |
| `src/schema/objects/modalities.yaml` | Added `survey` modality |
| `src/schema/objects/suffixes.yaml` | Added `survey` suffix |
| `src/schema/rules/files/raw/survey.yaml` | New file rule using `task` entity template; `.tsv` and `.json` extensions |
| `src/schema/rules/modalities.yaml` | Mapped `survey` modality → `survey` datatype |
| `mkdocs.yml` | Added navigation entry |

## Example Structure

Below is the subject-, session-, and run-resolved structure this PR supports. A root-level sidecar applies to all matching files via the [Inheritance Principle](https://bids-specification.readthedocs.io/en/stable/common-principles.html#the-inheritance-principle), avoiding duplication across sessions and runs. Per-file sidecars remain valid when instrument versions or languages differ between sessions.

```
study/
├── dataset_description.json
├── participants.tsv
├── task-pss_survey.json          ← shared sidecar via inheritance
├── sub-01/
│   ├── ses-baseline/
│   │   └── survey/
│   │       └── sub-01_ses-baseline_task-pss_run-01_survey.tsv
│   └── ses-week04/
│       └── survey/
│           ├── sub-01_ses-week04_task-pss_run-01_survey.tsv
│           └── sub-01_ses-week04_task-pss_run-02_survey.tsv
└── phenotype/
    └── pss_summary.tsv           ← optional aggregated downstream view
```

The `phenotype/` directory in this example is deliberate. It shows that aggregated outputs remain compatible with this proposal, but they are downstream views of structured data rather than the only canonical form.

## Reference Implementation

This proposal is grounded in an existing toolchain rather than a purely abstract design discussion:

- [PRISM Studio repository](https://github.com/MRI-Lab-Graz/prism-studio)
- [PRISM Studio documentation](https://prism-studio.readthedocs.io/)
- [Survey import template](https://github.com/MRI-Lab-Graz/prism-studio/blob/main/docs/examples/survey_import_template.xlsx)
- [Survey schema and modality schemas](https://github.com/MRI-Lab-Graz/prism-studio/tree/main/app/schemas)
- [Template validation documentation](https://github.com/MRI-Lab-Graz/prism-studio/blob/main/docs/TEMPLATE_VALIDATION.md)
- [Dataset validation documentation](https://github.com/MRI-Lab-Graz/prism-studio/blob/main/docs/VALIDATOR.md)

These references show that the proposed structure is already practical for curation, metadata authoring, template-based reuse, and validation.

## Scope

This PR does not propose removing aggregated phenotype tables. It proposes that BIDS should also recognize a canonical modality-style structure for instrument-based phenotypic data, especially when those data are acquired repeatedly across sessions and runs.

## Questions for Reviewers

- Should instrument-based phenotypic data be representable in a subject-, session-, and run-resolved modality structure in the schema?
- If so, should `phenotype/` remain an optional aggregate or derived representation rather than the only primary representation?
- Is `survey` the right directory and suffix label, or should the working group prefer another term such as `pheno`, `assess`, `form`, `inst`, or `meas`?
- The current sidecar spec uses the `task` entity to identify the instrument (e.g. `task-pss`). Should a dedicated `instrument` entity be considered instead?
- The sidecar fields (`TaskName`, `OriginalName`, `StimulusType`, `Respondent`, etc.) follow existing BIDS conventions — are there missing fields, or any that should move between REQUIRED and OPTIONAL?

### DIFF
--- a/src/modality-specific-files/survey.md
+++ b/src/modality-specific-files/survey.md
@@ -1,0 +1,145 @@
+# Survey data (instrument-based phenotypic)
+
+Survey data files contain responses collected from standardized questionnaires or
+assessment instruments.
+They are organized in the same subject- and session-resolved directory hierarchy
+as other BIDS modalities, with one `.tsv` data file per instrument administration
+and a corresponding `.json` sidecar that documents the instrument and its items.
+
+This representation complements the aggregated `phenotype/` tables described in
+[Phenotypic and assessment data](../modality-agnostic-files/phenotypic-and-assessment-data.md).
+Aggregated phenotype tables MAY be generated from subject-resolved survey data
+and stored under `phenotype/`,
+but the subject-resolved structure is the primary canonical form.
+
+<!--
+This block generates a filename templates.
+The inputs for this macro can be found in the directory
+  src/schema/rules/files/raw
+and a guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_filename_template("raw", datatypes=["survey"]) }}
+
+Survey files MUST be stored in a `survey` subdirectory of the subject (and
+optionally session) directory.
+The `task` entity identifies the instrument (for example, `task-pss` for the
+Perceived Stress Scale).
+The `run` entity MAY be used when the same instrument is administered more than
+once within a session.
+
+## Sidecar JSON (`*_survey.json`)
+
+Each survey `.tsv` file MUST be accompanied by a JSON sidecar with the same
+filename stem.
+The sidecar documents the instrument-level metadata and the individual items.
+
+### Instrument metadata
+
+The following top-level fields describe the survey instrument and its
+administration context.
+
+| Field | Requirement | Description |
+| --- | --- | --- |
+| `TaskName` | REQUIRED | Short label identifying the instrument as used in this dataset (MUST match the `task` entity value). |
+| `OriginalName` | REQUIRED | Full canonical name of the instrument. |
+| `StimulusType` | REQUIRED | MUST be `"Questionnaire"`. |
+| `FileFormat` | REQUIRED | MUST be `"tsv"`. |
+| `Language` | REQUIRED | Language code for the administered version (for example, `"en"`, `"de-AT"`). |
+| `Respondent` | REQUIRED | Who provided the responses (`"self"`, `"clinician"`, `"parent"`, or similar). |
+| `ShortName` | OPTIONAL | Common abbreviation of the instrument (for example, `"PSS-10"`). |
+| `Version` | OPTIONAL | Version or edition of the instrument. |
+| `Authors` | OPTIONAL | Array of instrument authors. |
+| `DOI` | OPTIONAL | DOI for the instrument or its primary validation paper. |
+| `License` | OPTIONAL | License under which the instrument may be used. |
+| `Construct` | OPTIONAL | Psychological or clinical construct measured (for example, `"perceived stress"`). |
+| `Instructions` | OPTIONAL | Instructions given to the participant. |
+| `AdministrationMethod` | OPTIONAL | How the instrument was administered (`"online"`, `"paper"`, `"interview"`, `"phone"`, or `"mixed"`). |
+| `SoftwarePlatform` | OPTIONAL | Software used to administer the instrument (for example, `"LimeSurvey"`, `"REDCap"`). |
+
+### Item-level metadata
+
+Any top-level key in the sidecar that is not one of the instrument-level fields
+listed above is treated as a column-level variable definition (for example, `"Q01"`,
+`"item_1"`).
+
+| Field | Requirement | Description |
+| --- | --- | --- |
+| `Description` | REQUIRED | The exact text of the question or item. |
+| `Levels` | OPTIONAL | Mapping of response values to their labels (for example, `{"0": "Not at all", "4": "Very often"}`). |
+| `MinValue` | OPTIONAL | Minimum expected value. |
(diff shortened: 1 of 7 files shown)


src/modality-specific-files/survey.md
```diff
@@ -0,0 +1,150 @@
...
+but the subject-resolved structure is the primary canonical form.
+
+<!--
+This block generates a filename templates.
+The inputs for this macro can be found in the directory
+  src/schema/rules/files/raw
+and a guide for using macros can be found at
+ https://github.com/bids-standard/bids-specification/blob/master/macros_doc.md
+-->
+{{ MACROS___make_filename_template("raw", datatypes=["survey"]) }}
```
yarikoptic, line 22: quick one (edited/expanded 2026/05/12):
Overall, I like this approach but immediate question -- is there semantical difference between "phenotype" and "survey" in this PR?  if not, I would have preferred to stay consistent and make it `phenotype/` (or `pheno/`) (I think I expressed smth like that elsewhere TODO: find refs here).  It would then be a consistent principle we already have overall and distilling more for BEPs, e.g. for `stimuli/` (@bids-standard/bep044 ; TODO: find/add refs). See:

- #2022
  - [rendered section on `stimuli/`](https://bids-specification--2022.org.readthedocs.build/en/2022/modality-agnostic-files/stimuli.html#file-organization)
- #1809    
- #2283 

and potentially others, one way or another hinting on it, e.g

- #1828

Then, if there is a desire to generalize "phenotype" into "survey" -- could be done for bids 2.0... WDYT?

effigies:
I'm curious why this isn't just a `beh.tsv` file? Apologies for the brevity, but the `participant_id` column is redundant with the file name, and `beh.tsv` would allow the data that is being described in this PR.

The advantage to a new suffix/datatype would be that you can add requirements, but BEP36 does demonstrate a way to opt in to novel requirements as well. As a very minimally thought out counterproposal:

```
sub-01/ses-baseline/phenotype/sub-01_ses-baseline_task-pss_run-01_beh.tsv
```

This reuses the existing datatype `phenotype`, entity `task` and suffix `beh`. If we do want to go that way, we would probably want to update BEP36 to use `task`/`beh` instead of introducing `tool`/`phenotype`. Alternately, this PR could adopt BEP36 terms and use:

```
sub-01/ses-baseline/phenotype/sub-01_ses-baseline_tool-pss_run-01_phenotype.tsv
```

karl-koschutnig:
@effigies Thanks, this gets to the central design question.

The motivation for this PR is that questionnaires and structured assessments are a regular primary data type in many studies, but BIDS currently represents them mainly as aggregated dataset-level phenotypic tables. That differs from how BIDS represents other acquired data: the canonical records are subject-, session-, and run-resolved, with metadata colocated with the data.

I am proposing survey as an additional modality for that reason. A survey administration is not necessarily behavioral data produced by an experimental task, and may have no connection to MRI, EEG, physiology, or another acquisition. For example, a participant may complete the PSS at baseline and follow-up, or several times within one visit, independently of any task. The task-pss label in the current proposal identifies the instrument, not an experimental task.

A subject/session/run-resolved survey record also carries instrument-specific provenance that is central to interpreting responses: the instrument and version/form, item definitions and response scales, language/translation, respondent, administration method/platform, and potentially different item sets or scale ranges across administrations.

phenotype/ remains important, but I see it as the aggregated view of these records, analogous to a downstream tabular representation. PRISM already supports that conversion for compatibility. The point of this proposal is that aggregation should not be the only canonical BIDS representation for questionnaire data.

I agree that participant_id is redundant in a one-participant file and should be removed from the example.

I am less convinced by beh.tsv: it can technically encode the values, but it suggests behavioral output from a task, while the use case here is an independently administered questionnaire or assessment. Is there an established BIDS interpretation of beh that includes these non-task instrument administrations? If so, I would like to understand that precedent; otherwise, I think a distinct survey modality better communicates the data’s role.

PR #2409, opened 2026-04-21
[ENH] Allow rawbids/ in derivative datasets for the raw BIDS source

yarikoptic:
PR #2191 added rawbids/ to the study DatasetType but not to derivative, even though the same convention -- "rawbids/ holds the raw BIDS dataset" -- is equally applicable when a standalone derivative dataset includes its raw source.  I think everyone was just too tired and ready to move on with their lives, thus missed that omission, although the wise @effigies [said in his final word](https://github.com/bids-standard/bids-specification/pull/2191#issuecomment-4225841301)

> Finally, for the sake of consistency, rawbids/ is permitted for derivative datasets, while it is not meaningful for raw datasets.

but we (attn @julia-pfarr @nikhil153) all missed that it was not defined for the `derivative` datasets.  Here me and claude are addressing that:

- Add rawbids as an optional opaque subdirectory of the derivative DatasetType in src/schema/rules/directories.yaml.
- Update the derivative example in common-principles.md to use rawbids/ instead of sourcedata/raw/.
- Clarify that rawbids/ is reserved for raw BIDS datasets in both study and derivative cases, and that derivatives of derivatives MUST place their source derivative under sourcedata/ (not rawbids/).

Marking it for the same milestone as the #2191

### DIFF
--- a/src/common-principles.md
+++ b/src/common-principles.md
@@ -369,9 +369,14 @@ Derivatives can be stored/distributed in two ways:
     dataset is provided with read-only access, for publishing derivatives as
     independent bodies of work, or for describing derivatives that were created
     from more than one source dataset.
-    The `sourcedata/` subdirectory MAY be used to include the source dataset(s)
-    that were used to generate the derivatives.
-    Likewise, any code used to generate the derivatives from the source data
+    If the source is a raw BIDS dataset,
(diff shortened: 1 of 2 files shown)


julia-pfarr (approved):

effigies (approved):
LGTM. I think  we should give it the usual week before merging, since there's a reasonable amount of text that people might want to discuss.

yarikoptic:
was 2 weeks ago, week passed, let's proceed!

PR #2441, opened 2026-06-04
[ENH] Support for concurrently-acquired data with scans.tsv

### LINKED ISSUE #2362: Explicitly support concurrently-acquired data with scans.tsv
# Context and motivation

We are proposing minor changes to the BIDS specification which would better facilitate and document instances of concurrently acquired data.

An example use case is the simultaneous acquisition of EEG, fMRI, and physiological files. Currently, concurrent physiological data are stored in the same file as the modality with which it has been acquired, but in the case of an EEG-fMRI acquisition with cardiac and respiratory time series, the physiological data could be stored in either the EEG or fMRI directory. While the naming of these data could suggest their concurrency, in the case of a repeated task there is no obvious way to distinguish an fMRI scan followed by an EEG scan from a simultaneous fMRI-EEG scan. The physiological data would in turn be assumed to be concurrent with whichever modality they accompany, rather than both modalities.

It seems clear that handling such concurrency - while possible using various metadata and scan agnostic files within the current BIDS specification - raises ongoing confusion within this community. For example, see this full but possibly not exhaustive list of previous discussions on how to organize or denote simultaneous data:

- https://github.com/bids-standard/bids-specification/issues/86
- https://github.com/bids-standard/bids-specification/issues/836
- https://github.com/bids-standard/bids-specification/issues/1324
- https://github.com/bids-standard/bids-specification/issues/1796#issuecomment-2074142796
- https://github.com/bids-standard/bids-specification/issues/2296 

# Proposed changes

Following previous discussions on how to indicate whether files were acquired concurrently (see issues linked above), it seems that clearly defined use of the scans.tsv is the best option for addressing this issue across all modalities where it might arise. While such use of scans.tsv is already technically possible within the current BIDS specification, several changes to the current specification would make this practice more accessible and reproducible. We therefore propose the following:

1. Adding support for instances of concurrency in BIDS documentation
2. Changing requirement levels for scans.tsv
3. Using ‘duration’ column to facilitate machine readability for identifying concurrent files in scans.tsv

## 1. Adding support for instances of concurrency in BIDS documentation

We propose explicitly providing guidance for using scans.tsv to reflect the relative timing of scans in cases of concurrently acquired data for different modalities.

We note that the naming of physiological files and their inclusion within another modality’s directory implies concurrency in the current BIDS specification; however, we note that we propose defining physiological data as its own modality in BEP 045 (#2267). Thus, the treatment of these data would not be different to other modalities (e.g. fMRI, EEG, fNIRS, motion) with the changes which we propose in points 2 & 3.

Since the temporal alignment of files typically requires more accuracy than a simple start point, the documentation could fruitfully note that scans.tsv is better used to *find* concurrent files than to *align* concurrent files, and that it is RECOMMENDED to record ongoing markers of timing (e.g. triggers) within modalities to be used to align concurrent data.

## 2. Changing requirement levels for scans.tsv

The purpose of scans.tsv is to describe timing and other properties of each recording file within one session; it is however OPTIONAL, as is the inclusion of the acq_time column. We propose making the scans.tsv file generally RECOMMENDED. In the case of concurrent data, we suggest making the scans.tsv REQUIRED, alongside REQUIRED use of the acq_time column and REQUIRED use of the duration column (see following section). 

This will remove ambiguity in whether or not files from different modalities have been acquired concurrently, and is minimally burdensome to specify. It also has the advantage of providing information about scan order which is useful (e.g. for assessing reproducibility or changes in participant state) even in cases of nonconcurrent data. We further note that this could be useful for cases where documenting cross-participant concurrency may be relevant, i.e. hyperscanning.

## 3. Adding a ‘duration’ column to facilitate machine readability for identifying concurrent files in scans.tsv

Although the use of acq_time will clarify the relative onset of files listed in scans.tsv, it does not provide a simple way to check whether files are concurrent with each other. Thus, we suggest defining a REQUIRED duration column, either by extending the definition of the current [duration](https://bids-specification.readthedocs.io/en/stable/glossary.html#duration-columns)  object to reflect the entire duration of a scan file, or by adding a new sense definition for this term. For files which are too short to specify a duration of acquisition, a duration of zero may be specified (as in the current duration usage).

An alternative, if seen as desirable, could be the definition of an acq_end column.

# Final points

One potential complication introduced by this approach is the duplication of information: if the duration of a file as defined in scans.tsv does not match the length of a file as defined in its metadata. We suggest that the lowest level of metadata be regarded as most reliable. 

We note that the usage of [StartTime](https://bids-specification.readthedocs.io/en/stable/glossary.html#starttime-metadata) to synchronize files presents a similar complication. To keep this discussion modality-agnostic we have not fully explored its usage or potential adaptation here, although its ongoing ambiguity in the problem of relating multiple data files has motivated our characterization of this issue.

We also note that the usage of scans.tsv in the instance of [split files](https://bids-specification.readthedocs.io/en/stable/appendices/entities.html#split) should be examined for compatibility with the proposed changes.

# Moving forward

This would require changes to the BIDS validator. From conversation with more experienced BIDS-ers (@neuromechanist), it is our understanding that a conditional requirement for the scans.tsv file is feasible. However, if this is not the case, we suggest downgrading the instances we suggest a usage should be REQUIRED to instead be RECOMMENDED, with clear documentation of how to handle concurrency in the main specification.

—
This suggestion was formulated in the context of [BEP 045](https://github.com/physiopy/bids-specification-physio/discussions/1), in conversation with [@me-pic @smoia @ShiyuBelleWang @mrikasper @Sourav-Kulkarni @neuromechanist] and following many physio-centric framings of this issue (https://github.com/physiopy/bids-specification-physio/issues/8; https://github.com/physiopy/bids-specification-physio/issues/23; https://github.com/physiopy/bids-specification-physio/issues/42)

Thank you in advance for your feedback!

m-miedema:
Improves guidelines for using the scans.tsv file to identify simultaneously-acquired data. This PR closes #2362, following helpful discussion with @neuromechanist. 

**Notable changes:**
- Adds documentation for usage of scans.tsv file in cases where multiple modalities or recordings are simultaneously acquired.
- Facilitates the use of the `duration` column along with `acq_time` to identify overlapping recordings by changing the definition of `duration` to explicitly cover recording length as well as event length
- Changes requirement levels of `scans.tsv`, `acq_time`, and  `duration` from OPTIONAL to REQUIRED when multiple recordings are acquired concurrently.

**Comments:**
In proposing to make `scans.tsv` conditionally REQUIRED, we (i.e. the BEP 045 working group) acknowledge the additional complexity that is introduced for validation. However, we see the requirement of `scans.tsv` files as minimally burdensome for BIDS users and highly rewarding to facilitate automated processing of multimodal data. We are happy to discuss alternative approaches should this be deemed infeasible.

An example implementing these changes can be found [here](https://github.com/physiopy/bids-examples/tree/bep045_examples/physio_rest_fmri).

Thank you in advance for your attention and feedback!


yarikoptic:
Woohoo @m-miedema -- it had been only a few months since you submitted this little PR!  So it must be a hot topic and "great minds think alike!" since I decided to do half of the job you did here in a dedicated PR:

- #2508

since I think 'duration' is as useful as `acq_time` and regardless either there is concurrent, overlapping or sequential acquisition.  Then, if BIDS gods bless it faster than yours, your PR could reduce to succinct depiction of comments/instructions for concurrent acquisition.

Note that this PR needs a little TLC to make CI green ! Have time to finish it up?

PR #2466, opened 2026-07-10
[FIX] Replace non-existent `len` with existent `length`

psadil:
Closes https://github.com/bids-standard/bids-schema/issues/13.

(And sorry if branching from master was not the right choice! Happy to rebase if needed)


effigies:
Yes, please rebase off the target branch. It will get merged into master.

effigies (approved):

PR #2467, opened 2026-07-16
[SCHEMA] Warn when *_beh.tsv contains onset and duration

### LINKED ISSUE #2213: Throw warning if `*_beh.tsv` has `onset` and `duration` column
A file with those columns should probably be an `events.tsv` in this case.

Issue raised in this PR
https://github.com/bids-standard/bids-validator/pull/1367#issue-1029095653

viranovskaya:
## Summary

Add a schema warning when a `*_beh.tsv` file contains both `onset` and `duration`. A file containing only one of these columns does not trigger the warning.

The check uses direct column access because `columns` is exposed as a mapping in the current validator, so `intersects(columns, ...)` does not inspect the column names.

Closes #2213

## Checks

- `yamllint` passes
- full `bidsschematools` test suite: 81 passed, 1 skipped
- checked with the current Deno validator against five small cases: both columns, onset only, duration only, neither, and a valid `events.tsv`
- checked two existing BIDS example datasets for false positives

### DIFF
--- a/src/schema/rules/checks/events.yaml
+++ b/src/schema/rules/checks/events.yaml
@@ -23,8 +23,8 @@ BehavioralOnsetAndDuration:
   issue:
     code: BEH_ONSET_DURATION
     message: |
-      A '*_beh.tsv' file contains both 'onset' and 'duration' columns.
-      Rename this file to use the '*_events.tsv' suffix.
+      This behavioral file contains onset and duration columns.
+      Consider renaming to use the `events` suffix.
     level: warning
   selectors:
     - suffix == "beh"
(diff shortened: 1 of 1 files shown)


src/schema/rules/checks/events.yaml
```diff
@@ -19,6 +19,19 @@ EventsMissing:
   checks:
     - '"events" in associations'
 
+BehavioralOnsetAndDuration:
+  issue:
+    code: BEH_ONSET_DURATION
+    message: |
+      A '*_beh.tsv' file contains both 'onset' and 'duration' columns.
+      Rename this file to use the '*_events.tsv' suffix.
```
effigies:
```suggestion
      This behavioral file contains onset and duration columns.
      Consider renaming to use the `events` suffix.
```

effigies (approved):
LGTM.

effigies:
@Remi-Gau Bugging you for a review since this is in response to your issue.

Remi-Gau (approved):

PR #2468, opened 2026-07-17
[FIX][SCHEMA] Phenotype subjects are a subset of participants.tsv

effigies:
The existing check required a superset relation. Probably not frequently hit because equality will be the norm.

Note that the error message implies a subset relation, and that was the intent in #2044.

> A phenotype/ .tsv file lists subjects that were not found in the participant_id column found in the participants.tsv file.

Bug introduced in 1ff73de71f06191b131dc3bc34ea38b1ecdb19b6 when moving away from an aggregated `dataset.subjects.phenotype` field.

### DIFF
--- a/src/schema/rules/checks/phenotype.yaml
+++ b/src/schema/rules/checks/phenotype.yaml
@@ -14,8 +14,9 @@ PhenotypeSubjectsMissing:
     - suffix == '.tsv'
     - type(dataset.subjects.participant_id) != 'null'
   checks:
+    # A∩B == A iff A⊆B
     - |
       allequal(
         sorted(intersects(columns.participant_id, dataset.subjects.participant_id)),
-        sorted(dataset.subjects.participant_id)
+        sorted(columns.participant_id)
       )


psadil (commented):
lgtm regarding this change now expressing a subset. But, doesn't the message imply a set difference? I interpret the message as being equivalent to "The phenotype/ .tsv file contains subjects absent from participants.tsv (participant_id column)." Should it instead be something like "The phenotype/ .tsv file contains some (possibly all) participants listed in participants.tsv (participant_id column)."

viranovskaya:
I think the existing diagnostic is consistent with the corrected relation. The validity condition is P_phenotype ⊆ P_participants, while the failure condition is P_phenotype \ P_participants ≠ ∅. So saying that the phenotype TSV contains subjects absent from participants.tsv describes the error correctly.
The proposed alternative (“contains some or all participants listed in participants.tsv”) would also describe a valid file.

“A phenotype TSV contains participant_id values that are absent from participants.tsv.”
It may also be worth testing both directions explicitly, since equality would hide the original reversal: {01,02} ⊆ {01,02,03} should pass, whereas phenotype {01,03} against participants {01,02} should fail.

psadil:
> I think the existing diagnostic is consistent with the corrected relation. The validity condition is P_phenotype ⊆ P_participants, while the failure condition is P_phenotype \ P_participants ≠ ∅. So saying that the phenotype TSV contains subjects absent from participants.tsv describes the error correctly. The proposed alternative (“contains some or all participants listed in participants.tsv”) would also describe a valid file.

Ah, of course! Thanks for spelling that out.

effigies:
Thanks for having a look, both!

PR #2469, opened 2026-07-17
[INFRA] Skip link check for encodeproject.org URLs

effigies:
Captcha added, so links cannot be visited by an automated tool.

LLM scraping has a wide blast radius, and maybe not every web page needs to be dynamically generated.


effigies:
@bendhouseart I'm going to take your support as read and merge to get other PRs to have green checks.

PR #2470, opened 2026-07-17
[INFRA] Add Bluesky social link, drop Twitter

effigies:
We squat a twitter handle, but do not post, so we should stop advertising it. We do post updates to bluesky.


tsalo (approved):

effigies:
This was also merged on the website, and has no impact on the spec, so I will waive a second review. Thanks @tsalo!

PR #2479, opened 2026-07-28
enh(metaschema): require and constrain `inherit` in associations + provide it for atlas _description.json

yarikoptic:
Enforce that every entry under `meta.associations` declares `inherit`, and that its value is `true` when the target extension is `.json`, `.tsv`, `.bval`, or `.bvec` and `false` otherwise. Also add the previously-missing `inherit: true` to `atlas_description`.

This is an alternative to a pure human product of
- #2478

### DIFF
--- a/src/metaschema.json
+++ b/src/metaschema.json
@@ -47,8 +47,43 @@
                   "type": "boolean"
                 }
               },
-              "required": ["target"],
-              "additionalProperties": false
+              "required": ["target", "inherit"],
+              "additionalProperties": false,
+              "if": {
+                "properties": {
+                  "target": {
+                    "properties": {
+                      "extension": {
+                        "oneOf": [
+                          {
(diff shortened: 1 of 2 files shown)


src/metaschema.json
```diff
@@ -47,8 +47,43 @@
                   "type": "boolean"
                 }
               },
-              "required": ["target"],
-              "additionalProperties": false
+              "required": ["target", "inherit"],
```
effigies:
```suggestion
              "required": ["selectors", "target", "inherit"],
```

effigies (approved):
LGTM. I pulled locally and verified that switching a JSON associated file to `inherit: false` triggered a schema validation error, so this does do what it is supposed to.

Waiving second review, as this has no impact on spec rendering or semantics.

PR #2480, opened 2026-07-29
[ENH] BEP032: Require the sample entity for icephys and enforce samples.tsv

bendichter:
Implements the decisions we reached in the working group meeting on 2026-07-29 about the `sample-<label>` entity, addressing https://github.com/bids-standard/bids-specification/pull/2307#discussion_r3539661198. This targets the `bep032-review` branch, so it merges into #2307 rather than into master.

## Requirement Level

The entity is now REQUIRED for `icephys` and remains OPTIONAL for `ecephys`. We agreed that an intracellular recording is always made from a discrete sample, so the requirement holds even when a subject yields only one sample, and even when the recording is made in vivo, since the patched cell is itself the sample. Requiring it in the single-sample case also avoids having to rename files when a second sample is added later. This mirrors how microscopy treats the entity.

In the schema this meant splitting the `microephys` file rule, since it previously covered both datatypes in one block with `sample: optional`. The `icephys` rule now inherits from the `ecephys` rule and overrides the datatype, suffix, and requirement level, following the pattern already used for `photo__micr`.

## Documentation

The entity was previously introduced only in the Photos section, framed as being for histological photos, even though the intracellular examples use it in every filename. It now has a dedicated section near the top of the page covering the requirement levels, the granularity convention, and the relationship to `samples.tsv`. The mention in the Photos section is now a back-reference to that section.

On granularity, the text states that a sample is the finest unit of biological material from which signals are measured: the individual cell for patch clamp, and typically the slice or cell population for in vitro extracellular recordings. Where several samples share a source, for example several cells patched from one slice, the source may also be described in `samples.tsv` and linked with `derived_from`. Such a parent sample generally does not appear as an entity in recording filenames, since it is not itself what was measured.

## Enforcing `samples.tsv`

The specification already states, in the description of the samples file, that `samples.tsv` is REQUIRED if `sample-<label>` appears in any filename in the dataset. The only check enforcing that was keyed on `"micr" in dataset.modalities`, so it never fired for microephys data. Following Sylvain's review, the check is now keyed on the presence of the entity instead, which covers `micr`, `icephys`, and `ecephys` with a single rule and matches the spec text exactly. The error is now reported per file rather than once at the dataset level.

## Verification

`bidsschematools` tests pass (81 passed, 1 skipped), as do `pre-commit` on the changed files, `npm run remark`, and the Latin-phrase check. I confirmed the rendered filename templates show `_sample-<label>` as required for `icephys` and optional for `ecephys`, and that every `icephys` file in the existing intracellular example already carries the entity, so that example remains valid.

@lzehl, this is the first of the items from the meeting. Happy to adjust the wording of the new section if it does not match what we discussed.


SylvainTakerkart:
Hi @bendichter , I can review this PR if that's ok with you...

bendichter:
@SylvainTakerkart that would be great!

bendichter:
@SylvainTakerkart are you still interested in reviewing this PR?

lzehl:
@bendichter reads well. For your request (*Please also check whether the sample_type vocabulary from #2484 covers the openMINDS cases you had in mind.*) I think #2484 integrates for now all we need. Minor comment for #2484 `derived_from` should it not state (?): 


```
"derived_from": {
        "Description": "sample_id or subject_id from which the sample is derived"
```

SylvainTakerkart:
> @SylvainTakerkart are you still interested in reviewing this PR?

sorry, I was off for a few weeks... I'm on it!

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -37,6 +37,37 @@ and analysis pipelines. The terms are established and used in [Neurodata Without
...
+to the samples derived from it using the `derived_from` column.
+Such a parent sample generally does not appear as a `sample-<label>` entity in recording filenames,
+since it is not itself what was measured.
+
+Whenever the `sample-<label>` entity is present in a filename, the dataset MUST include a
+[`samples.tsv` file](../modality-agnostic-files/data-summary-files.md#samples-file) describing each sample.
+This applies to both `icephys` and `ecephys` data.
+See the [Intracellular Electrophysiology example](#intracellular-electrophysiology-patch) below
+for a dataset organized this way.
+
```
SylvainTakerkart, line 70: Hi, the added text is clear and makes everything that was decided in the last group meeting explicit, which is great!

But I think most of this is not specific to `icephys` (except the reference to the example, which are useful, and should be kept here of cours), and I believe there would be an added value to also makes thing explicit for `micr` (because for now, it's not the case and the text is a bit scarse: https://bids-specification.readthedocs.io/en/stable/modality-specific-files/microscopy.html#filename-entities). WDYT? What would be the best way so that this text directly applies to all modalities that use the sample entity?

src/schema/rules/checks/dataset.yaml
```diff
@@ -72,6 +72,22 @@ SamplesTSVMissing:
...
+      The compulsory file '/samples.tsv' is missing.
+      This file is REQUIRED if the `sample-<label>` entity is present in any
+      filename within the dataset.
+      See 'Modality agnostic files' section of the BIDS specification.
+    level: error
+  selectors:
+    - '"sample" in entities'
+  checks:
+    - exists('samples.tsv', 'dataset')
+
```
SylvainTakerkart, line 90: my feedback:
- adding this does the job for icephys!
- but somehow, it's a parallel rule to the existing one just above that applies just for `micr`... imagine having a `micr` dataset where the `samples.tsv` file is missing, both  SamplesTSVMissing and SamplesTSVMissingForSampleEntity will be triggered, which is probably not desired...

should we stick to a set of selectors that are of the same nature, i.e that test the modality / datatypes?

src/schema/rules/files/raw/microephys.yaml
```diff
@@ -20,6 +18,18 @@ microephys:
...
+microephys__icephys:
+  $ref: rules.files.raw.microephys.microephys
+  suffixes:
+    - icephys
+  datatypes:
+    - icephys
+  entities:
+    $ref: rules.files.raw.microephys.microephys.entities
+    sample: required
+
```
SylvainTakerkart, line 32: Hi!

Here again, this clearly does the job for icephys, thanks! But I'm not sure making the `icephys` rule inherit from the `microephys` rule is the best implementation: it might also be possible to have to rules for 'ecephys' and 'icephys' without inheritance and that might make this code more easily readable? I dunno what's the best solution, if it exists elsewhere in BIDS, we might just use the same solution to be consistent?

Anyhow, this solution does the job!

SylvainTakerkart (commented):
Hi, overall, this does the job, thanks @bendichter !

However, there might be room for improvment, but I'm not that sure: I made three comments which are aimed at opening the discussion, because for all three of them, I'm not sure what's the best solution, so we might want to ask another review from somebody with a more general view on / more experience with the BIDS specs?

src/schema/rules/checks/dataset.yaml
bendichter: Good catch. I went the other direction and merged the two into one rule keyed on the entity. Microscopy requires `sample-` on every file, so the entity selector covers `micr` as well, and the single check now matches the sentence in the samples file description exactly. The one difference is that the error is reported per file rather than once at the dataset level, which I think is acceptable.

src/schema/rules/files/raw/microephys.yaml
bendichter: I would keep the inheritance. It is the established pattern in this schema: `photo__micr` and `photo__microephys` both inherit from `photo`, and the channels rules do the same for meg, emg, and microephys. Two independent rules would duplicate the extension list and the entity block, and those would drift over time.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Agreed that most of this applies to microscopy too, and the microscopy text is thin. I would rather not edit another modality's page inside the BEP review, since Chris has asked that substantial changes go in separate PRs. I will open an issue to generalize this text once BEP032 is merged, and keep the section here for now.

PR #2481, opened 2026-07-29
[ENH] BEP032: Move anatomical location to the electrode level

bendichter:
Implements the decisions we reached in the working group meeting on 2026-07-29 about anatomical location, addressing the threads on [Uberon intent](https://github.com/bids-standard/bids-specification/pull/2307#discussion_r3648196690) and [`(estimated|proof)` rendering](https://github.com/bids-standard/bids-specification/pull/2307#discussion_r3647855473), and following up on https://github.com/bids-standard/bids-specification/pull/2307#issuecomment-5110429004. This targets `bep032-review`, so it merges into #2307 rather than into master. It is independent of #2480 and the two do not touch the same lines.

## `probes.tsv`

`associated_brain_region` and `associated_brain_region_id` become `anatomical_location` and `anatomical_location_id`. The columns now describe the probe as a whole, so the structure named there must be one that contains every recording site on the probe, and may be as coarse as the brain or a single hemisphere. The renaming also drops the assumption that a probe is in the brain, which does not hold for a probe placed in peripheral tissue, muscle, or spinal cord.

Note that this differs from what I proposed in the comment on #2307. I had suggested reading the probe-level columns as the intended surgical target. The meeting landed somewhere better: the probe-level value is whatever level of anatomical detail actually holds for the whole probe, which is a claim you can make with confidence rather than a statement of intent.

## `electrodes.tsv`

`location` becomes `anatomical_location`, and `anatomical_location_id` and `localization_method` are added. Per-contact localization belongs here, since a long shank probe passes through several structures and only the electrode table can describe that. Putting the identifier here also means that different electrode tables, distinguished by the `space-<label>` entity, can carry localizations from different atlases produced by different methods.

## Removals

`associated_brain_region_quality_type` is removed. We could not draw a clear line between its `estimated` and `proof` values in practice, and knowing the actual method, for example whether the location was histologically confirmed, is both more useful and easier to record. The free-text `localization_method` column replaces it. This also resolves the rendering problem @effigies reported, since the `(estimated|proof)` notation that was truncating the table is gone.

`reference_atlas` is removed from `probes.tsv`. The atlas is now named in a `ReferenceAtlas` field in the electrodes sidecar, which is where the terms it applies to live. A `LocalizationMethod` field is available in the same sidecar for the case where one method applies to every electrode in the file. Both are optional, and the specification states that the method must not be given in both the column and the sidecar.

## Ontology Guidance

The guidance no longer prefers species-independent terms. Terms from a species-specific atlas are acceptable and are often the better choice, and identifiers should be resolvable, either as a URI or as a prefixed identifier whose prefix identifies the source. This is what @effigies' question about whether Uberon was required or merely an example was getting at, and the answer is now stated explicitly rather than implied by an example.

## Requirement Levels

The two probe columns keep the requirement level of the columns they replace (RECOMMENDED). On the electrodes table, `anatomical_location` keeps the RECOMMENDED level that `location` had, and the two new columns are OPTIONAL, as are both sidecar fields. Happy to adjust if the group would rather set these differently.

## Verification

`bidsschematools` tests pass (81 passed, 1 skipped), as do `pre-commit` on the changed files and `npm run remark`. I confirmed that both the probes and electrodes column tables and the new sidecar table render, and that no references to the removed columns remain anywhere in `src/`.

One thing worth a second pair of eyes: I updated the example tables to use Allen Mouse Brain Atlas identifiers (`MBA:315` for isocortex, `MBA:985` for MOp, `MBA:382` for CA1). I believe those structure identifiers are right, but someone who works with the atlas daily should confirm them before this merges.

@lzehl, this is the second item from the meeting. The first is #2480.


src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,59 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
 
 To specify electrode positions in surgical space, individual anatomical space, or a common coordinate system (such as the Allen CCF), use an additional `*_electrodes.tsv` file with a [`space-<label>`](../appendices/entities.md#space) entity. See the [`*_coordsystem.json` section](#coordinate-system-json-_coordsystemjson) for details on defining these coordinate systems.
 
+### Anatomical Location
+
+Anatomical location is described at two levels of detail.
+The `anatomical_location` column of `*_probes.tsv` names the structure in which the probe as a whole
+is placed, which for a probe spanning several structures MUST be a structure containing all of them,
```
effigies, line 333: The way I'm reading this MUST is as a directive to tools that *generate* this column to enforce the superstructure-substructure relationship between probes and electrodes, as they will presumably have knowledge of/access to the specific ontology being referenced.

I don't think that this can translate to a BIDS validator check that errors if the superstructure-substructure relatinship doesn't hold. I don't think we want to try to load arbitrary ontologies and resolve term relationships. Of course, nothing stops a microephys-specific validation tool from enforcing this and being adopted by archives/tools that need the enforcement.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,59 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+The `anatomical_location` column of `*_electrodes.tsv` names the structure in which an individual
+recording site is located, and is where a per-contact localization belongs.
+Only the electrode-level column can describe a probe that passes through several structures,
+which is common for long shank probes.
+
+Both tables also accept an `anatomical_location_id` column giving an identifier for the structure.
+Identifiers SHOULD be resolvable, either as a URI or as a prefixed identifier whose prefix
+identifies the source, so that a reader can look the term up.
+Terms may be taken from a species-independent ontology such as Uberon or from a species-specific
+atlas, and a species-specific atlas is often the better choice.
```
effigies: https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files already allows columns to be given a TermURL, and the levels for enumerable values can also given a TermURL.

For the `electrodes.tsv` below, you might have an `electrodes.json` snippet like:

```yaml
  "anatomical_location_id": {
    "TermURL": "https://URL/OF/ONTOLOGY",
    "Levels": {
      "MBA:985": {
        "Description": "Some description",
        "TermURL": "https://URL/OF/ONTOLOGY#985"
      }
    }
  }
```

Now, if you're providing URIs for some column, arguably you could do the same thing for `anatomical_location`, and then the ID is looked up in the JSON. I think you have more of a sense of how people actually want to engage with these fields, so this is merely a note of the potential for duplication for your consideration.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,59 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+
+The method used to determine the location SHOULD be recorded.
+Where it differs between recording sites, use the `localization_method` column of
+`*_electrodes.tsv`.
+Where the same method applies to every electrode in the file, it MAY be given once in the
+`LocalizationMethod` field of the corresponding `*_electrodes.json` file instead,
+but it MUST NOT be given in both places.
```
effigies, line 353: This needs a check, like:

```yaml
RuleName:
  issue: ...
  selectors:
    - intersects(datatype, ['icephys', 'ecephys'])
    - suffix == 'electrodes'
    - extension == '.tsv'
    - type(columns.localization_method) != 'null'
  checks:
    - '!("LocalizationMethod" in sidecar)'
```

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,59 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+The atlas from which the terms are taken SHOULD be named in the `ReferenceAtlas` field of the same
+sidecar file, which matters most for atlases that provide no resolvable identifiers.
```
effigies: Do you need to promote `ReferenceAtlas` to RECOMMENDED then? Or under some condition?

I don't know what "atlases that provide no resolvable identifiers" means.

Again, an example `electrodes.json` seems like it would be helpful.

effigies (commented):
Some notes. Apologies for the hit and run. Will be out-of-office for the next week.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: I agree, I don't want to add anatomical partology logic to BIDS, for a number of reasons. The MUST was intended to be a directive to the user, not a statement about validation. Maybe SHOULD would be better?

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Added in 0ff9a5c, essentially as you wrote it. I kept `intersects([datatype], [...])` to match the existing check in that file, and phrased the message so it names both places the value was found.

src/modality-specific-files/microelectrode-electrophysiology.md
effigies: Either way, just as long as it's understood that this is out-of-scope for the BIDS validator. Other tools will need their own validation in order to rely on this relationship, but as an instruction to a user, this is signalling the importance, e.g., tools are likely to break (MUST) or may behave in unexpected ways (SHOULD).

src/modality-specific-files/microelectrode-electrophysiology.md
effigies: Sorry, I think I wrote that last paragraph unclearly. The missing explicit sentence is: "If you provided this JSON description of the `anatomical_location` field, it may become redundant with the `anatomical_location_id` field."

Again, this is meant as a note, not a critique/request for changes.

bendichter: You have convinced me. I have dropped `anatomical_location_id` from `electrodes.tsv` and `probes.tsv` here, and from `samples.tsv` in #2484, and the terms are now documented in the sidecar with `TermURL` and `Levels` in the ordinary way.

What settled it was looking at the other identifier columns in BIDS. `participant_id`, `sample_id`, `session_id`, `desc_id`, `shank_id`, and `stream_id` all identify an entity within the dataset. `anatomical_location_id` would have been the only column whose values are external ontology references, so it was introducing a pattern that does not otherwise exist rather than following one.

`anatomical_location` stays as a plain human readable name, which a curator can fill in without knowing any ontology, and I have added an example showing the `Levels` form.

One thing this gives up, for the record: `TermURL` needs a URL, so an atlas that mints identifiers but publishes no per-term URL can no longer have them recorded in the table. That is the case `ReferenceAtlas` exists for, and I have narrowed its description to exactly it.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: You are right that these disagreed. The sidecar rule declared the field OPTIONAL while the description called it RECOMMENDED, so the recommendation was not backed by anything. That is the same problem we just fixed on `anatomical_reference_point`, so thank you for catching it here too.

I have kept it OPTIONAL rather than promoting it, because a flat RECOMMENDED would warn on every `electrodes.json`, including the many that carry probe-relative coordinates and no anatomical columns at all. Instead there is now a warning-level check that asks the question directly: if `anatomical_location` is given, but the sidecar neither resolves the terms through `TermURL` or `Levels` nor names the atlas in `ReferenceAtlas`, then the terms cannot be interpreted.

On the wording, that was unclear and I have replaced it. What I meant by an atlas with no resolvable identifiers is one that publishes region names but nothing you can look them up by. D99 is the example that came up in our working group meeting, as against Waxholm, which is ontologized. The description now says that directly, and ties it to there being no `TermURL` to give.

I have also added two `electrodes.json` examples, one for each case.

src/modality-specific-files/microelectrode-electrophysiology.md
effigies: In general, TermURLs would be better named TermURIs, which could be a URN and does not need to be resolveable. The BIDS Validator only checks that URLs are structurally valid URIs and does not attempt to resolve them.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,97 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+The `anatomical_location` column of `*_probes.tsv` names the structure in which the probe as a whole
+is placed, which for a probe spanning several structures MUST be a structure containing all of them,
+and may be as coarse as `brain` or a single hemisphere.
+The `anatomical_location` column of `*_electrodes.tsv` names the structure in which an individual
+recording site is located, and is where a per-contact localization belongs.
+Only the electrode-level column can describe a probe that passes through several structures,
+which is common for long shank probes.
+
+Terms may be taken from a species-independent ontology such as Uberon or from a species-specific
+atlas, and a species-specific atlas is often the better choice.
```
bendichter:
```suggestion
atlas or ontology such as the [Mouse Brain Atlas Ontology](https://www.ebi.ac.uk/ols4/ontologies/mba), and a species-specific is often the better choice.
```

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,97 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+The source the terms come from SHOULD be documented in the sidecar file, in the way BIDS
+documents any other tabular column, as described in
+[Tabular files](../common-principles.md#tabular-files).
+Where the individual terms resolve, give each one its own `TermURL` under `Levels`.
+Where they do not, name the atlas in the `ReferenceAtlas` field of the sidecar,
+which is then the only thing that makes the terms interpretable.
```
bendichter:
```suggestion
Where they do not, name the atlas in the `ReferenceAtlas` field of the sidecar.
```

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: After sitting with this, I am going to keep the MUST. The sentence defines what the probe-level value means: a structure that contains every recording site on the probe. A value that covers only part of the probe is incorrect data rather than a discouraged practice, and the spec uses MUST for requirements like this even when checking them is out of scope for the BIDS validator, which I agree it is here. Resolving on that understanding.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Understood on TermURL vs TermURI, and good to know the validator only checks that the value is a structurally valid URI. Since the identifier columns are gone as of b8afb5d and the terms are now documented in the sidecar with TermURL and Levels, I believe this thread is fully addressed, so I am resolving it.

src/modality-specific-files/microelectrode-electrophysiology.md
bendichter: Closing this out. ReferenceAtlas stays OPTIONAL in the sidecar rule, and ac5c1b0a added the warning-level check that fires when anatomical_location is given but the sidecar neither documents the terms with TermURL or Levels nor names the atlas in ReferenceAtlas. The confusing wording about atlases that provide no resolvable identifiers is gone from the text.

lzehl:
@bendichter you had two questions for me regarding this PR which I'm going to answer here. 

**1. Is a free-text method column enough, or should we keep a controlled quality-type field alongside it, given the openMINDS criteria quality types?** 
The quality type of the localization method is not necessarily inferrable from the provided free-text method (e.g. in particular if the method name is not specific enough). However the question is if we want to encourage/enforce a quick classification according to the quality type (asserted vs processive in openMINDS lingo) or if this is less relevant here/for now. Since this theoretically is a reusable quality assessment across all claimed methods within the BIDS it is maybe worth to not introduce it right now but discuss this in the broader BIDS context (maybe as part of how quality of data, methods, etc is handled in BIDS). @effigies maybe you could provide feedback here ?

2. And are the Allen MBA identifiers in the examples (MBA:315 isocortex, MBA:985 MOp, MBA:382 CA1) correct?
These are all correct variants of the short versions of the identifiers for the AMBA brain regions as registered by Allen Institute. However MBA_315 etc is also used in some cases. Problem with Allen Institute is that they actually do not define resolvable, globally unique PIDs to their brain regions and ontology registries built their own wrappers for those (several of those also not directly resolvable). So... not sure if we want to go through all the possibilities of best identification of brain regions? For me the example is efficient enough considering the missing resolvable, globally unique PID. Where region IDs properly resolve is for UBERON parcellations (e.g. http://purl.obolibrary.org/obo/UBERON_0001950).

bendichter:
Thanks @lzehl, this is very helpful.

On the quality type question, I agree with deferring it. A free-text method column captures what most experimenters can readily provide, and a controlled asserted/processive classification seems like something that should be designed once for BIDS as a whole rather than introduced piecemeal here. I'll leave that to the broader discussion and would also welcome @effigies's take.

On the identifiers, thanks for confirming the short forms are correct. One small addendum to the resolvability picture: while I agree that the Allen Institute itself has never minted persistent resolvable identifiers, there is a registered third-party route. The [Bioregistry](https://bioregistry.io) (a community-maintained registry of identifier prefixes for the life sciences, similar in spirit to identifiers.org, that maps compact identifiers like `mba:985` to resolvable URLs) registers the `mba` prefix and points it at the purls minted by the BICAN project (`https://purl.brain-bican.org/ontology/mbao/MBA_985`), which are in turn configured to resolve to the corresponding EBI OLS4 term pages. At the moment the BICAN purl server is misconfigured and every purl redirects to their config repo instead (I filed brain-bican/purl.brain-bican.org#73 about it), so today the form that reliably resolves is the OLS term page itself. This actually illustrates your point about wrapper fragility fairly well.

Given that, I plan to keep the example as is, with the Allen atlas viewer URLs as the `TermURL` values, since those resolve and land somewhere genuinely informative. The sidecar mechanism in this PR already covers both situations: per-term `TermURL` where something resolvable exists, and `ReferenceAtlas` alone where it does not.

src/modality-specific-files/microelectrode-electrophysiology.md
```diff
@@ -326,26 +326,96 @@ This rule is different from the electrodes.tsv table of the [iEEG modality](intr
...
+Anatomical location is described at two levels of detail.
+The `anatomical_location` column of `*_probes.tsv` names the structure in which the probe as a whole
+is placed, which for a probe spanning several structures MUST be a structure containing all of them,
+and may be as coarse as `brain` or a single hemisphere.
+The `anatomical_location` column of `*_electrodes.tsv` names the structure in which an individual
+recording site is located, and is where a per-contact localization belongs.
+Only the electrode-level column can describe a probe that passes through several structures,
+which is common for long shank probes.
+
+Terms may be taken from a species-independent ontology such as Uberon or from a species-specific
```
yarikoptic:
```suggestion
Terms may be taken from a species-independent ontology such as [Uberon](https://obophenotype.github.io/uberon/) or from a species-specific
```

I took url from their github page https://github.com/obophenotype/uberon

PR #2483, opened 2026-07-29
[ENH] BEP032: Remove the channel_label column

bendichter:
Implements the decision we reached in the working group meeting on 2026-07-29 about `channel_label`, addressing https://github.com/bids-standard/bids-specification/pull/2307#discussion_r3648204150 and following up on https://github.com/bids-standard/bids-specification/pull/2307#issuecomment-5111406144. This targets `bep032-review`, so it merges into #2307 rather than into master.

## The Change

`channel_label` is removed from `channels.tsv`. The column was defined as a human readable identifier for specifying the content of signals not generated by electrodes, but a channel already has an identifier in the required `name` column, and the categories its examples gave are values of the `type` column. This BEP added `SYNC` and `BEH` to that enum, and `SYSCLOCK`, `TRIG`, `STIM`, and `MISC` were there already.

Anything the `type` column does not capture belongs in `description`, so rather than deleting the column and leaving nothing in its place, this adds a description addendum to `channels.tsv` pointing at that use:

> For a channel that is not generated by an electrode, and so has `n/a` in the `electrode_name` column, use this column to state what the signal is, for example, `Synchronization pulse from the behavioral control system`. The category of such a channel is given by the `type` column, for example, `SYNC` or `BEH`.

The example phrasing there is deliberate. @lzehl made the point in the meeting that `behavioral signals` and `behavioral cues`, the examples the removed column carried, are terms rather than descriptions, so the addendum shows a description that actually describes something.

## Scope

This is a small change. `channel_label` appeared only in `objects/columns.yaml` and the `microephysChannels` rule, and was not used in any example table or referenced in any prose, so nothing else needed updating.

## Verification

`bidsschematools` tests pass (81 passed, 1 skipped), as do `pre-commit` on the changed files and `npm run remark`. I confirmed no references to `channel_label` remain anywhere in `src/`, and that the rendered channels table now carries the addendum on `description`.

@lzehl, this is the fourth item from the meeting. The others are #2480, #2481, and #2482.


CodyCBakerPhD (approved):

CodyCBakerPhD:
sounds fair

PR #2487, opened 2026-08-04
fix(ci): Restore green Validation CI under ruff 0.16 + add ruff to pre-commit

yarikoptic:
**Disclaimer**: done with claude code and diff eyeballed by me.  Among alternatives we had just to pin ruff to prior version with less checks but I decided that we better address them while at it!

## Summary

The [`Validation` workflow](https://github.com/bids-standard/bids-specification/actions/workflows/validation.yml) has been red on **every push, PR, and merge since 2026-07-23**. The failure is in the `python-style` job's `uvx ruff format --diff` step, and it's a rot introduced by **ruff 0.16.0 breaking changes**, not by any BIDS PR.

- Last green master run: 2026-07-20 (4523097).
- First red master run: 2026-07-28 (5b2cf70, PR #2479 was merged with `python-style` red).
- Root cause: [ruff 0.16.0 release, 2026-07-23](https://github.com/astral-sh/ruff/releases/tag/0.16.0).

Two independent breaking changes in ruff 0.16:

1. **`ruff format` now formats Python code blocks inside Markdown files by default.** `macros_doc.md` documents Jinja2 macro syntax inside <code>```python</code> fenced blocks, which ruff tries to re-indent as Python. This is the visible failure.
2. **`ruff check`'s default rule set expanded from 59 to 413 rules.** Twenty-two additional lint violations surface across `tools/` and `pdf_build_src/`. They were hidden until now because `ruff format` fails first and short-circuits the job.

<details><summary>Which PRs are affected (they are not the cause; ruff is)</summary>

All open PRs since 2026-07-23 have a red `python-style` check for this reason. Confirmed rerunning against several PRs picks up the same ruff 0.16.x and fails identically. Once this lands on master, rebases / retriggers of those PRs should go green.

</details>

## Changes

- Retag the four Jinja2 examples in `macros_doc.md` as <code>```jinja</code> rather than <code>```python</code> — they are macro calls, not Python. This addresses issue (1) properly rather than pinning ruff or excluding `*.md` from the formatter.
- Apply `ruff check --fix` + `--unsafe-fixes` for the 34 auto-fixable violations across `tools/` and `pdf_build_src/` (import ordering, f-strings, `OSError` alias for `IOError`, unused `noqa`, `set()`/`dict()` literals, `sorted()` redundant `list()`, implicit `Optional` → `X | None`, `_`-prefix for unused unpacks, `datetime.fromisoformat` handles `Z` natively in 3.11+, assert-with-print, dict `.items()`, mutable default argument).
- Manual fixes for the 9 residual violations with no auto-fix:

<details><summary>Detail of manual fixes</summary>

| File                                | Rule       | Change                                                                                    |
| ----------------------------------- | ---------- | ----------------------------------------------------------------------------------------- |
| `tools/add_contributors.py:156`     | `B015`     | `contributions is None` → `contributions = None` (was a no-op — real bug, meant assignment) |
| `pdf_build_src/pandoc_script.py:80` | `PLW1510`  | Add explicit `check=False` to `subprocess.run(cmd)`                                       |
| `pdf_build_src/process_markdowns.py:89` | `DTZ002` | `datetime.today()` → `datetime.now().astimezone()`                                       |
| `tools/add_contributors.py:264`     | `PLR1704`  | Rename inner context manager var to avoid shadowing `tributors_file` parameter            |
| `tools/add_contributors.py:288`     | `PLC0206`  | `for key in d:` + `d[key]` → `for key, value in d.items():`                               |
| `tools/add_contributors.py:590`     | `SIM102`   | Combine nested `if` + walrus into a single `if` with `and`                                |
| `tools/make_changelog.py:165`       | `ASYNC230` | `# noqa: ASYNC230` on the one-shot sync `open()` at end of async `main()` — not a hot path |
| `tools/no-bad-latin.py:132`         | `B008`     | Move `os.path.join(ABSOLUTE_HERE, "src")` into module-level `DEFAULT_SRC_DIR`             |
| `tools/no-bad-latin.py:165`         | `TRY002`   | `raise Exception(...)` → `raise RuntimeError(...)`                                        |

</details>

Test-touching change worth flagging: the `ISC004` fix in `tools/schemacode/src/bidsschematools/tests/test_validator.py` adds parens around an implicit string concatenation (was likely a missing-comma-looks-like-a-bug pattern — the strings were meant to be one long path, and now that intent is explicit).

## Test plan

- [x] `uvx ruff format --diff` locally on the resulting tree → clean (122 files already formatted).
- [x] `uvx ruff check` locally on the resulting tree → clean.
- [x] `pytest tools/schemacode/src/bidsschematools/tests/test_validator.py` → 22 passed.
- [x] CI on this PR — `python-style` job to go green.
- [x] Review rendered site around those spots where we changed spec from python to jinja for macros invocations etc
- [ ] After merge, other open PRs' `python-style` should also go green once rebased.

### DIFF
--- a/tools/make_changelog.py
+++ b/tools/make_changelog.py
@@ -19,10 +19,10 @@ import re
 import subprocess as sp
 import sys
 from dataclasses import dataclass
-from datetime import date, datetime, timedelta, UTC
+from datetime import UTC, date, datetime, timedelta
 from pathlib import Path
 
-from gql import gql, Client
+from gql import Client, gql
 from gql.transport.aiohttp import AIOHTTPTransport
 
 SEARCH_MERGED_PRS_QUERY = gql("""\
@@ -122,7 +122,7 @@ async def get_merged_prs(
             if not pr:
                 continue
 
-            merged_date = datetime.fromisoformat(pr["mergedAt"].replace("Z", "+00:00"))
+            merged_date = datetime.fromisoformat(pr["mergedAt"])
 
             prs.append(
                 PR(
@@ -162,7 +162,7 @@ async def main() -> None:
         client, repository="bids-standard/bids-specification", merged_after=start_date
     )
 
-    with open(repo / "src/CHANGES.md", "w") as changelog:
+    with open(repo / "src/CHANGES.md", "w") as changelog:  # noqa: ASYNC230
         changelog.write("# Changelog\n\n## Upcoming\n\n")
         changelog.write("\n".join([*map(str, prs), "", changes]))
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,6 +50,13 @@ repos:
     hooks:
       - id: codespell
         args: ["--config=.codespellrc", "--dictionary=-", "--dictionary=.codespell_dict"]
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Keep in sync with the ruff version used by .github/workflows/validation.yml
+    rev: v0.16.1
+    hooks:
+      - id: ruff-check
+        args: [--fix]
+      - id: ruff-format
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v2.1.0
     hooks:
--- a/macros_doc.md
+++ b/macros_doc.md
@@ -29,7 +29,7 @@ layouts in the documentation. The macro takes a single parameter, the directory
 tree to be displayed in JSON format. If you insert the following in the BIDS
 markdown document:
 
-```python
+```jinja
 {{ MACROS___make_filetree_example(
 
    {
@@ -230,7 +230,7 @@ The macro will create the different columns of the table:
 
 A general description of that macro call would look like this:
 
-```python
+```jinja
 {{ MACROS___make_metadata_table(
    {
       "TermToRender": "REQUIREMENT_LEVEL plus anything else after", "Extra content you want to append after the description of that term."
@@ -264,7 +264,7 @@ If you wanted to add some extra content to that table, but without modifying the
 definition in the schema, then you could just add some extra content into the
 macro call.
 
-```python
+```jinja
 #### Reconstruction
 
 {{ MACROS___make_metadata_table(
@@ -293,7 +293,7 @@ contains an entry for `MoonPhase`. If this is the case because the term already
 exists and is used somewhere in the BIDS specification, you are in luck and you
 can just stop there.
 
-```python
+```jinja
 #### Reconstruction
 
 {{ MACROS___make_metadata_table(
--- a/pdf_build_src/pandoc_script.py
+++ b/pdf_build_src/pandoc_script.py
@@ -4,8 +4,9 @@ This is done once the duplicate src directory is processed.
 """
 
 import subprocess
-import yaml
 from pathlib import Path
+
+import yaml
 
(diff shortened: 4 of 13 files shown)


tools/make_changelog.py
```diff
@@ -122,7 +122,7 @@ async def get_merged_prs(
             if not pr:
                 continue
 
-            merged_date = datetime.fromisoformat(pr["mergedAt"].replace("Z", "+00:00"))
+            merged_date = datetime.fromisoformat(pr["mergedAt"])
```
yarikoptic, line 125: yet to get details on the effect here...

yarikoptic: seems legit and indeed a new ruff check   https://docs.astral.sh/ruff/rules/fromisoformat-replace-z/

<details>
<summary>more of AI blurb</summary> 

  When it was added: commit a6cbf0395 chore: Update changelog script by @effigies on 2025-08-26. That commit was a substantial rewrite of tools/make_changelog.py (+132/-62), and the merged_date = 
  datetime.fromisoformat(pr["mergedAt"].replace("Z", "+00:00")) line was introduced fresh there — not carried over from earlier code. So it's not load-bearing legacy; it's a defensive Python idiom the author reached for.

  Why the idiom exists: datetime.fromisoformat() did not parse the Z suffix until Python 3.11. Pre-3.11 code that consumed GitHub API timestamps (which are ISO-8601 with Z) had to swap Z for +00:00 first.

  Why it's redundant here:
  - tools/make_changelog.py:3 declares # requires-python = ">=3.13" in its PEP-723 inline metadata block, and the script already uses datetime features that are 3.11+ only (from datetime import UTC on line 22).
  - The top-level pyproject.toml also has requires-python = ">=3.11".
  - Verified locally: even Python 3.11.2 parses datetime.fromisoformat("2026-07-28T19:44:51Z") correctly, returning +00:00 tzinfo.
  - GitHub GraphQL's DateTime scalar is always ISO-8601 with a Z suffix, so no format-variation edge cases.

  Verdict: Fully legitimate — the transformation is a no-op on every Python version that can even run this script. Ruff's FURB162 (fromisoformat-replace-z) is exactly the rule for this case, and the removal doesn't change
  behavior.

</details>

yarikoptic:
I think this is ready for review and merge .  @bids-standard/maintainers please review (`@effigies` I believe is somewhere in the woods) and merge to make other PRs green!  Since it is not changes to the spec, I think even a single approval should be enough IIRC.

yarikoptic:
and links check is failing for  https://www.json.org/ which seems indeed down ATM. I get

```
The page you are looking for is temporarily unavailable.
Please try again later.
eb067428129b39f6326605d8c56cde2f bce05e7ab16d792f825027da69e7d662 136eaeff0fa23a9859a5ac999e225427
```

Remi-Gau:
suggestion: run ruff via pre-commit so the version of ruff can be controlled and autoupdated monthly

yarikoptic:
@Remi-Gau good idea but I am just trying to get CI green ATM... subsequent PR could do extras... approve? ;)

yarikoptic:
damn, claude was still active, and it was too easy to add... so "we" did, and I checked that now running on `--all` does nothing, so I guess we might be good to go with it

effigies (approved):
Will patch my changes over this for separate review.

PR #2494, opened 2026-08-07
chore: Run ruff through pre-commit, avoiding drift

effigies:
The [Validation](https://github.com/bids-standard/bids-specification/actions/workflows/validation.yml) workflow is failing because ruff is trying to reformat things. This results from using `uvx ruff` instead of a pinned version. The simplest way to handle this is to pin in one place (pre-commit). Failures will still show up in CI, and failures related to changes in ruff will appear only in version update PRs.

@Remi-Gau Could I bother you for a review of this one?

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -41,15 +41,6 @@ jobs:
       - name: Lint yml files in src/schema
         run: uvx yamllint -f standard src/schema/ -c .yamllint.yml
 
-  # Python best practices
-  python-style:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v7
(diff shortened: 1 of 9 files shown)


yarikoptic:
Welcome back @effigies ! did you spot also

- https://github.com/bids-standard/bids-specification/pull/2487

?

effigies:
@yarikoptic Rebased the non-trivial alternatives after merging your PR. The first commit is the more consequential, the second is mostly stylistic choices.

yarikoptic (approved):
Looks to me in the right direction, and just to tooling and CI is happy, so let's proceed

Remi-Gau:
> @Remi-Gau Could I bother you for a review of this one?

late to the party (was on holiday) but good with me

PR #2495, opened 2026-08-07
[FIX] Update dbGaP URL

effigies:
The dbGaP search builder (https://web.archive.org/web/20250205225946/https://www.ncbi.nlm.nih.gov/gap/advanced/) was removed without redirecting when https://www.ncbi.nlm.nih.gov/gap/ began to 301 to https://dbgap.ncbi.nlm.nih.gov/home/.

Note that the nearest equivalent search page is https://dbgap.ncbi.nlm.nih.gov/beta/search/, but I don't want to link to a beta page, since that's almost certainly setting up a future URL failure.

The NIH's commitment to persistent URLs is something to behold.


effigies:
#2502 was better.

PR #2501, opened 2026-08-11
[SCHEMA] Require complete BrainVision file triplets

### LINKED ISSUE #2500: Create a schema rule enforcing that .vhdr/.vmrk/.eeg all exist simultaneously.
Currently exists only as an issue code:
https://github.com/bids-standard/bids-specification/blob/master/src/schema/rules/errors.yaml#170

Thank you to @viranovskaya for raising this as an issue with datasets on Openneuro.

viranovskaya:
Closes #2500.

This adds a schema check requiring matching `.eeg`, `.vhdr`, and `.vmrk` files for raw EEG and iEEG BrainVision recordings.

The check reports one error per incomplete recording. I tested complete triplets, every missing extension, files with different recording stems, datasets without an explicit `DatasetType`, and derivatives. It also catches the missing `.eeg` file in OpenNeuro dataset `ds005815`.

Validation:

- `83 passed, 1 skipped` in `bidsschematools`
- `18 passed` in the focused validator checks
- no new failures in the full BIDS Validator suite compared with baseline

The remaining `check_links` failure is caused by the existing dbGaP URL and is addressed separately in #2502.

This checks companion-file existence only. It does not validate the internal `DataFile` and `MarkerFile` references.

### DIFF
--- a/src/schema/rules/checks/general.yaml
+++ b/src/schema/rules/checks/general.yaml
@@ -35,3 +35,30 @@ DuplicateReadmes:
     - match(path, '^/README.*')
   checks:
     - exists(["README", "README.md", "README.rst", "README.txt"], "dataset") == 1
+
+# BIDS Validator Original Issue Code #100
+BrainvisionLinksBroken:
+  issue:
+    code: BRAINVISION_LINKS_BROKEN
+    message: |
+      A BrainVision recording must include matching '.eeg', '.vhdr', and '.vmrk' files.
+    level: error
+  selectors:
+    - dataset.dataset_description.DatasetType == "raw"
+    - intersects([suffix], ["eeg", "ieeg"])
+    - intersects([extension], [".eeg", ".vhdr", ".vmrk"])
+    # Report once per recording: prefer the header, then the marker, then the data file.
+    - |
+      extension == ".vhdr" ||
+      (extension == ".vmrk" &&
+       exists(substr(path, 1, length(path) - length(extension)) + ".vhdr", "dataset") == 0) ||
+      (extension == ".eeg" &&
+       exists(substr(path, 1, length(path) - length(extension)) + ".vhdr", "dataset") == 0 &&
+       exists(substr(path, 1, length(path) - length(extension)) + ".vmrk", "dataset") == 0)
+  checks:
+    - |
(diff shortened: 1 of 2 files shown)


src/schema/rules/checks/general.yaml
```diff
@@ -35,3 +35,30 @@ DuplicateReadmes:
...
+# BIDS Validator Original Issue Code #100
+BrainvisionLinksBroken:
+  issue:
+    code: BRAINVISION_LINKS_BROKEN
+    message: |
+      A BrainVision recording must include matching '.eeg', '.vhdr', and '.vmrk' files.
+    level: error
+  selectors:
+    - dataset.dataset_description.DatasetType == "raw"
+    - intersects([suffix], ["eeg", "ieeg"])
```
effigies: I think this applies whenever we see a brainvision file, not just when it shows up in a raw EEG/iEEG dataset. Or is there a derivative/non-(i)EEG case where we wouldn't want to apply this?

```suggestion
```

src/schema/rules/checks/general.yaml
```diff
@@ -35,3 +35,30 @@ DuplicateReadmes:
...
+    - intersects([extension], [".eeg", ".vhdr", ".vmrk"])
```
effigies: I think this is actually redundant with the next selector.

```suggestion
```

effigies (approved):
Hey, this looks great! Two comments.

rwblair (approved):
Verified it works as expected on some examples. I agree with @effigies if we can open it up to derivatives and ieeg we should.

src/schema/rules/checks/general.yaml
viranovskaya: Agreed. I removed both restrictions, so the check now applies whenever BrainVision files are present, including derivatives. I also reran the raw and derivative test matrix after rebasing.

src/schema/rules/checks/general.yaml
viranovskaya: Agreed. The priority selector already limits this rule to `.vhdr`, `.vmrk`, and `.eeg`, so I removed the redundant extension selector.

PR #2508, opened 2026-09-03
[ENH] Add optional 'duration' column to `_scans.tsv`

yarikoptic:
## Summary

Add an OPTIONAL `duration` column to `_scans.tsv` files, recording the wallclock duration (in seconds) of each recording from first to last acquired data sample.   The column level is OPTIONAL, matching `acq_time__scans`, to avoid introducing warnings for the large number of existing BIDS datasets that do not include it.

Also added schema infrastructure to validate consistency between this column and the (deprecated) `AcquisitionDuration` sidecar metadata field when both are present for the same MRI file.  But we could  drop it if deemed excessive.

**Relationship to existing work:**  

- Related to issue #2362 
- and PR #2441 which propose adding `duration` primarily for concurrent-recording identification with `duration` conditionally REQUIRED.

This PR scoped more generally!  And our initial need with @vmdocua is to get clean information on each MRI volume  for  https://github.com/ReproNim/reprostim/ to align collection audiovideo of stimuli.  Since otherwise, please correct me if I am wrong, **In BIDS we do not have a uniform way to tell how long (in time) any particular data file is**.

We will need to add that to heudiconv, although may be we would agree to agree with @neurolabusc for dcm2niix to somehow extract it may be in continuation to https://github.com/rordenlab/dcm2niix/issues/991#issuecomment-4239706491 discussion.

To implement validation, claude said we need to extend the 'context'!  @effigies @rwblair -- is that true?  should I just not bother for this one?:

> **Note for validator implementers:** the `associations.scans` context requires
> a new lookup pattern — find the `_scans.tsv` at the appropriate directory level
> and return the column values for the row whose `filename` matches the current
> file's relative path (as a scalar, not an array of all values).

## Test plan

- [ ] Decide if worth keeping the check of equality to to `AcquisitionDuration`
- [ ] Validator implementations populate `associations.scans` and exercise
      `ScansAcquisitionDurationConsistency`.
- [x] Schema loads correctly (verified with `bidsschematools.schema.load_schema`).
- [x] All pre-commit hooks pass (yamllint, codespell, prettier, check-yaml).
- [x] CI (`validate_schema` job) on this PR.
- [ ] Finalize examples

### DIFF
--- a/src/modality-agnostic-files/data-summary-files.md
+++ b/src/modality-agnostic-files/data-summary-files.md
@@ -190,11 +190,11 @@ All such included additional fields SHOULD be documented in an accompanying
 Example `_scans.tsv`:
 
 ```tsv
-filename	acq_time
-func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
-func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
-meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27
-meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27
+filename	acq_time	duration
+func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30	420
+func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33	390
+meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27	600
+meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27	600
 ```
 
 ## Sessions file
--- a/src/schema/meta/context.yaml
+++ b/src/schema/meta/context.yaml
@@ -321,6 +321,21 @@ properties:
           path:
             description: 'Path to associated atlas description file'
             type: string
+      scans:
+        description: 'Row from the scans.tsv file corresponding to the current file'
+        type: object
+        required: [path]
+        additionalProperties: false
+        properties:
+          path:
+            description: 'Path to associated scans file'
+            type: string
+          duration:
+            description: 'Value of the duration column for the current file, if present'
+            type: number
+          acq_time:
+            description: 'Value of the acq_time column for the current file, if present'
+            type: string
   # The following properties are populated if the current file is of an appropriate type
   columns:
     description: 'TSV columns, indexed by column header, values are arrays with column contents'
--- a/src/schema/meta/associations.yaml
+++ b/src/schema/meta/associations.yaml
@@ -129,3 +129,12 @@ atlas_description:
     suffix: description
     extension: .json
   inherit: true
+
+scans:
+  selectors:
+    - suffix != 'scans'
+    - extension != '.json'
+  target:
+    suffix: scans
+    extension: .tsv
+  inherit: true
--- a/src/schema/objects/columns.yaml
+++ b/src/schema/objects/columns.yaml
@@ -173,6 +173,18 @@ duration:
   type: number
   unit: s
   minimum: 0
+duration__scans:
+  name: duration
+  display_name: Scan duration
+  description: |
+    Wallclock duration of the recording in seconds,
+    from the acquisition of the first data sample to the end of the acquisition of the last data sample.
+    MUST be strictly positive.
+    For MRI data, this is equivalent to the `AcquisitionDuration` sidecar metadata field;
+    if both are present for the same file they SHOULD be equal.
+  type: number
+  unit: s
+  exclusiveMinimum: 0
 x_coordinate:
   name: x_coordinate
   display_name: Gaze x-coordinate
--- a/src/schema/rules/checks/mri.yaml
+++ b/src/schema/rules/checks/mri.yaml
@@ -173,3 +173,23 @@ SmallDeltaNotConsistent:
     - type(nifti_header) != "null"
   checks:
     - length(sidecar.SmallDelta) == nifti_header.dim[4]
+
+ScansAcquisitionDurationConsistency:
+  issue:
+    code: SCANS_DURATION_MISMATCH_ACQUISITION_DURATION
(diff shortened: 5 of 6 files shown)


src/modality-agnostic-files/data-summary-files.md
```diff
@@ -190,11 +190,11 @@ All such included additional fields SHOULD be documented in an accompanying
...
-filename	acq_time
-func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30
-func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33
-meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27
-meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27
+filename	acq_time	duration
+func/sub-control01_task-nback_bold.nii.gz	1877-06-15T13:45:30	420
+func/sub-control01_task-motor_bold.nii.gz	1877-06-15T13:55:33	390
+meg/sub-control01_task-rest_split-01_meg.nii.gz	1877-06-15T12:15:27	600
+meg/sub-control01_task-rest_split-02_meg.nii.gz	1877-06-15T12:15:27	600
```
yarikoptic, line 197: this one immediately brings a question on the meaning of `duration` for `_split-`'ed files.  Since IIRC it is some mechanistic split, I would say that then it must be the same across `splits` and thus be validated such?

rwblair:
For the `AcquisitionDuration` check, it is currently set as a warning. What would you expect a downstream dataset consumer to do if there is a mismatch here? May be worth making it an error if there is no conceivable reason for them to be different.

Another option would be to make the fields mutually exclusive, to nudge the dataset creator to drop the deprecated field and have a single source of truth.

As for the `split-` question. Scans going into the context is fine. The limitation of the schema expression language is on fine display for enforcing that specific rows have the same value in a given column. We have no filter or list comprehension to extract just the relevant `split-*` files and the rest of their rows. I may be lacking the imagination right now to see how this can be enforced at the moment, I'll bring it up next schema meeting to see if we can figure something out.

Just for fun, the one dataset on openneuro I found using `split-` has the same `acq_time` for each `split-` file, but also has the same `acq_time` for each run of a datafile as well. From ds004998:
```
meg/sub-PuPVlx_ses-PeriOp_task-MoveL_acq-MedOff_run-1_split-01_meg.fif  1925-01-01T09:16:12.112615Z
meg/sub-PuPVlx_ses-PeriOp_task-MoveL_acq-MedOff_run-1_split-02_meg.fif  1925-01-01T09:16:12.112615Z
meg/sub-PuPVlx_ses-PeriOp_task-MoveL_acq-MedOff_run-2_meg.fif   1925-01-01T09:16:12.112615Z
```

yarikoptic:
> What would you expect a downstream dataset consumer to do if there is a mismatch here? May be worth making it an error if there is no conceivable reason for them to be different.

since both are indeed floats serialized to text, as long as we keep underlying type precision the same should be numerically identical I think. We can make it an error.  I just thought to be "gentle" but since we are just introducing it -- there should be no breakage.

> Another option would be to make the fields mutually exclusive, to nudge the dataset creator to drop the deprecated field and have a single source of truth.

well, that AcquisitionDuration is already deprecated for `_bold.nii.gz` in favor of a [complicated combinatorics of others](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#timing-parameters_1). It is still defined [overall](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#timing-parameters) for MRI. In principle, I do not mind deprecating down the road but it touches on overall ambiguity of [metadata "life-cycle"](https://github.com/bids-standard/bids-2-devel/issues/85) so we might first stretch our brain to streamline it overall. For now IMHO duplication is ok'ish, but would be good to fascilitate uniform application. What about making `duration` in `_scans.tsv` to be recommended if defined as `AcquisitionDuration` or another (if there is) sidecar metadata?

> Just for fun, the one dataset on openneuro I found using `split-` has the same `acq_time` for each `split-` file, but also has the same `acq_time` for each run of a datafile as well. From ds004998:

I think it is a bug for the other run to have the same `acq_time`. Makes sense for splits since they are just "binary level" splits and thus not really multiple, potentially usable separately, chunks of the recording(s).

rwblair:
Consensus of the schema hack meeting today is that validating a set of given `split-` entries in a scans.tsv all have the same `acq_time` is not possible with the current expression language.

yarikoptic:
roger than @rwblair !  I should have joined, ... may be next week ;-)  So -- anything to do here? I am ok to strip addition of changes to context and new rules. Now there is some example(s) too: 
- https://github.com/bids-standard/bids-examples/pull/563

src/schema/meta/context.yaml
```diff
@@ -321,6 +321,21 @@ properties:
           path:
             description: 'Path to associated atlas description file'
             type: string
+      scans:
```
yarikoptic, line 324: this will require bids-validator PR, correct @rwblair ?

yarikoptic: ping on this... I wonder if I should simplify (drop any changes to validation/context) to ease this addition "approval" ;-)

effigies: Yes, this needs a matched validator PR.

yarikoptic: before I waste my and silicon friends' tokens -- what would be your desire here @effigies : should I proceed  that direction of exposing scans files in the context ( :+1: ), or just strip this extra validation ( :-1: ) ?

effigies: Saved you the tokens: https://github.com/bids-standard/bids-validator/pull/449

I think it's reasonable validation. @rwblair Any objection?

yarikoptic: pointed https://github.com/bids-standard/bids-examples/pull/563 to that PR now!

PR #2517, opened 2026-09-15
[FIX] Add definitions of continous/ asynchronous data to phys recordings

sinaklg:
Added definitions of continous and asynchronous data to physiological recordings

These are the proposed offical definitions of continous and asynchronous data as used in the BEP020 manuscript and consistent with BIDS standard (its just clearer now).


src/modality-specific-files/physiological-recordings.md
```diff
@@ -11,14 +11,21 @@
...
 [compressed tabular files](../common-principles.md#compressed-tabular-files)
 ([TSV.GZ file](../glossary.md#tsvgz-extensions)).
 TSV.GZ files MUST be accompanied by a JSON file with the same name as their
 corresponding tabular file but with a `.json` extension.
 
+**Continuous data**: regularly sampled over time at a fixed frequency.
+
+**Asynchronous data**: data or messages generated by a device independently of the sampling
+clock governing an associated continuous recording. Asynchronous events carry no fixed
+temporal relationship to that sampling rate. Their timing precision only needs to be
```
yarikoptic: I am wondering: would eye tracking data sampled continuously but independently of neural data constitute "asynchronous" recording?

Overall I wonder if worth making more explicit relationship that it should be relative to some continuous recording, since likely all those "asynchronous" events etc are internally guided by some clock/sampling rate, just not continuous recording? 

```suggestion
temporal relationship to the sampling rate of the continuous recording(s). Their timing precision only needs to be
```

